PR #2, opened 2024-11-16
Add list, options and conditions.proto files

sallycr:
Add three basic protobuf files for api server 
- list.proto  
- conditions.proto 
- options.proto


idl/michelangelo/api/conditions.proto
```diff
@@ -0,0 +1,66 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// Source: github.com/michelangelo-ai/michelangelo/api
+package michelangelo.api;
```
sallycr, line 4: @mincai @ying-zheng it seems uncommon to have source path as a package name. so changed to michelangelo.api here.

idl/michelangelo/api/conditions.proto
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
```
sallycr: @mincai @jonathanbaker7 We are still under the same copyright. right?

ying-zheng (approved):

idl/michelangelo/api/conditions.proto
sallycr: I see other proto files have the license at the top. e.g. Ray https://github.com/ray-project/ray/blob/master/src/ray/protobuf/gcs_service.proto

.bazelrc
```diff
@@ -0,0 +1,2 @@
+build --disk_cache=~/.bazelcache
```
linzhp:
```suggestion
build --disk_cache=~/.cache/bazel/disk_cache
```
We used `~/.bazelcache` for historical reasons. I think `~/.cache/bazel/disk_cache` is a better place, because we can potentially have `~/.cache/bazel/repository_cache` later. It's better to explicitly call out this is a directory for disk cache.

idl/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+# gazelle:prefix gogoproto
```
linzhp: Some go formatters don't like "gogoproto" prefix. We have to put a patch internally to get around with that. Let's have the prefix matching the repo name, so we people see the import, they immediately know where it is from

```suggestion
# gazelle:prefix gogoproto
```
```suggestion
# gazelle:prefix github.com/michelangelo-ai/michelangelo/gogoproto
```

idl/michelangelo/api/conditions.proto
```diff
@@ -0,0 +1,66 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// Source: github.com/michelangelo-ai/michelangelo/api
+package michelangelo.api;
+
+option go_package = "github.com/michelangelo-ai/michelangelo/api;apipb";
```
linzhp:
```suggestion
option go_package = "apipb";
```
Let's remove the full import path here and rely on Gazelle to generate it based on the repo structure. If we move this proto file in the future, Gazelle can automatically update the import path.

idl/michelangelo/api/list.proto
```diff
@@ -0,0 +1,117 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+// Source: github.com/michelangelo-ai/michelangelo/api
+package michelangelo.api;
+
+import "google/protobuf/any.proto";
+
+option go_package = "github.com/michelangelo-ai/michelangelo/api;apipb";
```
linzhp:
```suggestion
option go_package = "apipb";
```

ditto.

idl/michelangelo/api/options.proto
```diff
@@ -0,0 +1,336 @@
...
+// declarative Michelangelo API resource similar to K8s CRD resource.
+
+syntax = "proto3";
+
+// Source: github.com/michelangelo-ai/michelangelo/api
+package michelangelo.api;
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "github.com/michelangelo-ai/michelangelo/api;apipb";
```
linzhp:
```suggestion
option go_package = "apipb";
```
ditto

idl/michelangelo/api/conditions.proto
sallycr: ty

idl/michelangelo/api/conditions.proto
ying-zheng: This new proto package name, 'michelnagelo.api', is better than the old one, 'github.com.uber.michelangelo.api'

The problem is that we already have the old package names in our michelangelo servers, clients, and metadata (CRD data in etcd and mysql). Changing package name will break backward compatibility. For example, a client tries to get michelnagelo.api.Project will get nothing, as our server can only find github.com.uber.michelangelo.api.Project in our database (etcd / mysql).

I'm trying to set up a mechanism to allow package name and version number changes. This will take a few days.

For now, please keep the original package name

idl/michelangelo/api/options.proto
```diff
@@ -0,0 +1,336 @@
...
+
+// Defines a proto annotation that describes a string field that refers to an API resource.
+message ResourceReference {
+
+  // The resource type that the annotated field references.
+  //
+  // Example:
+  //
+  //     message Model {
+  //       ResourceIdentifier performance_report = 1 [(github.com.michelangelo-ai.michelangelo.api.resource_reference) = {
```
ying-zheng, line 314: Please don't change the protobuf package name for now.
Should be github.com.uber.michelangelo.api.resource_reference not github.com.**michelangelo-ai**.michelangelo.api.resource_reference

idl/michelangelo/api/options.proto
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// An annotation that describes an API resource definition
+// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
+// A message having this option is treated as a CRD
+//
+// Example:
+// message Project {
+//  // Indicates a CRD type. All the options are set to default values.
+//  option (github.com.michelangelo-ai.michelangelo.api.resource) = {};
```
ying-zheng, line 29: Please don't change the protobuf package name for now.
Should be github.com.uber.michelangelo.api.resource not github.com.**michelangelo-ai**.michelangelo.api.resource

linzhp (approved):

idl/michelangelo/api/conditions.proto
sallycr: @ying-zheng old package name is 'package github.com.uber.michelangelo.api;' we need to  figure out the change instead of putting incorrect package name here?

sallycr: note : @ying-zheng and I talk about we go with new package name and figure out the change in internal dependencies.

idl/michelangelo/api/options.proto
sallycr: note : @ying-zheng and I talk about we go with new package name and figure out the change in internal dependencies.

idl/michelangelo/api/options.proto
sallycr: note : @ying-zheng and I talk about we go with new package name and figure out the change in internal dependencies.

idl/michelangelo/api/conditions.proto
ying-zheng: @sallycr You can land this pull request with package name 'michelangelo.api'. I will create another pull request to fix the references

ying-zheng (approved):

PR #5, opened 2024-12-04
Add bazelisk to make bazel hermetic

mincai:
Add tools/bazel symlink to tools/bazelisk.py so that we can automatically install bazel from the repo locally. We are using the python version bazelisk from https://github.com/bazelbuild/bazelisk/blob/master/bazelisk.py.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,8 +1,7 @@
 # gazelle:exclude .git/
 # gazelle:exclude pkg/
 # gazelle:exclude build/
-# gazelle:exclude tools/
-# gazelle:exclude examples/
+# gazelle:exclude tools/bazelisk.py
 # gazelle:proto default
 # gazelle:go_naming_convention go_default_library
 # gazelle:go_naming_convention_external go_default_library
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,5 +1,8 @@
 # gazelle:exclude .git/
+# gazelle:exclude pkg/
 # gazelle:exclude build/
+# gazelle:exclude tools/
+# gazelle:exclude examples/
 # gazelle:proto default
 # gazelle:go_naming_convention go_default_library
 # gazelle:go_naming_convention_external go_default_library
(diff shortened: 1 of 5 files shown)


BUILD.bazel
```diff
@@ -1,5 +1,8 @@
 # gazelle:exclude .git/
+# gazelle:exclude pkg/
 # gazelle:exclude build/
+# gazelle:exclude tools/
```
linzhp: We don't have to exclude `tools`, just in case we want to put some more Python scripts in it

BUILD.bazel
```diff
@@ -1,5 +1,8 @@
...
+# gazelle:exclude examples/
```
linzhp: examples maybe ok to keep too?

BUILD.bazel
mincai: I actually hit an error since gazelle is trying to find dependencies for bazelisk.py ;-) Maybe only exclude that explicitly ?

linzhp: You need this in BUILD.bazel file for gazelle to work with bazelisk, which is what we have internally:

```
# gazelle:python_ignore_dependencies urlparse
```

Or exclude bazelisk explicitly. We don't have tests for bazelisk. So we don't need generate any bazel targets for it.

mincai: Actually internally we disabled python for gazelle. `# gazelle:python_extension disabled` otherwise, gazelle will generate py_binary and py_package targets for `bazelisk` which is unnecessary. 

```
+py_binary(
+    name = "bazelisk",
+    srcs = ["tools/bazelisk.py"],
+    visibility = ["//:__subpackages__"],
+)
+
+py_library(
+    name = "michelangelo",
+    srcs = ["tools/bazelisk.py"],
+    visibility = ["//:__subpackages__"],
+)
```

BUILD.bazel
mincai: Fixed.

PR #6, opened 2024-12-06
Update package names in proto files

ying-zheng:
1. Move protobuf files from idl/michelangelo/api to idl/api. Since this is in michelangelo repo, we don't need the extra "michelangelo" directory
2. Change go package name from "github.com/michelangelo-ai/michelangelo/gogoproto/michelangelo/api" to "github.com/michelangelo-ai/michelangelo/apipb"
3. Set go package name in proto files instead of gazelle / bazel. This is common practice of all the open source proto files I have seen (i.e. google cloud api, etcd, terraform)
4. Remove java options, as we don't currently have any Java client
5. Fix the protobuf package name in the examples (should be michelangel.api instead of github.com.michelangelo-ai.michelangelo.api)


idl/api/conditions.proto
```diff
@@ -1,18 +1,14 @@
...
 package michelangelo.api;
 
-option go_package = "apipb";
-option java_outer_classname = "ConditionsProto";
-option java_multiple_files = true;
-option java_package = "com.github.michelangelo-ai.michelangelo.api";
-
 import "google/protobuf/any.proto";
 
+option go_package = "github.com/michelangelo-ai/michelangelo/api/apipb";
```
linzhp: This means all proto files have to be changed if we want to use something like go.uber.org/michelangelo as the prefix in the future. With the Gazelle prefix approach, only the BUILD.bazel files change

ying-zheng: either "replace in files" in an ide or a sed command can do the work

linzhp: Sounds like additional work without any benefit? We can totally avoid this if we have `go_package = "apipb"` and have Gazelle to assign the fully qualified name.

mincai: +1 to `go_package = "apipb"`

ying-zheng: I tried to set "gazelle:prefix github.com/michelangelo-ai/michelangelo" in BUILD.bazel and  "option go_package = "apipb";" in proto, the result is a invalid go package.
The .go files are generated in "bazel-out/k8-fastbuild/bin/idl/api/apipb_go_proto_/github.com/michelangelo-ai/michelangelo/api/" directory, but inside the .go file the package is set to "package apipb"

My conclusion is "option go_package = "apipb";" does not work. You must set the full package in go_package option

ying-zheng: I updated the diff, set go package name to "github.com/michelangelo-ai/michelangelo/apipb" in gazelle / bazel

idl/api/BUILD.bazel
```diff
@@ -1,3 +1,4 @@
+# gazelle:prefix github.com/michelangelo-ai/michelangelo/apipb
```
linzhp: Are you planning to set `gazelle:prefix` for every package under `idl/`? Wouldn't it be easier if you set it in `idl/BUILD.bazel` file once and for all?

ying-zheng: I can't set apibp in the idl directory
there will be only two packages: github.com/michelangelo-ai/michelangelo/apipb and github.com/michelangelo-ai/michelangelo/api/v2pb

ying-zheng: Alternatively, I can rename the idl directory to proto (we don't plan to have any other idl other than proto), and then set the package name to github.com/michelangelo-ai/michelangelo/proto/api. This makes the go package name perfectly match the actual path.

linzhp: > This makes the go package name perfectly match the actual path

I like this, but it's your call, given there will be only two packages

linzhp (approved):

PR #12, opened 2024-12-10
Add the protobuf for Project CRD 

mincai:
This PR adds the first version of the Project CRD in v2 namespace and the corresponding bazel and go.mod files to make the compilation work. Will do a second pass the remove Uber specific validation for proto fields like owning_team, tier etc.


ying-zheng:
1. Should we remove owning_team in OSS?
   // Owning team UUID such as an uOwn asset UUID.
  string owning_team = 1 [([github.com](http://github.com/).uber.michelangelo.api.validation) = {uuid: true}];

2. In Uber, we have tier 5 projects. Why did you change the max project tier to 4 in the pull request?
-  int32 tier = 3 [(github.com.uber.michelangelo.api.validation) = {min: "1", max: "5"}];
+  int32 tier = 3 [(michelangelo.api.validation) = {min: "1", max: "4"}];

3. Why did you comment job spec instead of deleting it?
// Default job specs for Spark and Ray.
-  JobSpecs job_defaults = 6 [([github.com](http://github.com/).uber.michelangelo.api.validation) = {optional: true}];
+  // JobSpecs job_defaults = 6 [(michelangelo.api.validation) = {optional: true}];

4. Should we reserve the field ids and field names for the fields removed in oss but still needed in Uber?

mincai:
> 1. Should we remove owning_team in OSS?
>    // Owning team UUID such as an uOwn asset UUID.
>    string owning_team = 1 [([github.com](https://github.com/).uber.michelangelo.api.validation) = {uuid: true}];

We can keep owning_team since everything in proto3 is optional anyway. We can remove the validation in OSS  since other companies might use team name instead of UUID.

> 2. In Uber, we have tier 5 projects. Why did you change the max project tier to 4 in the pull request?
> 
> * int32 tier = 3 [(github.com.uber.michelangelo.api.validation) = {min: "1", max: "5"}];
> 
> * int32 tier = 3 [(michelangelo.api.validation) = {min: "1", max: "4"}];

Same here. We should consider remove the validation. However, we need to figure out how to patch it internally. Can we move the validation and index to a different file ?

> 
> 3. Why did you comment job spec instead of deleting it?
>    // Default job specs for Spark and Ray.
> 
> * JobSpecs job_defaults = 6 [([github.com](https://github.com/).uber.michelangelo.api.validation) = {optional: true}];
> 
> * // JobSpecs job_defaults = 6 [(michelangelo.api.validation) = {optional: true}];
> 

I haven't figured out the best way to model the default job specs yet. Need to iterate on this.

> 4. Should we reserve the field ids and field names for the fields removed in oss but still needed in Uber?

We should. Proto fields are monolithically increasing only.

ying-zheng:
I plan to keep a full copy of CRD proto files in Uber internal repo (instead of extension or patch). A "unit test" in Uber internal repo will check the OSS proto messages are subsets of the corresponding Uber internal proto message: the internal version may have extra fields (use reserved field ids / names, or with field id > 1000 and name starting with uber_), the internal version can also have extra validation.

ying-zheng (approved):

PR #13, opened 2024-12-12
api, auth, and logging package, Fix go.mod

ying-zheng:
1. Copy api, auth, logging packages
  These packages will be needed by the go code generated by kubeproto
  I removed most of the constants in api/api.go. Most of those constants are not needed by OSS michelangelo. We should keep those constants in internal repo
  For logging package, I only copied the audit log interface

2. Add go.mod file and empty.go file in proto directory to make 'go mod tidy' happy
'go mod tidy' is not working in the old uber/michelangelo repo, because 'go mod tidy' can't find the go packages generated from protobuf files. I fixed this issue using the method described in
https://stackoverflow.com/questions/75442156/go-mod-tidy-complains-that-the-bazel-generated-protobuf-package-is-missing
and
https://www.php.cn/faq/672765.html

3. remove unused.go, as "k8s.io/apimachinery/pkg/apis/meta/v1" package is already referred by api/api.go


linzhp (approved):

go/api/api.go
```diff
@@ -0,0 +1,115 @@
...
+	metav1.StatusReasonRequestEntityTooLarge: codes.InvalidArgument,
+	metav1.StatusReasonUnsupportedMediaType:  codes.InvalidArgument,
+	metav1.StatusReasonInternalError:         codes.Internal,
+	metav1.StatusReasonExpired:               codes.NotFound,
+	metav1.StatusReasonServiceUnavailable:    codes.Unavailable,
+}
+
+// GetGrpcStatusCode translates Kubernetes error to GRPC status code
+func GetGrpcStatusCode(err error) codes.Code {
+	if err != nil {
```
aud10slave: You can check the edge case first to improve the readability
```
if err == nil {
  return status.OK
}
...
```

go/api/api.go
```diff
@@ -0,0 +1,115 @@
...
+	"google.golang.org/grpc/status"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// IngesterFinalizer is used as the pre-delete hook for ingester controller.
+	// If Ingester finalizer is presented in a CRD object, ingester shall check if
+	// all the pre-delete actions are completed before removing this finalizer.
+	IngesterFinalizer = "michelangelo/Ingester"
```
andriika, line 30: Perhaps rename it to `IngesterAnnotation` to keep in sync with `DeletingAnnotation` and `ImmutableAnnotation`.

go/api/api.go
```diff
@@ -0,0 +1,115 @@
...
+
+// DefaultContextTimeout defines the default timeout for the context
+const DefaultContextTimeout = 30 * time.Second
+
+// K8sStatusReasonToGrpcError map tries to map the K8s api server's error code to
+// gRPC's status error code.  As there is no exact one to one mapping between this
+// two error systems, this map translates the mapping with the best effort.
+// For unlisted StatusReason, surfaceGrpcError() will translate that into codes.Unknown.
+// The api client shall use the detailed error message to determine the unknown error.
+var K8sStatusReasonToGrpcError = map[metav1.StatusReason]codes.Code{
```
andriika, line 51: Can this be private? I noticed there are quite a few public members across different packages in this PR. Do we really need all of them to be public?

go/api/api.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		return codes.Unknown
+	}
+
+	return codes.OK
+}
+
+// K8sError2GrpcError converts K8s error to GRPC error
+func K8sError2GrpcError(err error, msg string) error {
```
andriika, line 94: Perhaps rename to `K8sErrorToGrpcError ` to keep the name convention in sync with `K8sStatusReasonToGrpcError`

go/logging/audit_log_struct.go
```diff
@@ -0,0 +1,20 @@
+package logging
+
+type UAPIAuditLogEvent struct {
```
andriika: Perhaps rename to `AuditLogEvent` to keep in sync with `type AuditLog interface { }`

go/auth/auth.go
```diff
@@ -0,0 +1,51 @@
...
+	Create Action = "Create"
+	// Get ...
+	Get = "Get"
+	// Update ...
+	Update = "Update"
+	// Delete ...
+	Delete = "Delete"
+	// DeleteCollection ...
+	DeleteCollection = "DeleteCollection"
+	// List ...
```
andriika: It's better to remove comments that are not helpful.

go/logging/audit_log_struct.go
```diff
@@ -0,0 +1,20 @@
...
+	Namespace  *string `heatpipe:"namespace"`
```
andriika: 1. Can we do `string` instead of `*string`?
2. `heatpipe` is Uber's internal thing. Can we remove it?

andriika (commented):
There are many utility functions in this pull request that are currently not being used. It seems these utilities are intended to be called in future pull requests. However, it might be better to migrate them gradually, one by one, as they are needed, rather than including all of them in a single pull request.

go/api/api.go
ying-zheng: done

go/api/api.go
ying-zheng: Finalizer and annotations are different things in k8s
https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/

IngesterFinalizer is a finalizer not an annotation

go/api/api.go
ying-zheng: This map is used by src/code.uber.internal/uberai/michelangelo/apiserver/internal/api/api.go, which is in another package in Uber internal repo.

I will make this map private, and let src/code.uber.internal/uberai/michelangelo/apiserver/internal/api/api.go call GetGrpcStatusCode() instead of directly use this map.

ying-zheng:
> There are many utility functions in this pull request that are currently not being used. It seems these utilities are intended to be called in future pull requests. However, it might be better to migrate them gradually, one by one, as they are needed, rather than including all of them in a single pull request.

I was trying to keep the files unchanged while copying to this repo. So that, it's easier to find out what is copied and what is not.

But, since we plan to polish and refactor the code while moving to this repo, the files are going to be changed anyway. I have removed most of the util functions

go/logging/audit_log_struct.go
ying-zheng: done

go/auth/auth.go
ying-zheng: done

go/api/api.go
ying-zheng: done

go/logging/audit_log_struct.go
ying-zheng: done

andriika (approved):

PR #15, opened 2024-12-17
Add Spark and Ray job CRD protos from v2beta1

mincai:
Add Spark and Ray job CRD protos from v2beta1 to v1 unchanged. 

A few TODO items:
- Replace PodSpec with k8s/api/corev1/ PodTemplate
- Split RayJob to RayJob and RayCluster

### DIFF
--- a/proto/api/v2/pod.proto
+++ b/proto/api/v2/pod.proto
@@ -1,0 +1,68 @@
+// This file defines the Pod related messages in Michelangelo API.
+
+syntax = "proto3";
+
+package michelangelo.api.v2;
+
+option go_package = "api/v2";
+
+import "michelangelo/api/options.proto";
+
+// Resource requirements of a Kubernetes Pod or Peloton task
+// This proto is shared among project, pipeline and job. Default value is legitimate for input validation.
+// TODO: Replace with K8s Quantity in ResourceRequirements
+message ResourceSpec {
+  // CPU limit in number of CPU cores
+  int32 cpu = 1 [(michelangelo.api.validation) = {min: "0"}];
+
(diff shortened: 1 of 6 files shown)


proto/api/v2/pod.proto
```diff
@@ -0,0 +1,68 @@
...
+
+  // Custom docker image name
+  string image = 3;
+
+  // Environment variables
+  repeated Environment env = 4;
+
+  // Security token used to read/write to secure HDFS.
+  // TODO: replace this with K8s or Peloton Secrets
+  string hdfs_delegation_token = 5;
```
badcount, line 60: This is uber specific, should we remove this?

mincai: Yes. but prefer to do that in next few diffs.

mincai: Actually we will replace the whole PodSpec with PodTemplate in k8s api. 

https://github.com/kubernetes/api/blob/master/core/v1/generated.proto#L4651

badcount (approved):

proto/api/v2/spark_job.proto
```diff
@@ -0,0 +1,133 @@
...
+  repeated string files = 2;
+
+  // PyFiles is a list of Python files the Spark application depends on.
+  repeated string py_files = 3;
+}
+
+// Defines the status of a Spark job
+message SparkJobStatus {
+  // Deprecated - use status_conditions instead
+  // Conditions that describe states
```
bhjoshi1, line 87: Should we remove all the fields that are marked as Deprecated?

PR #16, opened 2024-12-17
Fix broken bazel build for go

mincai:


badcount (approved):
Skip the go_default_library to unblock

PR #18, opened 2024-12-17
CI/CD: Add git action for ci/cd bazel test

sishilong:
Test git action for ci/cd


sallycr:
good to see that build runs.

PR #19, opened 2024-12-17
Migrate sandbox from uber/michelangeo to michelangelo-ai/michelangelo

sallycr:
Migrate sandbox script to new open source repo.

### DIFF
--- a/tools/sandbox.sh
+++ b/tools/sandbox.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+set -euo pipefail # strict mode
+
+DIR="$(dirname "$(readlink -f "${0}")")/sandbox"
+CLUSTER="ma-sandbox"
+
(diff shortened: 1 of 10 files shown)


mincai (approved):

mincai:
We might want to rename sandbox and move to a different directory. But we can make the changes separately.

andriika:
How do users use it outside of this monorepo?

sallycr:
> We might want to rename sandbox and move to a different directory. But we can make the changes separately.

any alternative way do you want to propose? I browse many others and everyone has own way. but I like the approach of 'mactl demo start' in flytectl does or mlflow does 'mlflow server ...'

sallycr:
> How do users use it outside of this monorepo?

That should be the nexts step. And outside of this monorepo should use APIs. For SDK path, we need to design.

PR #20, opened 2024-12-18
add protobuf

badcount:
Create ray cluster and make ray job kubeRay compatible

1. Create new ray cluster, 
2. Modify existing ray job for job-only specs, 
3. Referring cluster in RayJob for association"

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,7 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="Eslint" enabled="true" level="WARNING" enabled_by_default="true" />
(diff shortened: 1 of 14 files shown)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,7 @@
+<component name="InspectionProjectProfileManager">
```
sallycr, line 1: @badcount Do we need this file in git?

proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
-
-  // Ray head specification
-  HeadSpec head = 3;
-
-  // Ray worker specification
-  WorkerSpec worker = 4;
-
   // User-specified Ray configurations
-  map<string, string> ray_conf = 5;
+  map<string, string> ray_conf = 2;
```
sallycr, line 22: @badcount should we keep the same identifier for the compatibility with internal merge?

sallycr:
fyi @aud10slave Can you also review this diff? thanks

PR #21, opened 2024-12-18
Create ray cluster and make ray job kubeRay compatible

badcount:
1. Create new ray cluster, 
2. Modify existing ray job for job-only specs, 
3. Referring cluster in RayJob for association

### DIFF
--- a/proto/api/v2/ray_cluster.proto
+++ b/proto/api/v2/ray_cluster.proto
@@ -9,6 +9,7 @@ option go_package = "api/v2";
 import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
 import "michelangelo/api/options.proto";
 import "michelangelo/api/v2/pod.proto";
+import "michelangelo/api/v2/job.proto";
 import "michelangelo/api/v2/user.proto";
 
 // Specification for Ray cluster head node
@@ -37,8 +38,8 @@ message RayWorkerSpec {
   // Node type for a worker in a heterogeneous cluster
   string node_type = 4;
 
-  // Ray object store memory will be allocated based on this
-  // ratio of the overall worker memory
+  // Ray object store memory will be allocated based on this ratio of the overall worker memory.
+  // This ratio is specified as an entrypoint parameter to configure the memory allocated for the object store during worker node startup.
   double object_store_memory_ratio = 5;
 
   // Ray start parameters for the worker nodes
@@ -63,11 +64,11 @@ message RayClusterSpec {
   // Additional Ray cluster-wide configurations
   map<string, string> ray_conf = 5;
 
-  // Enable or disable in-tree autoscaling
-  bool enable_in_tree_autoscaling = 6;
+  // Termination spec
+  TerminationSpec termination = 6;
 
-  // True if this is a kill request, used to terminate the cluster
-  bool kill = 7;
+  // Scheduling spec
+  SchedulingSpec scheduling = 7;
 }
 
 // Describes the state of a ray cluster.
@@ -115,45 +116,150 @@ message Endpoints {
   string serve = 5;
 }
 
-// Specification for the head/driver in a Ray cluster
-message Head {
-  // IP and port for accessing the Ray head node
-  string port_ip = 1;
+// Information on the head node
+message RayHeadNodeInfo  {
+  // Name of the pod.
+  // This is the unique identifier for the head node's pod within the Kubernetes cluster.
+  string name = 1;
 
-  // IP address of the head service
-  string service_ip = 2;
+  // Namespace of the pod.
+  // Specifies the Kubernetes namespace where the head node's pod is deployed.
+  // Namespaces allow logical separation of resources within a Kubernetes cluster.
+  string namespace = 2;
+
+  // IP address of the pod.
+  // This is the internal cluster IP assigned to the head node's pod, used for inter-node communication.
+  string ip = 3;
+
+  // Port for clients to connect.
+  // This port is used by external clients to connect to the Ray cluster's head node.
+  // It serves as the primary entry point for submitting jobs and managing cluster state.
+  int32 client_port = 4;
+
+  // Dashboard port for the Ray cluster.
+  // Provides a web-based interface to monitor and manage the Ray cluster.
+  // The dashboard displays metrics, logs, and node details for operational insights.
+  int32 dashboard_port = 5;
+
+  // Jupyter Notebook port for the Ray cluster.
+  // This port is used if a Jupyter Notebook server is running on the head node.
+  // While optional, it allows users to interactively develop and run Ray-based applications.
+  int32 jupyter_notebook_port = 6;
+}
+
+// Information on the pod errors
+message PodErrors {
+  // Name of the failed pod.
+  // Identifies the specific pod within the cluster that encountered an error.
+  string name = 1;
+
+  // Name of the failed container.
+  // Specifies the container within the pod that failed. Pods can have multiple containers, each serving different purposes.
+  string container_name = 2;
+
+  // Exit code of the failed container in the pod.
+  // Indicates the status code returned by the container's process upon failure.
+  // Non-zero exit codes usually signify an error or abnormal termination.
+  int32 exit_code = 3;
+
+  // Reason for the failure.
+  // Provides a brief description of why the container or pod failed.
+  // This is often derived from Kubernetes events and can include reasons like "OOMKilled" or "CrashLoopBackOff".
+  string reason = 4;
+
+  // Message describing the failure.
+  // Offers additional details about the error, including system-level messages or logs.
+  // This field is useful for debugging and determining the root cause of the failure.
+  string message = 5;
 }
 
 // Defines the status of a Ray cluster
-message RayClusterStatus {
+message RayJobStatus {
 
-  // Current state of the Ray cluster
+  // Current state of the Ray cluster.
+  // This field reflects the operational state of the cluster, such as "Running", "Pending", or "Failed".
+  // It is updated based on the observed state transitions in the Ray cluster managed by the Kubernetes operator.
   RayClusterState state = 1;
 
-  // Unique identifier for the generated Ray cluster in the Ray operator
+  // Unique identifier for the generated Ray cluster in the Ray operator.
+  // This ID is critical for associating jobs with specific Ray clusters in Kubernetes.
+  // It ensures the operator can track and manage individual clusters efficiently.
   string cluster_id = 2;
 
-  // Observed number of worker replicas available in the cluster
+  // Observed number of worker replicas available in the cluster.
+  // This value is dynamically updated based on the current state of the cluster.
+  // It reflects the actual number of active worker nodes currently running and available to handle tasks.
   int32 available_worker_replicas = 3;
 
-  // Desired number of worker replicas across all groups
+  // Desired number of worker replicas across all groups.
+  // This value is derived from the desired specifications set in WorkerGroupSpec in raycluster_types.go.
+  // It represents the target number of workers defined during cluster creation or scaling operations.
   int32 desired_worker_replicas = 4;
 
-  // Minimum number of worker replicas across all groups
+  // Minimum number of worker replicas across all groups.
+  // It aligns with the minimum configuration provided in the cluster spec.
+  // This value ensures the cluster maintains a baseline number of workers to meet operational requirements.
   int32 min_worker_replicas = 5;
 
-  // Maximum number of worker replicas across all groups
+  // Maximum number of worker replicas across all groups.
+  // It reflects the upper limit for scaling workers, often specified in autoscaling configurations.
+  // This value prevents over-provisioning and helps control resource costs.
   int32 max_worker_replicas = 6;
 
-  // Endpoints to access the cluster services
+  // Endpoints to access the cluster services.
+  // These endpoints include the head service and potentially other ingress points for the cluster.
+  // They are used for communication with the head node and accessing cluster-level APIs.
   Endpoints endpoints = 7;
 
-  // Details about the Ray head node
-  Head head = 8;
-
-  // Last update time for the cluster status
+  // Last update time for the cluster status.
+  // This timestamp is managed by Kubernetes and helps track the most recent changes to the cluster.
+  // It provides a snapshot of when the cluster's status was last refreshed.
   k8s.io.apimachinery.pkg.apis.meta.v1.Time last_update_time = 9;
+
+  // Information on the head node.
+  // The head node is critical for coordinating worker nodes and running the Ray scheduler.
+  // This field includes details about the pod running the head node, such as its IP address and health status.
+  RayHeadNodeInfo head_node = 10;
+
+  // Cluster assignment details.
+  // This field is marked for deprecation and should be removed in the future.
+  // It currently provides metadata about the logical grouping of clusters, but its usage is being phased out.
+  // TODO: Mark this as reserved after removing the usage.
+  ClusterAssignment cluster_assignment = 11;
+
+  // Assignment information.
+  // Provides additional metadata about how the cluster is allocated for specific workloads.
+  // This field helps map resources to their intended tasks or tenants within a shared environment.
+  // For example, it can include priority levels, tenant-specific resource limits, or job-specific constraints.
+  // These details ensure efficient resource utilization and prevent conflicts in multi-tenant environments.
+  AssignmentInfo assignment = 12;
+
+  // Scheduling information.
+  // Contains data on how resources are allocated and scheduled within the cluster.
+  // This includes policies for distributing tasks and managing resource contention.
+  // Scheduling policies can be defined to prioritize critical jobs, balance workloads, or meet SLAs (Service Level Agreements).
+  // It also includes the scheduling status of each worker group, indicating readiness or pending states.
+  SchedulingInfo scheduling_info = 13;
+
+  // Conditions that describe the states.
+  // These conditions detail specific events or issues affecting the cluster.
+  // Examples include "PodsPending" or "AutoScalingInProgress" to indicate operational status.
+  repeated Condition status_conditions = 14;
+
+  // URL to view job-specific details.
+  // This field is used for debugging and monitoring purposes, offering a direct link to job metadata.
+  // It provides an entry point to view logs, metrics, and other details for a given Ray job.
+  string job_url = 15;
+
+  // Information on pod errors in the Ray cluster.
+  // Collected from pod events, this information is essential for diagnosing issues, though the presence of errors
+  // does not necessarily mean the cluster is unhealthy.
+  // Examples include issues like "ContainerImagePullBackOff" or "CrashLoopBackOff" that may require intervention.
+  // Each error entry provides detailed context, such as error type, associated pod name, and timestamps of occurrences.
+  // This information aids in root cause analysis and accelerates resolution of cluster failures.
+  repeated PodErrors pod_errors = 16;
 }
+
 
 // RayCluster refers to a Ray cluster.
 message RayCluster {
--- a/proto/api/v2/ray_cluster.proto
+++ b/proto/api/v2/ray_cluster.proto
@@ -172,70 +172,65 @@ message RayJobStatus {
   // It is updated based on the observed state transitions in the Ray cluster managed by the Kubernetes operator.
   RayClusterState state = 1;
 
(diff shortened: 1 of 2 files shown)


proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
-  ClusterAssignment cluster_assignment = 3;
+  string rayClusterStatus = 5;
 
-  // Assignment info
-  AssignmentInfo assignment = 4;
+  string dashboardURL = 6;
 
-  // Scheduling info
-  SchedulingInfo scheduling_info = 5;
+  string jobDeploymentStatus = 7;
```
sallycr: I see that mixed use of snake case vs carmel case. Is this mistakenly used? can we go with snake case as consistently?

proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
-  map<string, string> ray_conf = 5;
+  map<string, string> ray_conf = 2;
 
-  // Termination spec
-  TerminationSpec termination = 6;
+  string entrypoint = 3;
 
-  // Scheduling spec
-  SchedulingSpec scheduling = 7;
+  string jobId = 4;
```
sallycr: @badcount Do we have job id in job spec?

proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
-  // Ray worker specification
-  WorkerSpec worker = 4;
-
   // User-specified Ray configurations
-  map<string, string> ray_conf = 5;
+  map<string, string> ray_conf = 2;
 
-  // Termination spec
-  TerminationSpec termination = 6;
+  string entrypoint = 3;
```
mincai: Detailed comments ?

proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
 }
 
-// Information on the head node
-message RayHeadNodeInfo  {
-  // Name of the pod
-  string name = 1;
+// Describes the state of a dashboard.
+enum RayJobState {
+  // The dashboard state is invalid.
+  RAY_JOB_V2_STATE_INVALID = 0;
```
mincai: Why do we need `RAY_JOB_V2_XXX` instead of just `RAY_JOB_XXX` ?

proto/api/v2/ray_job.proto
```diff
@@ -12,100 +12,59 @@ import "michelangelo/api/v2/job.proto";
...
+  RAY_JOB_V2_STATE_FAILED = 4;
+
+  RAY_JOB_V2_STATE_KILLED = 5;
 }
 
 // Defines the status of a Ray job
-message RayJobStatus {
-  // Information on the head node
-  RayHeadNodeInfo head_node = 1;
+message RayJobV2Status {
```
mincai: Please add detailed comments for each fields in the proto definition.

mincai (changes requested):

proto/api/v2/ray_job.proto
badcount: Added

proto/api/v2/ray_job.proto
badcount: Good catch, updated

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,187 @@
...
+
+  // Maximum number of worker instances
+  int32 max_instances = 3;
+
+  // Node type for a worker in a heterogeneous cluster
+  string node_type = 4;
+
+  // Ray object store memory will be allocated based on this
+  // ratio of the overall worker memory
+  double object_store_memory_ratio = 5;
```
mincai: this is for the entrypoint param

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,187 @@
...
+  // The cluster creation or operation failed.
+  RAY_CLUSTER_STATE_FAILED = 6;
+
+  // The cluster state is unknown.
+  RAY_CLUSTER_STATE_UNKNOWN = 7;
+}
+
+// Specification for the endpoints
+message Endpoints {
+  // Endpoint for client communication with the Ray cluster
```
mincai: can you provide an example in the comment ? Is endpoint a name or URL ?

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,187 @@
...
+  RayHeadSpec head = 3;
+
+  // Specifications for the worker groups in the cluster
+  repeated RayWorkerSpec workers = 4;
+
+  // Additional Ray cluster-wide configurations
+  map<string, string> ray_conf = 5;
+
+  // Enable or disable in-tree autoscaling
+  bool enable_in_tree_autoscaling = 6;
```
mincai: Do we need this ? What's the corresponding field in KubeRay CRD ?

proto/api/v2/ray_job.proto
```diff
@@ -12,123 +12,109 @@ import "michelangelo/api/v2/job.proto";
...
+  string ray_cluster_status = 5;
 
-  // Assignment info
-  AssignmentInfo assignment = 4;
+  // URL to access the Ray job's dashboard
+  string dashboard_url = 6;
 
-  // Scheduling info
-  SchedulingInfo scheduling_info = 5;
+  // Deployment status of the Ray job
```
mincai: Can we provide some examples of possible status ?

proto/api/v2/ray_cluster.proto
badcount: Removed

proto/api/v2/ray_job.proto
badcount: Updated

proto/api/v2/ray_cluster.proto
badcount: updated

proto/api/v2/ray_cluster.proto
badcount: Provided examples for each Endponits

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "michelangelo/api/options.proto";
+import "michelangelo/api/v2/pod.proto";
+import "michelangelo/api/v2/job.proto";
+import "michelangelo/api/v2/user.proto";
+
+// Specification for Ray cluster head node
+message RayHeadSpec {
+  // Kubernetes service type for the head service
```
bhjoshi1, line 17: I don't think there's any plan for us to use this because of our networking stack. If we cannot use or test it, I'd vote to remove it.

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+  string service_type = 1;
+
+  // Pod spec for the Ray head
+  PodSpec pod = 2;
```
bhjoshi1, line 21: do we plan to change these to v1 native types?

proto/api/v2/ray_cluster.proto
bhjoshi1: hi, @badcount is the update made already? I don't see any change.

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+// Describes the state of a ray cluster.
+enum RayClusterState {
+  // The cluster state is invalid.
+  RAY_CLUSTER_STATE_INVALID = 0;
+
+  // The cluster is in the process of being created.
+  RAY_CLUSTER_STATE_CREATING = 1;
+
+  // The cluster is being provisioned.
+  RAY_CLUSTER_STATE_PROVISIONING = 2;
```
bhjoshi1, line 79: what's the difference b/w these two?

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+  // Port for clients to connect.
+  // This port is used by external clients to connect to the Ray cluster's head node.
+  // It serves as the primary entry point for submitting jobs and managing cluster state.
+  int32 client_port = 4;
+
+  // Dashboard port for the Ray cluster.
+  // Provides a web-based interface to monitor and manage the Ray cluster.
+  // The dashboard displays metrics, logs, and node details for operational insights.
+  int32 dashboard_port = 5;
+
```
bhjoshi1: There's an overlap b/w this message and the Endpoints message.

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+  // This ID is critical for associating jobs with specific Ray clusters in Kubernetes.
+  // It ensures the operator can track and manage individual clusters efficiently.
+  string cluster_id = 2;
+
+  // Observed number of worker replicas available in the cluster.
+  // This value is dynamically updated based on the current state of the cluster.
+  // It reflects the actual number of active worker nodes currently running and available to handle tasks.
+  int32 available_worker_replicas = 3;
+
+  // Desired number of worker replicas across all groups.
```
bhjoshi1: why are these part of the status? Shouldn't these instead be in the spec?

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,337 @@
...
+
+  // Endpoints to access the cluster services.
+  // These endpoints include the head service and potentially other ingress points for the cluster.
+  // They are used for communication with the head node and accessing cluster-level APIs.
+  Endpoints endpoints = 7;
+
+  // Last update time for the cluster status.
+  // This timestamp is managed by Kubernetes and helps track the most recent changes to the cluster.
+  // It provides a snapshot of when the cluster's status was last refreshed.
+  k8s.io.apimachinery.pkg.apis.meta.v1.Time last_update_time = 9;
```
bhjoshi1: are we sure we can support this? Do we want to be consistent with other CRDs where we're not using this because of serialization issues?

bhjoshi1: @ying-zheng FYI ^

proto/api/v2/ray_job.proto
```diff
@@ -12,123 +12,127 @@ import "michelangelo/api/v2/job.proto";
...
-  // Scheduling spec
-  SchedulingSpec scheduling = 7;
-
-  // Ray worker specification. This field is introduced to
-  // support clusters that have a heterogeneous set of workers.
-  // Eventually we should deprecate the worker field above because
-  // it can be represent by this field.
-  repeated WorkerSpec workers = 8;
+  // Key-value pairs specifying user-defined Ray configurations
+  map<string, string> ray_conf = 2;
```
bhjoshi1: should this be present in both cluster and job CRDs?

proto/api/v2/ray_cluster.proto
badcount: I don't think this can be updated automatically, you can check the file here: https://github.com/michelangelo-ai/michelangelo/pull/21/files

proto/api/v2/ray_cluster.proto
badcount: Yes, we can use it as head node spec, that's very similar to kuberay head spec anyway: https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L42

proto/api/v2/ray_cluster.proto
badcount: Yes, we do, this is just for now to unblock. Once Ying get the proto ready, we will use native v1 proto

proto/api/v2/ray_cluster.proto
badcount: creating is to create the cluster, and then we go to provisioning state for the pod to get ready, this need to be a two seprate state so we avoid to create the cluster again.

proto/api/v2/ray_cluster.proto
badcount: Kuberay returns the available worker in the status.
https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go#L169

proto/api/v2/ray_cluster.proto
badcount: Agree, removed and use HeadnoteInfo instead

proto/api/v2/ray_cluster.proto
badcount: This returned from KubeRay, i added this for updated time just nice to have it.

proto/api/v2/ray_job.proto
badcount: Good call, removed it.

proto/api/v2/ray_cluster.proto
bhjoshi1: What is creating conveying? Creating the CRD?

badcount: Yes, it creates the ray.io/RayCluster CRD

proto/api/v2/ray_cluster.proto
bhjoshi1: How will we test any code based on this?

proto/api/v2/ray_cluster.proto
bhjoshi1: I see the update to the comment. What I mean to say is that this should not be present here at all since it is part of the entrypoint.

proto/api/v2/ray_cluster.proto
bhjoshi1: Then should it be called LAUNCHED instead? CREATING is a one shot operation. It's not a long running operation.

proto/api/v2/ray_cluster.proto
bhjoshi1: I don't think having the same field on both spec and status is correct. Our operator does not need to set it.

proto/api/v2/ray_cluster.proto
```diff
@@ -0,0 +1,283 @@
...
+
+// Defines the status of a Ray cluster
+message RayJobStatus {
+
+  // Current state of the Ray cluster.
+  // This field reflects the operational state of the cluster, such as "Running", "Pending", or "Failed".
+  // It is updated based on the observed state transitions in the Ray cluster managed by the Kubernetes operator.
+  RayClusterState state = 1;
+
+  // Unique identifier for the generated Ray cluster in the Ray operator.
```
bhjoshi1: why do we need this? namespace + name should be the unique identifier right?

proto/api/v2/ray_cluster.proto
bhjoshi1: I understand. But we cannot serialize it in our protobuf.

proto/api/v2/ray_cluster.proto
badcount: this is not final version, we can continue iterate this on the apiserver and controllerMgr.

proto/api/v2/ray_cluster.proto
badcount: Got it, moved it to the RayJob

proto/api/v2/ray_cluster.proto
badcount: Creating is just shot one indeed, the long one is "provisoning". I updated to launched.

proto/api/v2/ray_cluster.proto
badcount: This is kuberay standard, any reason why we don't need this?

proto/api/v2/ray_cluster.proto
badcount: removed.

proto/api/v2/ray_cluster.proto
bhjoshi1: Because these are part of the spec. Repeating them in the status looks incorrect to me. As part of the input, we specify a min and a max number of replicas. The CRD can tell us the current number of replicas in the status. But why does it need to repeat desired, min and max in the status?

proto/api/v2/ray_job.proto
```diff
@@ -12,123 +12,129 @@ import "michelangelo/api/v2/job.proto";
...
+  // - RUNNING: Indicates the job is currently in progress and tasks are being executed.
+  // This field provides a textual representation of the current job state for debugging
+  // and monitoring purposes.
+  string job_status = 3;
+
+  // Additional message describing the current state or issues
+  string message = 4;
+
+  // Status of the associated Ray cluster
+  string ray_cluster_status = 5;
```
bhjoshi1: why does the job have the status of the cluster? We should refer to the cluster and get the status from the cluster CRD. Otherwise, there will be an issues of keeping them in sync.

proto/api/v2/ray_job.proto
```diff
@@ -12,123 +12,129 @@ import "michelangelo/api/v2/job.proto";
...
+  //   - "Pending": The job is created but not yet started, waiting for resources or prerequisites.
+  //   - "Running": The job is actively executing tasks.
+  //   - "Succeeded": The job has completed successfully without errors.
+  //   - "Failed": The job terminated prematurely due to an error.
+  //   - "Terminating": The job is in the process of being shut down.
+  //   - "Cancelled": The job was manually stopped before completion.
+  //   - "Scaling": Resources are being adjusted to meet the job's demands.
+  //   - "Initializing": The job is setting up dependencies or preparing resources.
+  //   - "ResourcePending": The job is waiting for sufficient resources to execute.
+  //   - "Error: <specific_error>": A custom error message, such as "Error: Insufficient Memory".
```
bhjoshi1, line 103: should this be an enum instead?

proto/api/v2/ray_job.proto
```diff
@@ -12,123 +12,129 @@ import "michelangelo/api/v2/job.proto";
...
+  // and monitoring purposes.
+  string job_status = 3;
+
+  // Additional message describing the current state or issues
+  string message = 4;
+
+  // Status of the associated Ray cluster
+  string ray_cluster_status = 5;
+
+  // URL to access the Ray job's dashboard
```
bhjoshi1, line 89: This will be a property of the cluster. It doesn't depend on the job.

badcount: I just checked, somehow cluster endpoint return the port only, but the job status return full url like:
status:
  dashboardURL: rc-uf-ray-h24lq-lkgzw-head-svc.default.svc.cluster.local:8265
  endTime: "2024-12-21T01:29:00Z"
  failed: 0
  jobDeploymentStatus: Complete
  jobId: ray-uf-rj-uf-ray-h24lq-7tqsm-sxwlx
  jobStatus: SUCCEEDED
  message: Job finished successfully.
  rayClusterName: rc-uf-ray-h24lq-lkgzw
  rayClusterStatus:
    availableWorkerReplicas: 1
    desiredCPU: "2"
    desiredGPU: "0"
    desiredMemory: 2Gi
    desiredTPU: "0"
    desiredWorkerReplicas: 1
    endpoints:
      client: "10001"
      dashboard: "8265"
      metrics: "8080"
      redis: "6379"
      serve: "8000"
    head:
      podIP: 10.244.0.39
      podName: rc-uf-ray-h24lq-lkgzw-head-qqdkn
      serviceIP: 10.244.0.39
      serviceName: rc-uf-ray-h24lq-lkgzw-head-svc
    lastUpdateTime: "2024-12-21T01:28:31Z"
    maxWorkerReplicas: 1
    minWorkerReplicas: 1
    observedGeneration: 1
    readyWorkerReplicas: 1
    state: ready
    stateTransitionTimes:
      ready: "2024-12-21T01:28:31Z"
  startTime: "2024-12-21T01:28:51Z"
  succeeded: 1

proto/api/v2/ray_job.proto
badcount: I think it's better to keep it as string before once we change version, we don't have to update this status

proto/api/v2/ray_cluster.proto
badcount: I don't know, i guess there could be reason in KubeRay side? But I think this is not a blocker, remove this for now

proto/api/v2/ray_job.proto
badcount: removed it for now, will add it later if we find out we need it. Again, this is all from existing rayJob status

sallycr (approved):
Min Cai
I am OOTO today. You can ask him to merge first
Min Cai
I will do a post commit review later

PR #25, opened 2024-12-19
Initial Python code structure using Poetry

andriika:
Initial Python code structure.

## Proposal for Poetry

This PR advocates for adopting Poetry over Bazel for Python packaging and dependency management. Recent trends and developer feedback show that Poetry is more favored within the Python ecosystem for modern projects.

## A note on gRPC

However, Bazel should still be used to generate gRPC proto stubs and store the output in the `python/michelangelo/proto` directory. We can then add these stubs from (`python/michelangelo/**/*_pb2.py`) to `.gitignore` and configure the packaging script to run Bazel proto generation before running `poetry build`.


python/README.md
```diff
@@ -0,0 +1,13 @@
+## Development
+
+Preprequisites:
+
+- Poetry: https://python-poetry.org
+- Python 3.9
```
badcount, line 15: Are we still using 3.9? shouldn't be 3.10 more popular?

andriika: Think of it as the minimal Python version we will support, assuming later Python versions are backward-compatible with 3.9.

python/poetry.lock
sallycr, line 1: @andriikali delete this file?

ying-zheng:
why gRPC proto stubs still need to use bazel?

python/README.md
mincai: We should start with the latest python version so that we can use latest pytorch etc.

python/poetry.lock
andriika: Usually, it's a good practice to keep a lock file checked in.

python/examples/bert_cola/__main__.py
```diff
@@ -0,0 +1,12 @@
+import michelangelo as ma
+
+
+@ma.task(config="test")
```
mincai: shouldn't this be `@uniflow.task` ?

mincai (approved):

python/README.md
andriika: The latest torch (2.5.1) works for Python 3.9, according to their official guide: https://pytorch.org/get-started/locally/
> It is recommended that you use Python 3.9 - 3.12.

We should agree on the minimum supported Python version and dev/test things against it first. 3.9 is a good minimum version to stick to.

python/examples/bert_cola/__main__.py
sallycr: @mincai I think we would go with ma? Do you think we show the product as uniflow or ma?

python/poetry.lock
sallycr: ack. thanks.

sishilong (approved):

andriika:
> why gRPC proto stubs still need to use bazel?

@ying-zheng, as you suggested in our recent conversation, I will look into https://github.com/fedej/poetry-grpc-plugin to see if it can be used as an alternative to Bazel.

python/examples/bert_cola/__main__.py
andriika: Changed to:
```
import michelangelo.uniflow as uniflow

@uniflow.task( ... )
```

PR #28, opened 2024-12-21
make gazelle ingore python directory

ying-zheng:
Make gazelle ingore python directory as we use Poetry to manage python dependencies & build

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,5 +1,6 @@
 # gazelle:exclude .git/
 # gazelle:exclude pkg/
+# gazelle:exclude python/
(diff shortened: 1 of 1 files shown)


andriika:
We should add the gazelle run in GitHub actions as well.

ying-zheng:
> We should add the gazelle run in GitHub actions as well.

Yes. We have to add gazelle check. But, currently the go code generated from CRD proto files can't compile (the k8s go library doesn't contain some functions that are expected by the standard proto compiler). So, we manually removed the go_library part from the BUILD.bazel file of proto/api/v2. I have to fix this issue before adding gazelle check

PR #30, opened 2024-12-23
kubeproto

ying-zheng:
Protobuf compiler that generates go code for Kubernetes CRDs defined in protobuf files

The go code generated by this compiler include:
1. go struct types (with protobuf / json serialization / deserialization functions) of CRD or CRD list types (if michelangelo.api.resource or michelangelo.api.resource_list option is set in the proto mesage)
2. go code to register k8s GroupVersion schemas and CRD schemas
3. CRD schema in yaml form

The go code generated by this compiler is customized to be compatible with k8s go libraries (e.g. protobuf message types do not have Equal() function, the Spec and Status fields of CRD types are values instead of pointers)


badcount (approved):

proto/test/kubeproto/testobject.proto
sallycr, line 1: duplicated with [proto/test/kubeproto/protocreq/testobject.proto](https://github.com/michelangelo-ai/michelangelo/pull/30/files#diff-575e51bddd573809b91fd45efa233f21a2e5552363794b6cb9ff5f917e7cd905)? @ying-zheng

ying-zheng: this file is used in two ways in the unit test cases:
1. In code gen tests, use the original testobject.proto an input of protobuf compiler
2. In other tests, use the compiled go code as a go class

In the old repo, I manually configured 2 different bazel targets. But this is too fancy for gazelle to understand.
This time, I copied the file in 2 directories. The copy in kubeproto directory is for the compiled go library. The copy in the kubeproto/protocreq directory is used as an input proto file

The 2 copies are used independently. The content of the two copies does need to be the same

PR #32, opened 2024-12-23
Add Cache for CI/CD github action

sishilong:
Add Cache for CI/CD github action. I saw the repeated run after first run is 1 min https://github.com/michelangelo-ai/michelangelo/actions/runs/12471137926

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: Bazel Test
 # Trigger the workflow on push and pull request events for the main branch
 on:
   push:
-    branches: ['*']
+    branches: ['main']
   pull_request:
     branches: [main]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,9 +16,21 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
+      # Cache Bazel outputs
+      - name: Cache Bazel Outputs
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/bazel
+            bazel-out
+          key: ${{ runner.os }}-bazel-${{ hashFiles('WORKSPACE', '**/BUILD.bazel') }}
+          restore-keys: |
+            ${{ runner.os }}-bazel-
+
       # Run Bazel tests
       - name: Run Bazel Test
-        run: ./tools/bazel test ...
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -16,10 +16,22 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
+      # Cache Bazel outputs
+      - name: Cache Bazel Outputs
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/bazel
+            bazel-out
```
linzhp: bazel-out is actually a symlink pointing to `~/.cache/bazel/_bazel_<user>`. If we cache the whole `~/.cache/bazel`, it already incldues bazel-out

.github/workflows/main.yml
```diff
@@ -16,10 +16,22 @@ jobs:
...
+          key: ${{ runner.os }}-bazel-${{ hashFiles('WORKSPACE', '**/BUILD.bazel') }}
```
linzhp: Does this mean whenever `WORKSPACE` or any BUILD.bazel files change, the whole `~/.cache/bazel` is nuked?

If that's the case, we can put a static key like "bazel_cache_root" so it always has a warm cache.

.github/workflows/main.yml
```diff
@@ -16,10 +16,22 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v3
 
+      # Cache Bazel outputs
+      - name: Cache Bazel Outputs
+        uses: actions/cache@v3
```
linzhp: I saw the latest version is `actions/cache@v4`

.github/workflows/main.yml
```diff
@@ -16,10 +16,22 @@ jobs:
...
+      # Cache Bazel outputs
+      - name: Cache Bazel Outputs
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/bazel
+          key: bazel_cache_root
+          restore-keys: |
+            bazel_cache_root
+            ${{ runner.os }}-bazel-
```
linzhp: we probably don't need the second key?

linzhp (approved):

.github/workflows/main.yml
sishilong: I added the second key for fallback. but can remove it now. we can monitor how the ci/cd behavior

PR #36, opened 2024-12-23
ignore /pkg in .bazelignore

ying-zheng:
"go mod tidy" creates a directory pkg in the root directory. Without bazelignore, 'bazel ...' will try to build in this directory and fail


linzhp:
I think this is caused by this: https://github.com/michelangelo-ai/michelangelo/blob/6e94f17824938459bde416eadee500dc505387f4/.envrc#L12

We should probably remove that instead

ying-zheng:
> I think this is caused by this:
> 
> https://github.com/michelangelo-ai/michelangelo/blob/6e94f17824938459bde416eadee500dc505387f4/.envrc#L12
> 
> We should probably remove that instead

After removed the GOPATH environment variable, bazel still tries to build pkg and still fails

PR #41, opened 2024-12-24
protobuf validation compiler

ying-zheng:
Protobuf compiler that generates go validation code based on the validation rules defined in protobuf files

This protobuf compiler generates a function called "Validate" for each go struct generated by protobuf message. The validation logic is based on the validation rules (michelangelo.api.validation options) in the protobuf file


proto/test/kubeproto/validation_ut.proto
sallycr, line 1: any reason we keep duplicated codes/files?

PR #44, opened 2024-12-24
Kubeyarpc compiler

ying-zheng:
Protobuf compiler that generates YARPC service code for k8s CRDs defined in protobuf files

Go code generated by this compiler, include:

- A k8s client for each api group version
- A fx module for each YARPC service (each k8s CRD)
- Create / Get / Update / Delete YARPC functions
- API Hook, Audit log, Validation, etc.


go/kubeproto/templates/crd_svc.tmpl
```diff
@@ -0,0 +1,722 @@
...
+	defer c.auditLogEmitter.Emit(ctx, c.build{{.KindName}}AuditLogEventForDeleteCollection(
+												ctx,
+												request,
+												resp,
+												err,
+												maUUID,
+												),
+				)
+
+	// Authentication
```
sallycr, line 478: We need to implement this. did we capture this in any future milestone? @ying-zheng

sallycr: Ying, did we capture this in the future one? or create an issue to get an attention after?

sallycr (approved):

PR #46, opened 2024-12-28
Initial Python-based Sandbox code

### LINKED ISSUE #24: As a users, use sandbox cluster outside of this monorepo
> How do users use it outside of this monorepo?

That should be the nexts step. And outside of this monorepo should use APIs. For SDK path, we need to design.

_Originally posted by @sallycr in https://github.com/michelangelo-ai/michelangelo/issues/19#issuecomment-2552577464_

andriika:
Initial code for the Python-based sandbox CLI. Unlike `tools/sandbox.sh`, this sandbox code is written in Python and will be included in the Python package, enabling users to pip-install and use it outside of this repository.

### Choosing k3d for the local Kubernetes cluster

The sandbox runs a lightweight Kubernetes instance, installing all Michelangelo components along with their dependencies on top of it. Various products are available for running lightweight Kubernetes for local development and testing. We considered the following options: minikube, kind, k3d, Docker Desktop, and microk8s. After an assessment, we chose k3d over the others due to these features:

- Fastest startup: Quick startup time of less than 5 seconds on most machines, which is crucial for rapid development cycles.
- Minimal resource usage: Being based on K3s, k3d inherits its lightweight nature, making it efficient for local development.

Some more info: https://www.perplexity.ai/search/we-aim-to-run-a-lightweight-ku-5WkHrQSCTlOfC74IIL0fkA

### Next Steps

- Currently, the sandbox implementation installs MySQL and Cadence. Additional Michelangelo components will be installed as needed and when ready.
- Refactor `tools/sandbox.sh` to utilize this Python-based sandbox instead of the current bash script.


python/michelangelo/cli/sandbox/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## User Guide
+
+### Prerequisites
+
+Please install the following software before using the sandbox:
+
+- Docker: https://docs.docker.com/get-started/get-docker/
+- kubectl: https://kubernetes.io/docs/tasks/tools/#kubectl
+- k3d: https://k3d.io
```
ying-zheng: There are many different tools to run K8s locally. I remeber we used Kind before. Could you document the reason why we choose k3d rather than Kind / minikube / MicroK8s?

Can we let the users to choose how to run k8s? I think as long as there is a k8s environment (either local or a real k8s cluster), we should be able to run our sandbox in it

python/.gitignore
```diff
@@ -1,3 +1,5 @@
 .venv
 __pycache__
 dist
+.idea
```
ying-zheng: we already added .idea to the .gitignore file in the root directory. I think it's better to ignore **/.idea in the root .gitignore file, instead of ignore .idea in each sub directory

python/michelangelo/cli/cli.py
```diff
@@ -1,13 +1,31 @@
 import sys
 import argparse
+import logging
+from .sandbox import sandbox
```
hwaminkim: Can we use absolute path import instead of using relative path? (e.g. `.sandbox` --> `michelangelo.cli.sandbox`)
Relative imports sometimes hard to read and make the import architecture messy


> Remember that you should generally opt for absolute imports over relative ones, unless the path is complex and would make the statement too long.
> https://realpython.com/absolute-vs-relative-python-imports/

python/.gitignore
hwaminkim: @ying-zheng It seems like each IDE will set the rood directory as each language's root directory. Is it right? Otherwise, is there any possibility to use the repository root directory as IDE's root?

python/michelangelo/cli/sandbox/README.md
andriika: > Could you document the reason why we choose k3d rather than Kind / minikube / MicroK8s?

Added a section in the PR description - "Choosing k3d for the local Kubernetes cluster"

> Can we let the users to choose how to run k8s?

It is better to manage kube clusters for users, as it involves not only creating and deleting clusters but also pre-configuring ports to expose various Michelangelo components, such as the API, Studio UI, Cadence API, Cadence UI, etc.

python/.gitignore
ying-zheng: I think this is more or less a personal preference. When using GoLand, I currently set the project root directory to the git repo root directory (rather than the go directory), because GoLand also understands protobuf and bazel.

For PyCharm, it might be better to set the project root to the python directory, if you don't plan to edit protobuf files in PyCharm

For VS code, as it doesn't distinguish different languages, I think it's straightforward to open the git repo root directory in VS code

python/michelangelo/cli/config/config.py
```diff
@@ -0,0 +1,41 @@
+import sys
+import argparse
+
+description = "Config CLI"
```
ying-zheng: What's this command for? What exactly does this command do? It seems this command is not fully implemented yet?

python/.gitignore
andriika: Removed.

python/michelangelo/cli/cli.py
andriika: Changed to absolute import path.

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -0,0 +1,151 @@
+import sys
+import argparse
+import subprocess
+import time
+import uuid
+from pathlib import Path
+
+description = "Sandbox CLI"
```
ying-zheng: We need a better description and user guide for the command, and also help strings for each argument

python/michelangelo/cli/config/config.py
andriika: Yes, it was a stub. I have removed it from this PR.

python/michelangelo/cli/sandbox/sandbox.py
andriika: Agree. Please see an updated description and help strings for each argument.

ying-zheng (approved):

python/michelangelo/cli/sandbox/README.md
```diff
@@ -0,0 +1,22 @@
+Sandbox is a lightweight Michelangelo cluster for development and testing.
+It's also a great way for users to quickly try the platform and check out its interface.
+
+> Note: The sandbox deployment is not suitable for production environments. 
+> For instructions on how to create a production-ready Michelangelo deployment, checkout the Deployment Guide.
+
+## User Guide
```
sallycr, line 7: can we link to the wiki? and manage this user guide in wiki? @andriikali

sallycr (approved):
Made user messages

python/michelangelo/cli/sandbox/README.md
andriika: I'll link once the wiki is ready.

PR #47, opened 2024-12-31
Refactor and cleanup logging

ying-zheng:
1. Remove MA UUID and UnifiedLog
MA UUID is a uuid used for tracing requests across multiple services and stages
UnifiedLog is a log tag to mark log messages that are useful for Michelangelo users and log messages that are only for Michelangelo developers.
Both were not fully implemented. Based on our experience in the last 2 years, these features are not very useful and hard to maintain.

2. Remove the tally and log tags for user email, caller, source
This information is added to yarpc context by Uber's internal libraries. We can't get this info in open source code. We will retrieve this information from context in Uber's internal implementation of audit logger and metrics emit functions

3. Rename tally sub scope name from 'uapi_yarpc' to 'ma_api'
In OSS Michelangelo we don't use the term "unified API", as there is only one Michelanelo API

4. Rename tally tag name from "project" to "namespace"
Although, in most cases, k8s namespaces corresponds to michelangelo projects. There are a few cases the namespaces are not projects. Calling this tag "namespace" is more accurate

5. Do not truncate JSON string in the generated yarpc service code
Different logging services have different maximum string length limits. We will do the truncation in the custom implementations of audit log and log sink, rater than setting a global limit.

6. Remove the fields of AuditLogEvent that are not set / used in OSS michelangelo

7. Remove "Unified Log"
We used to use 2 loggers: zap sugar logger and logr logger. "Unified Log" was added as an abstract layer on top of these 2 loggers.  Now, we should only use logr. "Unified Log" is no longer needed

7. added go/vendor in .bazelignore
This directory is generated by 'go mod vendor' command

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,1 +1,2 @@
 pkg
+go/vendor
--- a/go/logging/audit_log_struct.go
+++ b/go/logging/audit_log_struct.go
@@ -5,16 +5,8 @@ type AuditLogEvent struct {
 	EntityType string
 	EntityName string
 	Procedure  string
-	UserEmail  string
-	Source     string
-	Headers    string
 	Request    string
 	Response   string
 	Error      string
-	MAUUID     string
-	RuntimeEnv string
-	CRDDiff    string
 	PreCRD     string
-	TimeStamp  int64
-	Dimension  string
 }
--- a/go/logging/constants.go
+++ b/go/logging/constants.go
@@ -1,0 +1,13 @@
+package logging
+
+// MichelangeloAPIScopeName is the metrics sub-scope name for all Michelangelo API calls
+const MichelangeloAPIScopeName = "ma_api"
+
+// APIProcedureTag is the metrics tag key for yarpc procedure name
+const APIProcedureTag = "api_procedure"
+
+// NamespaceTag is the metrics tag key for Michelangelo project name
+const NamespaceTag = "namespace"
+
+// EntityNameTag is the Tally metrics tag key for entity name
+const EntityNameTag = "name"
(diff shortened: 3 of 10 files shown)


.bazelignore
```diff
@@ -1 +1,2 @@
 pkg
+go/vendor
```
linzhp, line 2: how is this populated?

go/logging/audit_log_struct.go
```diff
@@ -5,16 +5,8 @@ type AuditLogEvent struct {
...
-	Source     string
-	Headers    string
 	Request    string
 	Response   string
 	Error      string
-	MAUUID     string
-	RuntimeEnv string
-	CRDDiff    string
 	PreCRD     string
-	TimeStamp  int64
```
sallycr, line 18: Isn't this timestamp useful?

go/logging/constants.go
```diff
@@ -0,0 +1,13 @@
+package logging
+
+// MichelangeloAPIScopeName is the metrics sub-scope name for all Michelangelo API calls
+const MichelangeloAPIScopeName = "ma_api"
```
sallycr: can we just use 'api'?

go/logging/audit_log_struct.go
sallycr: hm. this was not used before.

ying-zheng: Some of these fields were from heatpipe schema. We don't set these fields in our code

sallycr: got it.

sallycr (approved):
Thanks for refactoring for logging.

sallycr:
@ying-zheng thanks for the description of the pull reqs. It helps to review the diff.

PR #49, opened 2025-01-02
Add controller manager

badcount:
Add Controller Manager with configurations, Bazel rules, runtime dependencies, and documentation.


go/cmd/controllermgr/main.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/go-logr/logr"
+
+	"github.com/michelangelo-ai/michelangelo/go/controllermgr"
+	"go.uber.org/fx"
+	"k8s.io/apimachinery/pkg/runtime"
+	kubescheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// scheme provides a Kubernetes runtime.Scheme object.
```
sallycr, line 14: @badcount added function doc. please review it.

badcount: Thanks Sally!

proto/api/v2/empty.go
```diff
@@ -0,0 +1 @@
+package kubeproto
```
sallycr, line 1: @badcount @ying-zheng Why do we need this empty go file?

sallycr (approved):

proto/api/v2/empty.go
ying-zheng: this empty go file makes 'go mod tidy' treat this directory as a go package, otherwise 'go mod tidy' will fail

https://stackoverflow.com/questions/75442156/go-mod-tidy-complains-that-the-bazel-generated-protobuf-package-is-missing
https://www.php.cn/faq/672765.html

I will add this to wiki page

PR #50, opened 2025-01-02
Add `.gitignore` file with more general coverage

hwaminkim:
Add `.gitignore` file with more general coverage. The predefined component is originated from [https://gitignore.io/]()

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -1,3 +1,179 @@
+# Created by https://www.toptal.com/developers/gitignore/api/python
+# Edit at https://www.toptal.com/developers/gitignore?templates=python
+
+### Python ###
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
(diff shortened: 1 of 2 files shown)


python/.gitignore
```diff
@@ -1,3 +1,179 @@
...
+#pdm.lock
+#   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
+#   in version control.
+#   https://pdm.fming.dev/#use-with-ide
+.pdm.toml
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
+__pypackages__/
+
+# Celery stuff
```
sallycr, line 119: why celerybeat here? @hwaminkim

hwaminkim: [https://gitignore.io/]() has commonly used library sets under python category set, so I guess they decided to add it. It helps to add popular or potential usage in advance, so prevent to add multiple times.
I'm good to remove this part, but please mind that we would need manual change in the future when we add more category with this generator.

sallycr: I see. Thanks for adding this.

sallycr: @hwaminkim can you add this information in the header of this file?

sallycr: I added it in the header. thx.

sallycr (approved):

PR #52, opened 2025-01-05
Add logging, config and env to controller manager

badcount:
Configuration Management Enhancements:

1. Introduced YAML-based configuration management with environment-specific overrides and secret handling (config.go, yaml.go).
2. Implemented a flexible logging module supporting environment-based configurations (logging.go, config.go).
3. Application Modules Integration: Integrated environment, configuration, and logging modules into the main application (main.go, BUILD.bazel).

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -30,8 +30,12 @@ go_sdk.nogo(
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
 oci.pull(
     name = "distroless_base",
-    digest = "sha256:0530d193888bcd7bd0376c8b34178ea03ddb0b2b18caf265135b6d3a393c8d05",
+    digest = "sha256:ccaef5ee2f1850270d453fdf700a5392534f8d1a8ca2acda391fbb6a06b81c86",
     image = "gcr.io/distroless/base",
+    platforms = [
+            "linux/amd64",
+            "linux/arm64",
+        ],
 )
 use_repo(oci, "distroless_base")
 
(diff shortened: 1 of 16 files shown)


ying-zheng:
I think there are somethings about config and env can be improved:

- We should avoid predefine environment names in go code, such as production, staging, development. OSS users may use different environment names.
- It's better to put default configurations into yaml files instead of go code (i.e. logging. defaultProdConfig()). Instead of hard code some environment names and hard code the corresponding configurations in go, we can provide a set of example configuration yaml files (e.g. base.yaml, production.yaml, development.yaml etc.), and allow users manage yaml configuration files by themselves.
- Uber internal MA code uses cfgparse, which seems not compatible with go.uber.org/config. We have to make a decision if we want to use go.uber.org/config and rewrite the existing config structs, or we will open source cfgparse

Since these are existing issues from old github MA code and we need some discussions to decide the solution, I will approve this pull request for now to unblock you. We can refactor the code later

MODULE.bazel
```diff
@@ -27,6 +27,14 @@ go_sdk.nogo(
 ###########################################
 ############# Third-party DEPENDENCIES ####
 ###########################################
+oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+oci.pull(
+    name = "distroless_base",
+    digest = "sha256:0530d193888bcd7bd0376c8b34178ea03ddb0b2b18caf265135b6d3a393c8d05",
```
ying-zheng: This is a very old version
You can find out the digest of the latest version following the instructions in https://github.com/bazel-contrib/rules_oci/releases
"# 'latest' is not reproducible, but it's convenient.
    # During the build we print a WARNING message that includes recommended 'digest' and 'platforms'
    # values which you can use here in place of 'tag' to pin for reproducibility."

badcount: Thanks for catching this, updated

ying-zheng (approved):

PR #55, opened 2025-01-09
Add ray-cluster controllers

badcount:
This PR is still WIP
1. to add tests
2. tests with ray.io


MODULE.bazel
ying-zheng, line 1: I already made this change in https://github.com/michelangelo-ai/michelangelo/pull/54

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	_requestTimeout      = 60
+	_requeueAfterSeconds = 10
+)
+
+// Reconciler reconciles a Ray CRD object
```
ying-zheng: I think it's better to call it RayCluster object

I want to avoid using the word "CRD", because
1. In k8s documents, CR = Custom Resource, CRD = Custom Resource Definition = yaml schema. A "CRD object" is a registered yaml schema. In our case, the RayCluster Kind itself is a CRD object, each instance of RayCluster is a CR object. We wrongly used the term "CRD" in many places. I want to fix this when open source the code / doc.

2. We actually have a plan to convert Michelangelo API types from CRDs (stored as yaml string in ETCD) into Kubernetes native types (stored as protobuf binary in ETCD) for better performance. So, I prefer to use more generic words like API Kind (type) or API object, instead of CRD / CR Kind or CR object.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the RayCluster object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.2/pkg/reconcile
```
ying-zheng: This comment sounds like a generic instruction for k8s controller developers. We need a more specific comment for RayCluster controller. e.g. Reconcile is called by k8s controller runtime in reconciliation loop to move the current state of RayCluster to the desired state. ...

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctx, cancel := context.WithTimeout(ctx, _requestTimeout*time.Second)
+	logger := log.FromContext(ctx)
+	logger.Info("reconcile")
+	defer cancel()
+
+	logger.Info("Reconciling ray cluster v2")
+
+	// retrieve the dashboard
```
ying-zheng: retrieve "dashboard"?

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+// the RayCluster object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.2/pkg/reconcile
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctx, cancel := context.WithTimeout(ctx, _requestTimeout*time.Second)
+	logger := log.FromContext(ctx)
+	logger.Info("reconcile")
```
ying-zheng: Duplicated with line 57?

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+	defer cancel()
+
+	logger.Info("Reconciling ray cluster v2")
+
+	// retrieve the dashboard
+	var rayCluster v2pb.RayCluster
+	if err := r.Get(ctx, req.NamespacedName, &rayCluster); err != nil {
+		// TODO when the ray cluster is not found, means it has been deleted
+		// we also need to delete the evaluation report as well
```
ying-zheng: evaluation report ...

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+// reconcile launches the cluster if it is not already launched
+func (r *Reconciler) reconcile(
+	ctx context.Context,
+	log logr.Logger,
+	rayCluster *v2pb.RayCluster,
+) (ctrl.Result, error) {
+
+	// Here we check the state, take an action and then transition to next state.
+	// Only terminal state SUCCEED, FAILED and KILLED remove request from queue by returning empty ctrl.Result{}
+	// For non-terminal state, we should always requeue because we don't have a watch on the dashboard state.
```
ying-zheng: dashboard

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+			rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_FAILED
+			break StateMachine
+		}
+		// we check it back to provioning for checking the status
+		rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_PROVISIONING
+	case v2pb.RAY_CLUSTER_STATE_READY:
+		log.Info("cluster is ready, we do nothing but continue requeue until the job finishes and received termination signal")
+		if rayCluster.Spec.Termination.Type != v2pb.TERMINATION_TYPE_INVALID {
+			rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_TERMINATING
+			return ctrl.Result{}, nil
```
ying-zheng: There has to be a delay before the next reconcile
ctrl.Result{RequeueAfter: time.Second * _requeueAfterSeconds}?
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+		}
+	default:
+		if rayCluster.Spec.Termination.Type != v2pb.TERMINATION_TYPE_INVALID {
```
ying-zheng: What's rayCluster.Spec.Termination for? Is this a flag for users to terminate the RayCluster? If yes, should we check this before any other states?

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,427 @@
...
+) (ctrl.Result, error) {
+
+	// Here we check the state, take an action and then transition to next state.
+	// Only terminal state SUCCEED, FAILED and KILLED remove request from queue by returning empty ctrl.Result{}
+	// For non-terminal state, we should always requeue because we don't have a watch on the dashboard state.
+StateMachine:
+	switch rayCluster.Status.State {
+	case v2pb.RAY_CLUSTER_STATE_INVALID:
+		log.Info("RAY_CLUSTER_STATE_INVALID")
+		rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_LAUNCHED
```
ying-zheng: it seems the transition from RAY_CLUSTER_STATE_INVALID to RAY_CLUSTER_STATE_LAUNCHED is noop.
Can we remove "RAY_CLUSTER_STATE_LAUNCHED" state?

case RAY_CLUSTER_STATE_INVALID:
    r.createCluster(...)
    rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_PROVISIONING

go/controllers/raycluster/controller.go
badcount: Yes, this means user want to terminate it for a reason, we do check this for every state.

go/controllers/raycluster/controller.go
badcount: Launched is no noop, launched means create a resource in ray cluster, and provisionning is checking the status, they need to be separated to avoid duplicated effort.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,350 @@
...
+func (r *Reconciler) reconcile(
+	ctx context.Context,
+	log logr.Logger,
+	rayCluster *v2pb.RayCluster,
+) (ctrl.Result, error) {
+
+	// Here we check the state, take an action and then transition to next state.
+	// Only terminal state SUCCEED, FAILED and KILLED remove request from queue by returning empty ctrl.Result{}
+	// For non-terminal state, we should always requeue because we don't have a watch on the dashboard state.
+StateMachine:
```
andriika: We should avoid creating controller logic that's based on the status check. As suggested in the kubebuilder tutorial: https://book.kubebuilder.io/cronjob-tutorial/controller-implementation 

> Remember, status should be able to be reconstituted from the state of the world, so it’s generally not a good idea to read from the status of the root object. Instead, you should reconstruct it every run.

badcount: Updated this to use spec logic instead of controller, plz review it again

go/controllers/raycluster/controller.go
badcount: Updated, and removed launched

go/controllers/raycluster/controller.go
badcount: Updated

go/controllers/raycluster/controller.go
badcount: Updated

go/controllers/raycluster/controller.go
badcount: Updated

go/controllers/raycluster/controller.go
badcount: Updated

go/controllers/raycluster/controller.go
badcount: Updated

go/controllers/raycluster/controller.go
badcount: Updated

go/cmd/controllermgr/config/base.yaml
```diff
@@ -6,8 +6,9 @@ controllermgr:
   port: 9443
 
 controllers:
-  pipelineRun:
-    cadenceTaskList: default
+  rayCluster:
+    k8s_qps: 300
+    k8s_burst: 600
```
andriika: - Let's avoid mixing snake_case with camelCase. It seems we use camelCase in other places, so better names are `k8sQPS`, `k8sBurst`, etc.
- The `k8s` prefix might be a bit superfluous. Simply using `qps` and `burst` clearly expresses the intention.

go/controllers/raycluster/module.go
```diff
@@ -0,0 +1,29 @@
...
+var (
+	// Module FX
+	Module = fx.Options(
+		fx.Provide(newConfig),
+		fx.Invoke(register),
+	)
+)
+
+func register(
+	conf Config,
```
andriika, line 20: This `config` is not used. I guess we need to read `burst`, `qps`, and other configuration properties here?

go/controllers/raycluster/module.go
```diff
@@ -0,0 +1,29 @@
...
+	mgr manager.Manager,
+) error {
+	rayClient, _ := rayv1.NewForConfig(mgr.GetConfig())
+	return (&Reconciler{
+		Client:      mgr.GetClient(),
+		env:         e.New().Environment,
```
andriika: We don't need to create a new environment instance here. It should be available in the FX registry.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	e "github.com/michelangelo-ai/michelangelo/go/base/env"
+	v2pb "github.com/michelangelo-ai/michelangelo/proto/api/v2"
+	rayv1 "github.com/ray-project/kuberay/ray-operator/pkg/client/clientset/versioned/typed/ray/v1"
+)
+
+const (
+	_requeueAfterSeconds = time.Second * 10
```
andriika: - It's better to remove the unit-specific suffix `Seconds` since this variable is a `time.Duration`, which provides a more general representation of time than, for instance, an int, typically used for representing seconds.
- Let's avoid using excessive leading "_" in var names. Just `requeueAfter = time.Second * 10`

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+// Reconciler reconciles a Ray Cluster object
+type Reconciler struct {
+	client.Client
+
+	env e.Context
+
+	rayV1Client *rayv1.RayV1Client
+}
+
+const _apiVersion = "ray.io/v1"
```
andriika: - Let's consolidate all constants in one location in the constant block above.
- Remove the excessive leading "_" in the constant name.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+	_requeueAfterSeconds = time.Second * 10
+)
+
+// Reconciler reconciles a Ray Cluster object
+type Reconciler struct {
+	client.Client
+
+	env e.Context
+
+	rayV1Client *rayv1.RayV1Client
```
andriika, line 33: nit: remove blank lines between struct fields.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return res, nil
+}
+
+func (r *Reconciler) createCluster(log logr.Logger, cluster *v2pb.RayCluster) error {
+	rayV1Cluster := &v1.RayCluster{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "RayCluster",
+			APIVersion: _apiVersion,
+		},
```
andriika: If I'm not mistaken, we don't need this `TypeMeta`. Things should work without it because type information is preserved by the `v1.RayCluster` struct itself, and the Kubernetes client knows how to derive it.

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+	rayV1Client *rayv1.RayV1Client
+}
+
+const _apiVersion = "ray.io/v1"
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info(fmt.Sprintf("Reconciling ray cluster %s", req.NamespacedName))
```
andriika: - Typically, with Logr, rather than using a format string, we use a fixed message and add extra information through key-value pairs.
- Some log messages start with a capital letter, while others do not. For consistency, we could stick to a lowercase convention.

E.g.:

```
logger.Info("reconciling ray cluster", "namespace", req.Namespace, "name", req.Name)
```

go/controllers/raycluster/controller.go
ying-zheng: go naming convention doesn't like _
this constant should be called apiVersion
https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+							convertPodSpecToContainer(cluster.Spec.Head.Pod),
+						},
+					},
+				},
+			},
+			RayVersion:       cluster.Spec.RayVersion,
+			WorkerGroupSpecs: convertWorkerGroupSpecsToWorkerSpec(cluster.Name, cluster.Spec.Workers),
+		},
+	}
+	createdRayCluster, err := r.rayV1Client.RayClusters(cluster.Namespace).Create(context.TODO(), rayV1Cluster, metav1.CreateOptions{})
```
ying-zheng: I guess you should use the context from Reconcile(), rather than context.TODO()

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+		envVars = append(envVars, newEnv)
+	}
+	return envVars
+}
+
+func convertPodSpecToContainer(pod *v2pb.PodSpec) corev1.Container {
+	return corev1.Container{
+		Name:  pod.Name,
+		Image: pod.Image,
+		//ImagePullPolicy: "Never",
```
ying-zheng: what's this comment for?

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+			},
+		}
+		workerGroupSpecsJson[i] = workerGroupMap
+	}
+	return workerGroupSpecsJson
+}
+
+// IsNotFoundError checks if the error is not found error
+func IsNotFoundError(err error) bool {
+	if strings.Contains(err.Error(), "not found") {
```
ying-zheng: checking the contain of an error message seems weird to me. What if the error message is "function not found"?
shoudn't it be enough to just check the error code?

go/controllers/raycluster/controller.go
```diff
@@ -0,0 +1,304 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info(fmt.Sprintf("Reconciling ray cluster %s", req.NamespacedName))
+
+	// retrieve the ray cluster
+	var rayCluster v2pb.RayCluster
+	if err := r.Get(ctx, req.NamespacedName, &rayCluster); err != nil {
+		// Resource not found (resource deleted)
+		return ctrl.Result{}, nil
+	}
```
andriika: - We shouldn't ignore `err` - it might be a resource not found error, but we don't know for sure. Let's categorize the error using https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound and handle it accordingly. 
- I think we should delete the underlying cluster on resource not found. WDYT?

E.g.:

```
if err := r.Get(ctx, req.NamespacedName, &rayCluster); err != nil {
    if apierrors.IsNotFound(err) {
        // delete the underlying kuberay resource
        return deleteCluster(....)
    }
    return ctrl.Result{}, err
}
```

badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/module.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: updated

go/controllers/raycluster/controller.go
badcount: removed

go/cmd/controllermgr/config/base.yaml
badcount: updated

go/controllers/raycluster/module.go
badcount: Yes, updated

ying-zheng (approved):

PR #62, opened 2025-01-14
Set Pypi Package name as `michelangelo-ai`

hwaminkim:
Set Pypi Package name as `michelangelo-ai`

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "pymichelangelo"
+name = "michelangelo-ai"
 version = "0.1.0"
 description = "Michelangelo is an end-to-end model lifecycle management system at large scale"
 authors = ["Michelangelo Team <michelangelo-oss-group@uber.com>"]


python/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "pymichelangelo"
+name = "michelangelo-ai"
```
sallycr, line 2: @hwaminkim do you need pull request for these? can we just package and upload there?

hwaminkim: @sallycr Since we decided to use the pypi package name as `michelangelo-ai`, I think we need to land it. Isn't it?

andriika: You can easily upload PyPI packages without landing any code.

hwaminkim: @andriikali that's true, but I'd like to explicitly set our pypi package name is `michelangelo-ai`. It would let us not change the package name every time we deploy it.

sallycr (approved):

PR #63, opened 2025-01-15
API handler

ying-zheng:
Commit API handler to OSS repo

The original code of api/handler is from apiserver/internal and apiserver/pkg/api of uber internal repo
The original code of storage is from go/storage of old OSS repo and storage of uber internal repo

I refactored and cleaned up the code, rewrote unit test. There are some incompatible changes that we have to careful when bring this code back to uber internal use:
1. removed client version check and flipr related code
2. renamed storage and storageClient in configuration files and in fx params to metadataStorage
3. moved blob storage to storage package
4. upgraded controller runtime to the latest version. The new controller runtime version has some backward incompatible API changes

The blob storage related part has not been fully implemented yet.

I committed go mock generated code to git repo. We don't need to do this in uber internal repo, as the Uber internal version of gazelle did some magic work. In the old OSS repo, we had to manually write gomock bazel targets and still doesn't work with go mod. I think it's more convenient to use //go:generate and check in the generated code. This is actually a common practice for OSS projects using gomock.


go/api/apiutil/apiutil.go
```diff
@@ -1,10 +1,91 @@
 package apiutil
```
badcount: I want to use this in controller as well, should be mait it just util instead of apiutil?

go/tools.go
```diff
@@ -6,11 +6,14 @@ import (
 	_ "github.com/gogo/status"        // required by @org_uber_go_yarpc//encoding/protobuf/protoc-gen-yarpc-go
 	_ "github.com/golang/mock/gomock" // for mockgen
```
linzhp: This line is not needed anymore, as there are plenty of imports to this package already

linzhp (approved):

go/tools.go
ying-zheng: done

go/api/apiutil/apiutil.go
ying-zheng: renamed to utils

badcount (approved):

PR #64, opened 2025-01-15
Add ray job controller

### LINKED ISSUE #57: Ray Job controller

badcount:
Fixes #57 
Add ray job controller


go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,177 @@
...
+		res.RequeueAfter = requeueAfter
+		return res, err
+	}
+
+	originalRayJob := rayJob.DeepCopy()
+
+	if rayJob.Spec.Cluster == nil {
+		rayJob.Status.State = v2pb.RAY_JOB_STATE_FAILED
+		rayJob.Status.Message = "cluster is not set"
+		return res, nil
```
ying-zheng: You have to call r.Status().Update() after setting Status fields. Otherwise the change is not applied to the object on k8s server.

After MA api server is ready, this logic should be moved to protobuf validation, so the job can't be created if cluster is not specified

go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	rayCluster := &v2pb.RayCluster{}
+
+	err := r.Get(ctx, types.NamespacedName{
+		Namespace: rayJob.Spec.Cluster.Namespace,
+		Name:      rayJob.Spec.Cluster.Name,
+	}, rayCluster)
+	if err != nil {
```
ying-zheng: There can be transient errors like the k8s api server is being redeployed.

When creating a job, we have to check the cluster is valid in api hook
When deleting a cluster, we have to use finalizer or ownership to make sure all the jobs are terminated before the cluster info is deleted in k8s. https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

Failed to get cluster should be retried.

You can add todo here for now.

go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,177 @@
...
+		logger.Error(err, "error to get cluster")
+		rayJob.Status.State = v2pb.RAY_JOB_STATE_FAILED
+		rayJob.Status.Message = "cluster is not found"
+		return res, err
+	}
+
+	status, jobFailedReason, err := r.getJobStatus(ctx, logger, &rayJob)
+	if err != nil {
```
ying-zheng: Similar to the get cluster case, the error can be transient.
We need a condition in RayJob status to indicate if the job is already created. I guess, after creating a job, it may take awhile for the job status to become available.

go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,177 @@
...
+			rayJob.Status.State = v2pb.RAY_JOB_STATE_RUNNING
+			res.RequeueAfter = requeueAfter
+		}
+	} else {
+		logger.Info("unknown status, re-queuing")
+		res.RequeueAfter = requeueAfter
+	}
+
+	if !reflect.DeepEqual(originalRayJob, rayJob) {
+		err = r.Status().Update(ctx, &rayJob)
```
ying-zheng: mark the job as immutable when it reaches a final state

go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,177 @@
...
+		Spec: v1.RayJobSpec{
+			ClusterSelector: map[string]string{
+				"ray.io/cluster":      cluster.Name,
+				"rayClusterNamespace": cluster.Namespace,
+			},
+			Entrypoint: job.Spec.Entrypoint,
+		},
+	}
+
+	createdRayJob, err := r.rayV1Client.RayJobs(cluster.Namespace).Create(ctx, rayJob, metav1.CreateOptions{})
```
ying-zheng, line 161: After creating the ray job in ray cluster, we should keep the information of the createdRayJob in the MA RayJob's status

go/controllers/rayjob/controller.go
badcount: yes, this is retryable error, we keep retrying until get status

go/controllers/rayjob/controller.go
badcount: updated with checking cluster valid and add toto to clear up in cluster

go/controllers/rayjob/controller.go
badcount: updated

go/controllers/rayjob/controller.go
badcount: updated for all return statement.

go/controllers/rayjob/controller.go
badcount: updated to use the immutable function

go/controllers/rayjob/controller.go
```diff
@@ -0,0 +1,206 @@
...
+				// we start checking to see if the job has created by checking job status
+				status, jobFailedReason, jobErr := r.getJobStatus(ctx, logger, &rayJob)
+				if jobErr != nil {
+					logger.Error(jobErr, "error to get ray job")
+					err = r.createJob(ctx, logger, &rayJob, rayCluster)
+					if err != nil {
+						logger.Error(err, "failed to create the ray job in ray operator")
+						rayJob.Status.State = v2pb.RAY_JOB_STATE_FAILED
+						rayJob.Status.Message = fmt.Sprintf("failed to create the ray job in cluster %s/%s", rayCluster.Namespace, rayCluster.Name)
+						return res, nil
```
ying-zheng: update status

ying-zheng (approved):

PR #65, opened 2025-01-15
Change `ruff` instead of `black` as python linter (with poetry package fix)

hwaminkim:
Change `ruff` instead of `black` as python linter (with poetry package fix)

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -13,25 +13,102 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.9.0"
 fsspec = "^2024.10.0"
-ray = "^2.41.0"
-pydantic = "^2.10.6"
 pandas = "^2.2.3"
 pyarrow = "^19.0.0"
+pydantic = "^2.10.6"
+ray = "^2.41.0"
+typing-extensions = "^4.12.2"
 
 
 [tool.poetry.group.dev.dependencies]
-black = "^24.10.0"
 pytest = "^8.3.4"
+ruff = "^0.9.1"
 
 
 [tool.poetry.scripts]
 ma = 'michelangelo.cli.cli:main'
 
 
-[tool.black]
-line-length = 120
-
-
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".bzr",
+    ".direnv",
+    ".eggs",
+    ".git",
+    ".git-rewrite",
+    ".hg",
+    ".ipynb_checkpoints",
+    ".mypy_cache",
+    ".nox",
+    ".pants.d",
+    ".pyenv",
+    ".pytest_cache",
+    ".pytype",
+    ".ruff_cache",
+    ".svn",
+    ".tox",
+    ".venv",
+    ".vscode",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
+
+# https://peps.python.org/pep-0008/
+line-length = 88
+indent-width = 4
+
+# Assume Python 3.9
+target-version = "py39"
+
+[tool.ruff.lint]
+# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
+# Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
+# McCabe complexity (`C901`) by default.
+select = ["E4", "E7", "E9", "F"]
+ignore = []
+
+# Allow fix for all enabled rules (when `--fix`) is provided.
+fixable = ["ALL"]
+unfixable = []
+
+# Allow unused variables when underscore-prefixed.
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+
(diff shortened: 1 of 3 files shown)


python/pyproject.toml
```diff
@@ -13,21 +13,98 @@ packages = [
...
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
+
+# https://peps.python.org/pep-0008/
+line-length = 79
```
andriika: Let's keep the original line length of 120.

The "line length 80" convention originates from the limitations of IBM punch cards, which could only hold 80 characters per line. We don't need to follow this convention in the modern era with wide monitors. https://www.reddit.com/r/programming/comments/gt4wgn/linus_torvalds_on_80character_line_limit

andriika (dismissed):

python/pyproject.toml
hwaminkim: It's true the 79 is originated from the legacy system and we have wide monitor, but it still meaningful for keeping our code with better readability. PEP8 [explicitly states](https://peps.python.org/pep-0008/#maximum-line-length) the line length as 79, because it has benefits to read code side by side.

> Limit all lines to a maximum of 79 characters.
>
> Limiting the required editor window width makes it possible to have several files open side by side, and works well when using code review tools that present the two versions in adjacent columns.

In my opinion, we're doing open source now, so it would be great if we follow the python standard. It would cover various development styles with better readability. I understand it's hard to keep this rule and not persist to keep this with zero tolerance. However, many other open source projects, like [Ray](https://github.com/ray-project/ray/blob/master/.editorconfig#L25), are trying to keep this Pythonic coding style too.

hwaminkim: For you information, there are some trends in the other open source projects to set the max-line length **88**, which would be okay to see the code side-by-side. I'm not sure where this number originated, but it would be okay to use it since it still under 99 lines. This number is default config in `ruff` and `black` and popular on other open source projects.

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

andriika: Thank you for the information. It is known that PEP8 recommends 79 characters per line, while Black is set at 88. Still, the modern trend is leaning towards longer lines. Many open-source projects opt for 120: https://github.com/search?q=%2F%5Eline-length+%3D+120%2F+path%3Apyproject.toml&type=code, and we can be one of them. Besides, JetBrains IDEs default to 120, and Go recommends no limit at all.

If we zoom out a bit, I wouldn't make a big deal of it and just honor the existing rules, such as "line-length = 120". Let's spend our energy on more important things.

hwaminkim: Let's bring this topic up in today's team meeting to help us reach a faster decision.

For your information, I also checked projects that using [88](https://github.com/search?q=%2F%5Eline-length+%3D+88%2F+path%3Apyproject.toml&type=code) or [79](https://github.com/search?q=%2F%5Eline-length+%3D+79%2F+path%3Apyproject.toml&type=code) lines (27.5k and 9.5k # of projects), and there appear to be more projects following PEP 8 standards than using 120 characters. I also checked some popular projects, and they seem to follow one of these:

- [Ray](https://github.com/ray-project/ray/blob/master/.editorconfig#L25)
- [Poetry](https://github.com/python-poetry/poetry/blob/main/pyproject.toml#L110)
- [pytest](https://github.com/pytest-dev/pytest/blob/main/pyproject.toml#L92)
- [Tensorflow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/pylintrc#L188)
- [pyTorch](https://github.com/pytorch/pytorch/blob/main/pyproject.toml#L18)
- [click](https://github.com/pallets/click/blob/main/.editorconfig#L10)
- [Flask](https://github.com/pallets/flask/blob/main/.editorconfig#L10)
- [Django](https://github.com/django/django/blob/main/.flake8#L4)

I'm not sure what the existing rule is that you mentioned. However, PEP 8 is considered a standard for unifying Python coding styles to improve readability. Its purpose is to establish a consistent, “pythonic” style across various environments and user levels, whereas IDEs strive to be more general in order to appeal to the widest range of users, from beginners to experts. Since we use an auto-linter, changing the format should not create extra work.

ying-zheng:
Searched the line-length settings in gihub.

line-length = 120: 23k projects
line-length = 119: 4.8k projects
line-length = 100: 14.4k projects
line-length = 99: 5.1k projects
line-length = 90: 1.6k projects
line-length = 88: 25.3k projects
line-length = 80: 3.6k projects
line-length = 79: 9k projects

88 is the most popular, especially considering that 88 is the default of black, people don't need to explicitly set 88 when using black. The actual number of projects using line-length=88 should be much more than 25.3k

Let's use 88 for now, as it's still the most popular setting in the industry today, although I personally prefer to a longer line length. It's meaningless to debate which line length is best, as this is a personal preference thing.

A few things to note:
This line length only applies to python code. For go code, we have to follow [Uber's go style guide](https://github.com/uber-go/guide/blob/master/style.md): "We recommend a soft line length limit of 99 characters. Authors should aim to wrap lines before hitting this limit, but it is not a hard limit. Code is allowed to exceed this limit."

For the projects that use 88 line length, it's common to have exceptions. In Ray project, there are 126 exceptions: https://github.com/search?q=repo%3Aray-project%2Fray%20e501&type=code

ying-zheng (approved):

PR #69, opened 2025-01-20
Add Python-based tools/sandbox

andriika:
This PR replaces the bash-based tools/sandbox.sh with a new Python-based tools/sandbox.py, allowing Michelangelo developers and users to reuse the same sandbox code from the Michelangelo Python package.

Minor refactoring:
- Consolidate python/.gitignore into root .gitignore
- Add "python" dir to .vscode/settings.json, allowing vs-code users to navigate "from michelangelo import ..." imports.


.vscode/settings.json
ying-zheng, line 1: Shouldn't we ignore .vscode in .gitignore?

badcount (approved):

.vscode/settings.json
andriika: We should check in settings.json per our .gitignore file, which we auto-generate using [toptal's tool](https://www.toptal.com/developers/gitignore). 

```
### VisualStudioCode ###
.vscode/*
!.vscode/settings.json
```

AI thoughts: https://www.perplexity.ai/search/per-git-best-practices-should-mOhl0PF1S9mDoD2QUamYzA

ying-zheng: AI gave both "Arguments for Checking In" and "Arguments Against Checking In". Why do you think the conclusion is "checking in"?

If we think we should check in .vscode/settings.json, should we also check in configurations for all the other IDEs? e.g. .idea/?

ying-zheng: "ensure that all team members have consistent settings for the project, which can help maintain code style and formatting standards" Most people in our team actually using JetBrains IDEs. Checking in vscode settings doesn't ensure consistent settings

andriika: > Why do you think the conclusion is "checking in"?

@ying-zheng, I'm not drawing any new conclusions. I am following the existing conclusion in our .gitignore file, which clearly states that .vscode/settings.json should be checked in. (see `!.vscode/settings.json`)

> If we think we should check in .vscode/settings.json, should we also check in configurations for all the other IDEs? e.g. .idea/?

Yes, I believe we should.

> Checking in vscode settings doesn't ensure consistent settings.

It does guarantee consistent settings for VS Code users. We can also check in settings for other IDEs to create consistent settings for users of those IDEs.

PR #74, opened 2025-01-27
Dev Release of Michelangelo services

andriika:
- Added reusable Dockerfile for all Michelangelo services, e.g., controllermgr, api, worker, etc.
- Added GitHub workflow responsible for building container images of Michelangelo services and pushing them to the [Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) in org's packages: https://github.com/orgs/michelangelo-ai/packages

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -49,7 +49,7 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: [ controllermgr ] Extract Metadata
+      - name: controllermgr / Extract Metadata
         id: meta-controllermgr
         uses: docker/metadata-action@v5  # https://github.com/docker/metadata-action
         with:
@@ -61,7 +61,7 @@ jobs:
             type=ref,event=tag
             type=ref,prefix=pr-,event=pr
 
-      - name: [ worker ] Extract Metadata
+      - name: worker / Extract Metadata
         id: meta-worker
         uses: docker/metadata-action@v5  # https://github.com/docker/metadata-action
         with:
@@ -73,7 +73,7 @@ jobs:
             type=ref,event=tag
             type=ref,prefix=pr-,event=pr
 
-      - name: [ controllermgr ] Push Image
+      - name: controllermgr / Push Image
         id: push
         uses: docker/build-push-action@v6  # https://github.com/docker/build-push-action
         with:
@@ -87,7 +87,7 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
 
-      - name: [ worker ] Push Image
+      - name: worker / Push Image
         id: push
         uses: docker/build-push-action@v6  # https://github.com/docker/build-push-action
         with:
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -2,8 +2,6 @@ name: Dev Release
 
 on:
   workflow_dispatch:
-  pull_request:
-    branches: [ main ]
   push:
     branches: [ main ]
 
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -33,7 +33,6 @@ jobs:
         id: push
         uses: docker/build-push-action@v6
         with:
-          context: .
           build-args: |
             BAZEL_TARGET=//go/cmd/controllermgr
             BINARY_PATH=bazel-bin/go/cmd/controllermgr/controllermgr_/controllermgr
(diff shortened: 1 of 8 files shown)


.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,100 @@
...
+  push:
+    branches: [ main ]
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  push-service-container-images:
+    runs-on: ubuntu-24.04
+    permissions:
```
badcount, line 21: how about we use strategy here to have like this so we can avoid duplicate the steps?
strategy:
      matrix:
        service:
          - controllermgr
          - worker

andriika: I'm not sure if the strategy is applicable in this scenario, as some steps should be executed only once while others need to be executed for each service.

Do you think the strategy can handle this? Please share an example, I would be happy to incorporate it.

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Cache Bazel Outputs
+        uses: actions/cache@v4  # https://github.com/actions/cache
+        with:
+          path: ~/.cache/bazel
+          key: bazel_cache_root
+          restore-keys: bazel_cache_root
+
+      - name: Build Services
+        run: |
+          ./tools/bazel build //go/cmd/controllermgr
```
sishilong: not sure apiserver side is ready or not? @ying-zheng do we need to also build apiservice

.github/workflows/dev-release.yml
sishilong: how many services do we expect to have here? if there will be more and more services added, matrix sounds fine. if only a few, i think a single job is fine too.

badcount: I mean this: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategy
basically just make it a list, for the run once, i got some suggestion from GPT (I DIDN'T test it!)

name: Dev Release

on:
  workflow_dispatch:
  push:
    branches: [ main ]

env:
  REGISTRY: ghcr.io

jobs:
  setup:
    runs-on: ubuntu-24.04
    outputs:
      bazel-cache-key: ${{ steps.cache.outputs.cache-hit }}
    steps:
      - name: Set up QEMU
        uses: docker/setup-qemu-action@v3

      - name: Set up Buildx
        uses: docker/setup-buildx-action@v3

      - name: Checkout Repository
        uses: actions/checkout@v4

      - name: Cache Bazel Outputs
        id: cache
        uses: actions/cache@v4
        with:
          path: ~/.cache/bazel
          key: bazel_cache_root
          restore-keys: bazel_cache_root

      - name: Build and Save Artifacts
        run: |
          ./tools/bazel build //go/cmd/controllermgr
          cp bazel-bin/go/cmd/controllermgr/controllermgr_/controllermgr controllermgr
          ./tools/bazel build //go/cmd/worker
          cp bazel-bin/go/cmd/worker/worker_/worker worker
        # Save built binaries as artifacts for downstream jobs
        - name: Upload Artifacts
          uses: actions/upload-artifact@v3
          with:
            name: binaries
            path: |
              controllermgr
              worker

  push-service-container-images:
    needs: setup  # This ensures the setup job runs first
    strategy:
      matrix:
        service: 
          - controllermgr
          - worker
    runs-on: ubuntu-24.04
    steps:
      - name: Download Artifacts
        uses: actions/download-artifact@v3
        with:
          name: binaries

      - name: Login to Container Registry
        uses: docker/login-action@v3
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.repository_owner }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Extract Metadata
        id: meta
        uses: docker/metadata-action@v5
        with:
          images: ${{ env.REGISTRY }}/michelangelo-ai/${{ matrix.service }}
          tags: |
            type=schedule,pattern={{date 'YYYYMMDD-hhmmss' tz='UTC'}}
            type=sha
            type=ref,event=branch
            type=ref,event=tag
            type=ref,prefix=pr-,event=pr

      - name: Push Image
        id: push
        uses: docker/build-push-action@v6
        with:
          context: .
          file: docker/service.Dockerfile
          build-args: |
            BINARY_PATH=${{ matrix.service }}
            CONFIG_PATH=go/cmd/${{ matrix.service }}/config
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
          platforms: linux/amd64,linux/arm64
          push: true

.github/workflows/dev-release.yml
sallycr: @sishilong yes. we need to build apiservice. I can contribute it top of this diff.

.github/workflows/dev-release.yml
andriika: @badcount Thanks, matrix works. Updated.

badcount (approved):

PR #76, opened 2025-01-30
Enable container registry auth for sandbox

andriika:
Enable container registry auth for sandbox

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -1,6 +1,9 @@
+import json
+import os
 import sys
 import argparse
 import subprocess
+import tempfile
 import time
 import uuid
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -146,17 +197,64 @@ def _exec(
...
+                    _err_exit("command failed", code=e.returncode)
+
+            # Wait before the next attempt.
             print("retrying after", retry_delay_seconds, "seconds...")
             time.sleep(retry_delay_seconds)
 
 
+def _assert_command(command: str, err_message: str):
+    try:
+        _exec("command", "-v", command, raise_error=True)
```
badcount: Command is not very good on cross platform, how about shutil:
https://docs.python.org/3/library/shutil.html#shutil.which

```
import shutil
def _assert_command(command: str, err_message: str):
    if shutil.which(command) is None:
        _err_exit(err_message)
```

andriika: Good point. Updated.

badcount (approved):

sallycr (approved):
LGTM. note : can you create issues for these TODOs? especially for the code block for public readiness.

andriika:
@sallycr Issue created https://github.com/michelangelo-ai/michelangelo/issues/77

PR #78, opened 2025-01-31
Import Cadence starlark-worker

andriika:
- Import new 3rd party dep - https://github.com/cadence-workflow/starlark-worker
- Added a new 'starlark' module to the worker to let the worker serve the starlark workflow (cadence workflow with an embedded starlark interpreter)
- Added initial skeleton code for the Ray Starlark plugin
- Changed base image for Michelangelo services to extend from gcr.io/distroless/base-debian12 instead of gcr.io/distroless/static-debian12. This is needed because after importing https://github.com/cadence-workflow/starlark-worker, bazel can no longer build the worker service as a static binary (for some reason, further investigation is needed)


go/worker/plugins/ray/starlark_module.go
```diff
@@ -0,0 +1,69 @@
...
+func (r *module) createJob(t *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	ctx := cadstar.GetContext(t)
+	logger := workflow.GetLogger(ctx)
+
+	var spec *starlark.Dict
+	if err := starlark.UnpackArgs("create_job", args, kwargs, "spec", &spec); err != nil {
+		logger.Error("error", zap.Error(err))
+		return nil, err
+	}
+
```
badcount, line 67: do we missing the executeActivity here for the ray job?

go/worker/plugins/ray/starlark_module.go
```diff
@@ -0,0 +1,69 @@
...
+func (r *module) Hash() (uint32, error)                 { return 0, fmt.Errorf("no-hash") }
+func (r *module) Attr(n string) (starlark.Value, error) { return r.attributes[n], nil }
+func (r *module) AttrNames() []string                   { return ext.SortedKeys(r.attributes) }
+
+func (r *module) createCluster(t *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	ctx := cadstar.GetContext(t)
+	logger := workflow.GetLogger(ctx)
+
+	var spec *starlark.Dict
+	if err := starlark.UnpackArgs("create_cluster", args, kwargs, "spec", &spec); err != nil {
```
badcount, line 44: should the spec be struct rather than pointer?
var spec starlark.Dict

go/worker/plugins/ray/starlark_module.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return m
+}
+
+func (r *module) String() string                        { return pluginID }
+func (r *module) Type() string                          { return pluginID }
+func (r *module) Freeze()                               {}
+func (r *module) Truth() starlark.Bool                  { return true }
+func (r *module) Hash() (uint32, error)                 { return 0, fmt.Errorf("no-hash") }
+func (r *module) Attr(n string) (starlark.Value, error) { return r.attributes[n], nil }
```
badcount, line 36: Should do existence check like:
attr, ok := r.attributes[n]
if !ok {
    return nil, fmt.Errorf("attribute %s not found", n)
}
return attr, nil

go/worker/starlark/module.go
```diff
@@ -0,0 +1,26 @@
...
+func register(workers []worker.Worker) {
+
+	plugins := plugin.Registry
+	plugins[ray.Plugin.ID()] = ray.Plugin
+
+	service := &cadstar.Service{
+		Plugins: plugins,
+	}
+	for _, w := range workers {
+		service.Register(w)
```
badcount, line 30: should we have error handling:
if err := service.Register(w); err != nil {
        // Consider using structured logging instead of ignoring errors
        fmt.Printf("Failed to register worker: %v\n", err)
    }

go/worker/starlark/module.go
```diff
@@ -0,0 +1,26 @@
...
+	"go.uber.org/cadence/worker"
+	"go.uber.org/fx"
+)
+
+var Module = fx.Options(
+	fx.Invoke(register),
+)
+
+func register(workers []worker.Worker) {
+
```
badcount, line 22: how about add check at the beginning:
```

if len(workers) == 0 {
		panic("No workers available for registration") // Avoid nil pointer issues
	}
```

go/worker/plugins/ray/starlark_module.go
andriika: We should not. Per [HasAttr](https://pkg.go.dev/go.starlark.net/starlark#HasAttrs) docs, `Attr` returns (nil, nil) if the attribute is not present.

go/worker/plugins/ray/starlark_module.go
andriika: It should be a pointer because the `spec`'s type must be compatible with the [Value](https://pkg.go.dev/go.starlark.net/starlark#Value).

go/worker/plugins/ray/starlark_module.go
andriika: When comparing it to the `createCluster` method above, `createJob` misses the activity call. However, this is not a working code but rather just a skeleton. Both methods require further implementation.

go/worker/starlark/module.go
andriika: [Service.Register](https://pkg.go.dev/github.com/cadence-workflow/starlark-worker/cadstar#Service.Register) does not return an error, so error handling is not applicable.

go/worker/starlark/module.go
andriika: Check added.

badcount (approved):

PR #79, opened 2025-01-31
Uniflow: Initial code

andriika:
- Uniflow: Initial code
  - Core
  - Skeleton code for the Ray plugin
- Updated example

## Next Steps

- Make `examples/bert_cola.py` runnable in the Remote Mode (Cluster Mode)


python/michelangelo/uniflow/core/build.py
```diff
@@ -0,0 +1,384 @@
...
+
+class File:
+    def __init__(self):
+        self._functions: dict[str, ast.FunctionDef] = {}
+        self._loads: dict[str, dict[str, str]] = {}
+
+    def add_function(self, v: ast.FunctionDef):
+        functions = self._functions
+        if v.name in functions:
+            assert functions[v.name] == v
```
badcount, line 54: should we use ast.dump to verify the duplicatoin to make it more robust?
`assert ast.dump(self._functions[v.name]) == ast.dump(v)`

andriika: I’m not sure if adding this extra level of assurance is necessary. Do you think there could be a situation where `assert functions[v.name] == v` passes but `assert ast.dump(self._functions[v.name]) == ast.dump(v)` fails?

python/michelangelo/uniflow/core/context.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    # Validate the function's code.
+    try:
+        build(fn)
+    except Exception as err:
+        err_message = (
+            "Error in building the @workflow function. Ensure it meets all required workflow code specifications."
+        )
+        raise NameError(err_message) from err
```
badcount: NameError should be used when a variable/function is not defined, but the error here is a build failure, not an undefined name.
How about:
raise RuntimeError("Error in building the @workflow function.") from err

python/michelangelo/uniflow/core/decorator.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def write_task_result(url: str, value):
+    with fsspec.open(url, mode="wt") as f:
+        json.dump(value, f, default=encoder.default)
+
+
+def _unref_args(sig: inspect.Signature, args: tuple, kwargs: dict, io: IORegistry) -> tuple[tuple, dict]:
+    # Copy args and kwargs to avoid modifying the original
+    pos_args, key_args = list(args), dict(kwargs)
```
badcount, line 267: This copy is not enough i think:
`pos_args, key_args = copy.deepcopy(args), copy.deepcopy(kwargs)  
`

python/michelangelo/uniflow/core/decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            args=[ast.Constant(dot_path(origin_fn))],
+            keywords=keywords,
+        )
+
+
+def task(
+    config: TaskConfig,
+    alias: Optional[str] = None,
+    io: IORegistry = default_io,
+    cache_endabled: bool = False,
```
badcount: typo: cache_enabled

python/michelangelo/uniflow/core/decorator.py
andriika: A shallow copy suffices in this case since we never alter the values of the user-provided arguments. This function may replace an entire argument with a Ref value, but it never manipulates the argument value itself.

python/michelangelo/uniflow/core/context.py
andriika: Sounds good. Updated.

python/michelangelo/uniflow/core/decorator.py
andriika: Thanks. Updated.

python/michelangelo/uniflow/core/decorator.py
```diff
@@ -0,0 +1,298 @@
...
+
+    return tuple(pos_args), key_args
+
+
+def _replace_ref_args(sig: inspect.Signature, pos_args: list, key_args: dict) -> tuple[dict, dict]:
+    replaced_pos_args, replaced_key_args = {}, {}
+    for i, p in enumerate(sig.parameters.values()):
+        if p.annotation != Ref:
+            continue
+        if i < len(pos_args):
```
badcount, line 290: Do we want to replace even it's None?
should we do
`if i < len(pos_args) and pos_args[i] is not None:`

python/michelangelo/uniflow/core/decorator.py
```diff
@@ -0,0 +1,298 @@
...
+            replaced_pos_args[i] = pos_args[i]
+            pos_args[i] = None
+        else:
+            k = p.name
+            if k in key_args:
```
badcount, line 295: same here

python/michelangelo/uniflow/core/build.py
badcount: it just functions[v.name] == v is too strict and even a minor difference but not logical difference will fail.

andriika: I believe that the two assertions below are always in sync, meaning they either both succeed or both fail:
- assert functions[v.name] == v
- assert ast.dump(self._functions[v.name]) == ast.dump(v)

Because of this connection, using the "ast.dump-based" assertion might not be the best choice. It doesn't change any logic and doesn't make things clearer either.

python/michelangelo/uniflow/core/decorator.py
andriika: We don't need to. It works as expected, even with "None" values present. To keep things simple, we can just leave this code as it is without adding extra checks for "None".

python/michelangelo/uniflow/core/decorator.py
andriika: Replied previously in the `pos_args` thread.

badcount (approved):

PR #80, opened 2025-02-02
api server (part 2)

ying-zheng:
1. Sync CRD schemas when api server starts
   Refactored and clean up the code. 
   Rewrote the logic that that list / update / delete crds. Fixed a bug. Supported different groups and versions. Allowed deleting crds

2. Removed the remaining code of the kubernetes client in protobuf generated code
3. In kubeproto compiler, only generate CRD yaml schema for CRD messages, not CRD list messages


go/api/crd/gateway.go
```diff
@@ -0,0 +1,167 @@
...
+				ctx,
+				crd,
+				metav1.CreateOptions{},
+			)
+			if err != nil {
+				e := fmt.Errorf("failed to create CRD %s: %w", crd.Name, err)
+				r.logger.Error(e.Error())
+				return e
+			}
+			return err
```
badcount: should we just
return nil here?

go/api/crd/gateway.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Compare change, then apply update conditionally
+	r.logger.Info(fmt.Sprintf("CRD %s exist, compare CRD schema", crd.Name))
+	compareResult, err := CompareCRDSchemas(crdOnServer, crd)
+	if err != nil {
+		return err
+	}
+
+	if !compareResult.HasChange {
+		r.logger.Info(fmt.Sprintf("Skip schema update. No change in CRD %s", crd.Name))
```
badcount: I think we all use logr format like:
log.Info("ray cluster created", "namespace", createdRayCluster.Namespace, "name", createdRayCluster.Name)

so i think you can do 
r.logger.Info("Skip schema update. No change in CRD ", "name", crd.Name)

go/api/crd/schema_diff.go
```diff
@@ -0,0 +1,384 @@
...
+
+	if old.AllOf != nil || new.AllOf != nil {
+		allOfDiff, err := NewSchemaArrayDiff(old.AllOf, new.AllOf)
+		if err != nil {
+			return nil, err
+		}
+		diff.AllOfDiff = allOfDiff
+	}
+
+	if old.AllOf != nil || new.AllOf != nil {
```
badcount: Is this duplicated?

go/api/crd/schema_diff.go
```diff
@@ -0,0 +1,384 @@
...
+// Following changes are not compatible
+// 1) type change
+// 2) remove properties
+// 3) any 1) and 2) changes in nested schema hierarchy (recursive)
+func (d *SchemaDiff) Compatible() bool {
+	if d == nil {
+		return true
+	}
+
+	if !d.TypeDiff.Compatible() {
```
badcount: Should we have nil check? 
if d.TypeDiff != nil && d.TypeDiff.Compatible()

go/api/crd/schema_diff.go
```diff
@@ -0,0 +1,384 @@
...
+
+	return &CompareResult{
+		HasChange:  true,
+		Compatible: changeCompatible,
+	}, nil
+}
+
+func hasChange(oldCRD *apiextv1.CustomResourceDefinition, newCRD *apiextv1.CustomResourceDefinition) bool {
+	var oldSpec bytes.Buffer
+	json.NewEncoder(&oldSpec).Encode(oldCRD.Spec.Versions)
```
badcount: should we check error?
if err := json.NewEncoder(&oldSpec).Encode(oldCRD.Spec.Versions); err != nil {
    return false
}

go/api/crd/sync.go
```diff
@@ -0,0 +1,154 @@
...
+func upsertCRDs(ctx context.Context, logger *zap.Logger, gateway Gateway,
+	crds []*apiextv1.CustomResourceDefinition, enableIncompatibleUpdate bool) error {
+	logger.Info("Compare CRD in cluster with new CRD definition, and conditionally update CRDs")
+	for _, crd := range crds {
+		// retry on conflict
+		err := backoff.Retry(func() error {
+			err := gateway.ConditionalUpsert(ctx, crd, enableIncompatibleUpdate)
+			if err != nil {
+				// retry on OCC conflict
+				if k8sErrors.IsConflict(err) {
```
badcount: We only check on conflict? How about other errors like should retry:
k8sErrors.IsServerTimeout(err) and k8sErrors.IsTooManyRequests(err)

go/api/crd/sync.go
```diff
@@ -0,0 +1,154 @@
...
+			return syncCRDs(logger, groups, p.Config.EnableIncompatibleUpdate, p.Gateway, yamlSchemas...)
+		}
+
+		return nil
+	})
+}
+
+func syncCRDs(logger *zap.Logger, groups []string, enableIncompatibleUpdate bool, gateway Gateway,
+	yamlSchemas ...map[string]string) error {
+	ctx := context.Background()
```
badcount: should we get context from parameter?

go/api/crd/sync.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	existingCRDs, err := gateway.List(ctx)
+	if err != nil {
+		logger.Error("Failed to list existing CRDs", zap.Error(err))
+		return err
+	}
+
+	for _, crd := range existingCRDs.Items {
+		if !slices.Contains(groups, crd.Spec.Group) {
```
badcount, line 135: is this lookup efficient?
```
groupSet := make(map[string]struct{}, len(groups))
for _, g := range groups {
    groupSet[g] = struct{}{}
}
```
And then inside the loop
```
if _, found := groupSet[crd.Spec.Group]; !found {
    continue
}
```

go/api/crd/schema_diff.go
ying-zheng: go allows "this" pointer to be nil. All these compatible() functions have checked this != nil

go/api/crd/schema_diff.go
ying-zheng: Use reflect.DeepEqual() instead of json string comparison

ying-zheng:
1. Move CRD sync from fx initialization to fx OnStart. CRD sync has to talk to a remote server and may take a long time. It's better to be called OnStart hook
2. Use zap.String field, instead of fmt.Sprintf
3. Retry other temporary k8s errors
4. Use reflect.DeepEqual() to compare CRD versions
5. Remove duplicated / unused code
6. Change visibility of schema diff structs and functions to package level instead of global

go/api/crd/sync.go
ying-zheng: We only have tens of CRDs. The performance of this look up doesn't matter. Even if, in the future, we have 1000 CRDs, the look up will take less than 1 ms on a modern CPU

I prefer to keep the code simpler

badcount (approved):

sallycr (approved):
LGTM. @badcount please check Ying's response. thx

PR #84, opened 2025-02-10
Update Python test workflow & Add Dockerfile for example 

hwaminkim:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Add an example Dockerfile for user and test run
- Update Python test workflow as recommended in github doc (https://github.com/marketplace/actions/install-poetry-action)
- Add pyTorch in poetry pacakge

**Why?**

There were an report that poetry couldn't install pytorch package well in the docker image. It would be caused by a wrong poetry install way (e.g. `pip install poetry`). Added a Dockerfile for users.

**How did you test it?**
Ran the bert_cola example and a simple custom pytorch training job in the Docker image.

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**
N/A


python/Dockerfile
badcount, line 1: I have one question for this, is the file for examples? Because each example should have its own dockerfile right?

python/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ENV PATH="/root/.local/bin:$PATH"
+RUN poetry --version
+
+# Set working directory
+WORKDIR /app
+
+ENV PYTHONPATH=/app
+
+# Copy the application code
+COPY . .
```
badcount: we don't want to copy the env or venv folder under python as that's local env

python/Dockerfile
hwaminkim: Yes. It's for the example for users. I guess we can place it under each example

python/Dockerfile
hwaminkim: Discussed with @badcount in offline.

Currently we're putting all python package dependencies in the single poetry group, but we'd like to support multiple ML libraries exclusively. It will be supported by different groups in the poetry.

sallycr (approved):
Please add developer contribution wiki with "starting with docker image" section.

PR #87, opened 2025-02-12
Add unmarshalJson method to all proto generated go code

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Add unmarshalJson method to all proto generated go code.

**Why?**
This is due to error in controller manager:
W0211 16:56:01.791761   37339 reflector.go:569] external/gazelle~~go_deps~io_k8s_client_go/tools/cache/reflector.go:251: failed to list *v2.RayCluster: json: cannot unmarshal string into Go struct field RayCluster.items.spec of type v2.TerminationType

**How did you test it?**
Unit test and local cluster mode test

**Potential risks**
When directly using v2pb ENUM fields in any go-code place potentially will cause unmarshal issue, for example, defining struct for uniflow code using v2pb ENUM field will have issue for cannot unmarshal string into Go struct field. So we need to avoid direct using those ENUM in non-proto place and try to use string instead

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/go/cmd/kubeproto/protoc-gen-kubeproto/main.go
+++ b/go/cmd/kubeproto/protoc-gen-kubeproto/main.go
@@ -313,7 +313,7 @@ func updateGoFile(gofile *plugingo.CodeGeneratorResponse_File, protofile *protog
 							genImmutability(typeName, &crdFunctionsBuf, options)
 							genCRDObject(typeName, gInfo, &crdFunctionsBuf)
 							genCRDBlobFields(typeName, protoMsg, &crdFunctionsBuf, extTypes)
-							genCRDEnumFields(typeName, protoMsg, &crdFunctionsBuf, extTypes, processedGoPackageList)
+							genCRDEnumTypeFields(typeName, protoMsg, &crdFunctionsBuf, extTypes, processedGoPackageList)
 							crdFunctionsBuf.Write([]byte("func init() {\n\tYamlSchemas[\"" + typeName + "\"] = `"))
 							yamlStr := string(yaml.GenerateCRDYaml(protofile, extTypes, *gInfo))
 							crdFunctionsBuf.Write([]byte(strings.Replace(yamlStr, "`", "`+\"`\"+`", -1)))
@@ -489,11 +489,11 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
 	return resp
 }
 
-// findEnumFields recursively finds all enum fields within a given protobuf message and its nested messages.
+// findEnumTypeFields recursively finds all enum fields within a given protobuf message and its nested messages.
 //
 // This function:
 // - Traverses through all fields of the given message (`curMsg`).
-// - Identifies enum fields and adds them to `enumFields`, ensuring they are not duplicated.
+// - Identifies enum fields and adds them to `enumTypeFields`, ensuring they are not duplicated.
 // - Recursively processes nested messages to extract enum fields from them as well.
 // - Maintains a record of processed message types and Go package imports to avoid redundant processing.
 //
@@ -501,11 +501,11 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
 //
 //	curMsg                 *protogen.Message     - The current protobuf message being analyzed.
 //	pathPrefix             string               - The prefix representing the field hierarchy (used for recursion).
-//	enumFields             *[]string            - A list to store discovered enum field names.
+//	enumTypeFields             *[]string            - A list to store discovered enum type names.
 //	processedMessageTypes  *map[protogen.GoIdent]bool - Tracks already processed message types to prevent cycles.
 //	extTypes               *protoregistry.Types - A registry of external types (not used in this function but may be useful for extensions).
 //	processedGoPackageList *map[string]bool     - A map to track processed Go packages and prevent duplicate enum field entries.
-func findEnumFields(curMsg *protogen.Message, pathPrefix string, enumFields *[]string,
+func findEnumTypeFields(curMsg *protogen.Message, pathPrefix string, enumTypeFields *[]string,
 	processedMessageTypes *map[protogen.GoIdent]bool, extTypes *protoregistry.Types, processedGoPackageList *map[string]bool) {
 	for _, field := range curMsg.Fields {
 		// Check if this field is an enum
@@ -514,14 +514,14 @@ func findEnumFields(curMsg *protogen.Message, pathPrefix string, enumFields *[]s
 			if _, found := (*processedGoPackageList)[key]; !found {
 				(*processedGoPackageList)[key] = true
 				newPrefix := strings.Trim(field.Enum.GoIdent.GoName, ".")
-				*enumFields = append(*enumFields, newPrefix)
+				*enumTypeFields = append(*enumTypeFields, newPrefix)
 			}
 		}
 		// Continue processing nested messages
 		if field.Message != nil && field.Message.GoIdent.GoImportPath == curMsg.GoIdent.GoImportPath {
 			if _, found := (*processedMessageTypes)[field.Message.GoIdent]; !found {
 				(*processedMessageTypes)[field.Message.GoIdent] = true
-				findEnumFields(field.Message, pathPrefix+"."+field.GoName, enumFields, processedMessageTypes,
+				findEnumTypeFields(field.Message, pathPrefix+"."+field.GoName, enumTypeFields, processedMessageTypes,
 					extTypes, processedGoPackageList)
 				delete(*processedMessageTypes, field.Message.GoIdent)
 			}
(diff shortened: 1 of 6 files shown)


go/worker/plugins/ray/starlark_module.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/cadence-workflow/starlark-worker/cadstar"
+	"github.com/cadence-workflow/starlark-worker/ext"
+	"github.com/michelangelo-ai/michelangelo/go/worker/activities/ray"
+	"go.starlark.net/starlark"
+	"go.uber.org/cadence/workflow"
+	"go.uber.org/zap"
+)
+
+// TODO: andrii: implement Ray starlark plugin here
```
sallycr: @badcount Is this TODO?

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -471,17 +471,97 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
...
+//   processedMessageTypes  *map[protogen.GoIdent]bool - Tracks already processed message types to prevent cycles.
+//   extTypes               *protoregistry.Types - A registry of external types (not used in this function but may be useful for extensions).
+//   processedGoPackageList *map[string]bool     - A map to track processed Go packages and prevent duplicate enum field entries.
+func findEnumFields(curMsg *protogen.Message, pathPrefix string, enumFields *[]string,
+	processedMessageTypes *map[protogen.GoIdent]bool, extTypes *protoregistry.Types, processedGoPackageList *map[string]bool) {
+	for _, field := range curMsg.Fields {
+		// Check if this field is an enum
+		if field.Enum != nil {
+			key := fmt.Sprintf("%s.%s", field.Enum.GoIdent.GoImportPath, field.Enum.GoIdent.GoName)
+			if _, found := (*processedGoPackageList)[key]; !found {
```
yangfan0328, line 514: why not use `if !(*processedGoPackageList)[key]` directly?

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -471,17 +471,97 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
...
+				(*processedGoPackageList)[key] = true
+				newPrefix := strings.Trim(field.Enum.GoIdent.GoName, ".")
+				*enumFields = append(*enumFields, newPrefix)
+			}
+		}
+		// Continue processing nested messages
+		if field.Message != nil && field.Message.GoIdent.GoImportPath == curMsg.GoIdent.GoImportPath {
+			if _, found := (*processedMessageTypes)[field.Message.GoIdent]; !found {
```
yangfan0328, line 522: Same here.

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: no, when key not found, it throws error

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: Same

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -308,12 +313,13 @@ func updateGoFile(gofile *plugingo.CodeGeneratorResponse_File, protofile *protog
...
 							genCRDBlobFields(typeName, protoMsg, &crdFunctionsBuf, extTypes)
-							crdFunctionsBuf.Write([]byte("func init() {\n\tYamlSchemas[\"" + typeName + "\"] = `"))
-							yamlStr := string(yaml.GenerateCRDYaml(protofile, extTypes, *gInfo))
-							crdFunctionsBuf.Write([]byte(strings.Replace(yamlStr, "`", "`+\"`\"+`", -1)))
-							crdFunctionsBuf.Write([]byte("`\n"))
-							crdFunctionsBuf.Write([]byte("}"))
+							genCRDEnumFields(typeName, protoMsg, &crdFunctionsBuf, extTypes, processedGoPackageList)
 						}
+						crdFunctionsBuf.Write([]byte("func init() {\n\tYamlSchemas[\"" + typeName + "\"] = `"))
+						yamlStr := string(yaml.GenerateCRDYaml(protofile, extTypes, *gInfo))
```
ying-zheng: I think there is an incorrect merge with https://github.com/michelangelo-ai/michelangelo/pull/80/files#diff-558a20dbd03661f71c5658501965d13057b0d4d04e3286261ea0c36fdd3c4842

We should only generate CRD Yaml string for CRD types not CRD list types. The code that generates yamlStr should be inside "if options.Bool("has_resource") { }"

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -471,17 +477,94 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
...
+				*enumFields = append(*enumFields, newPrefix)
+			}
+		}
+		// Continue processing nested messages
+		if field.Message != nil && field.Message.GoIdent.GoImportPath == curMsg.GoIdent.GoImportPath {
+			if _, found := (*processedMessageTypes)[field.Message.GoIdent]; !found {
+				(*processedMessageTypes)[field.Message.GoIdent] = true
+				findEnumFields(field.Message, pathPrefix+"."+field.GoName, enumFields, processedMessageTypes,
+					extTypes, processedGoPackageList)
+				delete(*processedMessageTypes, field.Message.GoIdent)
```
uzbq, line 526: Why do we need to delete the message from the processed map? Do we need to precess it again?

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -471,17 +477,94 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
...
+			}
+		}
+	}
+}
+
+// genCRDEnumFields generates UnmarshalJSON methods for all enum fields in a given CRD (Custom Resource Definition) message.
```
uzbq: It is generating UnmarshalJson methods for enum fields, why do we call this func as genCRDEnumFields?

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -471,17 +477,94 @@ func generate(reqData []byte) *plugingo.CodeGeneratorResponse {
...
+func findEnumFields(curMsg *protogen.Message, pathPrefix string, enumFields *[]string,
+	processedMessageTypes *map[protogen.GoIdent]bool, extTypes *protoregistry.Types, processedGoPackageList *map[string]bool) {
+	for _, field := range curMsg.Fields {
+		// Check if this field is an enum
+		if field.Enum != nil {
+			key := fmt.Sprintf("%s.%s", field.Enum.GoIdent.GoImportPath, field.Enum.GoIdent.GoName)
+			if _, found := (*processedGoPackageList)[key]; !found {
+				(*processedGoPackageList)[key] = true
+				newPrefix := strings.Trim(field.Enum.GoIdent.GoName, ".")
+				*enumFields = append(*enumFields, newPrefix)
```
ying-zheng: This seems to be the go enum type name, not the field name.
Should we call the variable enumTypes instead of enumFields?

go/kubeproto/templates/templates.go
```diff
@@ -46,6 +46,24 @@ var CRDHasBlobFields = template.Must(template.New("crdHasBlobFields").Parse(`fun
 }
 `))
 
+// CRDUnmarshalEnum is a template for the HasEnumFields() function.
+var CRDUnmarshalEnum = template.Must(template.New("unmarshalJSON").Parse(`func (m *{{.EnumFieldName}}) UnmarshalJSON(b []byte) error {
```
ying-zheng: EnumFieldName should be EnumTypeName

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: Just make the name consistent with existing method genCRDBlobFields

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: Consider this structure:

MessageA
├── Field1 -> MessageB
└── Field2 -> MessageB
If processedMessageTypes is not reset (i.e., the key is not deleted after the first recursion), Field2 -> MessageB would be skipped entirely, leading to incomplete traversal.
Deleting field.Message.GoIdent ensures that MessageB can be processed again when encountered in a different branch

uzbq: I see! I thought for each message type, we only process once. Thanks @badcount for the detailed explanation!  It make senses to me. 

Just want to learn more about the logic.  If this is the case, why do we need to set 
**				
(*processedMessageTypes)[field.Message.GoIdent] = true
**

Will we have nested same-type message? Will we skip those inner messages?

badcount: this is just for traveling the tree structure, the decision of nested enum handle in the line 514, we based on the import_path + type name to identify the same type or not. And yes, we will have nested type, and I have added unit test for it, you can check it it there.

go/worker/plugins/ray/starlark_module.go
badcount: Removed

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: Reverted

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
badcount: Done

go/kubeproto/templates/templates.go
badcount: updated

yangfan0328 (approved):

uzbq (approved):

PR #89, opened 2025-02-13
Add ray activities and plugins

badcount:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What**
Add ray plugin to uniflow
Add ray plugin to starlark-worker
Add ray activities to starlark-worker

**Why**
Support ray task in uniflow

**Test**
Unit test and local cluster run

**Potential Risk**
This ray integration has been tested using latest [kubeRay](https://github.com/ray-project/kuberay/tree/release-1.2.2)

**Release notes**

**Documentation Changes**


go/controllers/raycluster/controller.go
```diff
@@ -94,33 +95,45 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+					rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_FAILED
+				}
+			} else {
+				// Fallback to deprecated state field
+				terminateStateMap := map[v1.ClusterState]v2pb.RayClusterState{
+					v1.Failed:    v2pb.RAY_CLUSTER_STATE_FAILED,
+					v1.Suspended: v2pb.RAY_CLUSTER_STATE_TERMINATED,
+					v1.Ready:     v2pb.RAY_CLUSTER_STATE_READY,
+				}
+				println("============state===========")
```
sallycr: @badcount please remove debugging message... and next line too.

badcount: Updated

go/controllers/raycluster/controller.go
```diff
@@ -94,38 +112,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-			}
-		} else if *status == "ready" {
-			logger.Info("cluster is ready, we continue to re-queue until receiving termination signal")
-			res.RequeueAfter = requeueAfter
-			rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_READY
 		} else {
-			res.RequeueAfter = requeueAfter
-			rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_FAILED
+			// Update the cluster state based on current conditions or legacy fields
+			if status.Conditions != nil {
```
uzbq, line 130: In which case will we use status.Conditions? Who set them?

go/controllers/raycluster/controller.go
```diff
@@ -94,38 +112,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+				if meta.IsStatusConditionTrue(status.Conditions, string(v1.HeadPodReady)) {
+					rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_READY
+					res.RequeueAfter = requeueAfter
+				} else if meta.IsStatusConditionTrue(status.Conditions, string(v1.RayClusterReplicaFailure)) {
+					rayCluster.Status.State = v2pb.RAY_CLUSTER_STATE_FAILED
+				}
+			} else {
+				terminateStateMap := map[v1.ClusterState]v2pb.RayClusterState{
```
uzbq, line 138: should we define this terminiateStateMap as a global constant? It looks like a constant which could be shared in the future?

go/controllers/raycluster/controller.go
badcount: This is the ray.io status, some of the ray has the condition set based on different versions. Going forward, the state field has been deprecated, ray will use conditions instead. https://github.com/ray-project/kuberay/blob/release-1.2.2/ray-operator/apis/ray/v1/raycluster_types.go#L126

go/controllers/raycluster/controller.go
badcount: Good point, but i will do this for now, and will move globally later on if needed. I think principle is we don't code for future but for current needs, and private > public as best practise.

uzbq (approved):

go/worker/activities/ray/ray_activities.go
```diff
@@ -2,25 +2,338 @@ package ray
...
+		}
+		getRayClusterResponse, err := r.rayClusterService.GetRayCluster(ctx, getRayClusterRequest)
+		if err != nil {
+			logger.Error(err, "activity-error")
+			return err
+		}
+
+		cluster = getRayClusterResponse.RayCluster
+		var terminateType v2pb.TerminationType
+		if request.Type == v2pb.TERMINATION_TYPE_SUCCEEDED.String() {
```
yangfan0328, line 240: why don't we do terminationTypeMap and check "if terminateType, exists := terminationTypeMap[request.Type]; exists" to make this more scalable?

go/worker/activities/ray/ray_activities.go
```diff
@@ -2,25 +2,338 @@ package ray
...
+			Type:   terminateType,
+			Reason: request.Reason,
+		}
+		updateRayClusterRequest := &v2pb.UpdateRayClusterRequest{
+			RayCluster:    cluster,
+			UpdateOptions: &metav1.UpdateOptions{},
+		}
+		updateRayClusterResponse, err := r.rayClusterService.UpdateRayCluster(ctx, updateRayClusterRequest)
+		if err != nil {
+			logger.Error(err, "activity-error")
```
yangfan0328, line 255: should we add retry info in logging?

go/worker/activities/ray/ray_activities.go
```diff
@@ -2,25 +2,338 @@ package ray
...
+			return err
+		}
+		if updateRayClusterResponse.RayCluster == nil {
+			return errors.New("failed to update ray cluster")
```
yangfan0328, line 259: same should we add retry info in logging?

go/worker/activities/ray/ray_activities.go
```diff
@@ -2,25 +2,338 @@ package ray
...
+	getRayJobResponse, err := r.rayJobService.GetRayJob(ctx, &request)
+	if err != nil {
+		logger.Error(err, "activity-error")
+		return nil, cadence.NewCustomError(err.Error())
+	}
+	rayJob := getRayJobResponse.RayJob
+	status := rayJob.Status
+
+	// Check if the job has reached a terminal state
+	terminal := status.State == v2pb.RAY_JOB_STATE_KILLED || status.State == v2pb.RAY_JOB_STATE_FAILED || status.State == v2pb.RAY_JOB_STATE_SUCCEEDED
```
yangfan0328, line 306: should this also be moved to a util function isRayJobTerminalState?

go/controllers/raycluster/controller.go
```diff
@@ -162,35 +196,40 @@ func (r *Reconciler) createCluster(ctx context.Context, log logr.Logger, cluster
...
 	log.Info("ray cluster created", "namespace", createdRayCluster.Namespace, "name", createdRayCluster.Name)
 	if err != nil {
 		log.Error(err, "Failed to submit RayCluster")
 		return err
 	}
+	// Update the cluster's head node information
 	cluster.Status.HeadNode = &v2pb.RayHeadNodeInfo{
-		Name: createdRayCluster.Name,
+		Name: createdRayCluster.Status.Head.PodName,
+		Ip:   createdRayCluster.Status.Head.PodIP,
```
yangfan0328, line 209: Just curious, why PodIP is a required field for head node info?

go/worker/activities/ray/ray_activities.go
badcount: what retry info you're referring to? All worker log has job metadata info in the log already

go/worker/activities/ray/ray_activities.go
badcount: same here

go/worker/activities/ray/ray_activities.go
badcount: this is only used here, why we need to move to util?

go/controllers/raycluster/controller.go
badcount: this is not required field, there isn't required field in the proto definition, all fields are optional

go/worker/activities/ray/ray_activities.go
badcount: I can't do arbitrary map as this need to handle specifically. As we only have succeed and failed states, hence i make it clear here.

badcount: And the succeeded and failed states are passed from starlark, to have clear logic handling instead of implicitly handling is better I think

go/worker/activities/ray/ray_activities.go
yangfan0328: Since it's in RetryOnConflict function, but if job metadata is in log then it should be good.

yangfan0328 (approved):

go/worker/activities/ray/ray_activities.go
badcount: Yes, agree

ying-zheng (approved):

PR #90, opened 2025-02-17
Add starlark code for ray

badcount:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Add ray plugin in uniflow so we can submit ray cluster/job to starlark worker. Also update the io to support s3 interface

**Why?**
For end to end training flow using uniflow

**How did you test it?**
For local cluster mode only, as we can't test python starlark code just in python env, but we will add integration test later on.

**Potential risks**
No uni test, need to have integration test for end to end test

**Release notes**
Support ray plugin in uniflow

**Documentation Changes**


python/michelangelo/uniflow/commons.star
```diff
@@ -0,0 +1,194 @@
...
+#    # the format of the path is <bucket>/<file_path>
+#    parts = path.split("/")
+#    if len(parts) != 2:
+#        fail("400: unsupported path:", path)
+#    b = s3.read(parts[0], parts[1])
+#    return b
+
+def io_read_json(url):
+    readers = {
+        #        "s3://": s3_read_json,
```
yangfan0328: Why comment this prefix? Also is this function used to trim the prefix on different storage?

badcount: Accidently commented. Uncommented

yangfan0328 (approved):

PR #92, opened 2025-02-19
Add minIO to sandbox storage

badcount:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
To use minIO as sandbox storage solution for local cluster run

**Why?**
This is for the end to end run in local demo model. MinIO implements S3 protocol as storage solution.

**How did you test it?**
Unit test and local cluster run

**Potential risks**
This is local demo purpose which should not used for production. MinIO is a simple implementation of S3 protocol, it's not a robust solutions for large distributed scenario

**Release notes**

**Documentation Changes**


go/worker/activities/s3/s3_activities.go
sallycr, line 1: shouldn't we change this to minio_activities instead of s3_activities?

sallycr: fyi @ying-zheng @mincai Ying, what is the plugin implementation in api / controller avoiding heavy dependencies? Can we refactor this to the same approach?

ying-zheng: I think we should add an abstraction layer on top of s3 / terrablob / gcs / azure blob

Here is my idea:
1. Add a generic uniflow plugin, blob. All the uniflow code should only talk to this blob plugin (instead of s3 / terrablob)
2. Declare a blob storage interface (not the one in go/storage/blob_storage.go, which is too specific to unified api) in go/base. The fx module of the uniflow blob plugin will take blob storage interfaces as fx input parameters
3. put the implementations of blob storage interface (s3 / gcs) into /go/module/blob/s3 and /go/module/blob/gcs. 
4. each implementation of blob storage interface is a fx module, which will initialize the module based on the configuration
5. The same blob storage interface and the implementations will also be used by api server and controller manager, replacing  go/storage/blob_storage.go

It's possible to have multiple blob implementations configured at the same time. When accessing blob storage, the actual implementation to be used will be determined by the url scheme (e.g. gs://, s3://, tb://)

go/worker/activities/storage/interface/interface.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+)
+
+// Storage is an interface that abstracts the operations for accessing
+// different storage systems. By defining this interface, we can implement
+// various storage backends (e.g., S3, local filesystem, etc.) that all adhere
+// to the same set of behaviors. This allows the rest of the application to
+// interact with any storage system in a uniform manner.
+type Storage interface {
```
uzbq, line 12: Do we need write function?

go/worker/activities/storage/minio/client.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/minio/minio-go/v7"
+)
+
+// minioClient is a wrapper around the MinIO S3 client.
+// It provides methods to interact with S3-compatible storage.
+type minioClient struct {
+	s3Client *minio.Client
```
uzbq, line 15: Is minio attached to s3? Does minio support other type of storage

Why dont we name it minioClient direclty?

go/worker/activities/storage/interface/interface.go
uzbq: It looks that we do not since results will be written in the spark/ray job with python.

go/worker/plugins/storage/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	var path string
+	if err := starlark.UnpackArgs("execute", args, kwargs,
+		"protocol", &protocol,
+		"path", &path,
+	); err != nil {
+		logger.Error("builtin-error", ext.ZapError(err)...)
+		return nil, err
+	}
+
+	// Call the S3 read activity
```
uzbq: The comment may need to be updated since here we should be able to call different protocol instead of s3 only.

go/worker/activities/storage/interface/interface.go
badcount: We do, the get is from starlark code, it uses this.

go/worker/activities/storage/minio/client.go
badcount: yes, minIO implements full s3 specs only: https://min.io/product/s3-compatibility

go/worker/plugins/storage/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	var res any
+	if err := workflow.ExecuteActivity(ctx, storage.Activities.Read, protocol, path).Get(ctx, &res); err != nil {
+		logger.Error("builtin-error", ext.ZapError(err)...)
```
yangfan0328, line 75: Just for future support, should we do retry if read has errors or classify retry-able and non retry-able errors?

go/worker/activities/storage/minio/client.go
yangfan0328: Why we choose minio as api client? What's the advantage of using it instead of s3 client directly? Is it any possibility we will use s3 client?

badcount: this is s3 client, this is a lightweight s3 client we choose. Just for demo and sandbox purpose.

go/worker/plugins/storage/plugin.go
badcount: The error handling should be in each implementation as each protocol has different error codes, so the generic handler can't handle this.

go/worker/plugins/storage/plugin.go
badcount: removed

badcount:
replied

go/worker/activities/s3/s3_activities.go
badcount: updated

yangfan0328 (approved):

PR #94, opened 2025-02-21
Add end to end flow for bert-cola examples and fix FX issues

badcount:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix FX issue for worker.Options so we can add DataConvertor for Starlark worker
Fix starlark code for some miss arguments
Fix the activities request type mismatch issues

**Why?**
This is to make the end to end train flow work from Uniflow to UnifiedAPI backend service so we can have local cluster mode work for Demo

**How did you test it?**
Unit test and local cluster mode run

**Potential risks**

**Release notes**

**Documentation Changes**


go/worker/activities/ray/ray_activities.go
```diff
@@ -175,12 +175,7 @@ type SensorRayJobReadinessResponse struct {
 func (r *activities) SensorRayClusterReadiness(ctx context.Context, request v2pb.GetRayClusterRequest) (*SensorRayClusterReadinessResponse, *cadence.CustomError) {
 	logger := log.FromContext(ctx)
 	logger.Info("activity-start", zap.Any("request", request))
-	getRayClusterRequest := &v2pb.GetRayClusterRequest{
-		Name:       request.Name,
-		Namespace:  "",
-		GetOptions: nil,
-	}
-	getRayClusterResponse, err := r.rayClusterService.GetRayCluster(ctx, getRayClusterRequest)
+	getRayClusterResponse, err := r.rayClusterService.GetRayCluster(ctx, &request)
```
badcount, line 178: Remove redundent class object creation, no change to unit test, exiting test covers this.

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -29,21 +29,61 @@ RAY_DEFAULT_ZONE = os.environ.get("RAY_DEFAULT_ZONE", "")
...
         worker_instances = RAY_DEFAULT_WORKER_INSTANCES,
+        gpu_sku = RAY_DEFAULT_GPU_SKU,
+        zone = RAY_DEFAULT_ZONE,
         breakpoint = False):
     def callable(*args, **kwargs):
         task_name = get_task_name(task_path, alias)
-        namespace = os.environ["MA_NAMESPACE"]
-        if not namespace:
-            namespace = "default"
+        namespace = os.environ.get("MA_NAMESPACE", "default")
```
uzbq, line 86: Just double confirm. Will we still use MA_NAMESPACE?

python/michelangelo/uniflow/commons.star
```diff
@@ -1,4 +1,4 @@
-load("@plugin", "cad", "hashlib", "json", "os", "progress", "uuid")
+load("@plugin", "cad", "hashlib", "json", "os", "progress", "s3", "uuid")
 
 ENV = {
     "UF_HACK_MA_33362": "1",
```
uzbq: Do we use this env anywhere?

python/michelangelo/uniflow/plugins/ray/task.star
badcount: Yes, we're we need to set the namspace for ray cluster job in unified API, what do you suggest?

python/michelangelo/uniflow/commons.star
badcount: Good catch, removed

python/michelangelo/uniflow/plugins/ray/task.star
uzbq: I was thinking MA_NAMESPACE is too MA, we may consider to use a generic env, e.g., NAMESPACE. 

But I think it should be fine given the package is also named michelangelo already. Besides, users can write their star and define their env if they want.

uzbq (approved):

python/michelangelo/uniflow/plugins/ray/task.star
badcount: Yes, correct. thanks for pointing out

PR #97, opened 2025-02-21
Python client

ying-zheng:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

1. The generator and base class of Michelangelo Python client. 
2. The generated code

**Why?**
This is the high level Michelangelo python client on top of the gRPC compiler generated client code. This code handles the gRPC connection to Michelangelo API server (e.g. get MA API server address from an environment variable, set the necessary headers, set retries)

**How did you test it?**
tested locally, with local Michelangelo API server and k3d

**Potential risks**

**Release notes**

**Documentation Changes**
I added an example in the README.md file


python/michelangelo/api/v2/client.py
badcount, line 1: Are those files from code-gen or it's your code?

python/michelangelo/api/v2/services/base.py
```diff
@@ -0,0 +1,218 @@
...
+                    ('grpc.max_receive_message_length', _MAX_MESSAGE_LENGTH),
+                ]
+            )
+            atexit.register(channel.close)
+            _channel = channel
+        return _channel
+
+    @staticmethod
+    def _get_default_header_provider():
+        return HeaderProvider()
```
badcount: This should be "no implemented" rather than create instance from abstraction class?

python/tools/client_generator/template.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        req = List{{crd_camel}}Request(namespace=namespace)
+        list_options = self._process_message_or_dict(list_options, ListOptions)
+        req.list_options.CopyFrom(list_options)
+
+        if list_options_ext is not None:
+            operation = CriterionOperation()
+            if isinstance(list_options_ext, dict) and 'operation' in list_options_ext:
+                operation = self._process_criterion_operation(list_options_ext['operation'])
+                del list_options_ext['operation']
```
badcount, line 198: @ying-zheng just curious where we add the 'operation' to the option? User's code?

python/michelangelo/api/v2/services/base.py
ying-zheng: Good catch! This function is not in use. The default header provider is created in Context.header_provider function. I have removed this code

python/tools/client_generator/template.py
ying-zheng: This code is for the indexed search. For the field index in mysql, users can search with complex conditions, like pipeline run owner="xxx" and state="failed" and end_time > '...'

This is used in ma studio and mactl. Some users also use this to find their models and pipeline runs.

python/michelangelo/api/v2/client.py
ying-zheng: only the code in michelangelo/api/v2/services/gen and michelangelo/gen are generated code

badcount (approved):

PR #98, opened 2025-02-27
Add nomic-ai model for fine-tune example

badcount:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Add nomic-ai fine-tune LLM model as an example

**Why?**
nomic-ai that is used in Uber as an example to share in github

**How did you test it?**
Verified local mode and cluster mode in sandbox env

**Potential risks**
N/A

**Release notes**

**Documentation Changes**


python/examples/bert_cola/model.py
badcount, line 1: Remove not used file

sallycr (approved):

python/examples/nomic_ai/train.py
```diff
@@ -0,0 +1,73 @@
...
+    ),
+)
+def train(
+        train_data: Dataset,
+        validation_data: Dataset,
+        test_data: Dataset,
+        model_name="nomic-ai/nomic-bert-2048",
+        # breakpoint=True,
+) -> dict:
+    log.info("Starting training...")
```
hwaminkim, line 32: `log` is not defined but used.

hwaminkim (changes requested):
Found there is a bug in `train.py`. Would you take a look?
Also, I wonder where is `log` is defined

python/examples/nomic_ai/train.py
sallycr: hm. interesting. Job was successfully run without that.

sallycr: btw, good catch. thx

sallycr:
> Found there is a bug in `train.py`. Would you take a look? Also, I wonder where is `log` is defined

Fixed it. thx

hwaminkim:
Thanks for fixing it. I checked the local run w/o the Docker image I faced the following error like below:
```
  File "/Users/hwamin/.pyenv/versions/3.9.21/envs/mi-py39-3/lib/python3.9/site-packages/pyarrow/dataset.py", line 441, in _ensure_single_source                                                                                                                                                                                                  raise FileNotFoundError(path)
FileNotFoundError: /Users/hwamin/uf_storage/5a0e31010a484f28852235b375b5e664
```

After I cleaned up all directories under `uf_storage`, now the local run succeed. I leave this just in case when someone face the same issue later.

hwaminkim (approved):
Checked the local run works fine. You can land the diff. I'll check the remote run later

sallycr:
Thanks @hwaminkim for checking it again.

PR #112, opened 2025-03-12
To fix sandbox env for running example

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
1. Add dockerfile for example image build
2. Add type check in uniflow to fix missing "__class__" for type object when parse from json string.
3. Fix minio config and port
4. Add readme for docker build

**Why?**
This is to make sure we can ran end to end train job in sandbox

**How did you test it?**
local cluster run

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -1,0 +1,52 @@
+# Use a base image with Python installed
+FROM nvidia/cuda:12.8.0-base-ubuntu22.04
+
+# Install dependencies required for building Python
+RUN apt-get update && apt-get install -y \
+    software-properties-common \
+    wget \
+    build-essential \
+    libssl-dev \
+    libreadline-dev \
+    libbz2-dev \
+    libsqlite3-dev \
+    libffi-dev \
+    liblzma-dev \
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set Python version
+ENV PYTHON_VERSION=3.9.18
+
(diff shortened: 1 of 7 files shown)


python/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+ENV PYTHONPATH=/app
+
+# Copy pyproject.toml and poetry.lock before copying full source
+COPY pyproject.toml poetry.lock ./
+
+# Copy the application code
+COPY ../.. .
+
+RUN poetry config virtualenvs.create false && poetry install --all-extras --no-interaction --no-ansi && pip install datasets ray[default] numpy==1.23.4 torch==2.0.1 transformers==4.36.2 pandas==1.5.1 pyarrow s3fs==2024.3.0
```
badcount, line 49: @hwaminkim i need to add this here to make pip install those packages, otherwise, the training job fails

yangfan0328 (approved):

python/Dockerfile
ying-zheng: We already have these libraries in pyproject.toml. We shouldn't need to add the libraries here again

ying-zheng: What's this docker image for? Is it for the example? If yes, it's better to put the docker file into the example directory

PR #114, opened 2025-03-16
Add json non-inline support for JSON unmarshal

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix the issue for json only support inline field which is not supported by jsonpb. All the CRDs registered in the kubernetes are not inline format due to we're using jsonpb, but as we're use json as fallback plan, hence we need to fix the json string before unmarshal using json

**Why?**
Without this fix, all inline fields will not be parsed successfuly

**How did you test it?**
Local cluster run

**Potential risks**
1. This is not recursively due to gjson and sjson not supported
2. Performance issue, we can save the results to code-gen to increase performance

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/kubeproto/util/util.go
+++ b/go/kubeproto/util/util.go
@@ -1,13 +1,12 @@
 package util
 
 import (
-	"bytes"
 	"fmt"
-	"github.com/gogo/protobuf/jsonpb"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


go/kubeproto/util/util.go
```diff
@@ -167,3 +171,165 @@ func SetPackageAlias(file *dst.File, pkgPath string, newAlias string) {
 		}
 	}
 }
+
```
badcount, line 175: The following code is tested in main_test.go as end to end test

ying-zheng (approved):

PR #145, opened 2025-04-10
Migrate view container component

craigmarker:
Added aliasing for components within core, such that all imports from src can be imported from "@/". This caused an issue in the dev environment because files within app/ could not resolve this alias.

Instead of using @michelangelo/core in imports within packages/core, which is pretty long, I explicitly registered the unique paths to app's vite config. "@" was choosen because it aligns with trends in other vite/TypeScript projects. And because packages/core will be the primary export from our project, and the primary import into our dev environment, app/

In the spirit of small import paths, renamed @michelangelo/gen-* to @ma/gen-*.

With this new aliasing approach, the import sorting sourced from internal does not correctly categorize @ma/ and @/ as "internal" imports. Updated the regex to address this issue.

Updated a default TypeScript eslint rule that will be incompatible with Michelangelo's open source migration -- internally we prefer Types to Interfaces.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Screenshot 2025-04-10 at 14 47 33

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/app/vite.config.ts
+++ b/javascript/app/vite.config.ts
@@ -8,9 +8,14 @@ export default defineConfig({
   plugins: [react()],
   resolve: {
     alias: {
+      // Multiple aliases are required for the core package, such that
+      // the core package can import from itself using "@" and app can
+      // import from the core package using "@michelangelo/core"
+      '@': path.resolve(__dirname, '../packages/core/src'),
       '@michelangelo/core': path.resolve(__dirname, '../packages/core/src'),
-      '@michelangelo/gen-k8s': path.resolve(__dirname, '../gen/grpc/k8s.io/'),
-      '@michelangelo/gen-api': path.resolve(__dirname, '../gen/grpc/michelangelo/api/'),
+
+      '@ma/gen-k8s': path.resolve(__dirname, '../gen/grpc/k8s.io/'),
+      '@ma/gen-api': path.resolve(__dirname, '../gen/grpc/michelangelo/api/'),
     },
   },
 });
--- a/javascript/eslint.config.js
+++ b/javascript/eslint.config.js
@@ -27,15 +27,17 @@ const sharedRules = {
     'error',
     {
       groups: [
-        ['^react', '^@?\\w'],
-        ['^\\.\\.(?!/?$)', '^\\.\\./?$', '^\\./(?=.*/)(?!/?$)', '^\\.(?!/?$)', '^\\./?$'],
+        ['^react', '^@(?!ma/|/)', '^[^@./]'],
+        ['^@ma/', '^@/'],
         ['^@?\\w.*\\u0000$', '^[^.].*\\u0000$', '^\\..*\\u0000$'],
+        ['^.*\\.(css|scss|sass|less)$'],
       ],
     },
   ],
   'baseui/deprecated-theme-api': 'warn',
   'baseui/deprecated-component-api': 'warn',
   'baseui/no-deep-imports': 'warn',
+  '@typescript-eslint/consistent-type-definitions': ['error', 'type'],
 };
 
 export default [
--- a/javascript/packages/core/src/components/views/constants.ts
+++ b/javascript/packages/core/src/components/views/constants.ts
@@ -1,0 +1,1 @@
+export const BREADCRUMB_BAR_HEIGHT = '64px';
--- a/javascript/packages/core/src/components/views/main-view-container.tsx
+++ b/javascript/packages/core/src/components/views/main-view-container.tsx
@@ -1,0 +1,48 @@
+import { Cell, Grid } from 'baseui/layout-grid';
+
+import type { Props } from './types';
+
+import { BREADCRUMB_BAR_HEIGHT } from './constants';
+
+/**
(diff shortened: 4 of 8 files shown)


trekhleb:
@craigmarker I'm curious, what is the main problem that is being effectively solved with `@` and `@ma` aliases?

From the PR description and from changed files, I see that they come with the cost:
- Additional level of complexity (touches 3 configs: Vite + Lint + TS) 
- Additional layer of abstraction (`@ma` means `@michelangleo` in some of those ^ configs)
- Any additional IDE configs (for those who don't use VSCode)?

And from the PR description, it seems like you start paying the price for this additional complexity immediately 🙂 
- _“This caused an issue in the dev environment because files within app/ could not resolve this alias.”_
- _“With this new aliasing approach, the import sorting sourced from internal does not correctly categorize @ma/ and @/ as "internal" imports.”_

The shortening of the imports (for now?) is not that drastical: 

```
import { ThemeProvider } from './themes/provider';
import { ThemeProvider } from '@/themes/provider';

import { MainViewContainer } from ‘./components/views/main-view-container';
import { MainViewContainer } from '@/components/views/main-view-container';

import type { Project } from '@michelangelo/gen-api/v2/project_pb';
import type { Project } from '@ma/gen-api/v2/project_pb';

import { ProjectService } from '@michelangelo/gen-api/v2/project_svc_pb';
import { ProjectService } from '@ma/gen-api/v2/project_svc_pb';
```

I might be missing the bigger picture here. Maybe you plan to use aliases as some sort of helpers when it comes to multiple `@ma` packages?

craigmarker:
> @craigmarker I'm curious, what is the main problem that is being effectively solved with `@` and `@ma` aliases?
> 
> From the PR description and from changed files, I see that they come with the cost:
> 
> * Additional level of complexity (touches 3 configs: Vite + Lint + TS)
> * Additional layer of abstraction (`@ma` means `@michelangleo` in some of those ^ configs)
> * Any additional IDE configs (for those who don't use VSCode)?
> 
> And from the PR description, it seems like you start paying the price for this additional complexity immediately 🙂
> 
> * _“This caused an issue in the dev environment because files within app/ could not resolve this alias.”_
> * _“With this new aliasing approach, the import sorting sourced from internal does not correctly categorize @ma/ and @/ as "internal" imports.”_
> 
> The shortening of the imports (for now?) is not that drastical:
> 
> ```
> import { ThemeProvider } from './themes/provider';
> import { ThemeProvider } from '@/themes/provider';
> 
> import { MainViewContainer } from ‘./components/views/main-view-container';
> import { MainViewContainer } from '@/components/views/main-view-container';
> 
> import type { Project } from '@michelangelo/gen-api/v2/project_pb';
> import type { Project } from '@ma/gen-api/v2/project_pb';
> 
> import { ProjectService } from '@michelangelo/gen-api/v2/project_svc_pb';
> import { ProjectService } from '@ma/gen-api/v2/project_svc_pb';
> ```
> 
> I might be missing the bigger picture here. Maybe you plan to use aliases as some sort of helpers when it comes to multiple `@ma` packages?

It's a good question, and I agree there is not much benefit in the current codebase.

> * Additional level of complexity (touches 3 configs: Vite + Lint + TS)
> * Additional layer of abstraction (`@ma` means `@michelangleo` in some of those ^ configs)

There may be an opportunity to remove the @michelangelo/core abstraction in app/, since app/ is only intended for local development whereas @michelangelo/core is intended to be a published package.

> * Any additional IDE configs (for those who don't use VSCode)?

I'm not aware of any additional IDE configs. What am I missing?

> I might be missing the bigger picture here. Maybe you plan to use aliases as some sort of helpers when it comes to multiple `@ma` packages?

I think the primary benefit of aliases is that imports are absolute instead of relative. There could be a better implementation to achieve this benefit.

For example, given a file at src/components/table/filters/filter-menu.tsx, if the component implements a relative import, any relocation of the filter-menu.tsx component will require import path changes:

```
import { TableProps } from '../../types'
```

Would become the following if filter-menu.tsx adds an additional directory between itself and src.
```
import { TableProps } from '../../../types'
```

Whereas absolute import (using aliasing) will only change if the import itself changes:

```
import { TableProps } from '@/components/table/types'
```

trekhleb:
Got it. Thanks for the clarifications 👍🏻

> I'm not aware of any additional IDE configs. What am I missing?

I meant that IDEs need to be configured. I believe most of modern IDEs will support alias in imports anyway, but it is still a bit fragile since config files need to be configured correctly and the IDE needs to be able to find them (if the repo structure is non-standard).

> There may be an opportunity to remove the @michelangelo/core abstraction in app/,

As we synced offline, maybe we could think about importing the actual `@michelangelo/core` as a dependency from `package.json` as npm [worspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) allows us to do (instead of aliasing).

But we can land this PR and see if the workspaces-based direct imports (as `import x from "@michelangleo/core"`) approach would be a better fit

trekhleb (approved):

PR #157, opened 2025-04-17
[go] Create a components directory and move all existing components u…

dragod812:
- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Created a components directory and moved all existing component files into it to improve organization and maintainability.

**Why?**
The Components folder holds domain/feature specific modules, Integrations and reusable building blocks
Each subfolder (e.g. ray/, spark/, pipeline/) encapsulates all relevant logic—controllers, apiHooks, workers, utilities—for that domain.
New features or integrations can be added under components/ as separate modules, keeping them self-contained and easy to extend.
Other modules (like apiserver or controllermgr) can reuse modules in components, reducing duplication and clarifying ownership.

**How did you test it?**
bazel build ...
bazel test ...
Created sandbox and made sure the controllers are registered properly.

**Potential risks**
NA, 

**Release notes**

**Documentation Changes**


ying-zheng:
it seems you didn't delete the old files
is it correct to put ray cluster code into job folder?

dragod812:
> it seems you didn't delete the old files is it correct to put ray cluster code into job folder?

The files have been moved to a new package, so the PR just shows that the files have been moved, not deleted and added. 

This was a suggestion from Anant Vyas - 
> cluster controller is common to ray and spark. we should do job->cluster/ray/spark.
> "Job" controller is basically 3 controllers and a scheduler.

go/components/ray/job/controller_test.go
```diff
@@ -1,4 +1,4 @@
-package rayjob
```
badcount, line 1: I still think we should have raycluster and rayjob as folder name

dragod812: It's generally not advisable to have repetitive package naming (like `ray/rayjob`), as it introduces redundancy and reduces clarity. To avoid this issue, one option could be to directly place modules such as `components/rayjob` and `components/raycluster` under the `components` folder. However, having a parent folder (`ray`) can be beneficial if you anticipate adding additional Ray-related modules in the future (e.g., `ray/trainers`, `ray/operators`). This hierarchical approach helps group all Ray-specific functionality clearly under a single domain, aligning with the overall goal of the **components** folder—to neatly organize domain-specific modules.

badcount (approved):

PR #164, opened 2025-04-22
Replace vite alias with package.json imports

craigmarker:
vite aliases were not sufficiently propagating across dependencies, such that depending on an aliased-packaged required awareness of the aliases.

Refactored the approach to leverage package.json imports. package.json imports need to be prefixed by # instead of @, so updated imports as well and the eslint sort order configuration.

Leveraged #core instead of just #, since there will likely be more packages in the future that would conflice the with the shared tsconfig.base.json paths configuration.

Also onboarded app/ to be a workspace, so it can fully take advantage of yarn workspace symlinks.

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**How did you test it?**
yarn dev
Screenshot 2025-04-22 at 12 09 54

**Potential risks**

**Release notes**

**Documentation Changes**


javascript/packages/core/components/views/project/project-detail.tsx
```diff
@@ -2,10 +2,9 @@ import { useEffect } from 'react';
 import { useState } from 'react';
 import { createClient, Interceptor } from '@connectrpc/connect';
 import { createGrpcWebTransport } from '@connectrpc/connect-web';
-
-import { Project } from '@ma/gen-api/v2/project_pb';
-import { ProjectService } from '@ma/gen-api/v2/project_svc_pb';
-import { useStudioParams } from '@/hooks/routing/use-studio-params/use-studio-params';
+import { useStudioParams } from '#core/hooks/routing/use-studio-params/use-studio-params';
+import { Project } from '#gen-api/v2/project_pb';
+import { ProjectService } from '#gen-api/v2/project_svc_pb';
```
trekhleb, line 7: @craigmarker what do you think about importing the autogenerated types from the workspace package, rather than from the import alias? Something similar to what we do with `import { XXX } from '@uber/idl-flow/...'` (where `@uber/idl-flow` can be replaced with something like `@michelangelo/gen`)? This way, we could continue using the workspace-based imports. Or are there some benefits of importing types directly via aliases that I'm missing?

trekhleb (approved):

javascript/packages/core/components/views/project/project-detail.tsx
craigmarker: To clarify, you're suggesting we add a package (@michelangelo/gen) that has Project and ProjectService? Then @michelangelo/core would depend on @michelangelo/gen?

trekhleb: Yes, if such a package would make sense.

So currently, if I'm reading the package dependencies correctly, the `@michelangelo/core` depends on the `Project` by importing its definition from _outside_ the package via `import { Project } from '#gen-api/v2/project_pb'`. Do I read it correctly that the `'#gen-api/v2/project_pb'` path is going to be resolved to the `/javascript/gen/...` folder, which is outside of `javascript/package/core`?

If this dependency is already in place and if this import in fact gets the files from the external (to `@michelangelo/core`) folder, then such import behaves/resembles a regular npm package dependency, and custom import prefixes like `#` try to simulate the import of the "npm" package.

**If** this understanding is correct, I was wondering, in this case, maybe it makes sense to actually extract the generated files to the external package (i.e., to the `@michelangelo/gen`) and import it in a more standard way using the workspaces like `import { Project } from '@michelangleo/gen`. Very similar to what we have in the `web-code`, when we import types from `@uber/idl-flow`.

Wdyt?

craigmarker: It's a good idea! One difference (that is not clear in this code) is that idl-flow is only exporting types (in this code, import type { Project }), whereas #ma/gen-api also exports ProjectService. I don't think it makes sense to have ProjectService in a package, what do you think?

Then if we don't have a package for ProjectService, we'll have to manually keep the types up to date with the service.

I'm not sure what the right solution is...

craigmarker: Putting ProjectService & Project into a package would help prevent untested proto changes from landing and breaking Michelangelo Studio, since MA Studio would have to update @michelangelo/gen before the proto changes take effect. 

I think I'd prefer to have proto changes fail code in `javascript/` before they can land. By relying on javascript/gen directly, typescript would fail on proto changes if the proto changes would cause an issue in the client code.

trekhleb: > One difference (that is not clear in this code) is that idl-flow is only exporting types (in this code, import type { Project }), whereas #ma/gen-api also exports ProjectService. I don't think it makes sense to have ProjectService in a package, what do you think?

Ah, yes, that's correct, the `@uber/idl-flow` only exports types. But there is a similar dedicated package `@uber/typed-rpc-client` that exports the clients themselves. With the current approach, when the ProjectService is imported from the `#gen-api/v2/project_svc_pb`, the `#gen-api/v2/project_svc_pb` is simulating the package anyway (importing something from _outside_ of the `@michelangleo/core` package), isn't it?

> I think I'd prefer to have proto changes fail code in javascript/ before they can land. By relying on javascript/gen directly, typescript would fail on proto changes if the proto changes would cause an issue in the client code.

This is a good point. Failing early on proto-change may indeed prevent a lot of bugs. I didn't get a chance to check how the CI pipeline works (how the tests and scripts are running). My assumption was that the objective of failing early may be achievable with the npm-worspace packages as well, since the package dependency is just a symlink to the package folder (similarly to the `javascript/gen` folder). But again, I might be missing many details about how CI works (I'm not sure what the process of updating the protos is, at what point the `generate proto` script is called, etc). So I'll definitely leave the decision to you here.

PR #166, opened 2025-04-23
Implement apihandler.handleUpdate()

ying-zheng:
- [ ] Refactor
- [X] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
1. Implement apihandler.handleUpdate()
2. do not return error when ListOptionsExt is specified but metadata storage is not enabled

**Why?**
1. handleUpdate() was not implemented in the previous pr
2. some clients always submit request with ListOptionsExt without knowing if the server side has MetadataStorage. It's better to ignore the ListOptionsExt option instead of returning an error

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/api/handler/handler_test.go
+++ b/go/api/handler/handler_test.go
@@ -227,17 +227,6 @@ func TestK8sOnly(t *testing.T) {
 		&metav1.ListOptions{LabelSelector: "bad label selector"}, nil, listJobs)
 	checkGrpcStatusCode(t, codes.InvalidArgument, err)
 
-	// ListOptionsExt is not supported
-	err = handler.List(context.Background(), "default", &metav1.ListOptions{}, &apipb.ListOptionsExt{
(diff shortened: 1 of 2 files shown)


go/api/handler/handler_test.go
```diff
@@ -227,17 +227,6 @@ func TestK8sOnly(t *testing.T) {
...
-	err = handler.List(context.Background(), "default", &metav1.ListOptions{}, &apipb.ListOptionsExt{
-		OrderBy: []*apipb.OrderBy{
-			{
-				Field: "test",
-			},
-		},
-	}, listJobs)
-	assert.Error(t, err)
-	checkGrpcStatusCode(t, codes.Unimplemented, err)
-
```
badcount, line 240: shouldn't we asset not error?

badcount (approved):

PR #172, opened 2025-04-27
Baoquan/pipeline controller

uzbq:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Added a controller for pipeline CRD

We need reconcile logic for pipeline CRD so that uniflow can be defined and executed as a pipeline. 

- Tested with sandbox using the images produced by the following workflow https://github.com/michelangelo-ai/michelangelo/actions/runs/14689004277 
- create a pipeline crd with the following yaml file 

```
kind: Pipeline
metadata:
  name: baoquan-pipeline-0426-01
  namespace: default
spec:
  commit:
    gitRef: 123456abcefg
    branch: main
```

The created pipeline is successfully reconciled and becomes ready state. 
- update the above pipeline again. The pipeline is successfully reconciled with the updated status.


go/cmd/controllermgr/config.go
```diff
@@ -0,0 +1,46 @@
+package main
```
badcount: Why we adding this?

go/cmd/controllermgr/config/base.yaml
```diff
@@ -4,6 +4,11 @@ controllermgr:
   leaderElection: false
   leaderElectionID: decaf1259.michelangelo.uber.com
   port: 9443
+  k8s:
```
badcount: You can move the ray cluster config out to make it generic

go/components/pipeline/controller.go
```diff
@@ -0,0 +1,107 @@
...
+	pipeline := &v2pb.Pipeline{}
+	if err := r.Get(ctx, req.Namespace, req.Name, &metav1.GetOptions{}, pipeline); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	originalPipeline := pipeline.DeepCopy()
+	state := pipeline.Status.State
+	logger.Info("Reconciling pipeline", zap.Any("PipelineStatusState", state.String()))
+	// TODO: add logic to handle different states
+	switch state {
+	case v2pb.PIPELINE_STATE_INVALID:
```
badcount: the reconciler logic shouldn't be state based, we should be able to construct state:
https://book.kubebuilder.io/cronjob-tutorial/controller-implementation

go/cmd/controllermgr/config.go
uzbq: The controller is using apiHandler which requires k8sConfig and storage config.

go/cmd/controllermgr/config/base.yaml
uzbq: This is not for raycluster. This is for the k8sclient in the apiHandler.

uzbq: IC. You mean also let ray cluster to use this field.

go/cmd/controllermgr/config.go
```diff
@@ -0,0 +1,46 @@
...
+type (
+	K8sConfig struct {
+		QPS   float32 `yaml:"qps"`
+		Burst int     `yaml:"burst"`
+	}
+)
+
+// getK8sRestConfig parses the configuration file and returns the k8s REST client configuration
+// for Michelangelo Controller Mananger.
+func getK8sRestConfig(provider config.Provider) (*rest.Config, error) {
```
ying-zheng: Why do we need this k8s rest config in controller manger?
In controller manager, k8s config is read by controller runtime at https://github.com/michelangelo-ai/michelangelo/blob/main/go/controllermgr/module.go#L53
Do you know which module is consuming this rest.Config?

uzbq: It is apiHandler

go/components/pipeline/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	originalPipeline := pipeline.DeepCopy()
+	state := pipeline.Status.State
+	logger.Info("Reconciling pipeline", zap.Any("PipelineStatusState", state.String()))
+	if shouldReconcile(pipeline) {
+		pipeline.Status.LatestRevision = &apipb.ResourceIdentifier{
+			Name:      formatRevisionName(pipeline),
+			Namespace: pipeline.Namespace,
+		}
+		pipeline.Status.State = v2pb.PIPELINE_STATE_READY
+		return r.updatePipelineStatus(ctx, pipeline, originalPipeline, logger)
```
yangfan0328: If there is updatePipelineStatus error, do we just regard this as retry-able error and enter next reconciliation loop? Should we update the pipeline state to PIPELINE_STATE_ERROR as  terminate state instead?

go/components/pipeline/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/michelangelo-ai/michelangelo/go/base/env"
+	apipb "github.com/michelangelo-ai/michelangelo/proto/api"
+	v2pb "github.com/michelangelo-ai/michelangelo/proto/api/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	reconcileInterval = 10 * time.Second
```
yangfan0328, line 23: Why we define reconcileInterval here? Does that mean every controller has different reconcileInterval?

go/components/ray/cluster/config.go
```diff
@@ -5,13 +5,13 @@ import (
 )
 
 const (
-	configKey = "controllers.rayCluster"
```
badcount: Sorry what i mentioned earlier was wrong, this configuration is for Kuberay not for our kubernetes cluster that's why we have sepearated config here.

go/components/pipeline/controller.go
uzbq: Yes. I think each controller should define the reconcile interval based on its requirement.

badcount (approved):

go/components/pipeline/controller.go
uzbq: The most common error for updating status is resource version conflict. instead of set pipeline state to error, I think it may be better to give retry in next reconcile.

PR #187, opened 2025-05-13
Fix example run with poetry

hwaminkim:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

* Add `__init__.py` on the packages to enable running example from `michalgenlo/python` directory.
* Fix README instruction to run example

**Why?**

* Local run with README command was not working

**How did you test it?**

* By running in local environment

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -70,7 +70,7 @@ print(project)
 - Run tests: `poetry run pytest`
 - Run examples:
   - Install dependencies for example (ML libs for BERT model): `poetry install -E example`
-  - Run example: `poetry run python ./examples/bert_cola/bert_cola.py`
+  - Run example: `$ PYTHONPATH="." poetry run python ./examples/bert_cola/bert_cola.py`
 - Format code: `poetry run ruff format .`
 - Run Michelangelo CLI: `poetry run ma --help`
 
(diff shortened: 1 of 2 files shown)


python/examples/bert_cola/__init__.py
badcount, line 1: do we need this? we didn't have this in other examples

hwaminkim: We need this to avoid package not found error, to use both -- running example from `michelangelo/python` directory and using absolute importing paths.
@TheGreatFrankie and me checked we need this on his environment.

hwaminkim: + Also @npuppala111 fixed issue with the same approach

hwaminkim: Synced in offline. Noticed that Eric was saying `examples/bert_cola/__init__.py` file only (not the one under `examples`). Move this file away because there is no problem without this file.

badcount (approved):

TheGreatFrankie (approved):

PR #193, opened 2025-05-21
Fix cache issue for non-found error

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Add error checking for non found error for cache, when not found, continue workflow without cache

**Why?**
So the workflow continue

**How did you test it?**
local cluster remote run

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/worker/activities/storage/activities.go
+++ b/go/worker/activities/storage/activities.go
@@ -3,6 +3,7 @@ package storage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/cadence-workflow/starlark-worker/activity"
@@ -12,6 +13,8 @@ import (
 
 	intf "github.com/michelangelo-ai/michelangelo/go/worker/activities/storage/interface"
 )
+
+var ErrNotFound = errors.New("not found")
 
 // Activities is a package-level variable that holds the activities implementation.
 var Activities = (*activities)(nil)
@@ -35,12 +38,7 @@ func (a *activities) Read(ctx context.Context, protocol string, path string) (an
 		// Attempt to read from the storage using the protocol's implementation.
 		result, err := impl.Read(ctx, path)
 		if err != nil {
-			if impl.IsNotFoundError(err) {
-				logger.Error("activity-error", zap.Error(err))
-				// If the error is a "not found" error, return empty result
-				// Due to starlark code can't handle errors from worker, so we return nil
-				return nil, nil
-			}
+			logger.Error("activity-error", zap.Error(err))
 			// Wrap the error in a Cadence CustomError using YARPC error codes.
 			return nil, workflow.NewCustomError(
 				ctx,
--- a/go/worker/activities/storage/activities.go
+++ b/go/worker/activities/storage/activities.go
@@ -4,7 +4,6 @@ package storage
 import (
 	"context"
 	"fmt"
-
 	"github.com/cadence-workflow/starlark-worker/activity"
 	"github.com/cadence-workflow/starlark-worker/workflow"
 	"go.uber.org/yarpc/yarpcerrors" // YARPC errors for standardized error codes.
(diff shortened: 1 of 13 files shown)


python/examples/boston_housing_xgb/boston_housing_xgb.py
```diff
@@ -86,6 +87,8 @@ def feature_prep(
     config=SparkTask(
         driver_cpu=1,
         executor_cpu=1,
+        driver_memory='1024',
```
badcount, line 90: make this smaller for local env

go/worker/activities/storage/activities.go
```diff
@@ -35,6 +35,12 @@ func (a *activities) Read(ctx context.Context, protocol string, path string) (an
 		// Attempt to read from the storage using the protocol's implementation.
 		result, err := impl.Read(ctx, path)
 		if err != nil {
+			if impl.IsNotFoundError(err) {
+				logger.Error("activity-error", zap.Error(err))
+				// If the error is a "not found" error, return empty result
+				// Due to starlark code can't handle errors from worker, so we return nil
+				return nil, nil
+			}
```
uzbq: why do we need to specially consider this not found error? 

Can we return nil for all error?

badcount: i left comment in the code that starlark worker can't deal with error return, all activiate error in starlark worker because immediate termination.

uzbq (approved):

python/michelangelo/uniflow/commons.star
```diff
@@ -37,7 +37,13 @@ def io_read_json(url):
     protocol = url_parts[0]
     path = url_parts[1]
 
-    return storage.read(protocol, path)
+    result = storage.read(protocol, path)
```
andriika: Using tuple unpacking syntax would be cleaner. Starlark should support this feature, similar to Python.

```
result, err = storage.read(protocol, path)
if err != None:
    ...
```

This code assumes that `storage.read` returns a tuple, which should always be true.

python/michelangelo/uniflow/commons.star
```diff
@@ -37,7 +37,13 @@ def io_read_json(url):
...
+    if len(result) == 2:
+        # result is a tuple (data, error)
+        if result[1]:
+            return None
```
andriika: Let’s try not to return None for any errors we encounter. It would be great to figure out how to categorize these errors so that we only return None when we specifically come across a 404 (Not Found) error. For example:
```
if err != None:
    if err.code == 404:
        return None
    fail(err)
```

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -108,8 +108,9 @@ def task(
                         output = cached_output.get("metadata", {}).get("name", ""),
                     )
                     result = io_read_json(cached_result_json_url)
-                    print("ray | cached", "result:", result)
-                    return result
+                    if result != None:
+                        print("ray | cached", "result:", result)
+                        return result
```
andriika, line 113: We need to handle the None `result` somehow (ie, not found). @uzbq do we need to fail? E.g.:
```
if result == None:
    fail("illegal state: cache not found")
```

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -264,8 +265,9 @@ def task(
             task_message = "Ray Task Completed Successfully",
         )
         result = io_read_json(result_url)
-        print("ray | caching", "result:", result)
-        return result
+        if result != None:
```
andriika: In this case, we should consider it a failure if the result comes back as None (meaning it wasn't found). The Starlark code must read the result JSON and ultimately return its content. If it doesn’t, the workflow won’t work as intended.

python/michelangelo/uniflow/plugins/ray/task.star
uzbq: I think @badcount 's purpose is to make the cache read as a best-effort process. That means, failed to get cache for any reason will not fail the workflow. 

I am OK to not fail the workflow in this case.

python/michelangelo/uniflow/plugins/ray/task.star
uzbq: yes. This is not cache related. We should fail the workflow if the reason is None

andriika (approved):

uzbq (approved):

PR #228, opened 2025-06-13
Migrate model manager from internal codebase to OSS (initial draft)

kenns29:
- [x] Migration

Migrate majority of model manager from internal codebase to OSS

**Why?**
We need the model manager library to package and upload models for deployment

**How did you test it?**
Fixed all the unit tests, but we still need to refactor the upload/download part by abstracting away internal file storage protocal. This needs to be done in a different PR

**Potential risks**
This code is not production ready yet. Need further update on the file upload/download part and we need some minor changes in the packager too.


python/michelangelo/_internal/gateways/terrablob_gateway/blob_info.py
badcount, line 1: Terrablob is not opensource solution

PR #258, opened 2025-07-14
Modify sandbox cli to create jobs cluster

dragod812:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Added an option - create-jobs-cluster to michelangelo sandbox CLI to create an additional k8s cluster for jobs.

**Why?**
This new jobs cluster will be used by jobs controllers to schedule jobs in a separate cluster instead of the same cluster as michelangelo.

**How did you test it?**
```poetry run ma sandbox create --create-jobs-cluster```
Successfully creates the additional cluster

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
Will update https://github.com/michelangelo-ai/michelangelo/wiki/Getting-Started#running-michelangelos-api-sandbox-environment
To create a separate jobs cluster along with michelangelo-sandbox  run ```poetry run ma sandbox create --create-jobs-cluster```

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -9,6 +9,7 @@ import time
 import uuid
 import yaml
 from pathlib import Path
+import base64
 
 short_description = "Manage the sandbox cluster."
 
@@ -19,7 +20,7 @@ This tool helps you create and manage a sandbox cluster directly on your machine
 
 _dir = Path(__file__).parent
 
-_kube_name = "michelangelo-sandbox"
+_michelangelo_sandbox_kube_cluster_name = "michelangelo-sandbox"
 _kube_ports = [
     "3306:30001",  # MySQL
     "9091:30007",  # MinIO
@@ -34,7 +35,15 @@ _cadence_ports = [
     "7933:30003",  # Cadence TChannel
     "8088:30004",  # Cadence Web
 ]
+
+# Ray framework ports
+_ray_ports = [
+    "10001:10001",  # Ray client port
+    "8265:8265",  # Ray dashboard
+]
+
 _cadence_domain = "default"
+_default_job_hosting_kube_cluster_name = "michelangelo-jobs-0"
 
 
 def init_arguments(p: argparse.ArgumentParser):
@@ -53,11 +62,28 @@ def init_arguments(p: argparse.ArgumentParser):
         default="cadence",
         help="Choose workflow engine: cadence or temporal (default: cadence).",
     )
+    create_p.add_argument(
+        "--create-jobs-cluster",
+        action="store_true",
+        help="Create an additional cluster for Ray jobs.",
+    )
+    create_p.add_argument(
+        "--jobs-cluster-name",
+        default=_default_job_hosting_kube_cluster_name,
+        help="Name of the jobs cluster to create when --create-jobs-cluster is used (default: %s)."
+        % _default_job_hosting_kube_cluster_name,
+    )
 
     _ = sp.add_parser(
         "demo", help="Create demo project and pipelines in the sandbox cluster."
     )
-    _ = sp.add_parser("delete", help="Delete the cluster.")
+    delete_p = sp.add_parser("delete", help="Delete the cluster.")
+    delete_p.add_argument(
+        "--jobs-cluster-name",
+        default=_default_job_hosting_kube_cluster_name,
+        help="Name of the jobs cluster to delete when --create-jobs-cluster is used (default: %s)."
+        % _default_job_hosting_kube_cluster_name,
+    )
     _ = sp.add_parser("start", help="Start the cluster.")
     _ = sp.add_parser("stop", help="Stop the cluster.")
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by badcount — APPROVED** [review]

**Review by badcount — APPROVED** [review]

File: python/michelangelo/cli/sandbox/sandbox.py [hunk 1]
```diff
@@ -35,6 +36,7 @@
     "8088:30004",  # Cadence Web
 ]
 _cadence_domain = "default"
+_jobs_cluster_name = "michelangelo-jobs-0"
```
bhjoshi1: should this also be an argument?

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -390,6 +403,7 @@ def _create_demo_crs(_: argparse.Namespace):
 def _delete(ns: argparse.Namespace):
     assert ns
     _exec("k3d", "cluster", "delete", _kube_name)
+    _exec("k3d", "cluster", "delete", _jobs_cluster_name)
```
bhjoshi1: shouldn't this happen before you tear down the underlying kube cluster?

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -497,5 +511,165 @@ def _err_exit(err_message: str, code: int = 1):
...
+        "--servers",
+        "1",
+        "--agents",
+        "2",  # More worker nodes for Ray
+        "--kubeconfig-switch-context=false",  # Don't switch kubectl context to this cluster
+    ]
+
+    # Add port mappings for Ray
+    ray_ports = [
+        "10001:10001",  # Ray client port
```
bhjoshi1: Can you add a README that tells about these default ports on the sandbox?

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -497,5 +511,165 @@ def _err_exit(err_message: str, code: int = 1):
...
+
+    # Extract certificate-authority-data from clusters[0].cluster
+    ca_data = kubeconfig_data["clusters"][0]["cluster"].get(
+        "certificate-authority-data"
+    )
+    if not ca_data:
+        raise ValueError("certificate-authority-data not found in kubeconfig")
+
+    # Create a secret for the certificate-authority-data
+    ca_secret = {
```
bhjoshi1, line 635: I'm wondering why we need to do this for sandbox?

python/michelangelo/cli/sandbox/sandbox.py
dragod812: sure let me add that, I will keep this as default if no args are specified.

badcount (approved):

python/michelangelo/cli/sandbox/sandbox.py
dragod812: This statement deletes / cleans up the jobs kubecluster. 
As part of sandbox, 
when user runs `poetry run ma create sandbox --create-jobs-cluster` we create two kubeclusters - sandbox and jobs. 
and when user runs `poetry run ma delete` we delete the two clusters. 
Order doesn't really matter, but i think makes sense that we delete the jobs cluster first and then the sandbox, making that change.

python/michelangelo/cli/sandbox/sandbox.py
dragod812: Sandbox cluster runs the Michelangelo control plane, and the Ray Jobs run in the jobs kubernetes cluster. 
We need to connect to the jobs cluster from the sandbox cluster. 
So we need to store the Jobs cluster secrets in sandbox cluster. 
_create_cluster_secrets creates secrets for the CA cert and the service account token for connecting to the jobs cluster.

python/michelangelo/cli/sandbox/sandbox.py
dragod812: We are using the github wikis for this. 
I added one here for MA Sandbox Ports and endpoints - https://github.com/michelangelo-ai/michelangelo/wiki/MA-Sandbox-Ports-and-Endpoints

python/michelangelo/cli/sandbox/sandbox.py
bhjoshi1: Can we rename `_jobs_cluster_name` to `_job_hosting_kube_cluster` and `_kube_name` to `michelangelo_sandbox_kube_cluster` so that the names are clear?

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -389,17 +415,19 @@ def _create_demo_crs(_: argparse.Namespace):
 
 def _delete(ns: argparse.Namespace):
     assert ns
-    _exec("k3d", "cluster", "delete", _kube_name)
+    if ns.create_jobs_cluster:
+        _exec("k3d", "cluster", "delete", ns.jobs_cluster_name)
+    _exec("k3d", "cluster", "delete", _michelangelo_sandbox_kube_cluster_name)
```
dragod812, line 428: @bhjoshi1 changed the constant names as per your suggestion, ptal.

bhjoshi1 (approved):

PR #292, opened 2025-08-05
Add PhaseCard component for project workflow navigation

craigmarker:
Add PhaseCard component that displays workflow phases with visual icons, descriptions, and direct entity access. This solution provides immediate visibility into available phases. Navigation is currently not configured.

Screenshot 2025-08-05 at 16 48 38

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/views/project/components/phase-list/phase-card.tsx
+++ b/javascript/packages/core/components/views/project/components/phase-list/phase-card.tsx
@@ -1,0 +1,67 @@
+import { useStyletron } from 'baseui';
+import { Button, KIND, SHAPE, SIZE } from 'baseui/button';
+
+import { Box } from '#core/components/box/box';
+import { Icon } from '#core/components/icon/icon';
+import { Link } from '#core/components/link/link';
+import { capitalizeFirstLetter } from '#core/utils/string-utils';
+
+import type { PhaseConfig } from '#core/types/common/studio-types';
+
+export function PhaseCard({ icon, name, description, docUrl, entities }: PhaseConfig) {
+  const [css, theme] = useStyletron();
+
+  return (
+    <Box
+      title={
+        <div className={css({ display: 'flex', alignItems: 'center', gap: theme.sizing.scale400 })}>
+          <Icon name={icon} size={theme.sizing.scale500} />
+          {name}
+        </div>
+      }
+      description={
+        description && (
+          <div className={css({ display: 'flex', alignItems: 'center' })}>
+            {description}
+            {docUrl && (
+              <Button
+                kind={KIND.tertiary}
+                onClick={() => window.open(docUrl, '_blank')}
+                shape={SHAPE.circle}
+                size={SIZE.mini}
+              >
+                <Icon name="arrowLaunch" title="Learn more" size={theme.sizing.scale500} />
+              </Button>
+            )}
+          </div>
+        )
+      }
+    >
+      {entities && entities.length > 0 && (
+        <div className={css({ display: 'flex', flexDirection: 'column' })}>
+          {entities.map((entity) => {
+            return (
+              <Link
+                key={entity.id}
+                href={entity.id}
+                overrides={{ Link: { style: theme.typography.ParagraphSmall } }}
+              >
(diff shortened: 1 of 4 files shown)


javascript/packages/core/components/views/project/components/phase-list/phase-card.tsx
```diff
@@ -0,0 +1,66 @@
...
+            )}
+          </div>
+        )
+      }
+    >
+      {entities && entities.length > 0 && (
+        <div className={css({ display: 'flex', flexDirection: 'column' })}>
+          {entities.map((entity) => {
+            return (
+              <Link
```
oleksii-trekhleb, line 44: Do we need to specify a `key` prop here?

javascript/packages/core/types/common/studio-types.ts
```diff
@@ -58,6 +58,52 @@ export enum Phase {
...
+   */
+  id: string;
+}
+
+/**
+ * Simplified phase configuration matching the original studio config structure
+ */
+export interface PhaseConfig {
+  /** Icon name from the application's icon provider system */
+  icon: string;
```
oleksii-trekhleb, line 90: nit: What do you think about allowing users to provide not only the icon name but also to render an arbitrary component `() => <MyCustomSVGIcon withSomeCustomProps>` (like BaseWeb does in their interfaces)? 

It could potentially make the `<PhaseCard>` component more loosely coupled (fewer files would be required to be changed to introduce some custom case)

oleksii-trekhleb (approved):

javascript/packages/core/types/common/studio-types.ts
craigmarker: Hmm, I'm not sure. I like the potential flexibility (though it does require additional testing and maintenance). Specifically with Icons, the core library is trying to be icon-provider-agnostic and it does so by leveraging a React context that maps icon name (string) to an icon component (e.g., MyCustomSVGIcon).

In other words, it'd probably be a violation of this icon-provider-agnostic design decision if an _icon_ prop could bypass the icon provider.

What do you think?

#208

javascript/packages/core/components/views/project/components/phase-list/phase-card.tsx
craigmarker: nice catch!

javascript/packages/core/types/common/studio-types.ts
oleksii-trekhleb: That was a `nit` comment, so I'm totally ok with icon name being a `string` that is mapped to an actual icon elsewhere.

>  the core library is trying to be icon-provider-agnostic and it does so by leveraging a React context that maps icon name (string) to an icon component (e.g., MyCustomSVGIcon)

This also makes sense, however, if we're talking about the solution that is _icon-provider-agnostic_, a possibility to provide an arbitrary component as an icon (similarly to BaseWeb's startEnhancer, endEnhancer) seems to be the most icon-provider-agnostic solution out there, since it doesn't care about the provider at all (no need to even register the icon in the context provider). 

But yes, then, with such generalization, the valid counter-argument might be, why calling the property an `icon` when it is so generic that it should be called a `startEnhancer` instead :)

craigmarker: I appreciate these kinds of comments, it's helpful!

> if we're talking about the solution that is icon-provider-agnostic, a possibility to provide an arbitrary component as an icon (similarly to BaseWeb's startEnhancer, endEnhancer) seems to be the most icon-provider-agnostic solution out there

I like this idea! I investigated porting PhaseCardProps to use *enhancer approach and ran into naming trouble, since `icon` really would translate to `titleStartEnhancer`. This started to get a little overwhelming for a level of reusability the component doesn't currently require.

I will definitely incorporate this philosophy (enhancer/leading/trailing/etc) into the code going forward!

PR #307, opened 2025-08-11
python wheel file build and release

ying-zheng:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
1. Setup poetry-dynamic-versioning plugin in pyproject.toml. This will make the michelangelo wheel file use the github release tag as version, instead of a hard coded version pyproject.toml
2. Setup github actions to automatically build and publish michelangelo wheel with github release

**Why?**
1. Michelangelo users will use the wheel file version to know which version of michelangelo python code is installed and if a update is needed. As we plan to give prerelease wheel files to external users, we have to make sure the version is set correctly
2.  With this PR, the release build will use the release version as its wheel file version (e.g. 0.1.0), the dev build's wheel file version have the git sha1 appended to the last release version (e.g. 0.1.0.dev2+gabc1234)
3. The new github actions will automatically build the wheel file and attach it to the release, when a version tag is pushed

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


.github/actions/setup-python-poetry/action.yml
```diff
@@ -0,0 +1,72 @@
...
+      run: poetry install --no-interaction --no-root -E dev
+
+    #----------------------------------------------
+    # install your root project, if required
+    #----------------------------------------------
+    - name: Install project
+      if: inputs.install-root-project == 'true'
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: poetry install --no-interaction -E dev
```
badcount, line 72: why we require -dev to setup release wheel?

ying-zheng: this command is just for installing all the dependencies in the build environment. It's the pyproject.toml file that determines which libraries the wheel file depends to. This github action is shared by both unit test action and wheel file build action. Building wheel file doesn't need dev dependencies. But unit test needs dev dependencies.

badcount (approved):

PR #322, opened 2025-08-20
Add Docker and Kubernetes integration for Michelangelo UI

craigmarker:
Integrate Michelangelo UI with existing containerized infrastructure to enable unified development workflow alongside Go services.

Previous development required separate UI server run via yarn dev in addition to sandbox setup, preventing consistent development patterns across team. Similarly, the yarn dev workflow did not allow for the UI to be included in Michelangelo's k8s cluster installations.

This solution implements dual Docker strategy to balance development needs with CI reliability. Full ui.Dockerfile enables self-contained review builds while ui-prebuilt approach aligns with Go service dockerfile convention. The prebuilt approach also speeds up ui release workflow.

Implemented a separate workflow from Go service deployments for UI release, since they have fundamentally different build dependencies and triggering requirements. For instance, UI requires yarn/node and only triggers on version increment.

Currently leveraging 8090 for Michelangelo UI to match the existing pattern of UIs (cadence/temporal) being within 80XX. Notably, 3000 is currently occupied by Uber's internal deployment of MA Studio, so it's likely not a good candidate.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**How did you test it?**
Ran ma sandbox create with Docker image built locally, successfully serving at localhost:8090 and connecting to sandbox apiserver

Screenshot 2025-08-20 at 16 43 16

Temporarily allowlisted workflow triggering on test branch. Successfully detected version change and created docker that I installed locally & used to serve UI at localhost:8090
https://github.com/michelangelo-ai/michelangelo/actions?query=branch%3Atest%2Fui-workflow

### DIFF
--- a/.github/workflows/ui-release.yml
+++ b/.github/workflows/ui-release.yml
@@ -104,5 +104,5 @@ jobs:
           file: docker/ui-prebuilt.Dockerfile
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          platforms: linux/amd64,linux/arm64  # Now both platforms work!
+          platforms: linux/amd64,linux/arm64
           push: true
--- a/.github/workflows/ui-release.yml
+++ b/.github/workflows/ui-release.yml
@@ -1,0 +1,108 @@
+name: UI Release
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+    paths:
+      - 'javascript/**'
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  check-version-change:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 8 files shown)


.github/workflows/ui-release.yml
```diff
@@ -0,0 +1,108 @@
+name: UI Release
```
badcount, line 1: why we don't ad this to dev-release?

craigmarker: I tried to capture this in the commit message, let me know how I can clarify:

> Implemented a separate workflow from Go service deployments for UI release, since they have fundamentally different build dependencies and triggering requirements. For instance, UI requires yarn/node and only triggers on version increment.

badcount (approved):

PR #357, opened 2025-09-03
[MaCTL] Fix MaCTL error - missing annotation field

hwaminkim:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Confirmed with @npuppala111 that E2E pipeline run finished successfully (bert cola example)

**What changed?**

Fixed to add pipeline metadata.annotation information. Those information will be automatically loaded from yaml file when schema is there (future plan).
Disabled json convert for result output, since there was an missing schema issue from gRPC reflection. Will be fixed later.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/pipeline/create.py
+++ b/python/michelangelo/cli/mactl/plugins/pipeline/create.py
@@ -22,9 +22,6 @@ from michelangelo.gen.api.typed_struct_pb2 import TypedStruct
 
 _LOG = getLogger(__name__)
 
-# Constant for uniflow image annotation
-_UNIFLOW_IMAGE_ANNOTATION_KEY = "michelangelo/uniflow-image"
-
 # Constants for registration output files
 _UNIFLOW_TAR_PATH_FILENAME = "uniflow_tar_path.txt"
 _UNIFLOW_INPUT_FILENAME = "uniflow_input.txt"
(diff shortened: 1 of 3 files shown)


npuppala111 (approved):

python/michelangelo/cli/mactl/plugins/pipeline/create.py
```diff
@@ -234,16 +231,8 @@ def convert_crd_metadata_pipeline_create(
         uniflow_tar_path = ""
 
     res = {"spec": deepcopy(yaml_dict["spec"])}
-
-    # At this time, we expect the uniflow image annotation to be present in the pipeline config file
-    annotations = deepcopy(yaml_dict.get("metadata", {}).get("annotations", {}))
-    if _UNIFLOW_IMAGE_ANNOTATION_KEY not in annotations:
-        error_message = f"expected {_UNIFLOW_IMAGE_ANNOTATION_KEY} annotation within pipeline config file"
-        _LOG.error(error_message)
-        raise KeyError(error_message)
```
aritra0227, line 243: @hwaminkim shouldn't we keep this exception in case the pipeline.yaml file does not contain of the `michelangelo/uniflow-image` annotation? Maybe we do not want to have such checks in the future, but aren't we currently relying on `michelangelo/uniflow-image` annotation to explicitly exist in the pipeline.yaml?

hwaminkim: @aritra0227 the main concept of the mactl is universal for different servers, which may have different protobuf schema or docker image flow. I think adding this checking would be too much for flexible schema design principal. If we think that this manual injecting docker image feature is temporal solution, I think it's regulating too much.

hwaminkim: As you pointed out, we're using the metadata field in hardcoded way, but it's not what we want for now. I'll make a change to accept more flexible schema. On the top of this functionality, we may add additional checklists on the per-repo or plugin settings

aritra0227: I agree that it doesn't make sense to keep the check if we want to support a flexible schema design. I was only suggesting that we keep it for the time-being (until the change to "accept more flexible schema" is made) since we currently rely on it and the exception would provide a useful message to current users. 

Regardless, I don't have any strong opinions against this, just thought I should point it out!

hwaminkim: Thanks for sharing the point!
I also think we need a more kind message for missing image for now, since auto image build is not implemented yet. However, the only reason I removed this is, I think it would be good to be placed in the API server or Controller manager side.

PR #363, opened 2025-09-04
Allow none return value in uniflow task

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix the None return value issue that considers uniflow task failure

**Why?**
User can return none value in the task and continue to the next task

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


badcount:
Screenshot 2025-09-05 at 12 40 41 PM

python/michelangelo/uniflow/core/decorator.py
```diff
@@ -286,6 +288,9 @@ def _in_the_box(txt):
 
 
 def write_task_result(url: str, value):
+    # Handle None return values from tasks without explicit return statements
+    if value is None:
+        value = {}
```
andriika: Is this change required? `json.dump` should work well for `None` values.

badcount: Agree, revert this changes, and tested without this and it works

andriika (approved):

PR #381, opened 2025-09-10
Implemented DevRun PipelineRun Controller Changes

npuppala111:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Implemented DevRun PipelineRun Controller side changes. On detection of a new PipelineRun CRD, PipelineRun controller in the SourcePipeline actor will check if the PipelineRun.Status.pipeline_spec is empty. If PipelineRun.Status.pipeline_spec is non-nil, controller will handle the pipeline run as a DevRun. For a DevRun, controller will retrieve uniflow task image from PipelineRun.metadata.annotations[michelangelo/uniflow-image], correctly pass on input and environment variables to workflow engine and override pipeline_spec.manifest.content.value with pipelinerun.spec.input.

**Why?**
This will enable users to run a pipeline without needing to register it first. This feature will boost pipeline development velocity, as developers won't have to go through the pipeline registration process to test-run their pipeline during development. [https://docs.google.com/document/d/1Jjip6T1yaFCqIOGWZfSqn5nhubUo-16wQia32V8IZyE/edit?usp=sharing](url)

**How did you test it?**

Adding unit tests to sourcepipeline_test.go. Additionally tested by running pipeline successfully end-to-end using the following command: 
`poetry run python ./michelangelo/cli/mactl/mactl.py dev-run --file=examples/bert_cola/pipeline.yaml --env=foo=bar --env=lorem=ipsum`

**Potential risks**
Pipeline runs can fail if there are incorrect changes.

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/components/pipelinerun/actors/executeworkflow.go
+++ b/go/components/pipelinerun/actors/executeworkflow.go
@@ -169,6 +169,15 @@ func getWorkflowInputs(pipelineRun *v2.PipelineRun) ([]interface{}, []interface{
 		}
 	}
 
+	// Apply DevRun environment overrides if present
+	if pipelineRun.Spec.Input != nil {
+		if environField := pipelineRun.Spec.Input.Fields["environ"]; environField != nil {
+			if environOverrides := environField.GetStructValue(); environOverrides != nil {
+				applyDevRunEnvironmentOverrides(envs, environOverrides)
+			}
+		}
+	}
+
 	envs["MA_NAMESPACE"] = pipelineRun.Namespace
 	envs["MA_PIPELINE_RUN_NAME"] = pipelineRun.Name
 	envs["UF_STORAGE_URL"] = DefaultWorkSpaceRootURL
@@ -215,4 +224,23 @@ func addTaskImageToEnv(pipelineRun *v2.PipelineRun, envs map[string]interface{})
 
 func (a *ExecuteWorkflowActor) GetType() string {
 	return ExecuteWorkflowType
+}
+
+// applyDevRunEnvironmentOverrides applies DevRun environment variable overrides to the base environment
+func applyDevRunEnvironmentOverrides(baseEnv map[string]interface{}, devInput *pbtypes.Struct) {
+	if devInput == nil || len(devInput.Fields) == 0 {
+		return // No overrides to apply
+	}
+
+	// Apply dev input overrides - convert all values to strings for environment variables
+	for key, value := range devInput.Fields {
+		switch value.GetKind().(type) {
+		case *pbtypes.Value_StringValue:
+			baseEnv[key] = value.GetStringValue()
+		case *pbtypes.Value_NumberValue:
+			baseEnv[key] = fmt.Sprintf("%g", value.GetNumberValue())
+		case *pbtypes.Value_BoolValue:
+			baseEnv[key] = fmt.Sprintf("%t", value.GetBoolValue())
+		}
+	}
 }
--- a/go/components/pipelinerun/actors/executeworkflow.go
+++ b/go/components/pipelinerun/actors/executeworkflow.go
@@ -172,8 +172,8 @@ func getWorkflowInputs(pipelineRun *v2.PipelineRun) ([]interface{}, []interface{
 	// Apply DevRun environment overrides if present
 	if pipelineRun.Spec.Input != nil {
 		if environField := pipelineRun.Spec.Input.Fields["environ"]; environField != nil {
-			if environOverrides := environField.GetStructValue(); environOverrides != nil {
-				applyDevRunEnvironmentOverrides(envs, environOverrides)
+			if err := applyDevRunEnvironmentOverrides(envs, environField.GetStructValue()); err != nil {
+				return nil, nil, nil, fmt.Errorf("failed to apply DevRun environment overrides: %w", err)
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


go/components/pipelinerun/actors/sourcepipeline.go
```diff
@@ -87,3 +109,159 @@ func (a *SourcePipelineActor) Run(ctx context.Context, pipelineRun *v2.PipelineR
...
+	// Update manifest content with merged environment
+	baseConfig["environ"] = baseEnv
+
+	// Re-encode manifest content
+	updatedContent, err := a.encodePipelineManifestContent(baseConfig)
+	if err != nil {
+		return fmt.Errorf("failed to encode updated manifest: %w", err)
+	}
+
+	pipeline.Spec.Manifest.Content = updatedContent
```
andriika: Can we avoid updating the original manifest content of the pipeline? Ideally, we'd like to keep it exactly as it's defined in the pipeline's source code (Git repository).

How about we apply input overrides in the [getWorkflowInputs](https://github.com/michelangelo-ai/michelangelo/blob/605c189afd8dcbb1f1548b409a0e4893b86c2174/go/components/pipelinerun/actors/executeworkflow.go#L148) instead?

npuppala111: Hi @andriikali thanks for pointing this out! So this pipeline, that we are modifying the manifest content of, is the source_pipeline we are creating in sourcepipeline.go. It is not the pipeline_spec that is passed in from user and it did not previously exist. So in this case, would it be okay to keep the logic here or should we move it to getWorkflowInputs instead?

andriika: Yes, it would be best to move the logic to the `getWorkflowInputs`. We shouldn't view the status.source_pipeline as part of the business logic. Instead, it should be viewed as a [denormalization](https://en.wikipedia.org/wiki/Denormalization) of the user-defined pipeline specification, with its ground truth stored in the Git repository.

npuppala111: Makes sense, I'll make this change, thanks Andrii!

go/components/pipelinerun/actors/executeworkflow.go
```diff
@@ -216,3 +225,22 @@ func addTaskImageToEnv(pipelineRun *v2.PipelineRun, envs map[string]interface{})
...
+	// Apply dev input overrides - convert all values to strings for environment variables
+	for key, value := range devInput.Fields {
+		switch value.GetKind().(type) {
+		case *pbtypes.Value_StringValue:
+			baseEnv[key] = value.GetStringValue()
+		case *pbtypes.Value_NumberValue:
+			baseEnv[key] = fmt.Sprintf("%g", value.GetNumberValue())
+		case *pbtypes.Value_BoolValue:
+			baseEnv[key] = fmt.Sprintf("%t", value.GetBoolValue())
+		}
```
andriika, line 244: What if there are other types, like struct, list, or nil? Should we just ignore them?

In general, environment variables are strings. Can we check that the value is actually a string and return an error if it isn't?

go/components/pipelinerun/actors/sourcepipeline.go
```diff
@@ -87,3 +105,48 @@ func (a *SourcePipelineActor) Run(ctx context.Context, pipelineRun *v2.PipelineR
...
+			Name:        fmt.Sprintf("devrun-%s", pipelineRun.Name),
+			Namespace:   pipelineRun.Namespace,
+			Annotations: make(map[string]string),
+		},
+		Spec: *pipelineSpec, // Use inline spec directly
+	}
+
+	// Copy annotations from pipeline run metadata (set by CLI during dev run creation)
+	// For dev runs, annotations like image-id are stored in the PipelineRun metadata
+	if pipelineRun.Annotations != nil {
```
andriika: This nil check is unnecessary. It's fine to apply `range` to `nil` value, which will skip it over.

go/components/pipelinerun/actors/executeworkflow.go
```diff
@@ -169,6 +169,15 @@ func getWorkflowInputs(pipelineRun *v2.PipelineRun) ([]interface{}, []interface{
 		}
 	}
 
+	// Apply DevRun environment overrides if present
+	if pipelineRun.Spec.Input != nil {
+		if environField := pipelineRun.Spec.Input.Fields["environ"]; environField != nil {
+			if environOverrides := environField.GetStructValue(); environOverrides != nil {
```
andriika: It seems this nil check is unnecessary since `applyDevRunEnvironmentOverrides` already handles nil values.

go/components/pipelinerun/actors/executeworkflow.go
```diff
@@ -216,3 +225,22 @@ func addTaskImageToEnv(pipelineRun *v2.PipelineRun, envs map[string]interface{})
 func (a *ExecuteWorkflowActor) GetType() string {
 	return ExecuteWorkflowType
 }
+
+// applyDevRunEnvironmentOverrides applies DevRun environment variable overrides to the base environment
+func applyDevRunEnvironmentOverrides(baseEnv map[string]interface{}, devInput *pbtypes.Struct) {
+	if devInput == nil || len(devInput.Fields) == 0 {
```
andriika: It seems this `len == 0` check is unnecessary because `range` will skip the execution block when there are no items.

go/components/pipelinerun/actors/sourcepipeline.go
```diff
@@ -87,3 +105,48 @@ func (a *SourcePipelineActor) Run(ctx context.Context, pipelineRun *v2.PipelineR
...
+
+// validateDevRunSpec validates required fields for dev run pipeline specs
+func (a *SourcePipelineActor) validateDevRunSpec(pipelineSpec *v2.PipelineSpec) error {
+	if pipelineSpec.Manifest == nil {
+		return fmt.Errorf("pipeline_spec.manifest is required for dev runs")
+	}
+	if pipelineSpec.Manifest.UniflowTar == "" {
+		return fmt.Errorf("pipeline_spec.manifest.uniflow_tar is required for dev runs")
+	}
+	return nil
```
andriika: `Manifest` and `Manifest.UniflowTar` are required fields, regardless of whether it's a dev run or not. I.e., it seems PipelineSpec validation logic should be universal.

Can we avoid creating a separate function for PipelineSpec validation just for the dev run? Maybe delete this function for now, or use an existing validation function if one exists.

go/components/pipelinerun/actors/executeworkflow.go
npuppala111: Ah okay makes sense!

andriika (approved):
Thank you @npuppala111, LGTM!

PR #390, opened 2025-09-13
Fix apihander factory Fx for controllerMgr

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix the fx ingestion for controlelrMgr calling ApiFactory

**Why?**
Bugfix

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


go/api/handler/builder.go
```diff
@@ -188,3 +188,77 @@ func NewCtrlManagerHandler(params Params) (api.Handler, error) {
...
+		builder = builder.WithBlobStorage(params.BlobStorage)
+	}
+
+	return builder.Build()
+}
+
+// Factory interface that controllers can use to create handlers with different K8s clients
+type Factory interface {
+	GetAPIHandler(client ctrlRTClient.Client) (api.Handler, error)
+}
```
dragod812, line 222: Should the factory be in a separate file ? 
We have named the file builder.go, so should this only have builders ?

go/api/handler/builder.go
```diff
@@ -188,3 +188,77 @@ func NewCtrlManagerHandler(params Params) (api.Handler, error) {
...
+
+// builderFactory implements Factory using the builder pattern internally
+type builderFactory struct {
```
dragod812: builderFactory name is misleading, means this is a factory for builders, but thats not the case. We should name it apiHandlerFactory or apiHandlerFactoryImpl.

go/api/handler/builder.go
```diff
@@ -188,3 +188,77 @@ func NewCtrlManagerHandler(params Params) (api.Handler, error) {
 
 	return builder.Build()
 }
+
+
+// newK8sAndMetadataStorageFactory creates an API handler with both K8s and metadata storage enabled.
+// This function maintains compatibility with the old factory pattern while using the builder internally.
+func newK8sAndMetadataStorageFactory(params Params) (api.Handler, error) {
```
dragod812, line 194: What is using this other than tests ?

dragod812 (commented):
We have a builder that helps in building APIHandler. 
And a factory that returns the correct implementation of the APIHandler. 
I think the Builder interface should be private to the go/api/handler module. 
And we use APIHandlerFactory outside the handler module anywhere needed.

go/api/handler/builder.go
badcount: this is moved from handler to here for code clean up

go/api/handler/builder.go
badcount: no one, just keep it the same as before

go/api/handler/builder.go
badcount: updated to apiHandlerFactory

go/api/handler/builder.go
dragod812: Shouldn't this be in test file then ? Its just test code.

go/api/handler/builder.go
dragod812: i know but this is not builder code, should this be in a separate file called factory.go ?

dragod812 (approved):

PR #408, opened 2025-09-17
PipelineRun "Continued Reconcile Re-queue after Terminal State" Bug Fix

npuppala111:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Added terminal condition check to execute workflow actor so that pipeline runs that have reached a terminal state are no longer re-queued for reconcilation.

**Why?**
This change protects the pipeline run controller reconciled queue from being overloaded with terminated pipeline runs.

**How did you test it?**
I have verified this change by running pipeline runs end to end, and checking condition engine and pipeline run controller logs.

**Potential risks**
Pipeline run controller can be impacted and no execute pipeline runs properly.

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/components/pipelinerun/actors/executeworkflow.go
+++ b/go/components/pipelinerun/actors/executeworkflow.go
@@ -60,6 +60,12 @@ func (a *ExecuteWorkflowActor) Run(ctx context.Context, pipelineRun *v2.Pipeline
 		}, nil
 	}
 
+	if previousCondition.Status != apipb.CONDITION_STATUS_UNKNOWN {
+		// the previous condition is terminal, so we don't need to run the actor again
+		logger.Info("pipeline run has a terminal condition, skipping")
+		return previousCondition, nil
+	}
+
 	executeWorkflowStep := pipelinerunutils.GetStep(pipelineRun, pipelinerunutils.ExecuteWorkflowStepName)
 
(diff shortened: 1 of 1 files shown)


go/components/pipelinerun/actors/executeworkflow.go
```diff
@@ -60,6 +60,12 @@ func (a *ExecuteWorkflowActor) Run(ctx context.Context, pipelineRun *v2.Pipeline
 		}, nil
 	}
 
+	if previousCondition.Status != apipb.CONDITION_STATUS_UNKNOWN {
```
uzbq, line 63: How can this resolve the issue of requeueing a terminaited CRD?

npuppala111: So every other actor has a terminal condition check which allows the actor to skip, but I believe execute workflow was running continuously since it was never checking for a terminal condition.

uzbq: Skipping is just make a no-op operation, but the CRD should not re-queue at all from beginning.  

Is the issue mentioned here  the reason why a terminated CRD keep being re-queued?

uzbq (approved):
talked offline. The above question is a separate issue. But the fix in this PR is valid in general.

PR #412, opened 2025-09-18
Replace build-time environment config with runtime injection

craigmarker:
Migrate RPC service initialization from build-time localhost:8081 hardcoding to runtime config.json fetching. Convert synchronous service creation to async initialization with singleton pattern.

Previous approach hardcoded API endpoints at build time, preventing operators from configuring different envoy proxy endpoints without rebuilding Docker images. This blocked flexible Kubernetes deployments where the same image needs to route to different backend services.

New runtime configuration fetches /config.json on first RPC call, enabling ConfigMap-based endpoint configuration while preserving localhost:8081 behavior for local development. Async refactoring aligns request() function implementation with existing usage patterns in use-studio-query hook, which was already calling request with await.

Singleton caching ensures configuration is fetched once per session with comprehensive error handling providing diagnostic messages for deployment troubleshooting.

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**
Docker UI configured with localhost:8081 API URL
docker-local-fetch

docker-local-8081

Docker UI configured with envoy:8081 API URL (fails to query, since envoy is listening on localhost:8081)
envoy-8081-fetch

envoy-8081

Local dev (yarn dev) finds localhost:8081 in app/public/config.json. Maintains existing behavior
local-8081

**Potential risks**
For local development, developers leveraging michelangelo-ui docker could observe empty data due to inability for michelangelo-ui to access michelangelo-apiserver

For local development, developers leveraging vite dev mode could observer empty data due to inability for vite application to access michelangelo-apiserver

For production deployments, e.g. Grab, users could observe empty data due to inability for michelangelo-ui to access michelangelo-apiserver

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/rpc/runtime-config.ts
+++ b/javascript/packages/rpc/runtime-config.ts
@@ -1,0 +1,46 @@
+interface RuntimeConfig {
+  apiBaseUrl: string;
+}
+
+// Fetches runtime configuration from /config.json.
+export async function getApiConfig(): Promise<string> {
+  let response: Response;
+  try {
+    response = await fetch('/config.json');
+  } catch (error) {
+    if (error instanceof TypeError && error.message.includes('fetch')) {
+      console.error(`Config network error: ${error.message}`);
+      throw createConfigError('Check network connectivity.', { cause: error });
+    }
+    console.error(
+      `Config fetch error: ${error instanceof Error ? error.message : 'Unknown error'}`
+    );
+    throw createConfigError('Check that config.json is properly mounted.', { cause: error });
+  }
+
+  if (!response.ok) {
+    console.error(`Config fetch failed: ${response.status} ${response.statusText}`);
+    throw createConfigError('Check that config.json is properly mounted.', { cause: response });
+  }
+
+  let config: RuntimeConfig;
+  try {
+    config = (await response.json()) as RuntimeConfig;
+  } catch (error) {
+    console.error(
+      `Config JSON parsing failed: ${error instanceof Error ? error.message : 'Invalid JSON'}`
+    );
+    throw createConfigError('Check that config.json contains valid JSON.', { cause: error });
+  }
+
+  if (!config.apiBaseUrl) {
+    console.error('Config missing apiBaseUrl field', JSON.stringify(config));
+    throw createConfigError('Check that config.json contains apiBaseUrl field.');
+  }
+
+  return config.apiBaseUrl;
+}
+
+function createConfigError(reason: string, options?: ErrorOptions): Error {
+  return new Error(`Failed to load runtime configuration. ${reason}`, options);
+}
--- a/javascript/packages/rpc/runtime-config.ts
+++ b/javascript/packages/rpc/runtime-config.ts
@@ -3,7 +3,7 @@ interface RuntimeConfig {
 }
 
 // Fetches runtime configuration from /config.json.
-export async function getApiConfig(): Promise<string> {
+export async function getRuntimeConfig(): Promise<RuntimeConfig> {
   let response: Response;
   try {
     response = await fetch('/config.json');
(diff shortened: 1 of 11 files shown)


badcount (approved):

javascript/packages/rpc/runtime-config.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+    throw createConfigError('Check that config.json contains valid JSON.', { cause: error });
+  }
+
+  if (!config.apiBaseUrl) {
+    console.error('Config missing apiBaseUrl field', JSON.stringify(config));
+    throw createConfigError('Check that config.json contains apiBaseUrl field.');
+  }
+
+  return config.apiBaseUrl;
```
oleksii-trekhleb: - I see that the function is called `getApiConfig()`; however, it returns just one `apiBaseUrl` field from the config. I was wondering if it would be better to either rename the function to `getApiBaseUrl` or to return the full config?

- Do we plan to expose any "private" information in this config (that we don't want the users to see)?

python/michelangelo/cli/sandbox/resources/michelangelo-ui.yaml
```diff
@@ -20,13 +30,21 @@ spec:
         imagePullPolicy: Always
         ports:
         - containerPort: 80
+        volumeMounts:
```
oleksii-trekhleb, line 33: I'm not fully aware of what kind of `yaml` configuration this is, but I was wondering if it supports passing the config via environment variables (similarly to what Docker Compose [does](https://docs.docker.com/compose/how-tos/environment-variables/set-environment-variables/)). Do you by any chance know?

What I'm thinking about is if it is possible to pass the host and the port via ENV config while starting the app, which _might_ be a more secure way to pass the sensitive data (if we will have one). Something like:

```
   environment:
        api_base_host: localhost
        api_base_port: 8081
``` 

With the Docker example above, I believe it doesn't require Docker containers to be rebuilt. And this approach should make it harder to expose the config publicly by mistake (i.e. under the <domain>/config.js path).

Do you think it makes sense to try this direction?

craigmarker: I'm still getting familiar with this stuff, so thank you for providing feedback here!

My understanding is that the environment variable solution doesn't work because vite only injects environment variables at build time.

The only workaround I've seen is string replacement run via script in Dockerfile, e.g., https://stackoverflow.com/a/77454537

craigmarker: Upon further consideration, I actually think this approach makes more sense. It's significantly simpler, albeit less explicit.

Let me experiment a bit.

oleksii-trekhleb (approved):

javascript/packages/rpc/runtime-config.ts
craigmarker: I updated the function to be called getRuntimeConfig; returns the entire configuration. The Kubernetes configuration has been renamed from ui-config to public-config to indicate that the configuration is publicly exposed.

python/michelangelo/cli/sandbox/resources/michelangelo-ui.yaml
craigmarker: I implemented the environment variable string substitution solution in https://github.com/michelangelo-ai/michelangelo/pull/416, and we decided this is the preferred solution.

1. There is no Node.js part of our web application
2. vite does not expose an ability to inject environment variables at runtime
3. String substitution is tricky
4. We believe we can update this approach to be explicit that config is public

oleksii-trekhleb (approved):

PR #451, opened 2025-10-12
Compute Cluster Controller for Michelangelo Jobs 

dragod812:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Added compute cluster controller and cache:
  - New `jobs/cluster` controller (`Reconciler`) that watches `v2pb.Cluster`, adds a deletion finalizer, and maintains a thread-safe in-memory cache of registered clusters.
  - Leader-only periodic health checks update cluster `Status`; non-leader nodes still reconcile reads.
  - Exposed `RegisteredClustersCache` interface with implementations on the controller to list/get clusters by readiness.
- Introduced controller utilities:
  - `controllerutil.NonLeaderReconciler` and `LeaderOnlyRunnable` wrappers to control leader election behavior per task.
  - `controllerutil.SafeReconciler` to recover from panics and improve resiliency.
- Tests and build definitions:
  - Added unit tests for cluster controller and types.
  - Added mocks (`handlermocks`, `clientmocks`).

**Why?**
- Provide a centralized, read-only, resilient controller to track registered compute clusters and surface health for scheduling/placement decisions.
- Improve reliability and operability via safe reconciler recovery and explicit leader/non-leader execution semantics.

**How did you test it?**
- Added unit tests for `jobs/cluster` controller and types.
- Verified reconciliation locally by creating compute cluster CRD for the jobs cluster in sandbox.py. The controller will be attached to controller manager service in a separate PR. 

**Potential risks**
- NA, cluster controller is not being used yet, will be used by the change in ray cluster controller 

**Release notes**
- Add compute cluster controller with registered-clusters cache and leader-only health monitoring.
- Add controller utilities for safe reconciliation and leader/non-leader runnables.
- No schema or data migrations required; no user-facing API changes.

**Documentation Changes**
- NA, cluster controller is not being used yet, will be used by the change in ray cluster controller


go/components/controllerutil/reconciler.go
```diff
@@ -0,0 +1,99 @@
+package controllerutil
```
dragod812, line 1: @bhjoshi1 unchanged.

go/components/jobs/cluster/controller.go
```diff
@@ -0,0 +1,330 @@
+package cluster
```
dragod812, line 1: @bhjoshi1 holds core logic for cluster controller, modified to remove the resource pool concept.

dragod812: Later when we finalize on the resource pool alternative / design in OSS, I can introduce that.

go/components/jobs/cluster/types.go
```diff
@@ -0,0 +1,142 @@
+package cluster
```
dragod812, line 1: @bhjoshi1 unchanged

go/components/jobs/cluster/controller.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %w", err)
+	}
+
+	// Watch for changes to the cluster resource on all nodes
+	cont := controllerutil.NonLeaderReconciler{Controller: controller}
+	err = cont.Watch(source.Kind(mgr.GetCache(), &v2pb.Cluster{}), &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return fmt.Errorf("failed to watch clusters: %w", err)
+	}
```
dragod812: @bhjoshi1 do you know why we need this ? 
Why don't we do the standard way of registering the controller - 
```
func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {

	return ctrl.NewControllerManagedBy(mgr).
		For(&v2beta1.Cluster{}).
		WithEventFilter(predicate.GenerationChangedPredicate{}).
		Complete(ctypes.MakeSafeReconciler(r, "cluster"))
}
```

go/components/controllerutil/reconciler.go
```diff
@@ -0,0 +1,99 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NonLeaderReconciler is a controller that does not need leader election.
+// This is a temporary solution until we upgrade to controller-runtime v0.14.X.
+// https://github.com/kubernetes-sigs/controller-runtime/pull/2117
+type NonLeaderReconciler struct {
+	controller.Controller
+}
```
bhjoshi1: Why are we introducing this complexity when OSS should ideally run on latest version and has this built-in?

bhjoshi1: It was required internally for a different purpose that does not exist in OSS. I'd suggest not taking this on.

go/components/jobs/cluster/controller.go
bhjoshi1: It's because of the non-leader thing that I pointed out above. I'd suggest dropping the whole thing and simplifying.

go/components/controllerutil/reconciler.go
dragod812: I see, will remove this and introduce the standard way of running this. Just wanted to confirm that we don't need this.

go/components/jobs/cluster/controller.go
dragod812: Sure.

go/components/controllerutil/reconciler.go
```diff
@@ -0,0 +1,35 @@
+package controllerutil
+
+import (
+	"context"
+)
+
+// LeaderOnlyRunnable is a Runnable that should only run on the leader node.
+// It's very important that the given function block
```
bhjoshi1, line 8: You don't need all this business. Follow the parent of this commit to go to what we want https://sg.uberinternal.com/code.uber.internal/uber-code/go-code/-/commit/1d0de3362df333acfce434031a464b3daa186b16?first=7 
This is more idiomatic and also consistent with other controllers in the OSS code base.

go/components/jobs/cluster/controller.go
bhjoshi1: Has a discussion with @mincai and we should introduce some concept that represents the concept of resource level abstraction.

dragod812: It is planned next, right now the aim is to introduce an interface that can be extended / implemented to execute complex scheduling logic.

go/components/jobs/cluster/controller.go
```diff
@@ -0,0 +1,312 @@
...
+	r.log = mgr.GetLogger().
+		WithName(_controllerName)
+	apiHandler, err := r.apiHandlerFactory.GetAPIHandler(mgr.GetClient())
+	if err != nil {
+		return err
+	}
+	r.Handler = apiHandler
+
+	// Update cluster health status only on the leader node
+	clusterStatusRunnable := controllerutil.LeaderOnlyRunnable(r.periodicallyMonitorCluster)
```
bhjoshi1, line 236: I looked at this more. The `Add` idea is good for it to be tracked by the manager. But doesn't look like you need an interface for it. See https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#example-Manager-Add

bhjoshi1: Talked offline. We need the interface for the Leader only runnable. There doesn't seem to be a `manager.LeaderRunnableFunc` similar to existing `manager.RunnableFunc`.

bhjoshi1 (approved):

PR #454, opened 2025-10-14
Added Ray Cluster Creation Timeout

npuppala111:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Added Ray cluster creation timeout to ensure that pipeline run does not get stuck if cluster creation has failed.

**Why?**
This bug fix is important for avoiding the pipeline run controller to be stuck when ray cluster creation is not successful and keeps retrying.

**How did you test it?**
I have tested this with end-to-end pipeline runs. Previously, when the incorrect image has been passed in, the execute workflow appears to be stuck. However, after adding the timeout, the execute workflow step now correctly fails after the timeout duration.

**Potential risks**
If these changes are incorrect, future pipeline runs can be affected.

**Release notes**

**Documentation Changes**


uzbq (approved):

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -292,7 +292,7 @@ def execute_ray_task(task_path, task_name, cluster, cluster_namespace, runtime_e
         retry_attempt_id = retry_attempt_id,
     )
 
-    cluster = ray.create_cluster(cluster)
+    cluster = ray.create_cluster(cluster, timeout_seconds = CREATE_CLUSTER_TIMEOUT_SECONDS)
```
badcount: Let's add DEFAULT to the name to make it consistent

PR #460, opened 2025-10-16
NEW FEATURE: Added `mactl triger_run kill` command

yaoyueyue:
- [ ] Refactor
- [*] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

  **What changed?**
  Added `mactl triger_run kill` command. Implemented generic kill functionality that sets `spec.kill=true` via gRPC Update
  method, enabling clean Cadence workflow termination.

  
  **Why?**
Users need a clean way to  terminate triggers without leaving orphaned processes. This provides the missing lifecycle management piece.

  
  **How did you test it?**
Screenshot 2025-10-16 at 2 04 28 PM

  - Created trigger → killed with mactl → verified state change to killed
  - Tested edge cases: killing non-existent resources, already killed resources

  
  **Potential risks**
  Low risk - purely additive feature. Only adds new "kill" action to existing mactl commands. Uses existing gRPC Update infrastructure. 

  
  **Release notes**
  **NEW FEATURE**: Added `mactl triger_run kill` command
  - `mactl trigger_run kill --namespace=<ns> --name=<name>`
  - Clean Cadence workflow termination without orphaned processes

### DIFF
--- a/python/michelangelo/cli/mactl/mactl.py
+++ b/python/michelangelo/cli/mactl/mactl.py
@@ -419,7 +419,9 @@ class CRD:
             _LOG.info("Retrieved resource for kill: %r", current_resource)
 
             # Create update request with kill flag set
-            current_dict = MessageToDict(current_resource, preserving_proto_field_name=True)
+            current_dict = MessageToDict(
+                current_resource, preserving_proto_field_name=True
+            )
 
             # Set kill flag in spec
             resource_name = _self.name
--- a/python/michelangelo/cli/mactl/mactl.py
+++ b/python/michelangelo/cli/mactl/mactl.py
@@ -386,6 +386,80 @@ class CRD:
         _LOG.info("Generate DEV RUN method for %r / %r", self.name, self.full_name)
         pass
 
+    def generate_kill(self, channel: Channel):
+        """
+        Generate kill function of this class.
+        """
+        # Ensure get method is available for retrieving current resource
+        self.generate_get(channel)
+
+        _LOG.info("Generate KILL method for %r / %r", self.name, self.full_name)
+        method_name, input_class, output_class = self._extract_method_info(
+            channel, self.full_name, "Update"
+        )
+
+        kill_func_signature = Signature(
+            [Parameter("self", Parameter.POSITIONAL_OR_KEYWORD)]
+            + [
+                Parameter(name, Parameter.POSITIONAL_OR_KEYWORD)
+                for name in ["namespace", "name"]
+            ]
+        )
+
+        @bind_signature(kill_func_signature)
+        def kill_func(bound_args: Signature) -> Message:
+            _LOG.info("Start kill_func for %r", self.full_name)
+            _LOG.info("Bound arguments: %r", bound_args.arguments)
+            _self: CRD = bound_args.arguments["self"]
+            _name = get_single_arg(bound_args.arguments, "name")
+            _namespace = get_single_arg(bound_args.arguments, "namespace")
+
+            # Get current resource
+            current_resource = _self.get(_namespace, _name)
+            _LOG.info("Retrieved resource for kill: %r", current_resource)
+
+            # Create update request with kill flag set
+            current_dict = MessageToDict(current_resource, preserving_proto_field_name=True)
+
+            # Set kill flag in spec
+            resource_name = _self.name
+            if resource_name in current_dict and "spec" in current_dict[resource_name]:
+                current_dict[resource_name]["spec"]["kill"] = True
+            else:
+                _LOG.error("Invalid resource structure for kill operation")
+                raise ValueError(f"Cannot set kill flag on {resource_name}")
+
+            # Convert back to protobuf message for update
+            request_input = input_class()
+            ParseDict(current_dict, request_input, ignore_unknown_fields=True)
+
+            _LOG.info(
+                "KILL Request input (%r) ready: %r",
+                type(request_input),
+                request_input,
+            )
+
+            method_fullname = f"/{_self.full_name}/{method_name}"
+            _LOG.info("Method fullname for gRPC call: %s", method_fullname)
+
+            stub_method = channel.unary_unary(
+                method_fullname,
+                request_serializer=input_class.SerializeToString,
+                response_deserializer=output_class.FromString,
+            )
+
+            response = stub_method(
+                request_input,
+                metadata=METADATA_STUB,
+                timeout=30,
+            )
+
+            _LOG.info("Kill operation completed (%r): %r", type(response), response)
+            return response
+
+        kill_func.__signature__ = kill_func_signature  # type: ignore[attr-defined]
+        self.kill = MethodType(kill_func, self)
+
     def generate_list(self, channel: Channel):
         """
         Generate list function of this class.
@@ -819,6 +893,7 @@ def handle_args() -> tuple[str, str, dict[str, list[str]]]:
         "list",
         "run",
         "dev-run",
+        "kill",
     ]
 
     # For file-based actions, validate file parameter exists (preserving original validation)
--- a/python/michelangelo/cli/mactl/mactl.py
+++ b/python/michelangelo/cli/mactl/mactl.py
@@ -519,7 +519,9 @@ def get_message_class_by_name(pool: DescriptorPool, message_name: str) -> type[M
     return MessageClass
 
 
-def read_plugins(crd: CRD, user_command_action: str, crds: dict[str, CRD], channel: Channel) -> None:
+def read_plugins(
+    crd: CRD, user_command_action: str, crds: dict[str, CRD], channel: Channel
+) -> None:
     """
     Read and apply plugins for a given crd.
     """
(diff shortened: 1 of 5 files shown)


python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
...
+                request_serializer=input_class.SerializeToString,
+                response_deserializer=output_class.FromString,
+            )
+
+            response = stub_method(
+                request_input,
+                metadata=METADATA_STUB,
+                timeout=30,
+            )
+
```
yangfan0328: We should do assert "kill operation" is successful instead of just checking response, here are 2 options you may proceed with:
option 1. check response spec if spec.kill = true is set correctly
option 2. run another get pipeline run request after some seconds and check if spec.kill = true is set correctly

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
...
+            _self: CRD = bound_args.arguments["self"]
+            _name = get_single_arg(bound_args.arguments, "name")
+            _namespace = get_single_arg(bound_args.arguments, "namespace")
+
+            # Get current resource
+            current_resource = _self.get(_namespace, _name)
+            _LOG.info("Retrieved resource for kill: %r", current_resource)
+
+            # Create update request with kill flag set
+            current_dict = MessageToDict(
```
yangfan0328: Can you create a "generate_update" function and move this part of logic there? So we could share and reuse update request code. Make the dict as function input and call that function here

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
...
+        kill_func_signature = Signature(
+            [Parameter("self", Parameter.POSITIONAL_OR_KEYWORD)]
+            + [
+                Parameter(name, Parameter.POSITIONAL_OR_KEYWORD)
+                for name in ["namespace", "name"]
+            ]
+        )
+
+        @bind_signature(kill_func_signature)
+        def kill_func(bound_args: Signature) -> Message:
```
yangfan0328: Are we able to add retry logic for kill func? Since grpc network call failures are retryable error. (not sure about other methods but it worths to take a look generally)

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
         _LOG.info("Generate DEV RUN method for %r / %r", self.name, self.full_name)
         pass
 
+    def generate_kill(self, channel: Channel):
+        """
+        Generate kill function of this class.
+        """
+        # Ensure get method is available for retrieving current resource
```
yangfan0328: Can we pop up some warning messages to user -> "Please confirm to kill trigger run/pipeline run {namespace.name}. Are you sure to go through this operation?" to avoid unattended kill operation.

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
         _LOG.info("Generate DEV RUN method for %r / %r", self.name, self.full_name)
         pass
 
+    def generate_kill(self, channel: Channel):
```
badcount, line 389: Please add details comments to the method and give enough examples, use chatGPT for help

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,82 @@ def generate_dev_run(self, channel: Channel):
...
+            current_dict = MessageToDict(
+                current_resource, preserving_proto_field_name=True
+            )
+
+            # Set kill flag in spec
+            resource_name = _self.name
+            if resource_name in current_dict and "spec" in current_dict[resource_name]:
+                current_dict[resource_name]["spec"]["kill"] = True
+            else:
+                _LOG.error("Invalid resource structure for kill operation")
```
badcount: why not saying "missing required spec field in the resource structure"

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -819,6 +895,7 @@ def handle_args() -> tuple[str, str, dict[str, list[str]]]:
         "list",
         "run",
         "dev-run",
+        "kill",
```
badcount, line 836: can you add comments to all the actions, like:
SUPPORTED_ACTIONS = [
      "get",      # Retrieve a resource
      "create",   # Create a new resource from file
      "apply",    # Update existing resource from file  
      "delete",   # Delete a resource
      "list",     # List resources in namespace
      "run",      # Execute a pipeline/trigger
      "dev-run",  # Execute in development mode
      "kill",     # Terminate a running resource
  ]

hwaminkim: @badcount this list will be removed in the future and CRD instance will substitute those information with argparse library. This list was mainly used when we were using verb-noun way

python/michelangelo/cli/mactl/mactl.py
hwaminkim, line 389: @yaoyueyue kill command is only for the pipeline, so it should be delivered under pipeline plugin

hwaminkim (changes requested):

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Good point about retry logic. I checked other mactl methods use the same gRPC call pattern without retry logic. The codebase already has gRPC-level retry configuration in base.py, but mactl uses basic insecure_channel() without it - this is an architectural decision that affects all operations. If we really want retry logic, maybe address it as a separate PR?

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Thanks for the suggestion. While that would be a good abstraction, in my understanding the kill operation has specific validation logic and logging error handling. Is there a specific scenario you had in mind where we'll to reuse this update pattern for other operations?

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Thanks for the suggestion! I thought the kill command remain in the base mactl.py since it's a generic operation that works identically for both trigger_run and pipeline_run (just sets spec.kill=true), following the same pattern as other base operations. Is there a specific situation you had in mind where pipeline-specific kill logic would be needed?

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: done

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: done

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Good point, added

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Good point, added

python/michelangelo/cli/mactl/mactl.py
hwaminkim: generic operations should be applicable for any types of CRD, so we only support 3 commands for now (GET, APPLY, DELETE). Though some CRDs are sharing almost same features, if it's not available for other CRDs, I think it should be moved under plugins.
Of course, the common utility functions can be shared.

badcount (approved):

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -386,6 +386,14 @@ def generate_dev_run(self, channel: Channel):
         _LOG.info("Generate DEV RUN method for %r / %r", self.name, self.full_name)
         pass
 
+    def generate_kill(self, channel: Channel):
```
hwaminkim, line 389: @yaoyueyue do we need this function to generate kill function? I think we inject this function in manual.

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -870,7 +885,7 @@ def main(channel: Channel):
         crds[user_command_crd], f"generate_{user_command_action}"
     )
     func_action_generator(channel)
-    read_plugins(crds[user_command_crd], user_command_action, crds)
+    read_plugins(crds[user_command_crd], user_command_action, crds, channel)
```
hwaminkim, line 888: Good change to pass channel argument in explicit way

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: > @yaoyueyue do we need this function to generate kill function? I think we inject this function in manual.

Yes, we need generate_kill() because line 880 asserts its existence. This follows the same pattern as generate_run() and generate_dev_run() which are also plugin-delegated actions.

hwaminkim: Synced with @yaoyueyue in offline -- since there are `assert` checking before the plugin command is generated. it would cause an error. we'll fix this issue in the separated PR.

hwaminkim (approved):

yangfan0328 (approved):

PR #464, opened 2025-10-21
Compute Cluster batch job scheduler with a simple “cluster-only” assignment strategy

dragod812:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Implemented initial compute-cluster batch job scheduler with a simple “cluster-only” assignment strategy.
- Added `ClusterOnlyAssignmentStrategy` and wired it into the scheduler framework for batch jobs.
- Introduced basic scheduling utilities and types for cluster selection.
- Added scheduler metrics scaffolding and instrumentation.
- Wrote targeted unit tests for the assignment engine and scheduler integration; added mocks for cluster interfaces.

**Why?**
- Establish an MVP scheduling path for compute-cluster batch jobs to enable safe, incremental rollout.
- Improve observability for scheduling behavior via metrics.

**How did you test it?**
- Added unit tests for the cluster-only assignment engine and scheduler integration (e.g., `cluster_only_assignment_engine_test.go`, `scheduler_test.go`).

**Potential risks**
- NA, will be tested in the next PR

**Release notes**
- Introduces initial compute-cluster batch job scheduler (cluster-scoped assignment). No schema or config migrations required.

**Documentation Changes**
- No API changes. Add a short wiki section describing the cluster-only batch scheduling behavior and exposed metrics after merge.


go/components/jobs/common/constants/constants.go
```diff
@@ -128,6 +128,14 @@ const (
 	ResourceNvidiaGPU corev1.ResourceName = "nvidia.com/gpu"
 )
 
+// Scheduler default resource SKU constants (OSS)
+const (
+	// DefaultCPU is the default resource SKU key for CPU-only jobs
+	DefaultCPU string = "defaultCPU"
+	// DefaultGPU is a placeholder SKU key for unlabelled GPU jobs (OSS)
```
bhjoshi1: why do we need this? And not the real GPU sku name?

go/components/jobs/scheduler/framework/cluster_only_assignment_engine.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+
+	"github.com/michelangelo-ai/michelangelo/go/components/jobs/cluster"
+	v2pb "github.com/michelangelo-ai/michelangelo/proto/api/v2"
+)
+
+var _ AssignmentEngine = ClusterOnlyEngine{}
+
+// ClusterOnlyEngine selects a cluster using simple affinity signals.
```
bhjoshi1: nit: "using affinity" suffices?

go/components/jobs/scheduler/framework/cluster_only_assignment_engine.go
```diff
@@ -0,0 +1,44 @@
...
+// It never assigns a resource pool.
```
bhjoshi1: Have we defined the concept of "resource pool" somewhere? If not, why refer it here?

go/components/jobs/scheduler/framework/interface.go
```diff
@@ -0,0 +1,15 @@
...
+package framework
+
+import (
+	"context"
+
+	v2pb "github.com/michelangelo-ai/michelangelo/proto/api/v2"
+)
+
+// AssignmentEngine decides assignment for a job.
+type AssignmentEngine interface {
```
bhjoshi1: Why do we call it `AssignmentEngine` and not `SchedulerQueue` or `JobQueue`? The queue model of scheduling is quite standard and popular. Used in both Kubernetes and Peloton.
I remember we discussed this before.

go/components/jobs/scheduler/framework/cluster_only_assignment_engine.go
```diff
@@ -0,0 +1,44 @@
...
+// Select implements Engine.
+func (e ClusterOnlyEngine) Select(_ context.Context, job BatchJob) (*v2pb.AssignmentInfo, bool, string, error) {
+	// For OSS MVP: choose the first available cluster, or a specific one
+	// if job affinity provides an explicit cluster name via resource selector label
+	// "resourcepool.michelangelo/cluster".
+
+	// Prefer explicit cluster by label, else first available.
+	selector := job.GetAffinity().GetResourceAffinity().GetSelector()
+	if selector != nil && selector.MatchLabels != nil {
+		if name, ok := selector.MatchLabels["resourcepool.michelangelo/cluster"]; ok && name != "" {
```
bhjoshi1: Can we define all affinity labels and constants?

go/components/jobs/scheduler/framework/interface.go
bhjoshi1: Also why is it call "Engine" if it does not run anything of it's own?

go/components/jobs/scheduler/framework/job.go
```diff
@@ -0,0 +1,366 @@
...
+// If a resource is not requested, it falls back to limits for that resource.
+func getResourceRequestsFromPodSpec(podSpec *v1.PodSpec) v1.ResourceList {
+	// Start with an empty map; only include resources that are explicitly requested/limited
+	summed := make(v1.ResourceList)
+
+	for _, c := range podSpec.Containers {
+		// Prefer requests; fallback to limits per resource
+		// CPU
+		if qty, ok := c.Resources.Requests[v1.ResourceCPU]; ok {
+			if existing, ok2 := summed[v1.ResourceCPU]; ok2 {
```
bhjoshi1: Instead of this, can you initialize with a `zeroList` and then always add? Like how you would do for a regular int variable? And uplevel it for `ResourceList`

go/components/jobs/scheduler/framework/job.go
```diff
@@ -0,0 +1,366 @@
...
+				summed[v1.ResourceEphemeralStorage] = q
+			} else {
+				summed[v1.ResourceEphemeralStorage] = qty.DeepCopy()
+			}
+		}
+	}
+	return summed
+}
+
+func getBucketKeyFromPodSpec(_ *v1.PodSpec) string {
```
bhjoshi1: Not sure I follow this concept. Are you taking inspiration from FedV2 pools? Given there are no pools in this model, I'm not sure I understand the use for this.

go/components/jobs/scheduler/scheduler.go
```diff
@@ -0,0 +1,406 @@
...
+	v2pb "github.com/michelangelo-ai/michelangelo/proto/api/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// JobQueue is the job queue for the scheduler.
+type JobQueue interface {
```
bhjoshi1, line 38: Not following why both this interface and `AssignmentEngine` exist

go/components/libs/goroutine/goroutine.go
```diff
@@ -0,0 +1,51 @@
+package goroutine
+
+import "sync"
+
+// RoutineExecutor executes the given routine
+type RoutineExecutor interface {
+	Execute(routine func())
+}
```
bhjoshi1: As with before, I'm reluctant to add interfaces that do not back any business logic. These things are hard to maintain and not necessarily the best implementation. Why don't we instead use the WaitGroup inline where required?

go/components/jobs/common/constants/constants.go
dragod812: Don't need it right now, this is a placeholder. We don’t ship GPUs in OSS yet and don’t have SKU/ResourcePool management wired up or testable. Hard-coding a “real” SKU would leak internal names and add config we can’t validate right now. This key simply lets us bucket GPU requests without committing to a SKU taxonomy. Once SKU/ResourcePool management lands, we’ll replace this with the real SKU mapping and make it configurable. I’ll add a TODO here and track it.

go/components/jobs/scheduler/framework/cluster_only_assignment_engine.go
dragod812: Yess, added a single label for now in jobs/common/constants.

go/components/jobs/scheduler/framework/interface.go
dragod812: Because it implements only the placement/assignment policy (deciding a Cluster for a job), not queueing. We already have a queue: the externally exposed JobQueue interface for producers to enqueue into the scheduler. Naming this component “Queue” would be misleading.

dragod812: We separate submission from placement so each can evolve independently.
1. Custom scheduler (queue + placement): Just implement the JobQueue interface, Implementing AssignmentStrategy is not necessary
2. Custom placement only: implement AssignmentStrategy; keep existing enqueueing and queue. The Current scheduler will get a different AssignmentStrategy implementation through DI.

go/components/jobs/scheduler/scheduler.go
dragod812: Responded with details in go/components/jobs/scheduler/framework/cluster_only_assignment_engine.go. 
Scheduler uses AssignmentStrategy to get the AssignmentInformation. 
I think an interal interface in framework would provide us flexibility. 
Right now in OSS, we only do local testing, once we deploy in GCP / AWS we might have some different implementations of AssignmentStrategy. 
We haven't explored the ResourcePool / Kueue implementation for OSS yet (TBD), but when we do that, I feel this interface would stay the same in that case too.

go/components/libs/goroutine/goroutine.go
dragod812: Makes sense, not needed. Removing it.

go/components/jobs/scheduler/framework/job.go
dragod812: Yess sort of. “Bucket” here is just a placeholder for a resource-class key (e.g., CPU-only, GPU:A100, etc.). Today it’s a no-op: everything maps to DefaultCPU, so it isn’t used for scheduling decisions. We kept the hook so that when we add resource-aware cluster assignment, we can derive this key from node selectors, tolerations, requests/limits, and extended resources. If the term “bucket” is confusing, I can rename it to “resourceClassKey” and add a TODO, or remove it for now and reintroduce it when we wire up resource-aware placement.

go/components/jobs/scheduler/framework/job.go
dragod812: Done, removed duplicated code.

go/components/jobs/scheduler/framework/job.go
```diff
@@ -0,0 +1,331 @@
...
+
+// getResourceRequestsFromPodSpec sums container resource requests from a core k8s PodSpec.
+// If a resource is not requested, it falls back to limits for that resource.
+func getResourceRequestsFromPodSpec(podSpec *v1.PodSpec) v1.ResourceList {
+	// Start with an empty map; only include resources that are explicitly requested/limited
+	summed := v1.ResourceList{
+		v1.ResourceCPU:              resource.MustParse("0"),
+		v1.ResourceMemory:           resource.MustParse("0"),
+		v1.ResourceEphemeralStorage: resource.MustParse("0"),
+		constants.ResourceNvidiaGPU: resource.MustParse("0"),
```
bhjoshi1, line 108: nit: make this a util like `getZeroResourceList`

bhjoshi1 (approved):

PR #475, opened 2025-10-28
File sync on docker container for both ray and spark task

yangfan0328:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Added `--file-sync` feature for Uniflow remote runs, enabling hot code reloading without Docker image rebuilds. When enabled, local Git changes are automatically packaged, uploaded to S3, and applied to remote containers before workflow execution via Python's sitecustomize.py mechanism. This dramatically speeds up development iteration cycles from minutes to seconds.

1. python/michelangelo/uniflow/core/file_sync.py: Creates and uploads tarballs of Git changes (modified + untracked files) to S3 storage, extracting Git SHA from Docker images and stripping python/ prefix for correct container paths.
2. python/michelangelo/uniflow/core/sitecustomize.py: Auto-runs on container startup to download and extract dev files from S3 to /app directory using fsspec, with comprehensive logging and graceful error handling.
3. python/michelangelo/uniflow/core/context.py: Added `--file-sync` CLI argument and passes it to RemoteRun.
4. python/michelangelo/uniflow/core/remote_run.py: Triggers tarball build/upload when file_sync=True and sets DEV_RUN_REMOTE_FILE_PATH + PYTHONPATH environment variables for remote containers.
5. python/tests/uniflow/core/test_file_sync.py: Tests tarball creation, Git SHA extraction from Docker images, S3 uploads, and error handling with mocked Docker/fsspec.
6. python/tests/uniflow/core/test_sitecustomize.py: Tests FsspecDownloader with mocked fsspec for various storage protocols, large files, and error scenarios.
7. python/examples/Dockerfile: Added Git SHA build args/labels and SSL certificate configuration for secure downloads.
8. python/michelangelo/uniflow/core/run_task.py: Added --file-sync flag in parameter argument.

**Why?**

**Problem:**
Developers iterating on ML workflows needed to rebuild Docker images for every code change, leading to 5-10 minute iteration cycles, poor developer experience, and wasted resources.

**Solution:**
The `--file-sync` feature enables hot code reloading by:
- Detecting local changes via Git (modified and untracked files)
- Uploading only changed files as a compressed tarball to S3
- Automatically applying changes in remote containers before workflow execution

**Future Support:**
- Disclaimer on mactl integration: This PR supports on remote run due to user requests, we will integrate it to `mactl pipeline dev-run` with `--file-sync` later to support it with unified api backend.
- Disclaimer on integration with auto image builder: This PR is only applicable to user scenario where there is a manually prebuilt image, and user changes files on local branch with checking local file changes with git command. Once auto image builder is launched, on creating pipeline run stage, we will check if this revision has a prebuilt image through image builder controller API call, then check if user integrates with `--file-sync` flag on kicking off dev-run -> skip auto image building, instead using same image with the corresponding base commit info then including this file change. 

**How did you test it?**
unit tests, e2e tests

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**
N/A


python/michelangelo/uniflow/core/apply_local_changes.py
badcount, line 1: some suggestions for the name:
If your apply_local_changes feature behaves like copying updated files into a pre-built image and restarting relevant processes (no rebuild, near-instant feedback), the most recognizable OSS-aligned names would be:

live_sync – used by Skaffold and Okteto.

hot_reload – widely understood across Python, web, and container tools.

file_sync – neutral and descriptive, matches Kubernetes tooling.

So, to align with open-source terminology, you could rename or describe it as:

“live_sync (apply_local_changes)” — a mechanism to hot-reload local code or configuration into an existing container image without triggering a rebuild.

python/michelangelo/uniflow/core/apply_local_changes.py
```diff
@@ -0,0 +1,229 @@
...
+import os
+import logging
+import subprocess
+import fsspec
+
+log = logging.getLogger(__name__)
+_DEFAULT_BASE_PROJECTS_PATH = "s3://default/uniflow"
+
+
+class UniflowDevRunFileBuilder(ABC):
```
badcount: we still will use this for remote-run right? why only called "dev-run"?

python/michelangelo/uniflow/core/apply_local_changes.py
```diff
@@ -0,0 +1,229 @@
...
+import tempfile
+import tarfile
+import io
+import os
+import logging
+import subprocess
+import fsspec
+
+log = logging.getLogger(__name__)
+_DEFAULT_BASE_PROJECTS_PATH = "s3://default/uniflow"
```
badcount: let's not over define this in many places, Uniflow should have single way to define this.

python/examples/Dockerfile
```diff
@@ -6,6 +6,13 @@
 # Use a base image with Python installed
 FROM nvidia/cuda:12.8.0-base-ubuntu22.04
 
+# Capture Git SHA build argument
+ARG GIT_SHA=unknown
+LABEL git.commit=${GIT_SHA}
+LABEL git.sha=${GIT_SHA}
+LABEL org.opencontainers.image.revision=${GIT_SHA}
+ENV GIT_SHA=${GIT_SHA}
```
andriika: It looks like the container is required to provide the git commit information. Could we consider relaxing this requirement a bit? If the container doesn't have the git commit, maybe we could use the commit from the local environment instead.

andriika (changes requested):

python/michelangelo/uniflow/core/apply_local_changes.py
yangfan0328: Imported from uniflow_tar.py, will create another PR to refactor with one central constant.py file.

python/examples/Dockerfile
yangfan0328: Updated with fallback: if no git commit info in docker file, just check local git repo with updated files.

python/michelangelo/uniflow/core/apply_local_changes.py
yangfan0328: I did some research, and figured out "file sync" is the most broadly used name on OSS community, e.g Skaffold https://skaffold.dev/docs/filesync and Okteto https://www.okteto.com/docs/1.38/development/containers/file-sync. I updated my diff with --file-sync and all related files/functions.

python/michelangelo/uniflow/core/apply_local_changes.py
yangfan0328: Updated with UniflowFileSyncBuilder.

python/michelangelo/uniflow/core/sitecustomize.py
```diff
@@ -0,0 +1,187 @@
...
+
+class FsspecDownloader(StorageDownloader):
+    """Downloader using fsspec for OSS S3-compatible storage."""
+
+    def download(
+        self, remote_path: str, local_path: Path, logger: logging.Logger
+    ) -> bool:
+        """Download using fsspec (works with S3, MinIO, etc)."""
+        try:
+            import fsspec
```
badcount: why do we need this? Uniflow require fsspec

yangfan0328: This is for sharing code between internal and OSS, so that we use download interface with different uber internal storage system.

python/michelangelo/uniflow/core/remote_run.py
```diff
@@ -13,6 +13,7 @@
 from michelangelo.uniflow.core.codec import encoder
 from michelangelo.uniflow.core.build import build
 from michelangelo.uniflow.core.utils import dot_path
+from michelangelo.uniflow.core.file_sync import UniflowFileSyncBuilderOSS
```
andriika: Can we please rename `UniflowFileSyncBuilderOSS` to `FileSyncBuilder`? The terms `Uniflow` and `OSS` seem a bit excessive.

badcount (approved):

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -2,6 +2,25 @@ load("@plugin", "atexit", "json", "os", "ray", "time")
 load("../../commons.star", "DEFAULT_RETRY_ATTEMPTS", "CACHE_OPERATION_GET", "CACHE_OPERATION_PUT", "TASK_STATE_FAILED", "TASK_STATE_KILLED", "TASK_STATE_PENDING", "TASK_STATE_RUNNING", "TASK_STATE_SKIPPED", "TASK_STATE_SUCCEEDED", "TIME_FOMART", "create_cached_output", "get_cache_enabled", "get_cache_keys", "get_cached_output", "get_result_url", "get_task_image", "get_task_name", "io_read_json", "report_progress", "resource_dict", COMMONS_ENV = "ENV")
 
 DEFAULT_CREATE_CLUSTER_TIMEOUT_SECONDS = 60 * 30  # Timeout duration for cluster creation in seconds.
+
+def _get_pythonpath():
```
andriika: Please move this function to the `commons.star` to avoid code duplication in the ray and spark tasks.

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -11,6 +30,7 @@ RAY_ENV = {
     # As a result, a worker node might start significantly earlier than the head node.
     # Calculate RAY_NUM_REDIS_GET_RETRIES to allow workers approximately DEFAULT_CREATE_CLUSTER_TIMEOUT_SECONDS to connect to the GCS, assuming 7 seconds per retry.
     "RAY_NUM_REDIS_GET_RETRIES": str((DEFAULT_CREATE_CLUSTER_TIMEOUT_SECONDS // 7) + 1),
+    "PYTHONPATH": _get_pythonpath(),
```
andriika: It seems that updating the `ENV` in the `commons.star` file would be sufficient, rather than updating the ray and spark environments specifically.

python/michelangelo/uniflow/core/remote_run.py
yangfan0328: https://github.com/michelangelo-ai/michelangelo/pull/475/files#diff-e8b79ae02e3585d70f6c64a4d9ad82d24384e1ddd8c6b5226e356680927702f5R35 It's already updated. That one is outdated code

python/michelangelo/uniflow/plugins/ray/task.star
yangfan0328: moved to commons.star and import back.

python/michelangelo/uniflow/plugins/ray/task.star
yangfan0328: moved to commons.star and import back.

andriika (approved):

PR #481, opened 2025-10-29
[mactl] remove non-default generator in base CRD

hwaminkim:
- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Remove non-default function generator in base CRD.

**Why?**

Now users don't need to change the main mactl file.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/pipeline/apply.py
+++ b/python/michelangelo/cli/mactl/plugins/pipeline/apply.py
@@ -1,24 +1,12 @@
 from copy import deepcopy
 from logging import getLogger
-from os import getenv
 from pathlib import Path
-from uuid import uuid4
 
 from git import Repo
 from google.protobuf.message import Message
-from grpc import Channel
-
-from mactl import CRD, PWD
 
 
 _LOG = getLogger(__name__)
-
-
-def generate_apply(crd: CRD, channel: Channel):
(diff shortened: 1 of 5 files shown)


python/michelangelo/cli/mactl/plugins/pipeline/apply.py
```diff
@@ -35,7 +23,9 @@ def convert_crd_metadata_pipeline_apply(
     repo = Repo(".", search_parent_directories=True)
     _LOG.info("Current git repository info: %r", repo)
 
+    # TODO: update path retrieval logic
     res = {"spec": deepcopy(yaml_dict["spec"])}
+    """
```
yaoyueyue, line 28: Is pipeline apply intentionally disabled? All converter logic is commented out. Will this break current functionality?

hwaminkim: Thanks for checking. I checked that pipeline apply worked fine with latest api server. Most of those fields would be handled by api server.
The only remaining concern is the path field. It would be handled in the next PR

yaoyueyue (approved):

PR #488, opened 2025-11-05
Fix spark tasks for pipeline run

npuppala111:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
  - Modified GetJobStatus() function signature to return an additional errorMessage
  parameter extracted from SparkApplication.status.applicationState.errorMessage
  - Added new setCondition() function that accepts message and reason parameters to
  populate condition fields with detailed error information
  - Updated controller's Reconcile() logic to handle different Spark states (RUNNING,
  COMPLETED, FAILED) with specific messages
  - For FAILED state, the controller now passes the SparkApplication error message to
  the condition instead of using generic defaults
  - Updated interface documentation and test mocks to match new function signatures

**Why?**
Pipeline runs were previously succeeded without the spark task reporting the correct step state. Also on failure, pipeline failures showed generic "unknown reason:Spark job failed" messages, making debugging extremely difficult

**How did you test it?**
Tested with end to end pipeline runs: 
post-spark-fix-fail
post-spark-fix-kill
post-spark-fix-succeed

**Potential risks**

**Release notes**

**Documentation Changes**


go/components/spark/job/client/client.go
```diff
@@ -82,7 +82,7 @@ func (r SparkClient) CreateJob(ctx context.Context, log logr.Logger, job *v2pb.S
 }
 
 // GetJobStatus retrieves the status of the Spark job
-func (r SparkClient) GetJobStatus(ctx context.Context, logger logr.Logger, job *v2pb.SparkJob) (*string, string, error) {
+func (r SparkClient) GetJobStatus(ctx context.Context, logger logr.Logger, job *v2pb.SparkJob) (*string, string, string, error) {
```
uzbq, line 86: can you add some comments to explain what are those return values?

go/components/spark/job/controller.go
```diff
@@ -62,10 +63,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			InvalidatingState     ApplicationStateType = "INVALIDATING"
+			SucceedingState       ApplicationStateType = "SUCCEEDING"
+			FailingState          ApplicationStateType = "FAILING"
+			UnknownState          ApplicationStateType = "UNKNOWN"
+		*/
+		switch *stateStr {
+		case "RUNNING":
+			setCondition(&sparkJob.Status.StatusConditions, "SparkAppRunning", apipb.CONDITION_STATUS_TRUE, "Spark application is running", "Running")
+		case "COMPLETED":
+			setCondition(&sparkJob.Status.StatusConditions, "SparkAppRunning", apipb.CONDITION_STATUS_FALSE, "Spark application completed", "Completed")
```
uzbq: we should not change the running condition after the spark app is completed.

go/components/spark/job/controller.go
```diff
@@ -62,10 +63,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			FailedSubmissionState ApplicationStateType = "SUBMISSION_FAILED"
+			PendingRerunState     ApplicationStateType = "PENDING_RERUN"
+			InvalidatingState     ApplicationStateType = "INVALIDATING"
+			SucceedingState       ApplicationStateType = "SUCCEEDING"
+			FailingState          ApplicationStateType = "FAILING"
+			UnknownState          ApplicationStateType = "UNKNOWN"
+		*/
+		switch *stateStr {
+		case "RUNNING":
+			setCondition(&sparkJob.Status.StatusConditions, "SparkAppRunning", apipb.CONDITION_STATUS_TRUE, "Spark application is running", "Running")
```
uzbq: For strings which is referred by different code pointers, we would better to make them as constants to ensure consistency

go/components/spark/job/controller.go
```diff
@@ -95,12 +125,41 @@ func (r *Reconciler) Register(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// setCondition sets or updates a condition in the status conditions slice.
+// If a condition with the same type already exists, it updates it only if the status changed.
+// Returns true if the condition was added or updated, false if it was already set to the same status.
+func setCondition(conditions *[]*apipb.Condition, conditionType string, status apipb.ConditionStatus, message string, reason string) bool {
```
uzbq, line 130: how ray job controller set condition? can they become a common function?

badcount (approved):

go/components/spark/job/controller.go
npuppala111: @uzbq seems like the ray job controller is setting the job status directly within the reconcile loop, I don't think this function can be shared between the two.

uzbq (approved):

PR #491, opened 2025-11-06
Ray Job Controller changes for Multi Cluster support for Ray Jobs

dragod812:
- [x] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR introduces multi-cluster support for Ray Jobs with the following key changes:

1. **Implemented Ray Job status retrieval**: The `GetJobStatus` method now supports fetching Ray Job statuses from remote Kubernetes clusters via the KubeRay API, which was previously unsupported.
2. **Generalized status interface**: Updated `GetJobStatus` interface to return a JobStatus object with v2pb.RayJobStatus as a field.
3. **Fixed submitter pod RestartPolicy**: Explicitly set the Ray Job submitter pod's `restartPolicy` to `Never` to comply with Kubernetes Job requirements (previously inherited from RayCluster head pod template).

**Why?**
To enable users to deploy and manage Ray Jobs across multiple Kubernetes clusters, enhancing scalability, resilience, and resource utilization. The previous implementation did not support retrieving Ray Job statuses, which was a critical gap for multi-cluster management. The interface refactoring makes the system more flexible and framework-agnostic, supporting both Ray and Spark jobs uniformly. The RestartPolicy fix ensures correct Kubernetes Job semantics and prevents invalid configurations that could cause job failures.

**How did you test it?**
- Added comprehensive unit tests for `GetJobStatus` covering various scenarios including successful status translation, error handling for client set acquisition, and unsupported job types.
- Added test for `GetJobClusterStatus` to ensure proper error handling.
- Added assertion in `TestMapper_MapGlobalToLocal` to verify correct RestartPolicy configuration.
- Tested locally by deploying Ray Jobs to michelangelo-compute cluster created through - `poetry run ma sandbox create --create-compute-cluster`                                                                         

**Potential risks**
- NA

**Release notes**
This PR introduces multi-cluster support for Ray Jobs, a significant new feature. Users can now configure and manage Ray workloads across multiple Kubernetes clusters. The changes are designed to be backward compatible for existing single-cluster deployments. Configuration updates will be required to specify target clusters for multi-cluster deployments.

**Documentation Changes**
Yes, this PR introduces user-facing and API changes. Documentation updates are required for:
- How to manage / add / delete compute clusters
- Multi-cluster Ray Job management guide

The changes are designed to be backward compatible for existing single-cluster deployments.


go/components/jobs/client/client.go
```diff
@@ -308,15 +310,42 @@ func (c *Client) DeleteSecret(
...
+			return "", "", fmt.Errorf("get client for cluster err:%v", err)
+		}
+
+		localNamespace, localName := c.mapper.GetLocalName(jobObject)
+
+		// Get the KubeRay RayJob resource
+		rayV1Job := &rayv1.RayJob{}
+		err = c.helper.GetResource(
+			ctx,
+			cs.Ray,
```
bhjoshi1, line 333: This should really happen via watcher.
1. We should not be retrieving the underlying Kuberay CRD in the hot path.
2. It may not even exist.

go/components/jobs/client/k8sengine/ray.go
```diff
@@ -21,6 +21,10 @@ func (m Mapper) mapRay(rayJob *v2pb.RayJob, jobClusterObject runtime.Object, clu
 		return nil, fmt.Errorf("expected *v2pb.RayCluster, got %T", jobClusterObject)
 	}
 	pod := rayCluster.GetSpec().Head.GetPod()
+	submitterPod := k8sptr.Deref(pod, corev1.PodTemplateSpec{})
```
bhjoshi1, line 23: Do we need this? Whether this pointer deref is safe should be taken up as part of validation? And would using a default value from deref make sense in this case?

go/components/jobs/client/client.go
dragod812: We will introduce watcher as part of this ticket - #605 . Then i can remove the use of this function from the controller.

go/components/jobs/client/k8sengine/ray.go
dragod812: Deref internally is nil safe - 
```
func Deref[T any](ptr *T, def T) T {
	if ptr != nil {
		return *ptr
	}
	return def
}
```
So no validation needed. 
Should we put a smart default here ?

go/components/jobs/client/client.go
```diff
@@ -309,15 +311,42 @@ func (c *Client) DeleteSecret(
...
 }
 
-// GetJobStatus gets the job status from the cluster
-func (c *Client) GetJobStatus(ctx context.Context, jobObject runtime.Object, cluster *v2pb.Cluster) (constants.SparkJobStatus, error) {
+// GetJobStatus retrieves the status of a Ray job from a remote cluster
+func (c *Client) GetJobStatus(
+	ctx context.Context,
+	jobObject runtime.Object,
+	kubeCluster *v2pb.Cluster,
+) (string, string, error) {
```
bhjoshi1: Looks like it went back to a previous commit?

go/components/jobs/client/client.go
bhjoshi1: You should still handle the case of job having completed / failed / not found.

dragod812: By default we don't delete the RayJob CR ( [Source](https://docs.ray.io/en/latest/cluster/kubernetes/getting-started/rayjob-quick-start.html#:~:text=DELETE_RAYJOB_CR_AFTER_JOB_FINISHES%20(Optional%2C%20added,the%20Kubernetes%20Job.)) )
We have to set this env variable DELETE_RAYJOB_CR_AFTER_JOB_FINISHES to delete the RayJob CR after completion, in OSS we have not set that. 
But i have added handling for NotFound Error in line 341. It treats it as Unknown state with Job Not found in compute cluster message. The Unknown state is a terminal state in Ray Job controller.

go/components/jobs/client/client.go
dragod812: Fixed.

go/components/jobs/client/client.go
bhjoshi1: All objects need to be deleted and cannot be kept forever. And "Unknown" cannot be a Terminal state.

bhjoshi1: But I'll stamp this now since you have already taken a follow-up for this.

go/components/ray/job/controller.go
```diff
@@ -144,75 +93,160 @@ func (r *Reconciler) Register(mgr ctrl.Manager) error {
...
 
-func (r *Reconciler) isRayJobTerminatedState(status v2pb.RayJobState) bool {
-	for _, state := range []v2pb.RayJobState{v2pb.RAY_JOB_STATE_FAILED, v2pb.RAY_JOB_STATE_SUCCEEDED, v2pb.RAY_JOB_STATE_KILLED} {
-		if status == state {
-			// Return OK. The job submission has reached a terminal status.
-			return true
-		}
+func isTerminalRayJobState(state v2pb.RayJobState) bool {
+	switch state {
+	case v2pb.RAY_JOB_STATE_FAILED, v2pb.RAY_JOB_STATE_SUCCEEDED, v2pb.RAY_JOB_STATE_KILLED, v2pb.RAY_JOB_STATE_UNKNOWN:
```
bhjoshi1: why would Unknown be a terminal state?

go/components/jobs/client/client.go
dragod812: Once RayJob reaches a terminal state "Success" / "Fail" we will not reconcile the RayJob again, so we won't try to do GetJobStatus "Forever". 
In the unlikely case that Job is running and we don't find the resource, it will be moved to unknown state and we will stop reconciling the controller forever.

go/components/ray/job/controller.go
dragod812: If we are not able to know the RayJobStatus for some reason ( RayJob CR is deleted ), instead of reconciling indefinitely we can terminate reconciliation. Maybe we can find a better name than "UNKNOWN". 
Maybe "NOT_FOUND" ?

bhjoshi1 (approved):

PR #509, opened 2025-11-11
feat(controllermgr): ConditionEngines Support Actor.Retrieve()

aritra0227:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR:
1. Adds the Retrieve method in the `ConditionActor[T client.Object]` interface
2. Updates the `defaultEngine`'s `Run` method to support it
3. Updates existing Actors to implement Retrieve() method
4. Updates Unit tests

***Summary of Changes to runPlugin and Run Methods***
1. Updates to `Engine.runPlugin(...)`: The `runPlugin` method runs actor.Retrieve(...) for all actors in a plugin but only runs actor.Run(...) for the first non-True condition returned from actor.Retrieve(...). We call this condition the criticalCondition. `runPlugin` always returns the criticalCondition when applicable. If it returns a nil condition (with nil error), then all conditions have been satisfied.
2. Updates to `Engine.Run(...)`:
Here is a state machine diagram to illustrate what `Engine.Run(...)` is meant to do based on the result of runPlugin:
Screenshot 2025-11-11 at 2 16 55 AM

Later on, retry strategies can be implemented to address the non-successful terminal states returned by `runPlugin`.

**Why?**
The Retrieve method is used heavily in the deployment/inference-server controller. More generally, a combination of retrieve+run would allow clever reconcile strategies to be implemented by plugin actors.

**How did you test it?**
1. Tested PipelineRun Functionalities within sandbox including run, dev-run, resume_from, kill
2. Unit tests


go/base/conditions/engine/defaultengine.go
```diff
@@ -84,24 +94,65 @@ func (e *DefaultEngine[T]) Run(ctx context.Context, plugin conditionInterfaces.P
 			}, nil
 		}
 	}
+
 	return defaultResult, nil
 }
 
+// runPlugin executes the plugin's actors against the given resource and returns:
+//  1. The "critical condition", specifically the condition yielded by the actor's Run() method
```
uzbq, line 102: is this critical condition also used internally or this is something new for OSS?

uzbq (approved):
In general, I do not block to have a retrieve and a run interfaces. Would love to learn more the use cases about the deployment controller.

go/base/conditions/engine/defaultengine.go
aritra0227: Nothing new, its used internally as well. In the internal implementation the exposed `Run` method has retry policies for how to handle terminal conditions.

aritra0227:
> In general, I do not block to have a retrieve and a run interfaces. Would love to learn more the use cases about the deployment controller.

While it's technically possible to encapsulate the `retrieve` method's logic within the `run` method (essentially having `run` handle both responsibilities), I believe keeping them separate maintains a cleaner interface. In the deployment controller specifically, actors make multiple types of API calls (e.g. to multiple inference server backends, Kubernetes clients, etc.). Combining both methods would increase the complexity of the `actor` logic, whereas keeping `run` and `retrieve` separate make responsibilities more clear and the interface simpler.

Of course, the above is just my opinion, but since the retrieve method is already used internally, I think introducing it in OSS will at very least reduce merge-back effort in the future. @badcount cc-ing you if you want to add anything.

PR #583, opened 2025-11-18
feat(inference-server): inference server gateway and model configmap utilities

aritra0227:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR introduces the following:
1. A `Gateway` interface which is used to communicate with inference server backends.
2. A `ModelConfigMapProvider` interface which is used to manage kubernetes config maps for models deployed to inference servers.
3. A default implementation for each of the above interfaces
4. Unit tests where appropriate

**Why?**
The deployment and inference server controllers need to communicate with the inference-server backends (e.g. Triton, LLM-D) and manage models through configmap manipulations. This PR is a starting point for abstracting away the model configmap management and inference-server backend communication into their own packages.

**How did you test it?**
1. The interfaces are used thoroughly within the deployment+inference-server controllers. Tested locally via sandbox.
2. Unit Tests.


go/components/inferenceserver/gateways/gateway.go
```diff
@@ -0,0 +1,152 @@
+package gateways
```
aritra0227, line 1: **Skipping tests for this file at the moment since all methods within the gateway function are either routing requests to the inference-server backend or the httproute Manager.**

Once more inference-server backends are added, if there are some common logic that can be abstracted away and placed in this package, then unit tests can be added for those.

go/components/inferenceserver/gateways/backends/triton.go
```diff
@@ -0,0 +1,976 @@
+package backends
```
aritra0227, line 1: **Skipped adding unit tests here since it didn't make much sense to add them.** Almost all methods are non-hermetic and would need to be heavily mocked. I think the functionality here is better tested via integration tests of the inference-server controller later on.

go/components/inferenceserver/gateways/backends/triton.go
```diff
@@ -0,0 +1,976 @@
...
+# Function to get currently loaded models from Triton
+get_loaded_models() {
+  curl -s http://localhost:8000/v2/models 2>/dev/null | jq -r '.models[]?.name // empty' 2>/dev/null || echo ""
+}
+
+# Function to load model in Triton
+load_model() {
+  local model_name=$1
+  echo "Loading model $model_name in Triton"
+  curl -s -X POST "http://localhost:8000/v2/repository/models/$model_name/load" -H "Content-Type: application/json" -d '{}'
```
aritra0227, line 561: Please correct me if I am wrong, but I believe the value of the url remains `http://localhost:8000` regardless of whether this is ran within a cluster or locally (i.e. bazel run).

go/components/inferenceserver/configmap/provider.go
```diff
@@ -0,0 +1,206 @@
...
+	modelListKey      = "model-list.json"
+	modelConfigSuffix = "model-config"
+)
+
+var addSuffixToString = func(str, suffix string) string {
+	return fmt.Sprintf("%s-%s", str, suffix)
+}
+
+// defaultModelConfigMapProvider implements the ModelConfigMapProvider interface.
+type defaultModelConfigMapProvider struct {
```
badcount, line 27: add "var _ ModelConfigMapProvider = defaultModelConfigMapProvider 
so we can check in compile level

go/components/inferenceserver/proxy/httproute.go
```diff
@@ -0,0 +1,586 @@
...
+		Version:  gatewayAPIVersion,
+		Resource: httpRouteResource,
+	}
+
+	addSuffixToString = func(str, suffix string) string {
+		return fmt.Sprintf("%s-%s", str, suffix)
+	}
+)
+
+type httpRouteManager struct {
```
badcount, line 41: You created interface for configMap, but not for httpRoute? So how should we implement routing internally?

go/components/inferenceserver/configmap/provider.go
aritra0227: done

go/components/inferenceserver/proxy/httproute.go
aritra0227: This was an oversight on my end. Updated implementation to include interface.

badcount (approved):

PR #639, opened 2025-12-03
Further migrate packaging utils in model manager to OSS

kenns29:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Migrate more critical utils needed for model packaging


krishpatel9 (approved):

python/michelangelo/lib/model_manager/interface/custom_model.py
```diff
@@ -1,17 +1,17 @@
...
 
 
 class Model(ABC):
-    """
-    The base class for custom models.
-    """
+    """The base class for custom models."""
 
     @abstractmethod
     def save(self, path: str):
```
badcount, line 12: we only have save and load? how do you know if save to raw model or deployed model?

kenns29: both raw and deployable model will contain this model wrapper. The path for save/load is determined according to the environment where the model is loaded. The path is automatically inferred, the user doesn't have to worry about the actual path.

Issue #636, opened 2025-11-26
[Inference Server] Add Kubernetes DaemonSet that syncs model config and performs model load/unload


aritra0227:
See comment here: https://github.com/michelangelo-ai/michelangelo/pull/632#discussion_r2565952167

PR #640, opened 2025-12-03
feat(inference): add sandbox support to demo inference

aritra0227:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR introduces:
1. Created a separate kubernetes deployment for a model-sync daemon. This will be responsible for loading/unloading models into the deployed inference server. Completes https://github.com/michelangelo-ai/michelangelo/issues/636.
2. Sandbox updates to support inference server demos. As part of this, we add support for setting up Istio infrastructure and routing via the Kubernetes Gateway API.
3. Breaks up the `ma sandbox demo` command into `ma sandbox demo pipeline` and `ma sandbox demo inference`. 
4. Restructures directories for demo CRs

**Next Steps**
When model packager/pusher is ready, as part of `ma sandbox demo inference` we can:
1. Run some pipeline to push the model into storage
2. Create a deployment CR which deploys the model to the inference server

**How did you test it?**
1. Created docker image against branch using dev-release workflow. Then ran `ma sandbox create` followed by `ma sandbox demo inference` and `ma sandbox demo pipeline`.


python/michelangelo/cli/sandbox/resources/model-sync.yaml
```diff
@@ -0,0 +1,328 @@
...
+        echo "Model $model_name unloaded successfully"
+      else
+        echo "Failed to unload model $model_name"
+      fi
+    }
+    
+    # Main sync loop
+    while true; do
+      echo "========================================"
+      echo "Starting UCS-style model sync cycle"
```
badcount: remove this comment

python/michelangelo/cli/sandbox/resources/model-sync.yaml
```diff
@@ -0,0 +1,328 @@
...
+        echo "--- Completed sync for $INFERENCE_SERVER ---"
+      done < "$INFERENCE_SERVERS_FILE"
+      
+      echo ""
+      echo "Model sync cycle completed for all servers, sleeping for 60 seconds"
+      sleep 60
+    done
+---
+apiVersion: apps/v1
+kind: Deployment
```
badcount, line 234: why this is a deployment but not a daemonSet?

python/michelangelo/cli/sandbox/resources/gateway-api-setup.yaml
```diff
@@ -0,0 +1,52 @@
...
+  listeners:
+  # Standard HTTP port - no hostname restriction for development
+  - name: http
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+  # Triton inference server port
+  - name: triton-http
```
badcount: why do we need this?

python/michelangelo/cli/sandbox/resources/gateway-api-setup.yaml
```diff
@@ -0,0 +1,52 @@
...
+    port: 8889
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+  # Additional ports for other services
+  - name: cadence-web
```
badcount: we don't need this, this is model serving cluster, it shouldn't need cadence

python/michelangelo/cli/sandbox/resources/gateway-api-setup.yaml
```diff
@@ -0,0 +1,52 @@
...
+    port: 8088
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+  - name: envoy-admin
```
badcount: Same here

python/michelangelo/cli/sandbox/resources/model-sync.yaml
aritra0227: I'm assuming the issue is with the term "UCS"? I modified the line to be "starting model sync cycle" instead.

python/michelangelo/cli/sandbox/resources/gateway-api-setup.yaml
aritra0227: We don't. This was an oversight on my end. Removed this and the rest of the entries below.

python/michelangelo/cli/sandbox/resources/model-sync.yaml
aritra0227: The default behavior of a DaemonSet is to schedule a `model-sync` pod on every node. However, not all nodes will have an inference-server deployment.

To correct this, I attempted to use podAffinity on the DaemonSet to target nodes labeled for the inference server. However, even with the affinity rule applied in the sandbox, two model-sync pods were created: one succeeded while the other remained in a pending state. Apparently this is the expected behavior for a DaemonSet but I thought it provided poor visibility on the pods.

That's why I went with a deployment for the time being.

badcount (approved):

PR #684, opened 2025-12-09
fix(inference): update httproute path matching

aritra0227:
- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR makes the following changes:
1. Updates path matching for deployment httproutes to route `http://localhost:8080/<inference-server-name>/<deployment-name>/*` to `http://localhost:8080/v2/models/<model-name>/*`. This allows the deployment httproute to be agnostic of the model-name and the inference-server name.
2. Updates path matching for inferenceserver httproutes to route `http://localhost:8080/<inference-server-name>/models/<model-name>*` to `http://localhost:8080/v2/models/<model-name>/*`. 
3. Updates inference-server CR name from `inference-server-bert-cola` to `inference-server-example`. As part of this, we also update the model-sync script to target the appropriate inference-server.

**How did you test it?**
1. Tested changes in sandbox
2. Unit tests

### DIFF
--- a/python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
+++ b/python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
@@ -1,7 +1,7 @@
 apiVersion: michelangelo.api/v2
 kind: InferenceServer
 metadata:
-  name: inference-server-bert-cola
+  name: inference-server-demo
   namespace: default
   labels:
     app: inference-server
--- a/python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
+++ b/python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
@@ -1,7 +1,7 @@
 apiVersion: michelangelo.api/v2
 kind: InferenceServer
 metadata:
-  name: inference-server-demo
+  name: inference-server-example
   namespace: default
   labels:
     app: inference-server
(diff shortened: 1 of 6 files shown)


python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: michelangelo.api/v2
 kind: InferenceServer
 metadata:
-  name: inference-server-bert-cola
+  name: inference-server-demo
```
badcount: this inference service is for examples, we can make it example

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -1196,13 +1196,13 @@ def _create_inference_demo_crs():
         "🌐 Deployment-agnostic endpoint:\
             Use the following URL to test the inference server"
     )
-    print("  http://localhost:8080/inference-server-bert-cola/v2")
+    print("  http://localhost:8080/inference-server-demo/v2")
```
badcount: why do we need v2?

aritra0227: we don't, just updated.

python/michelangelo/cli/sandbox/demo/inference/inferenceserver.yaml
aritra0227: done.

badcount (approved):

Issue #622, opened 2025-11-25
[deployment controller] Properly use GetState() and ParseStage() to set state/stage in controller


aritra0227:
Currently, we manually set the state and stages on the actor implementations. Instead, we should utilize the Plugin.GetState() and Plugin.ParseStage() functions properly to determine the state and stages based on the conditions and the deployment resource. This way, actors will only interact with the deployment spec and modify the conditions with necessary information and the controller can utilize `GetState()` and `ParseStage()` to set the appropriate status on the deployment resource. This method is similar to the internal implementation.

PR #732, opened 2026-01-13
refactor(deployment): condition actor improvements and correct usage of plugin.GetState() and plugin.ParseStage 

aritra0227:
- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**Summary**
This PR addresses the following:
1. **Properly implement and utilize plugin.GetState() and plugin.ParseStage()**. Actors no longer explicitly set the deployment CR's state and stage. Instead, actors set the CR's conditions, and the GetState() and ParseStage() functions derive the deployment's state and stage from those conditions. The deployment controller now uses these functions to orchestrate the deployment CR's lifecycle. This completes https://github.com/michelangelo-ai/michelangelo/issues/622.
2. **Implement utility functions for generating conditions within actors**: `GenerateTrueCondition`, `GenerateFalseCondition`, `GenerateUnknownCondition`.
3. **Traffic Switching now happens before old model cleanup**. When updating a deployment from `model-A` to `model-B`, we initiate cleanup/unloading of `model-A` only after the traffic is switched to `model-B`. **From a user's perspective, there is now zero downtime for making any inference requests even during model deployment updates**.
4. **Rollback actor properly performs model unload**. This replaces previous no-op solution.
5. **Reduce clutter in Actor logic**: Merged `model_sync_actor` into `rolling_actor`. Simplified `Run()` methods for actors where the run is a no-op (e.g., `ValidationActor`, `AssetPreparationActor` simply return the input condition unchanged).

Upcoming PRs will add multi-cluster deployment strategies. This PR reduces clutter in the existing plugin logic and aligns condition engine usage with Uber's internal patterns. These simplifications make it easier to introduce the multi-cluster strategies, which will add significant code changes.

**How did you test it?**
1. Tested changes on sandbox
2. Unit tests


go/components/deployment/plugins/oss/steadystate/steady_state.go
```diff
@@ -23,103 +24,47 @@ func (a *SteadyStateActor) GetType() string {
...
+	if err != nil {
+		a.logger.Error("failed to check health of inference server",
+			zap.Error(err),
+			zap.String("operation", "steady_state_health_check"),
+			zap.String("namespace", resource.Namespace),
+			zap.String("deployment", resource.Name),
+			zap.String("inference_server", resource.Spec.GetInferenceServer().Name))
+		return conditionsutil.GenerateFalseCondition(condition, "HealthCheckFailed", fmt.Sprintf("Failed to check health of inference server: %v", err)), nil
+	}
+	if !healthy {
```
badcount, line 43: Curious, where is the logic if the inference server is not healthy forever, and when it go to fail state?

go/components/inferenceserver/backends/triton.go
```diff
@@ -205,32 +205,31 @@ func (b *tritonBackend) IsHealthy(ctx context.Context, logger *zap.Logger, infer
 					logger.Info("Triton pods are ready", zap.String("server", inferenceServerName), zap.Int("readyReplicas", int(deployment.Status.ReadyReplicas)))
 					return true, nil
 				} else {
-					logger.Error("Triton deployment available but pods not ready",
+					logger.Info("Triton deployment available but pods not ready",
```
badcount: Should this be warning?

go/components/inferenceserver/backends/triton.go
```diff
@@ -205,32 +205,31 @@ func (b *tritonBackend) IsHealthy(ctx context.Context, logger *zap.Logger, infer
...
 						zap.String("server", inferenceServerName),
 						zap.Int("readyReplicas", int(deployment.Status.ReadyReplicas)),
 						zap.Int("totalReplicas", int(deployment.Status.Replicas)))
-					return false, fmt.Errorf("pods not ready for deployment %s/%s: %d/%d",
-						namespace, deploymentName, deployment.Status.ReadyReplicas, deployment.Status.Replicas)
+					return false, nil
 				}
 			} else {
-				logger.Error("Triton deployment not available",
+				logger.Info("Triton deployment not available",
```
badcount: Should this be warning?

go/components/inferenceserver/backends/triton.go
aritra0227: updated

go/components/inferenceserver/backends/triton.go
aritra0227: updated.

go/components/deployment/plugins/oss/steadystate/steady_state.go
aritra0227: Internally, we just emit errors. I think in OSS, we want to emit these errors to Prometheus. I'm unsure if we are properly doing that at the moment. I will take a look and update in a separate PR later. Created issue here: https://github.com/michelangelo-ai/michelangelo/issues/743

go/components/deployment/plugins/oss/rollout/strategies/strategy.go
badcount, line 1: why we call this interface.go but we don't define any interface here?

go/components/deployment/plugins/oss/steadystate/steady_state.go
badcount: can we do the same? ideally all the behaviors should be same as internal.

aritra0227: The controller does emit the errors in this case (expected behaviour).

What I am unsure about is whether the errors/counters are properly reflected in prometheus so that a user can hook up alerting with it. That is what I created the above issue for.

go/components/deployment/plugins/oss/rollout/strategies/strategy.go
aritra0227: good point. Just changed it to `strategy.go`.

badcount (approved):

PR #745, opened 2026-01-23
Remove the Any ext fields in protobuf messages

ying-zheng:
- [X] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Removed the Any ext fields in protobuf messages

**Why?**
We added these fields for open source users to customize or extend Michelangelo protobuf schema
But now we think this is not necessary. It's easier and more flexible to do the extension with patches or forks that simply add extra fields

**How did you test it?**
unit test

**Potential risks**
For users who are already using these fields, this can be a backward incompatible change
But, AFAIK, no one is using these these ext fields

**Release notes**

**Documentation Changes**


himanshk96 (approved):
LGTM

PR #757, opened 2026-01-29
Add generated proto-go module and tools for non‑Bazel users

ying-zheng:
- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

1. Introduced a generated protobuf Go module in proto-go/, including go.mod, go.sum, and a README explaining usage and version-sync policy.
2. Added tools/gen-proto-go.sh script to build protos via Bazel, copy generated .go files into proto-go, sync dependency versions from go/go.mod, and run go mod tidy.
3. Switched Go imports and Bazel proto importpath settings from .../proto/... to .../proto-go/..., and updated Bazel module deps accordingly.
4. Added a CI dirty-check to ensure proto-go stays up to date.
5. Removed the dummy go.mod and empty.go files in /proto. These files were used for making 'go mod tidy' treat the /proto directory as a go module. This is no longer needed as we added a real go module (/proto-go) for the protobuf generated go code
6. Updated developer guide: docs/contributing/how-to-write-apis.md
7. add a .gazelleignore file to avoid gazelle from generating BUILD.bazel files in /proto-go directory

**Why?**

Before this change, the go files generated from protobuf were not checked in to github repo, and were only generated in bazel cache directory during bazel build. This works well for bazel users. But open source users cannot use Michelangelo libraries in the go projects that are not build with bazel, because without bazel there was no way to get the protobuf generated go files.

This change commits the protobuf generated go code into github repo for non-Bazel users.

For bazel users, the go code is still automatically generated during bazle build (bazel build does not use the files in proto-go directory) whenever the protobuf files are modified.

**How did you test it?**

- tools/gen-proto-go.sh
- tools/bazel mod tidy
- tools/gazelle
- go mod tidy in go/
- go mod tidy in proto-go/
- bazel build ...
- bazel test ...

**Potential risks**

**Release notes**

**Documentation Changes**
Added a README.md file in proto-go directory
Added a section in docs/contributing/how-to-write-apis.md


tools/gen-proto-go.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+WORKSPACE_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+BAZEL="${WORKSPACE_ROOT}/tools/bazel"
+OUT_DIR="${WORKSPACE_ROOT}/proto-go"
+IMPORT_PREFIX="github.com/michelangelo-ai/michelangelo/proto-go"
```
badcount, line 10: can this be passed as input, i wonder if Grab want to add their own proto, so it won't be in this folder.

tools/gen-proto-go.sh
```diff
@@ -0,0 +1,77 @@
...
+for line in go_text.splitlines():
+    line = line.strip()
+    if not line or line.startswith("//"):
+        continue
+    m = re.match(r"^([\\w./-]+)\\s+(v\\S+)$", line)
+    if m:
+        versions[m.group(1)] = m.group(2)
+
+# Align Go version too.
+proto_text = re.sub(r"^go\\s+\\S+", "go 1.23.2", proto_text, flags=re.M)
```
badcount, line 55: Should we hardcode this? go 1.23.3? this could be missed if we upgrade go,

ying-zheng: good catch! I will fix this

tools/gen-proto-go.sh
ying-zheng: this is the go import path used in our go code. It cannot be changed with this script

badcount (approved):

PR #778, opened 2026-02-03
Ingester & Mysql for CRD backup

himanshk96:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

Commands Run (in order)

  # 1. Delete sandbox
  python3 python/michelangelo/cli/sandbox/sandbox.py delete

  # 2. Recreate sandbox (with cached images imported manually into k3d)
  python3 python/michelangelo/cli/sandbox/sandbox.py create

  # 3. Verify MySQL tables (39 tables from schema init job)
  kubectl exec pod/mysql -- mysql -uroot -proot michelangelo -e "SHOW
  TABLES;"

  # 4. Create namespace + 13 CRs
  kubectl create namespace ingester-test
  kubectl apply -f scripts/ingester-test-crs/

  # 5. Section 2 — Patch all CRs, verify MySQL updates
  kubectl patch project / modelfamily / model / pipeline / pipelinerun / etc.

**Potential risks**

**Release notes**

**Documentation Changes**


go/components/ingester/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	log.Info("Object is being deleted")
+
+	// Check if our finalizer is present
+	if !ctrlutil.ContainsFinalizer(object, api.IngesterFinalizer) {
+		log.Info("Finalizer not present, nothing to do")
+		return ctrl.Result{}, nil
+	}
+
+	// Check grace period
+	gracePeriodSeconds := object.GetDeletionGracePeriodSeconds()
```
badcount: Is this a bug? 
from claude code:
GetDeletionGracePeriodSeconds() returns the static value set at deletion time — it does not count down. So *gracePeriodSeconds > 10 just checks "was the grace period
  configured to more than 10 seconds?", not "has 10 seconds elapsed?". The condition never becomes false on its own, so the reconciler would requeue forever and never complete
  the deletion.

go/components/ingester/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	Log             logr.Logger
+	Scheme          *runtime.Scheme
+	TargetKind      client.Object
+	MetadataStorage storage.MetadataStorage
+	Config          Config
+}
+
+// Reconcile is the main reconciliation loop
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("namespace", req.Namespace, "name", req.Name)
```
badcount, line 70: Do we still timeout like: 
ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
  defer cancel()

badcount: Also we still need like EnableStorage toggle right?

go/components/ingester/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	log.Info("Reconciling object")
```
badcount, line 71: do we need to check IsImmutableKind?

go/components/ingester/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	log.Info("Successfully moved immutable object to metadata storage only")
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	kind := r.TargetKind.GetObjectKind().GroupVersionKind().Kind
+	controllerName := fmt.Sprintf("ingester_%s", kind)
+
+	concurrentReconciles := r.Config.ConcurrentReconciles
```
badcount, line 230: We don't do reconcileMap anymore?

go/components/ingester/controller.go
himanshk96: Good catch! GetDeletionGracePeriodSeconds() is indeed a static label, not a countdown —    
  removed the check entirely. The API Server already enforces the grace period before the 
  delete event reaches controllers, so by the time we see DeletionTimestamp != nil the grace 
  period has already elapsed. Fixed in 8b84a06d.

go/components/ingester/controller.go
himanshk96: Added in 4228dc6b. Wrapping the reconcile context with a 30s timeout so a hung MySQL       
  connection doesn't block the goroutine indefinitely.

himanshk96: Already handled — EnableMetadataStorage is checked at startup in ingester_providers.go. If 
  metadataStorage.enableMetadataStorage is false, provideMetadataStorage returns nil, and  
  module.go skips controller registration entirely. So Reconcile() is never called when      
  storage is disabled. No toggle needed inside Reconcile() itself.

go/components/ingester/controller.go
himanshk96: Good catch! Model and CachedOutput both have IsImmutableKind() == true, meaning they were  
  silently routing to handleSync and staying in etcd forever instead of being moved to       
  MySQL-only storage. Fixed in d4d6e241 — added isImmutableKind() helper that type-asserts   
  the object and routes to handleImmutableObject for both annotation-based and kind-based    
  immutability.

go/components/ingester/controller.go
himanshk96: I think reconcileMap was internal config. Is this new way okay?

badcount: We need to import back, with your changes, we can't have same features internally.

badcount (approved):

PR #798, opened 2026-02-09
CR version conversion functions

ying-zheng:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
1. Add a function to create empty object with specified group / version / kind
2. Add a function to do version conversion

**Why?**
When interact with k8s API server using k8s client library, the version conversion is automatically handled with WebHook
For MySQL storage, we need to do the version conversion by ourselves. These helper functions are needed to do the version conversion when retrieve / update CR objects in MySQL storage

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


go/api/crd/BUILD.bazel
```diff
@@ -7,6 +7,7 @@ go_library(
         "module.go",
         "schema_diff.go",
         "sync.go",
+        "versions.go",
```
TheGreatFrankie, line 10: Could you help me understand what is the use case of conversion function in MySQL?

I previously thought the MySQL is abstracted away behind the k8-apiserver. So all component would reach k8-apiserver first and the conversion webhook would do the conversion for us. What component directly interacts with MySQL?

TheGreatFrankie: Answered in Slack thread: https://uber.slack.com/archives/C036Y1N3MND/p1770665550422999?thread_ts=1770607802.974669&cid=C036Y1N3MND

TheGreatFrankie (approved):

PR #799, opened 2026-02-09
Fix retry cell: types, accessor, dev artifacts

austingreco:
## Summary

Cleanup fixes for the retry button implementation:

- **Fix `CellProps` → `CellRendererProps`**: `CellProps` doesn't exist in the types file; all cell renderers use `CellRendererProps<T>`
- **Simplify accessor**: The original `(record) => record?.activityId ? record : null` returned the entire record — `RetryCell` uses `props.record`, not `props.value`, so a simple `'activityId'` string accessor is correct
- **Restore `hideEmpty: true`**: Prevents an empty "Actions:" label from rendering on steps without a retry button
- **Replace `create(RetryInfoSchema)` with plain object**: Matches codebase pattern — the Connect client handles serialization via `OmitTypeName` type stripping
- **Replace `>= 3` with `TERMINATED_STATES` Set**: Explicit enumeration of terminated states (3-6) instead of a range that would incorrectly match future state values
- **Replace `alert()` with in-modal error display**: Uses `updatePipelineRunMutation.error.message` with `theme.colors.negative` styling
- **Remove `console.log` / `alert()` calls**: Development artifacts
- **Add local interfaces** (`PipelineRunStatus`, `PipelineRunData`, `StepRecord`): Replaces `(x as any)` casts with typed interfaces per CLAUDE.md guidelines

## Test plan

- [ ] Verify retry button only appears on steps with `activityId` when pipeline run is terminated
- [ ] Verify no empty "Actions:" label on steps without retry button
- [ ] Verify retry modal shows error message on failure (red text, not alert())
- [ ] Verify successful retry refreshes the pipeline run data


austingreco:
creating standalone PR instead

PR #802, opened 2026-02-09
refactor(inference+deployment): make user interfaces pluggable

aritra0227:
- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### **Summary**
**Most changes in this file are just updated function signatures and moving file locations**.

1. **This PR updates the interfaces that we expect most users to implement. In particular, the following files define the core changes in this PR:**
  a) `go/components/inferenceserver/backends/interface.go`
  b) `go/components/inferenceserver/modelconfig/interface.go`
  c) `go/components/deployment/route/interface.go`

2. **Other than interface changes, the PR also updates the deployment controller and plugins to directly call modelConfigProvider instead of using Gateway. As a result, the gateway interface was also updated to simply serve health check requests:**
  a) `go/components/inferenceserver/gateways/interface.go`

3. **We update the inference-server controller code to match exactly as internal controller. Before, the implementation expected, 1 backend -> 1 plugin (e.g. triton, dynamo all have their own controller plugin). Now, all backends will share the same plugin and have separate backend implementations. [See here for change.](https://github.com/michelangelo-ai/michelangelo/pull/802/changes#r2785393777)**

4. Finally, we also added an additional condition actor to explicitly monitor state of an inference-server model-config:
  a) `go/components/inferenceserver/plugins/oss/creation/model_config_provision.go`

**Why?**
1. This PR stabilizes the Go interfaces that will serve as public API contracts. With the repository being open-sourced soon, we need well-defined interfaces that remain backward-compatible.

**How did you test it?**
1. Unit tests
2. Ran changes in sandbox to verify


go/components/inferenceserver/plugins/oss/plugin.go
```diff
@@ -1,24 +1,182 @@
 package oss
```
aritra0227, line 1: No new logic, directly copied over from go/components/inferenceserver/plugins/oss/triton/plugin.go

go/components/inferenceserver/modelconfig/interface.go
```diff
@@ -0,0 +1,41 @@
+//go:generate mamockgen ModelConfigProvider
```
aritra0227, line 1: content is exactly the same. This was just moved from `go/components/inferenceserver/configmap`

go/components/inferenceserver/modelconfig/module.go
```diff
@@ -0,0 +1,15 @@
+package modelconfig
```
aritra0227, line 1: content is exactly the same. This was just moved from go/components/inferenceserver/configmap

go/components/inferenceserver/modelconfig/provider.go
```diff
@@ -0,0 +1,306 @@
+package modelconfig
```
aritra0227, line 1: content is exactly the same. This was just moved from go/components/inferenceserver/configmap

go/components/inferenceserver/plugins/oss/creation/backend_provision.go
```diff
@@ -0,0 +1,90 @@
+package creation
```
aritra0227, line 1: content is exactly the same, this was just moved from `go/components/inferenceserver/plugins/oss/creation/resource_creation.go`

go/components/inferenceserver/plugins/oss/creation/condition_plugin.go
```diff
@@ -0,0 +1,64 @@
+package creation
```
aritra0227, line 1: content is exactly the same, this was just moved from go/components/inferenceserver/plugins/oss/triton/

go/components/inferenceserver/gateways/interface.go
```diff
@@ -4,24 +4,18 @@ package gateways
...
 
 	"go.uber.org/zap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v2pb "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
 )
 
-// Gateway provides a unified interface for inteacting with inference servers across different backend types.
+// Gateway provides a unified interface for interacting with inference servers across different backend types.
 type Gateway interface {
```
aritra0227: Question for reviewer:

Is there a reason to keep this interface anymore? This is not something that a user will implement. Currently, the implementation simply gets the appropriate backend for the inference server, then routes to the appropriate function of that backend. See ERD [here](https://docs.google.com/document/d/1cLfXQfqCWRJYD8c5NL2-_5L8ZNTRl3skKNwoLjEUoi8/edit?tab=t.0#heading=h.flpiez89ndop) for explanation. This interface is only used by the deployment controller. 

**Alternative:** The deployment controller can directly call `backend.CheckModelStatus` and `backend.InferenceServerIsHealthy`. Since each backend will have a separate method of checking model status/performing health-checks, its important that each `backend` implements these methods separately. 

WDYT?

go/components/inferenceserver/backends/interface.go
```diff
@@ -4,31 +4,32 @@ package backends
...
-	Ready     bool
 	Endpoints []string
 }
 
 // Backend defines the interface for inference server backend implementations (Triton, vLLM, etc.).
 // Each backend provides platform-specific logic for server and model management.
 type Backend interface {
 	// CreateServer provisions backend-specific Kubernetes resources for an inference server.
-	CreateServer(ctx context.Context, logger *zap.Logger, inferenceServer *v2pb.InferenceServer) (*ServerStatus, error)
+	CreateServer(ctx context.Context, logger *zap.Logger, kubeClient client.Client, inferenceServer *v2pb.InferenceServer) (*ServerStatus, error)
```
aritra0227, line 26: By directly passing in the clients to all the methods, the control-plane (inference-server controller) can connect to the appropriate cluster and provide its client as an argument to this function. **With this change, this interface does not need any modifications for the multi-cluster case.**

go/components/inferenceserver/modelconfig/interface.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// ModelConfigProvider manages model configurations for inference servers.
+// This facilitates model management through a sidecar pattern, where a sidecar container
+// watches the config and loads/unloads models accordingly.
+// Configurations are stored in a backing store (e.g., Kubernetes ConfigMap, or other storage).
+// The InferenceServer controller creates/deletes the config, while the Deployment controller adds/removes model entries.
+type ModelConfigProvider interface {
+	// CreateModelConfig creates a new model config with model configurations.
+	CreateModelConfig(ctx context.Context, logger *zap.Logger, kubeclient client.Client, inferenceServerName string, namespace string, labels map[string]string, annotations map[string]string) error
```
aritra0227, line 25: By directly passing in the clients to all the methods, the control-plane (inference-server/deployment controller) can connect to the appropriate cluster and provide its client as an argument to this function. **With this change, this interface does not need any modifications for the multi-cluster case.**

go/cmd/controllermgr/main.go
```diff
@@ -92,10 +95,13 @@ func options() fx.Option {
 		pipeline.Module,
 		pipelinerun.Module,
 		controllermgr.Module,
```
aritra0227, line 97: The changes in this file are to directly provide the user's implementations for:
1. routeProvider (e.g. httproute, istio-virtualservice)
2. modelConfigProvider (e.g. configmap, s3)
3. backend (e.g. triton, dynamo)

go/components/inferenceserver/controller.go
```diff
@@ -210,15 +194,13 @@ func (r *Reconciler) reconcile(ctx context.Context, inferenceServer *v2pb.Infere
 			zap.String("name", inferenceServer.Name))
 	}
 
-	var conditionPlugin conditionInterfaces.Plugin[*v2pb.InferenceServer]
+	conditionPlugin := r.plugin.GetCreationPlugin()
```
aritra0227, line 197: With this change, the inferenceserver controller logic will be **identical** to the internal implementation. Before we were assuming that triton/llm-d/dynamo will have their own plugin. Now, we expect all backends to share the same plugin. However, each backend will have their own implementation of the `backend` interface here: go/components/inferenceserver/backends/interface.go

go/components/inferenceserver/backends/interface.go
badcount: Let's add more detail comment to each interface methods

go/components/inferenceserver/modelconfig/interface.go
badcount: Add more comment for each of interface method

go/components/inferenceserver/modelconfig/provider.go
badcount: the provider looks like a detail implementation using configmap, can we have configmap folder under the provider?

go/components/inferenceserver/gateways/interface.go
aritra0227: synced offline: keep it simple for now and directly route to `Backend`.

badcount (approved):

PR #814, opened 2026-02-12
fix sitecustomize log to avoid using basicConfig

yangfan0328:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Refactored `sitecustomize.py` logging configuration to eliminate `basicConfig()` usage and prevent root logger pollution:

1. **Removed `basicConfig()` call** - Replaced `_get_logger()` function that used `logging.basicConfig()` with a module-level logger
2. **Added module-level logger** with `propagate = False` to prevent logging hierarchy pollution
3. **Made logging configurable** - Added `UF_FILE_SYNC_DEBUG` environment variable to control logging (silent by default)
4. **Made `downloader` parameter required** in `file_sync_pre_run()` for explicit configuration
5. **Added single-execution guard** - `_file_sync_executed` flag prevents multiple executions per process
6. **Consistent logger fallback pattern** - Both `file_sync_pre_run()` and `download_and_extract_dev_files()` use `logger=None` with runtime fallback, which makes importing back to internal more clean

**Why**
The original code used `logging.basicConfig()`, which configures the root logger and affects ALL loggers in the Python process, causing:
- Log pollution in other modules (PySpark, libraries, application code)
- Unexpected log output format changes globally
- Cannot be disabled or controlled without affecting the entire application
- Printed debug info unconditionally on every import

**How did you test it?**
E2E test, follow steps on https://docs.google.com/document/d/1GA52ROSz3jWp1_TbPkE_MdjnCwVU9SGp-crOCtc1EbY/edit?usp=sharing
Update boston_housing_xgb.py file with local changes (log.info)
[boston_housing_xgb.py](https://github.com/user-attachments/files/25307696/boston_housing_xgb.py)

Tested by enabling logs, on uniflow/core/remote_run.py, line 73 adding 
environ["UF_FILE_SYNC_DEBUG"] = "1"
ray logs:
[ray-feature_prep-dk4lc-logs.txt](https://github.com/user-attachments/files/25307678/ray-feature_prep-dk4lc-logs.txt)
spark logs:
[spark-preprocess-52ws6-logs.txt](https://github.com/user-attachments/files/25307679/spark-preprocess-52ws6-logs.txt)

Tested by disabling logs,unset UF_FILE_SYNC_DEBUG flag
ray logs:
[ray-feature_prep-m2phl-2y2bv-logs.txt](https://github.com/user-attachments/files/25307682/ray-feature_prep-m2phl-2y2bv-logs.txt)
spark logs:
[spark-preprocess-s6t4n-2y2bv-logs.txt](https://github.com/user-attachments/files/25307683/spark-preprocess-s6t4n-2y2bv-logs.txt)

**Potential risks**
1. Visible changes: Debug output now requires UF_FILE_SYNC_DEBUG=1 (was unconditional before); Logs go to stderr instead of stdout
2. Migration path for internal implementations: Internal code needs to explicitly pass downloader parameter to file_sync_pre_run(). One-line code in main function: file_sync_pre_run(downloader=TerrablobDownloader())
3. Rollback plan: If issues arise, can be reverted without data loss with this change only affects logging

**Release notes**
N/A

**Documentation Changes**
N/A


sallycr (commented):
@yangfan0328 Please add the extended docstring for the public interfaces that is introduced in open source repo.

python/michelangelo/uniflow/core/sitecustomize.py
```diff
@@ -10,169 +10,45 @@
...
 
 # Run the file sync pre-run functionality automatically when this module is imported
-if __name__ != "__main__":
-    import sys
-
-    print(f"[sitecustomize] Python executable: {sys.executable}")
-    print(f"[sitecustomize] Working directory: {os.getcwd()}")
-    print(
-        f"[sitecustomize] UF_FILE_SYNC_TARBALL_URL: {os.environ.get('UF_FILE_SYNC_TARBALL_URL', 'NOT SET')}"
+if __name__ != "__main__":  # pragma: no cover
```
andriika: Why do we need this condition? Isn't it always `True`?

python/michelangelo/uniflow/core/sitecustomize.py
```diff
@@ -10,169 +10,45 @@
...
+    # Check if debug mode is enabled via environment variable
+    debug_mode = os.environ.get("UF_FILE_SYNC_DEBUG", "").lower() in (
+        "1",
+        "true",
+        "yes",
     )
 
+    if debug_mode:
```
andriika: Would it be possible to move all this code out of sitecustomize.py so we can easily import and reuse it in Uber's internal sitecustomize.py?

The idea is to keep this sitecustomize.py as simple as:

```
from michelangelo.uniflow.core import file_sync

file_sync.run(downloader=file_sync.FsspecDownloader())
```

Then, in Uber's internal sitecustomize.py, we could do something like:

```
from michelangelo.uniflow.core import file_sync

class TerrablobDownloader:
    ...
file_sync.run(downloader=TerrablobDownloader())
```

python/michelangelo/uniflow/core/sitecustomize.py
yangfan0328: It's a mechanism to ensure file sync only runs during automatic Python initialization, not when someone runs the file directly. It means the sitecustomize script is being imported as a module. So when Python starts in a container, it automatically imports sitecustomize.py
```
__name__ = "michelangelo.uniflow.core.sitecustomize"
__name__ != "__main__" → TRUE
```
And it won't be run file through python command while executed directly as a script on testing/debugging stage. 
Check python sitecustomize explanations https://docs.python.org/3/library/site.html
 **Any other exception causes a silent and perhaps mysterious failure of the process.**

python/michelangelo/uniflow/core/sitecustomize.py
yangfan0328: Updated.

yangfan0328:
> @yangfan0328 Please add the extended docstring for the public interfaces that is introduced in open source repo.

Updated.

python/michelangelo/uniflow/core/sitecustomize.py
andriika: What will we break if we remove this condition?

Why would someone want to run sitecustomize.py directly? If it's for testing, then the direct run should trigger file sync. Otherwise, what is the point of running sitecustomize.py directly if it doesn't actually do anything?

yangfan0328: It won't break the code, discussed offline and removed this line.

andriika (approved):
TY!

PR #835, opened 2026-02-13
[MaCTL] Support multiple plugins

hwaminkim:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Support multiple plugins in mactl

**Why?**

Support multiple plugins in mactl

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/mactl.py
+++ b/python/michelangelo/cli/mactl/mactl.py
@@ -27,11 +27,11 @@ from michelangelo.cli.mactl.grpc_tools import list_services
 
 # Load configuration
 # Priority: env vars (highest) > RC file > defaults (lowest)
-_config = load_config()
+_CONFIG = load_config()
 
-ADDRESS = _config["address"]
-USE_TLS: bool = _config["use_tls"]
-METADATA = list(_config["metadata"].items())
+ADDRESS = _CONFIG["address"]
+USE_TLS: bool = _CONFIG["use_tls"]
+METADATA = list(_CONFIG["metadata"].items())
 
 METADATA_STUB = [*METADATA, ("ttl", "600")]
 
@@ -66,11 +66,13 @@ def read_plugin_modules(crd_name: str, plugin_dirs: list[str]) -> list[object]:
         plugin_dirs (list[str]): List of directories to search for plugins.
             The later directories have higher priority.
     """
+    _LOG.info("Read plugin modules from directories: %r", plugin_dirs)
     plugin_modules = []
     for i, plugin_dir in enumerate(plugin_dirs):
         plugin_module = read_module_from_file(crd_name, Path(plugin_dir), i)
         if plugin_module is not None:
             plugin_modules.append(plugin_module)
+    _LOG.info("Total %d plugin modules are loaded.", len(plugin_modules))
     return plugin_modules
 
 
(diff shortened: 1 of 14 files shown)


python/michelangelo/cli/mactl/mactl.py
```diff
@@ -58,12 +58,40 @@ def kebab_to_snake(name: str) -> str:
...
+
+    Args:
+        crd_name (str): The name of the CRD.
+        plugin_dirs (list[str]): List of directories to search for plugins.
+            The later directories have higher priority.
+    """
+    _LOG.info("Read plugin modules from directories: %r", plugin_dirs)
+    plugin_modules = []
+    for i, plugin_dir in enumerate(plugin_dirs):
+        plugin_module = read_module_from_file(crd_name, Path(plugin_dir), i)
```
yaoyueyue, line 72: This loads plugins from file system paths (e.g., plugins = ["/path/to/custom/plugins"] in config.toml). For our import back, we need to support Python module paths instead (e.g., plugins = ["uber.ai.michelangelo.sdk.mactl_plugins"]). Would you be open to enhancing read_plugin_modules() to detect and handle module paths, or should we handle this in a follow-up PR?

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -58,12 +58,40 @@ def kebab_to_snake(name: str) -> str:
...
-    if not plugin_dir.exists():
-        _LOG.info("Plugin directory does not exist: %r", plugin_dir)
+    _LOG.info("Check Plugin directory: %r", plugin_dir)
+    if not plugin_dir.exists() or not plugin_dir.is_dir():
+        _LOG.warning(
+            "Plugin base directory does not exist (or not a directory): %r", plugin_dir
+        )
+        return
+
+    plugin_dir = plugin_dir / "entity" / crd_name
```
yaoyueyue, line 90: Another question: how should we handle general plugins that aren't tied to a specific CRD entity? For example,  authentication (usso tokens). Currently plugins are loaded per-CRD under entity/{crd_name}/, but we'd need something like channel/main.py or global/main.py that runs once at startup before any CRD operations. Should we add support for these plugin types in this PR, or handle separately?

yaoyueyue (commented):
Thanks for adding multi-plugin support! Left comments

python/michelangelo/cli/mactl/mactl.py
hwaminkim: Yes. it would be updated in the follow-up PR. I intentionally skipped this part since this PR is already getting large. Thanks for remind it.

python/michelangelo/cli/mactl/mactl.py
hwaminkim: @yaoyueyue Good question. As you mentioned, for now it only supports per-entity function and class.
I guess the plugin function override should be changed to cover custom function level overriding. For example, users can set up the authentication module with their plugins like below.

```
"plugin":
  "cli.mactl.auth": "my.custom.plugin.auth"
```

yaoyueyue:
LGTM

yaoyueyue (approved):

PR #847, opened 2026-02-20
perf(javascript): switch test environment from jsdom to happy-dom

austingreco:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
- Replaced `jsdom` with `happy-dom` as the test DOM environment in vitest config
- Updated one test assertion that relied on jsdom's CSS color normalization (`rgb(255, 0, 0)` → `red`)
- Updated ResizeObserver mock comment (happy-dom provides it natively, but we keep the mock to prevent real observation side effects)

**Why?**
happy-dom is a lighter-weight DOM implementation that provides faster test execution. Local benchmarks:

| Metric | jsdom | happy-dom | Change |
|--------|-------|-----------|--------|
| Total duration | 31.46s | 27.57s | **-12%** |
| Table integration tests | 31.19s | 23.30s | **-25%** |
| Test execution time | 51.03s | 44.82s | **-12%** |

CI environments (slower hardware) tend to amplify these differences further.

**How did you test it?**
- Ran full test suite locally: 865 tests, 105 files, all passing
- `yarn vitest --run`

**Potential risks**
- happy-dom has minor behavioral differences from jsdom (e.g., doesn't normalize CSS color values to rgb). One test was updated to account for this. Other tests may surface similar minor differences if they rely on jsdom-specific behavior.

**Release notes**
- N/A (test infrastructure only, no user-facing changes)

**Documentation Changes**
- N/A

### DIFF
--- a/javascript/app/__tests__/app.test.tsx
+++ b/javascript/app/__tests__/app.test.tsx
@@ -13,16 +13,13 @@ vi.mock('@michelangelo-ai/rpc', () => ({
   normalizeTranscoderError: vi.fn(() => null),
 }));
 
-// AppNavBar renders both its desktop and mobile menu variants at once and switches between them
-// with a `matchMedia` query, which jsdom doesn't implement — the inactive variant reports as
-// hidden to the accessibility tree, so role queries need `hidden: true`.
+// AppNavBar renders both its desktop and mobile menu variants at once. The inactive variant
+// reports as hidden to the accessibility tree, so role queries need `hidden: true`.
 const HIDDEN = { hidden: true };
-const originalLocation = window.location;
 
 afterEach(() => {
   window.localStorage.clear();
   vi.unstubAllGlobals();
-  Object.defineProperty(window, 'location', { configurable: true, value: originalLocation });
   window.history.replaceState(null, '', '/');
 });
 
@@ -59,13 +56,6 @@ it('signs out to clear the cached dev profile', async () => {
       avatarUrl: 'https://example.com/a.png',
     })
   );
-  // jsdom's window.location.reload is non-configurable, so it can't be spied on directly;
-  // swap in a plain copy with reload replaced instead.
-  Object.defineProperty(window, 'location', {
-    configurable: true,
-    value: { ...window.location, reload: vi.fn() },
-  });
-
   render(<App />);
   await userEvent.click(
     (await screen.findAllByRole('button', { name: /Ada Lovelace/, ...HIDDEN }))[0]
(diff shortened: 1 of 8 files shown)


craigmarker:
How are you weighing the cons of adopting the happy-dom environment? My concern is primarily about less realistic/less established. What do you thiink?

austingreco:
happy-dom is pretty established now, but of course it doesn't have as wide support of all dom apis like jsdom. I liked it because it has far fewer dependencies and runs faster. Since we don't use any exotic dom apis it seemed pretty safe. 

Side note: I also found this the other day https://github.com/stacksjs/very-happy-dom which looks even cooler 😃

paulzim (approved):
## Review: #847 — perf(javascript): switch test environment from jsdom to happy-dom

**Verdict: Approve**

This is a re-review — I nudged this one earlier for two open items, both of which are now resolved:

1. **Craig's Feb 24 question** ("how are you weighing the cons of happy-dom — less established, less realistic DOM emulation?") got a substantive reply from @austingreco on 2026-07-17: happy-dom's fewer dependencies and faster runs are worth it since the suite doesn't touch exotic DOM APIs. Reasonable answer, on the record now.
2. **`yarn.lock` drift** is gone — the branch is rebased exactly onto `upstream/main`'s current HEAD (`661bdd6f`, confirmed via `git merge-base`). The lockfile diff is now clean: 64 insertions / 238 deletions, just jsdom's CSS-color-parsing transitive deps (`@asamuzakjp/css-color`, `@csstools/*`) dropping out in favor of happy-dom's own tree. The one stray `debug@4` range-collapse I'd flagged before is just normal yarn lockfile churn, not a real conflict.

**Re-verified by actually running it, not just reading the diff:**
- `yarn install --frozen-lockfile` against the rebased lockfile succeeds cleanly (needed Node 24.14.1 + Corepack's yarn 1.22.22 to match `package.json` engines — installed via `nvm install 24.14.1`).
- Full suite: **1368/1368 tests passing, 146/146 files**, on happy-dom (up from 865 at the time of the original review, since `main` picked up more tests in the meantime — not a regression, just growth).
- The `rgb(255, 0, 0)` → `red` test-assertion change in `default-cell-renderer.tests.tsx` still checks out as a real, expected consequence of happy-dom's CSS color serialization differing from jsdom's.
- `test-setup.ts` comment update and `vitest.config.ts` environment swap are unchanged from the original review — still exactly as described.

No blocking concerns. Nice to see both loose ends tied off cleanly in the same rebase.

PR #866, opened 2026-02-25
[mactl] add pagination limit support 

yaoyueyue:
## **What type of PR is this? (check all applicable)**
  - [x] Feature
  - [ ] Refactor
  - [ ] Bug Fix
  - [x] Optimization
  - [ ] Documentation Update

  ## **What changed?**

  Added pagination support and `--limit` flag to get and list commands with a default limit of 100 items to prevent gRPC RESOURCE_EXHAUSTED errors.

  **Examples:**
  ```bash
  # List pipelines (defaults to 100 items)
  ma pipeline list -n my-namespace

  # List pipelines with custom limit
  ma pipeline list -n my-namespace --limit 10

  # Get pipelines without name (defaults to 100 items)
  ma pipeline get -n my-namespace

  # Get pipelines with custom limit
  ma pipeline get -n my-namespace --limit 10
  ```

  ## **Why?**

  Prevents `grpc.RpcError: RESOURCE_EXHAUSTED` errors when running commands like `ma pipeline get -n ma-integration-test` and `ma project get -n ma-integration-test`. The error occurs when response messages exceed the max size limit (e.g., "Received message larger thanmax (XXXXX vs. 104857600)"). By implementing pagination with a default 100-item limit, we keep message sizes within acceptable bounds.

  ## **How did you test it?**

  Tested on both staging and OSS servers:
  - **Staging** (`ma-integration-test`): Verified limits of 3, 10, 100 work correctly
  - **OSS** (`ma-dev-test`): Verified limits of 1, 2, 10 work correctly
  - **Message size reduction**: Confirmed ~1KB per pipeline (3KB for 3 items vs 102KB for 100 items)
  - **Warning behavior**: Verified warning appears when `result_count == limit`, not when `result_count < limit`

  ## **Potential risks**

  None 

  ## **Release notes**

ma get/list commands now default to 100 items per request to prevent RESOURCE_EXHAUSTED errors. Use pagination options to retrieve  additional results.

  ## **Documentation Changes**

  None required - CLI help text already includes `--limit` flag documentation.

### DIFF
--- a/python/michelangelo/cli/mactl/crd.py
+++ b/python/michelangelo/cli/mactl/crd.py
@@ -26,7 +26,6 @@ from michelangelo.cli.mactl.grpc_tools import (
     get_message_class_by_name,
     get_methods_from_service,
 )
-from michelangelo.gen.api.list_pb2 import ListOptionsExt, PaginationSpec
 
 _LOG = getLogger(__name__)
 METADATA_STUB = []
@@ -220,6 +219,10 @@ def crd_method_call(crd_method_info, request_input: Message) -> Message:
 
     method_fullname = f"/{crd_method_info.crd_full_name}/{crd_method_info.method_name}"
     _LOG.info("Method fullname for gRPC call: %s", method_fullname)
+
+    request_size = len(request_input.SerializeToString())
+    _LOG.info("gRPC request size: %d bytes", request_size)
+
     stub_method = crd_method_info.channel.unary_unary(
         method_fullname,
         request_serializer=crd_method_info.input_class.SerializeToString,
@@ -230,6 +233,9 @@ def crd_method_call(crd_method_info, request_input: Message) -> Message:
         metadata=METADATA_STUB,
         timeout=30,
     )
+
+    response_size = len(response.SerializeToString())
+    _LOG.info("gRPC response size: %d bytes (%d KB)", response_size, response_size // 1024)
     _LOG.info("Stub method completed (%r): %r", type(response), response)
     return response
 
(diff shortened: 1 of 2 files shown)


badcount (approved):

python/michelangelo/cli/mactl/crd.py
```diff
@@ -500,6 +526,49 @@ def __init__(self, name: str, full_name: str, metadata: list):
...
+                            "default": 100,
+                            "help": (
+                                "Maximum number of items to return when listing "
+                                "(default: 100)"
+                            ),
+                        },
+                    },
+                ],
+            },
+            "list": {
```
hwaminkim: @yaoyueyue We would not expose the list function for users, isn't it?
I remember that this `func_signature` attribute is only for exposed user sub commands

python/michelangelo/cli/mactl/crd.py
```diff
@@ -330,6 +346,16 @@ def list_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Mes
     raw_elems = results[next(iter(results))]
 
     print_list_formatted(raw_elems.items)
+
+    # Show warning if we got exactly the limit (there might be more)
+    if len(raw_elems.items) == limit:
```
hwaminkim, line 351: To me, this message is too verbose. would we skip this information message for now? as the same as the current mactl implementation?
Ideally, I think we can show `...` at the end of the list, only when there are more lists than configured limit.

python/michelangelo/cli/mactl/crd.py
```diff
@@ -500,6 +526,49 @@ def __init__(self, name: str, full_name: str, metadata: list):
...
                     },
+                    {
+                        "func_signature": Parameter(
+                            "limit", Parameter.POSITIONAL_OR_KEYWORD, default=100
+                        ),
+                        "args": ["--limit"],
+                        "kwargs": {
+                            "dest": "limit",
+                            "type": int,
+                            "default": 100,
```
hwaminkim, line 537: would we set the default value as `10`? (same as the current mactl)

hwaminkim (changes requested):

python/michelangelo/cli/mactl/crd.py
yaoyueyue: Good catch, updated

python/michelangelo/cli/mactl/crd.py
hwaminkim: Note: Discussed with team in offline and found that the current mactl is using the same message. decided to keep this message for now.

python/michelangelo/cli/mactl/crd.py
hwaminkim: same as above. keep as the same config as current mactl

hwaminkim (approved):

PR #875, opened 2026-03-04
Rename variables to remove cadence words

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Rename variables to remove cadence

**Why?**
Make it more generic

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


yangfan0328 (approved):

go/worker/plugins/utils/utils.go
```diff
@@ -28,8 +28,8 @@ const (
 	SparkAppFailedCondition  string = "SparkAppFailed"
 )
 
-const CadenceLongTimeout = time.Hour * 24 * 365 * 10 // 10 years, practically - no timeout
-const CadenceLongRetry = time.Hour * 24 * 365 * 10   // 10 years, practically - no timeout
+const LongTimeout = time.Hour * 24 * 365 * 10 // 10 years, practically - no timeout
+const LongRetry = time.Hour * 24 * 365 * 10   // 10 years, practically - no timeout
 
 var CadenceDefaultNonRetriableErrorReasons = []string{
```
andriika: You might want to rename this one as well.

go/worker/plugins/utils/utils.go
andriika, line 36: I'm not entirely sure, but it looks like we should also add temporal system errors to this list.

badcount: this is for cadence only, we keep it, for temporal they use application error, there is not fixed term for retry, so we don't do it.

andriika (approved):

PR #880, opened 2026-03-07
Make filter menu scrollable

donovanlopez:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

### Filter menu behind the header

Users had a hard time selecting filters as a result. This behavior would happen when there are too many filters. If the filter menu reaches the bottom of the viewport, it will jump to the top and be behind the header. 

The fix was to make the filter menu have a max height so that it'd be scrollable and open to the right of the "Add Filter" button. 

### The filter popover edges
All the filters have rounded edges so that they match the filter menu popover

**Why?**
The bug for having the filter menu behind the header is a customer pain point as they can't use the platform effectively. By making the filter menu smaller and scrollable, we avoid the issue as the filter menu won't be tall enough to reach the header. Also it'll take up less real estate on the screen, we can anticipate a growing list of filters and there's less jumpiness when going back and forth when between filters and the filter menu.

As for making the filter popover edges all rounded, that's just for consistency with the filter menu which always had rounded edges.

**How did you test it?**
I tested it locally. Because we're moving the placement and styling of the filter menu, I don't see a need for unit tests.

**Potential risks**
The worst case is the filter menu is somehow rendered incorrectly.

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/table/components/table-action-bar/components/table-active-filter-tag-list/table-active-filter-tag.tsx
+++ b/javascript/packages/core/components/table/components/table-action-bar/components/table-active-filter-tag-list/table-active-filter-tag.tsx
@@ -41,6 +41,14 @@ export function ActiveFilterTag<TData = unknown>(props: ActiveFilterTagProps<TDa
   return (
     <StatefulPopover
       placement={PLACEMENT.bottomLeft}
+      overrides={{
+        Body: {
+          style: ({ $theme }: { $theme: Theme }) => ({
+            borderRadius: $theme.borders.radius300,
+            overflow: 'hidden',
+          }),
+        },
+      }}
       content={({ close }) => (
         <FilterComponent
           column={column}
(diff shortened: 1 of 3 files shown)


craigmarker (approved):
LGTM, though I'm not seeing what code change results in the filter menu having rounded corners 🤔

donovanlopez:
> LGTM, though I'm not seeing what code change results in the filter menu having rounded corners 🤔

@craigmarker I was confused too. I was commenting the changes file by file and it was because `overflow: hidden` was added:

https://github.com/michelangelo-ai/michelangelo/blob/4c4a4ca2fc944b2d02cb07198613b8b8dd8451ea/javascript/packages/core/components/table/components/table-action-bar/components/table-filter-menu/table-filter-menu.tsx#L30-L32

Apparently it's because the content inside the filter menu was overflowing and resulted in the edges being sharp. This popover always had rounded edges so using `overflow: hidden` will show the rounded edges

craigmarker:
> > LGTM, though I'm not seeing what code change results in the filter menu having rounded corners 🤔
> 
> @craigmarker I was confused too. I was commenting the changes file by file and it was because `overflow: hidden` was added:
> 
> https://github.com/michelangelo-ai/michelangelo/blob/4c4a4ca2fc944b2d02cb07198613b8b8dd8451ea/javascript/packages/core/components/table/components/table-action-bar/components/table-filter-menu/table-filter-menu.tsx#L30-L32
> 
> Apparently it's because the content inside the filter menu was overflowing and resulted in the edges being sharp. This popover always had rounded edges so using `overflow: hidden` will show the rounded edges

Makes sense to me! Thanks for the clarification

PR #890, opened 2026-03-10
Extend select field component

konukselim:
Extended select-field component to support:

- Getting options asynchronously
- Arbitrary option id types(including objects) 
  -  Internally, IDs are serialized to strings for BaseUI compatibility and resolved back to their originals for form state via a lookup map.
- Auto-clear invalid field values
  - When the current form value (from defaultValue, initialValue, or initialValues) doesn't match any available option, it's automatically cleared. In multi-select mode, only the invalid entries are removed. Skipped when isLoading is true (to avoid clearing before async options arrive) or when creatable is true (where unmatched values are intentional). 
- Add maxOptions prop to limit the number of visible dropdown options

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/fields/select/__tests__/select-field.test.tsx
+++ b/javascript/packages/core/components/form/fields/select/__tests__/select-field.test.tsx
@@ -311,4 +311,142 @@ describe('SelectField', () => {
 
     expect(screen.getByText('Choose the appropriate priority level')).toBeInTheDocument();
   });
+
+  it('submits object id as the field value', async () => {
+    const user = userEvent.setup();
+    const onSubmit = vi.fn();
+
+    const objectOptions = [
+      { id: { tier: 'free', limit: 100 }, label: 'Free Tier' },
+      { id: { tier: 'pro', limit: 10000 }, label: 'Pro Tier' },
+    ];
+
+    render(
+      <>
+        <SelectField name="plan" label="Plan" options={objectOptions} />
+        <button type="submit">Submit</button>
+      </>,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({ onSubmit }),
+      ])
+    );
+
+    const select = screen.getByRole('combobox', { name: 'Plan' });
+    await user.click(select);
+    await user.click(screen.getByRole('option', { name: 'Pro Tier' }));
+
+    await user.click(screen.getByRole('button', { name: 'Submit' }));
+    await waitFor(() =>
+      expect(onSubmit).toHaveBeenCalledWith(
+        { plan: { tier: 'pro', limit: 10000 } },
+        expect.anything(),
+        expect.anything()
+      )
+    );
+  });
+
+  it('clears initial value that does not match any option', async () => {
+    const user = userEvent.setup();
+    const onSubmit = vi.fn();
+
+    render(
+      <>
+        <SelectField name="priority" label="Priority" options={options} />
+        <button type="submit">Submit</button>
+      </>,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({ initialValues: { priority: 'nonexistent' }, onSubmit }),
+      ])
+    );
+
+    expect(screen.queryByText('nonexistent')).not.toBeInTheDocument();
+
+    await user.click(screen.getByRole('button', { name: 'Submit' }));
+    await waitFor(() =>
+      expect(onSubmit).toHaveBeenCalledWith({}, expect.anything(), expect.anything())
+    );
+  });
+
+  it('keeps only valid values in multi-select when some are invalid', async () => {
+    const user = userEvent.setup();
+    const onSubmit = vi.fn();
+
+    render(
+      <>
+        <SelectField name="priority" label="Priority" options={options} multi />
+        <button type="submit">Submit</button>
+      </>,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({
+          initialValues: { priority: ['low', 'nonexistent', 'high'] },
+          onSubmit,
+        }),
+      ])
+    );
+
+    await waitFor(() => {
+      expect(screen.getByText('Low Priority')).toBeInTheDocument();
+      expect(screen.getByText('High Priority')).toBeInTheDocument();
+    });
+    expect(screen.queryByText('nonexistent')).not.toBeInTheDocument();
+
+    await user.click(screen.getByRole('button', { name: 'Submit' }));
+    await waitFor(() =>
+      expect(onSubmit).toHaveBeenCalledWith(
+        { priority: ['low', 'high'] },
+        expect.anything(),
+        expect.anything()
+      )
+    );
+  });
+
+  it('limits the number of visible options with maxOptions', async () => {
+    const user = userEvent.setup();
+    const manyOptions = Array.from({ length: 10 }, (_, i) => ({
+      id: `opt-${i}`,
+      label: `Option ${i}`,
+    }));
+
+    render(
+      <SelectField name="choice" label="Choice" options={manyOptions} maxOptions={3} />,
+      buildWrapper([getBaseProviderWrapper(), getIconProviderWrapper(), getFormProviderWrapper()])
+    );
+
+    const select = screen.getByRole('combobox', { name: 'Choice' });
+    await user.click(select);
+
+    expect(screen.getAllByRole('option')).toHaveLength(3);
+  });
+
+  it('does not clear value while options are loading', async () => {
+    const user = userEvent.setup();
+    const onSubmit = vi.fn();
+
+    render(
+      <>
+        <SelectField name="priority" label="Priority" options={[]} isLoading />
+        <button type="submit">Submit</button>
+      </>,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({ initialValues: { priority: 'low' }, onSubmit }),
+      ])
+    );
+
+    await user.click(screen.getByRole('button', { name: 'Submit' }));
+    await waitFor(() =>
+      expect(onSubmit).toHaveBeenCalledWith(
+        { priority: 'low' },
+        expect.anything(),
+        expect.anything()
+      )
+    );
+  });
 });
--- a/javascript/packages/core/components/form/fields/select/__tests__/select-field.test.tsx
+++ b/javascript/packages/core/components/form/fields/select/__tests__/select-field.test.tsx
@@ -450,6 +450,82 @@ describe('SelectField', () => {
     );
   });
 
+  it('matches object id regardless of key order', async () => {
+    const user = userEvent.setup();
+    const onSubmit = vi.fn();
+
+    const objectOptions = [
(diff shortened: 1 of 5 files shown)


javascript/packages/core/components/form/fields/select/__tests__/select-field.test.tsx
```diff
@@ -311,4 +311,142 @@ describe('SelectField', () => {
...
+      buildWrapper([getBaseProviderWrapper(), getIconProviderWrapper(), getFormProviderWrapper()])
+    );
+
+    const select = screen.getByRole('combobox', { name: 'Choice' });
+    await user.click(select);
+
+    expect(screen.getAllByRole('option')).toHaveLength(3);
+  });
+
+  it('does not clear value while options are loading', async () => {
```
craigmarker, line 427: maybe need another test to not clear value if it is creatable

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -21,40 +38,73 @@ export const SelectField: React.FC<SelectFieldProps> = ({
...
+  // Clear field value when it doesn't match any available option.
+  // Deps intentionally exclude input/multi to avoid re-running on every value change,
+  // which would loop since we call onChange inside.
+  useEffect(() => {
+    if (isLoading || creatable) return;
+
+    const currentValue = input.value;
+    if (!currentValue || (Array.isArray(currentValue) && currentValue.length === 0)) return;
+
+    if (multi) {
```
craigmarker, line 63: This applies to more than one place, but I'm confused by how `multi` interacts with typing value as `V | V[]`.

My understanding is that BaseWeb's Select component expects value as an array. We want to adhere to that but also support populating form values with a select field that sets value to a non-array value.

With this in mind, can we safely always type input.value as V when multi = false and input.value as V[] when multi = true?

javascript/packages/core/components/form/fields/select/types.ts
```diff
@@ -1,15 +1,17 @@
...
-export interface SelectFieldProps extends BaseFieldProps<string | string[] | number | number[]> {
-  options: SelectOption[];
+export interface SelectFieldProps<V = string | number> extends BaseFieldProps<V | V[]> {
+  options: SelectOption<V>[];
   clearable?: boolean;
   searchable?: boolean;
   multi?: boolean;
   creatable?: boolean;
+  isLoading?: boolean;
+  maxOptions?: number;
```
craigmarker: I think there could be some confusion between `maxOptions` and `options`. What do you think about `visibleOptionLimit` or `dropdownLimit`? Plus a JSDoc

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useCallback, useMemo } from 'react';
-import { OnChangeParams, Select } from 'baseui/select';
+import { useEffect, useMemo } from 'react';
+import { filterOptions as defaultFilterOptions, OnChangeParams, Select } from 'baseui/select';
```
craigmarker: I'm confused by the `as defaultFilterOptions` when there is only one filterOptions which is controlled by BaseUI. I think you could either just `import {filterOptions}` or if you think the aliasing should be maintained, perhaps `import {filterOptions as baseUIFilterOptions}`.

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -8,7 +8,24 @@ import { formatSelectedValue } from './format-selected-value';
...
 
-export const SelectField: React.FC<SelectFieldProps> = ({
+/**
+ * BaseUI requires option IDs to be string | number. To support arbitrary ID types
+ * (including objects), we serialize IDs to strings for BaseUI and resolve them back
+ * to their originals for form state via a lookup map.
+ */
+const serialize = JSON.stringify;
+
+function buildOptionLookup<V>(options: SelectOption<V>[]) {
```
craigmarker: I'm having a little trouble following the normalizing/lookup/serializing logic, and I have a proposal. Can you inline a memoized object into the SelectField component: 

```  
const { baseUIOptions, findByValue, findByKey } = useMemo(() => {
    const map = new Map<string, SelectOption<V>>();
    const adapted = options.map((opt) => {
      const key = JSON.stringify(opt.id);
      map.set(key, opt);
      return { id: key, label: opt.label, disabled: opt.disabled };
    });
    return {
      baseUIOptions: adapted,
      // Use findByValue when interacting with form state eg input.value
      findByValue: (v: V) => map.get(JSON.stringify(v)),

      // Use findByKey when interacting with SelectOption from BaseUI
      findByKey: (key: string) => map.get(key),
    };
  }, [options]);
```

This does a few things:

1. Removes the requirement for callers to know specifically how Base Select component expects options to be formatted.
2. Consolidates the lookup/resolveOriginalId/serialize functionality

What do you think?

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -21,40 +38,73 @@ export const SelectField: React.FC<SelectFieldProps> = ({
...
+      input.onChange(resolveOriginalId(String(params.value[0].id)) as V | V[]);
+    } else {
+      input.onChange('' as V | V[]);
+    }
+  };
 
-      return { id: item, label: String(item) }; // Creatable options will fall into this case
-    });
-  }, [input.value, options]);
+  const selectedValue = useMemo(() => {
```
craigmarker: Consider renaming to clarify the underlying formatting that is happening within the memoization

```suggestion
  const baseUIValue = useMemo(() => {
```

Or selectedItems to clarify that it becomes a list?

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -8,7 +8,24 @@ import { formatSelectedValue } from './format-selected-value';
 
 import type { SelectFieldProps, SelectOption } from './types';
 
-export const SelectField: React.FC<SelectFieldProps> = ({
+/**
+ * BaseUI requires option IDs to be string | number. To support arbitrary ID types
+ * (including objects), we serialize IDs to strings for BaseUI and resolve them back
+ * to their originals for form state via a lookup map.
+ */
+const serialize = JSON.stringify;
```
craigmarker: Something to consider with JSON.stringify is that JSON.stringify({a: 'b', c: 'd'}) !== JSON.stringify({c: 'd', a: 'b'})

craigmarker (changes requested):

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: Good call, I've refactored prop types into a discriminated union on  `multi`, so consumers now get the correct type for `defaultValue` and `initialValue` depending on the mode. Also removed redundant `as V | V[]` casts.

> With this in mind, can we safely always type input.value as V when multi = false and input.value as V[] when multi = true?

Yes, conceptually that is correct. When multi = false, the form value is always V. When multi = true, it's always V[].

However, we can't fully express this at the useField call site because useField takes a single type parameter and we can't conditionally call hooks. So useField<V | V[]> is the one place where the union persists.

The `as V | V[]` casts in each multi branch have been removed (they were redundant). The only remaining cast is `as V | V[]` for clearing the field, which is a react-final-form convention where empty string signals "no value."

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: That is a stale import - I imported `filterOptions` with as since there was another filterOptions logic in the beginning. Thanks for catching that. Removing `as`.

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: I actually named it exactly as `baseUIValue` in the beginning but I didn't want to specify design language inside the code. I'd prefer `baseUIValue` over `selectedItems`.

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: Couldn't think of a better approach then JSON.stringify to support arbitrary option types. I think the responsibility of eliminating duplicate options for arbitrary types should be on consumers. I'd love to hear your opinions on this

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: I tried to convert the implementation to your suggestion. I also found the logic hard to read today. But I'm not sure if the updated one is more readable and consolidated.

Also I looked this comment last so the changes I applied for the previous comments might not be as good as my intentions. It'd be great if you review my changes as a whole rather than focusing on my replies separately for each comment.

javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -21,40 +21,85 @@ export const SelectField: React.FC<SelectFieldProps> = ({
...
-      if (multi) {
-        const values = params.value.map((item) => item.id);
-        input.onChange(values as string[] | number[]);
-      } else {
-        const value = params.value.length > 0 ? params.value[0].id : '';
-        input.onChange(value);
+  const { baseUIOptions, findByValue, findByKey } = useMemo(() => {
+    const map = new Map<string, SelectOption<V>>();
+    const adapted = options.map((opt) => {
+      const key = JSON.stringify(opt.id);
```
craigmarker: I think the code becomes a little more clear when we establish that our `OnChangeParams` _only_ supports string options. Then we can safely remove `resolveOriginalId` and use `findByKey().id` directly.

javascript/packages/core/components/form/fields/select/select-field.tsx
craigmarker: seems reasonable. i suppose i was wondering if the options list would ever diverge from form state. For example, end-user creates a resource and selects option[0] for the field random.object.field (which equals `{a: 'b', c: 'd'})`). When the end-user goes to update that resource, a Get request returns `random.object.field = {c: 'd', a: 'b'}`

In this case, would the select-field want to remove the user's selection?

javascript/packages/core/components/form/fields/select/select-field.tsx
craigmarker: Thanks for the taking a look at this change. I'm somewhat partial to this particular change, but I think there is still room for improvement in consolidating the baseui/select <--> react-final-form values. I'm just not exactly sure how to do that.

Feel free to keep this change or revert to your original implementation. Maybe something will come to us in the future!

craigmarker (approved):

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: I liked your suggestion - I also think there is a room for improvement but for now will leave it as is.

javascript/packages/core/components/form/fields/select/select-field.tsx
konukselim: That's a real bug. I was only considering object key order of the options previously, but your example is a realistic scenario we should definitely handle.

I created a util to serialize keys using JSON.stringify but it also sorts the objects before turning them into json strings. This way the component will be more robust.

Also added the test-scenarios.

craigmarker (approved):

PR #891, opened 2026-03-10
Add StickyFooter component with Form footer prop

craigmarker:
Add StickyFooter, a fixed viewport-bottom bar with left and right content slots, sized dynamically to prevent page content from being obscured when scrolled to the bottom.

Expose this through a new `footer` prop on Form. The shorthand { left, right } object automatically wraps content in StickyFooter, covering the common case of a submit button pinned to the viewport. A raw ReactNode is also accepted as an escape hatch for full control over the footer element.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

https://github.com/user-attachments/assets/172b4564-9673-4180-8762-1b4f7e1a9730

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/components/sticky-footer/sticky-footer.tsx
+++ b/javascript/packages/core/components/form/components/sticky-footer/sticky-footer.tsx
@@ -1,0 +1,34 @@
+import { useEffect, useRef, useState } from 'react';
+
+import { StickyFooterContainer, StickyFooterSlot, StickyFooterSpacer } from './styled-components';
+
+import type { StickyFooterProps } from './types';
+
+export function StickyFooter({ leftContent, rightContent }: StickyFooterProps) {
+  const footerRef = useRef<HTMLElement>(null);
+  const [footerHeight, setFooterHeight] = useState(0);
+
+  // Dynamically track footer height so the spacer always matches, preventing
+  // content from being hidden behind the fixed footer when scrolled to bottom.
+  useEffect(() => {
+    const footer = footerRef.current;
+    if (!footer) return;
+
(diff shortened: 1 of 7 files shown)


javascript/packages/core/components/form/components/sticky-footer/sticky-footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+    observer.observe(footer);
+    return () => observer.disconnect();
+  }, []);
+
+  return (
+    <>
+      <StickyFooterSpacer $height={footerHeight} />
+      <StickyFooterContainer ref={footerRef}>
+        <StickyFooterSlot>{leftContent}</StickyFooterSlot>
+        <StickyFooterSlot>{rightContent}</StickyFooterSlot>
```
donovanlopez, line 30: would we ever want to be more flexible with what the content of the sticky footer could be since only takes left and right content?

donovanlopez (approved):
LGTM! I had one question but I wouldn't call it a blocker.

javascript/packages/core/components/form/components/sticky-footer/sticky-footer.tsx
craigmarker: i could envision some scenarios where more flexibility would be beneficial! it's a good idea! I don't know of a use case at the moment. And if a requirement emerges, it may be best for the use case to leverage a custom footer unless there's a clear abstraction that can be implemented 👍🏻

PR #892, opened 2026-03-10
Replace broad JavaScript style-guide skill with focused Claude skills

craigmarker:
The previous approach used a single 330-line style-guide skill that loaded into context for every code task. skill-creator evaluation revealted that 35-50% of its content was irrelevant for its specific task. Its broad "reviewing code" trigger made it fire even when no code would be written.

Replaces it with 6 granular skills (testing, React, TypeScript, UI principles, file structure, documentation) each with intent-based triggers that fire at task planning time. The key change: descriptions are phrased around what the user is trying to do ("implementing a hook") rather than the file action already underway ("writing a .ts file"), so skills engage before decisions are made, not after.

Trims CLAUDE.md from 198 to ~75 lines. Style-guide content moves into skills; CLAUDE.md now covers architecture (component-driven first, config-driven layered on top, provider pattern), the full yarn command reference, and a skills index. Adds evals.json to the skills workspace so trigger behavior can be re-validated as skills evolve.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Results

| # | Prompt | Old skill | New skills |
|---|--------|-----------|------------|
| 1 | Write tests for `struct-utils.ts` | fires — ~50% irrelevant content | `testing-standards`, `typescript-cookbook`, `file-directory-structure` |
| 2 | Author form error test cases in `form.test.tsx` | fires — ~30% irrelevant content | `testing-standards`, `ui-coding-principles`, `react-component-development`, `typescript-cookbook` |
| 3 | Implement `ArrayField` component | fires | all 6 skills |
| 4 | Add `use-field-dirty.ts` hook | fires | all 6 skills |
| 5 | Refactor shared logic between components | fires | 5/6 skills (`testing-standards` correctly skipped — no tests requested) |
| 6 | Explain `use-error-list.ts` *(read-only)* | ❌ **false positive** | ✅ no skills |

### Key findings

**Nested directory discovery.** [Claude skills documentation](https://code.claude.com/docs/en/skills#automatic-discovery-from-nested-directories) suggests that skills are automatically discovered when modifying files in nested directories, "For example, if you’re editing a file in packages/frontend/, Claude Code also looks for skills in packages/frontend/.claude/skills/. This supports monorepo setups where packages have their own skills." Relocating javascript-specific skills to the `javascript/` monorepo root directory ensures that javascript-specific rules are applied to all javascript code changes regardless of where the Claude code session is started.

**Selective triggering.** On eval 5 (a refactoring question, no tests requested), `testing-standards` correctly does not fire with the new skills. The old skill fired on all coding tasks regardless of whether testing was relevant.

**False positive on read-only tasks.** The old skill's trigger — `"Use when writing, reviewing, or refactoring code"` is broad enough to fire on a pure code explanation request ("explain how each line works"). The new `ui-coding-principles` skill explicitly excludes this: `"Skip only for pure read-only questions like explaining or debugging without changes."`

**Proactive triggering.** The new `testing-standards` description fires when *implementing* code that will need tests, not only when explicitly asked to write tests. On eval 3 (implement a component), it correctly fires because new components need test coverage — the old skill had no equivalent distinction.


javascript/.claude/skills/file-directory-structure/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+user-invocable: false
+---
+
+# File & Import Organization
+
+## Import Strategy
+
+- **Direct imports only** — no `index.ts` barrel exports; import directly from source files
+- **Co-locate related code** — types, context, hooks, tests, and utils live alongside the component they belong to
+- **Flat over nested** — prefer flat structures when file count is manageable (~10 files)
```
pixelgrid: this might be ambiguous when the number of files exceeds 10.
needs some testing to see if it will fallback to creating folders at that point

same for hooks bellow. will it refactor all current hooks when it reaches the limit or introduce two different patterns of directory structure

javascript/.claude/skills/style-guide/SKILL.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Michelangelo JavaScript/TypeScript Style Guide
+
+Entry point for coding standards. Claude loads the relevant sub-skill automatically based on what you're working on.
+
+## Sub-Skills
+
+| Working on | Skill |
+|---|---|
+| Any code | `principles` |
```
pixelgrid: should the skills in the second column match a file path?

javascript/.claude/skills/testing-standards/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+| Mock                                          | Don't mock                                  |
+| --------------------------------------------- | ------------------------------------------- |
+| External APIs, RPC calls, server dependencies | Internal hooks and components               |
+|                                               | React context                               |
+|                                               | Well-tested utilities                       |
+|                                               | Dependencies already comprehensively tested |
+
+## Test Organization
+
+- Place tests in the closest `__tests__/` directory to the tested code
```
pixelgrid: "create if it does not exist" to avoid cases where it traverses the tree up until it finds the folder?

javascript/.claude/skills/testing-standards/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+- Use `describe` blocks only when tests benefit from grouping (shared setup, related assertions)
+- Keep tests flat when individual test names provide enough context
+- Prefer `expect(fn(args)).toEqual(result)` over assigning to intermediate variables
+
+## Anti-Patterns
+
+- ❌ Testing implementation details (`container.firstChild`, verifying props passed to children)
+- ❌ Duplicating behavior across unit and integration tests — pick the right level
+- ❌ Testing scenarios already covered by the dependency being used
+- ❌ Module-level wrapper constants — don't define `const wrapper = buildWrapper([...])` at the top of a test file. It creates hidden coupling; if one test needs a different provider, diverging becomes awkward. Inline the wrapper per test instead:
```
pixelgrid: for a big test file where you introduce a new provider that now needs a wrapper this will create a big change in the file that is a no-op as far a testing goes

javascript/.claude/skills/typescript-cookbook/SKILL.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Type Design
+
+- **`unknown` over `any`** — better compile-time safety for new types
+- **Create focused types** — map from generated types, include only the properties you need
+- **No type suppression** — unless stress-testing with invalid input
+
+## Naming
+
+- **No T suffix** — use `Props`, `User`, not `TProps`, `TUser`
```
pixelgrid: suffix -> PropsT, UserT
wording needs to change

pixelgrid (commented):
we should also add a rule for preferring named exports instead of default exports

javascript/.claude/skills/file-directory-structure/SKILL.md
craigmarker: I should theoretically be able to test this with the skill-creator eval framework. I'll take a look.

My guess right now is that, in actual day-to-day practice, Claude will just ignore this.

javascript/.claude/skills/testing-standards/SKILL.md
craigmarker: seems reasonable to me!

javascript/.claude/skills/testing-standards/SKILL.md
craigmarker: Thanks for the feedback!

The churn tradeoff is real, but intentional — inline wrappers make it clear what a component actually needs, and each test is self-contained. When you need to diverge (one test needs extra context), it's straightforward.

For the "new provider, lots of components need it" case — that's a good candidate for an agentic solution 🤓 . Claude can find all affected test files and add the wrapper in one pass. Hopefully something like this would reduce the mechanical work in practice.

What do you think?

javascript/.claude/skills/style-guide/SKILL.md
craigmarker: oh, i actually think i can just remove the style-guide, since this should all be contained in CLAUDE.md. good catch!

javascript/.claude/skills/file-directory-structure/SKILL.md
craigmarker: I updated the instruction to try to capture instances where claude is adding several files/several hooks and utilities:

```
**Flat over nested** — start flat; introduce a subdirectory only when you have multiple files of the same type (e.g., several hooks → `hooks/`, several utils → `utils/`
```

I highly doubt that the agent is going to count the number of files in a directory.

javascript/.claude/skills/testing-standards/SKILL.md
```diff
@@ -0,0 +1,52 @@
...
+name: Testing standards
+description: 'Trigger when testing is any part of the task — writing tests, adding coverage, adding test cases, or implementing code that will need tests. Project-specific mocking rules that OVERRIDE standard defaults.'
+user-invocable: false
+---
+
+# Testing Patterns
+
+## What to Test
+
+- **User-facing behavior** — `screen.getByText()`, `screen.getByRole()`, not `container.firstChild`
```
donovanlopez: Are we strict on not using `getByTestId`? I've seen recommendations that are against using it as much as possible since it doesn't reflect actual user behavior.

donovanlopez (approved):

javascript/.claude/skills/testing-standards/SKILL.md
konukselim: personally I find `data-test-{x}` attributes quite polluting so would vote against using `getByTestId` as much as possible too

konukselim (approved):
How about adding `explain-code` skill as well? Not sure how would we want to structure `SKILL.md` though.

javascript/.claude/skills/testing-standards/SKILL.md
pixelgrid: i think the best practice is that you use test-ids if there is no other way to match the same way a user would ByText, ByLabel, ByRole are usually preferred

pixelgrid (approved):

javascript/.claude/skills/testing-standards/SKILL.md
craigmarker: thanks for calling this out! I like to avoid getByTestId as well, and I agree with @pixelgrid that sometimes there is no way to use another query type.

I added a **Querying Element** section to testing-standards. PTAL! 

https://github.com/michelangelo-ai/michelangelo/pull/892/changes/a4bdcdb6cf972e61d0024b4b0b2eb57f843169a4

konukselim: LGTM! Thanks for adding that section.

konukselim (approved):

PR #893, opened 2026-03-10
Add hook for programmatic form mutations

craigmarker:
Form consumers need a way to imperatively set field values and trigger submission from outside standard form controls, For example, when actions like state transitions mutate multiple fields before submitting.

The new useForm() hook exposes change() and submit() from the underlying form library, alongside the existing fieldRegistry, through a single unified hook. The abstraction is kept at the hook boundary so the underlying final-form library remains an implementation detail.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/hooks/use-form.ts
+++ b/javascript/packages/core/components/form/hooks/use-form.ts
@@ -1,0 +1,15 @@
+import { useForm as useReactFinalForm } from 'react-final-form';
+
+import { useFormContext } from '#core/components/form/form-context';
+
+import type { FormApi } from '../types';
+
+export function useForm(): FormApi {
+  const { fieldRegistry } = useFormContext();
+  const form = useReactFinalForm();
+  return {
+    fieldRegistry,
+    change: form.change,
+    submit: form.submit,
+  };
+}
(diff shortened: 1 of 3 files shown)


javascript/packages/core/components/form/hooks/use-form.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { FormApi } from '../types';
+
+export function useForm(): FormApi {
+  const { fieldRegistry } = useFormContext();
+  const form = useReactFinalForm();
+  return {
+    fieldRegistry,
+    change: form.change,
+    submit: form.submit,
+  };
```
konukselim, line 14: I feel like this hook will be the only export or are we going to export `use-field` separately?  Do you think if it makes to include field subscriptions to `useForm` API?

craigmarker: I'm not sure I fully understand the question. What are the field subscriptions you're referring to?

konukselim (approved):

javascript/packages/core/components/form/hooks/use-form.ts
craigmarker: We'll export both useField and useForm for form consumers to build forms with

PR #894, opened 2026-03-11
Add support for repeatable form fields

craigmarker:
Introduce components and a hook for managing variable-length arrays of structured form fields.

useArrayField centralizes minItems enforcement and exposes an `entries` array — each entry pairs a stable UUID `id` with an `indexedFieldPath`. UUIDs are used as React key props so that removing an item causes React to destroy and recreate the shifted components rather than reuse DOM nodes that carry stale input state from their previous position.

ArrayFormGroup renders repeated card-style sections with numbered titles and a trash button. ArrayFormRow renders inline field rows with a compact remove button. Both use a render-prop children API so consumers control field content, and each implements its own add/remove buttons since the differences between them don't justify a shared abstraction.

FormGroup previously rendered endEnhancer inline next to the title with a small gap. It now uses space-between to push action buttons like the remove button to the far right of the card header regardless of title length. FormGroup also gains an overrides prop (BoxOverrides) so ArrayFormGroup can suppress the default bottom margin, which is instead provided by the add button below.

FormRow now provides support for unwrapping fragments. This allows ArrayFromRow to reuse FormRow component. This ensures ArrayFormRow renders fields in a row instead of a stack.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

https://github.com/user-attachments/assets/2264f0d3-d417-40d3-a86e-4e9e350cfef8

### DIFF
--- a/javascript/packages/core/components/form/hooks/use-array-field.ts
+++ b/javascript/packages/core/components/form/hooks/use-array-field.ts
@@ -1,0 +1,76 @@
+import { useEffect, useRef } from 'react';
+import { useFieldArray } from 'react-final-form-arrays';
+
+import type { ArrayFieldOptions } from '#core/components/form/types';
+
+/**
+ * @param rootFieldPath - Dot-notation path to the root field of the array.
+ *
+ * @returns An object containing the following properties:
+ * - `entries`: one entry per array item, each with a stable `id` (UUID) for use as a React
+ * `key` prop and an `indexedFieldPath` for constructing nested field names. Guaranteed to have at
+ * least `minItems` entries after the initial effect runs.
+ *
+ * - `add`: Pushes one new empty item to the array.
+ *
+ * - `remove`: Removes an item from the array.
+ *
+ * - `isRemovable`: Indicates if an array item can be removed.
+ *
+ * @example
+ * ```tsx
+ * const { entries, add, remove, isRemovable } = useArrayField('contacts', { minItems: 1 });
+ *
+ * return (
+ *   <>
+ *     {entries.map(({ id, indexedFieldPath }, index) => (
+ *       <div key={id}>
+ *         <StringField name={`${indexedFieldPath}.email`} label="Email" />
+ *         {isRemovable && <button onClick={() => remove(index)}>Remove</button>}
+ *       </div>
+ *     ))}
+ *     <button onClick={add}>Add more</button>
+ *   </>
+ * );
+ * ```
+ */
+export function useArrayField(
+  rootFieldPath: string,
+  { minItems = 0, readOnly = false }: ArrayFieldOptions = {}
+) {
+  const { fields } = useFieldArray(rootFieldPath);
+
+  const entryIds = useRef<string[]>(
+    Array.from({ length: fields.length ?? 0 }, () => crypto.randomUUID())
+  );
+
+  useEffect(() => {
+    const currentLength = fields.length ?? 0;
+    if (currentLength < minItems) {
+      for (let i = currentLength; i < minItems; i++) {
+        entryIds.current.push(crypto.randomUUID());
+        fields.push({});
+      }
+    }
+  }, [fields, minItems]);
+
+  const add = () => {
+    entryIds.current.push(crypto.randomUUID());
+    fields.push({});
+  };
+
+  const remove = (index: number) => {
+    entryIds.current.splice(index, 1);
+    fields.remove(index);
+  };
+
+  return {
+    entries: fields.map((indexedFieldPath, index) => ({
+      id: entryIds.current[index],
+      indexedFieldPath,
+    })),
+    add,
+    remove,
+    isRemovable: (fields.length ?? 0) > minItems && !readOnly,
+  };
+}
--- a/javascript/packages/core/components/form/hooks/use-array-field.ts
+++ b/javascript/packages/core/components/form/hooks/use-array-field.ts
@@ -54,12 +54,15 @@ export function useArrayField(
     }
   }, [fields, minItems]);
 
+  const isRemovable = (fields.length ?? 0) > minItems && !readOnly;
+
   const add = () => {
     entryIds.current.push(crypto.randomUUID());
     fields.push({});
   };
 
   const remove = (index: number) => {
+    if (!isRemovable) return;
     entryIds.current.splice(index, 1);
     fields.remove(index);
   };
@@ -71,6 +74,6 @@ export function useArrayField(
     })),
     add,
     remove,
-    isRemovable: (fields.length ?? 0) > minItems && !readOnly,
+    isRemovable,
   };
 }
--- a/javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
+++ b/javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
@@ -12,6 +12,7 @@ import type { ArrayFormGroupProps } from './types';
 export function ArrayFormGroup({
   rootFieldPath,
   groupLabel,
+  addLabel: addLabelProp,
   minItems = 0,
   readOnly = false,
   children,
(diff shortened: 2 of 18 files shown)


javascript/packages/core/components/form/hooks/use-array-field.ts
```diff
@@ -0,0 +1,76 @@
...
+  );
+
+  useEffect(() => {
+    const currentLength = fields.length ?? 0;
+    if (currentLength < minItems) {
+      for (let i = currentLength; i < minItems; i++) {
+        entryIds.current.push(crypto.randomUUID());
+        fields.push({});
+      }
+    }
```
konukselim, line 54: This condition is redundant but makes code more readable.

javascript/packages/core/components/form/hooks/use-array-field.ts
```diff
@@ -0,0 +1,76 @@
...
+  const remove = (index: number) => {
+    entryIds.current.splice(index, 1);
+    fields.remove(index);
+  };
+
+  return {
+    entries: fields.map((indexedFieldPath, index) => ({
+      id: entryIds.current[index],
+      indexedFieldPath,
+    })),
```
konukselim, line 74: Asking out of curiosity - have you considered returning `entries`  as a function instead of directly returning the actual array? If so, wondering why you made the decision returning the array directly.

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
```diff
@@ -0,0 +1,83 @@
...
+  description,
+  tooltip,
+  collapsible,
+}: ArrayFormGroupProps) {
+  const [, theme] = useStyletron();
+  const { entries, add, remove, isRemovable } = useArrayField(rootFieldPath, {
+    minItems,
+    readOnly,
+  });
+  const addLabel = groupLabel ? `Add ${groupLabel.toLowerCase()}` : 'Add more';
```
konukselim: Should we expose an optional prop to allow for more context specific customization?

konukselim (approved):

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
craigmarker: What do you have in mind?

javascript/packages/core/components/form/hooks/use-array-field.ts
craigmarker: Agreed — the `if (currentLength < minItems)` guard is technically redundant since the loop condition handles it, but it makes the intent immediately clear. Happy to leave as-is for readability.

javascript/packages/core/components/form/hooks/use-array-field.ts
craigmarker: I didn't consider using a function. What did you have in mind?

javascript/packages/core/components/form/hooks/use-array-field.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const add = () => {
+    entryIds.current.push(crypto.randomUUID());
+    fields.push({});
+  };
+
+  const remove = (index: number) => {
+    entryIds.current.splice(index, 1);
+    fields.remove(index);
+  };
```
donovanlopez, line 68: Should we add a check if `isRemovable` is `true`?

javascript/packages/core/components/form/hooks/use-array-field.ts
```diff
@@ -0,0 +1,76 @@
...
+        entryIds.current.push(crypto.randomUUID());
+        fields.push({});
+      }
+    }
+  }, [fields, minItems]);
+
+  const add = () => {
+    entryIds.current.push(crypto.randomUUID());
+    fields.push({});
+  };
```
donovanlopez, line 62: Should we ever be allowed to add a new array field if `isRemovable` is false?

javascript/packages/core/components/form/hooks/use-array-field.ts
```diff
@@ -0,0 +1,76 @@
...
+ *         <StringField name={`${indexedFieldPath}.email`} label="Email" />
+ *         {isRemovable && <button onClick={() => remove(index)}>Remove</button>}
+ *       </div>
+ *     ))}
+ *     <button onClick={add}>Add more</button>
+ *   </>
+ * );
+ * ```
+ */
+export function useArrayField(
```
donovanlopez, line 37: nit: It's not a huge deal but I feel like the name of this hook is inaccurate. Maybe `useEntryArray` since we return `entries` (and yes I just changed `field` with `entry` from `import { useFieldArray } from 'react-final-form-arrays'`?

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
donovanlopez: nit: Would we ever want this not to be lowercase? I'm thinking if we ever have a group that has an acronym, i.e. `ML something` or `AI something`

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
```diff
@@ -0,0 +1,83 @@
...
+                  shape={SHAPE.pill}
+                  size={SIZE.compact}
+                  startEnhancer={<Icon name="trashCan" kind={IconKind.PRIMARY} />}
+                  aria-label="Remove"
+                  onClick={() => remove(index)}
+                >
+                  Remove
+                </Button>
+              ) : undefined
+            }
```
donovanlopez, line 53: I might be misremembering my React but I think we might be able to simplify this to:

```
           endEnhancer={
              isRemovable && (
                <Button
                  type="button"
                  kind={KIND.secondary}
                  shape={SHAPE.pill}
                  size={SIZE.compact}
                  startEnhancer={<Icon name="trashCan" kind={IconKind.PRIMARY} />}
                  aria-label="Remove"
                  onClick={() => remove(index)}
                >
                  Remove
                </Button>
              )
            }

```

donovanlopez (approved):
LGTM! Just some small comments

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
craigmarker: Done — added an addLabel prop that overrides the derived label.

craigmarker: Good call! Done — added an addLabel prop that overrides the derived label.

javascript/packages/core/components/form/layout/array-form-group/array-form-group.tsx
craigmarker: Done — switched to isRemovable && <Button>…</Button>. The endEnhancer prop accepts ReactNode so false is safe here.

javascript/packages/core/components/form/hooks/use-array-field.ts
craigmarker: Hmmm, I think that isRemovable and add ability are independent. isRemovable is false when at the minimum count or in read-only mode — neither should block adding. 

The add button visibility/disabling is handled at the component layer (it's not rendered in read-only mode). We could add a separate canAdd / maxItems guard if needed, but we don't have a use case for that at the. moment.

javascript/packages/core/components/form/hooks/use-array-field.ts
craigmarker: Good catch — added a guard: if (!isRemovable) return; at the top of remove. Also added test cases for use-array-field directly that aren't covered by integration-style tests within array-form-group, array-form-row.

javascript/packages/core/components/form/hooks/use-array-field.ts
craigmarker: I like the reasoning, but I like useArrayField — it reads as "hook for an array-typed form field", which mirrors how our [other field hook is named ](https://github.com/michelangelo-ai/michelangelo/blob/2fe9b8f4bba7b1e9e0a65627d72d1cc7f8a38c6e/javascript/packages/core/components/form/hooks/use-field.ts#L10). The fact that we call the items entries internally is an implementation detail. Open to discussing if others feel strongly!

donovanlopez (approved):
Approving new changes

PR #897, opened 2026-03-12
Add miscellaneous form fields & layouts

craigmarker:
[Wire onFocus through all form fields](https://github.com/michelangelo-ai/michelangelo/commit/7b2cf6ae54013438552e996840a390f6fa80f28e)
[Add CheckboxField, NumberField, and UrlField](https://github.com/michelangelo-ai/michelangelo/commit/e3431335dd2370f231bf6c58bc3a2d041d38a109)
[Add FormColumn, FormGrid, and FormNote layouts](https://github.com/michelangelo-ai/michelangelo/commit/84ec17ab5b2fe9dfe04da2c4dfb434be8e409bb3)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

https://github.com/user-attachments/assets/c9a84622-7988-40e0-97c0-355649f83fae

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/fields/number/number-field.tsx
+++ b/javascript/packages/core/components/form/fields/number/number-field.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import { Input } from 'baseui/input';
+
+import { FormControl } from '#core/components/form/components/form-control';
+import { useField } from '#core/components/form/hooks/use-field';
+
+import type { BaseFieldProps } from '#core/components/form/fields/types';
+
+export const NumberField: React.FC<BaseFieldProps<number | undefined>> = ({
+  name,
+  label,
+  defaultValue,
+  initialValue,
+  required,
+  validate,
+  readOnly,
+  disabled,
+  placeholder,
+  description,
+  caption,
+}) => {
+  const { input, meta } = useField<number | undefined>(name, {
+    required,
+    validate,
+    defaultValue,
+    initialValue,
+    label,
+  });
+
+  return (
+    <FormControl
+      label={label}
+      required={required}
+      description={description}
+      caption={caption}
+      error={meta.touched && meta.error ? meta.error : undefined}
+    >
+      <Input
+        id={input.name}
+        type="number"
+        value={input.value == null ? '' : String(input.value)}
+        name={input.name}
+        onChange={(e) => input.onChange(parseNumber(e.currentTarget.value))}
+        onBlur={input.onBlur}
+        onFocus={input.onFocus}
+        placeholder={placeholder}
+        readOnly={readOnly}
+        disabled={disabled}
+      />
+    </FormControl>
+  );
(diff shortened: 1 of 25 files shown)


javascript/packages/core/components/form/layout/form-note/__tests__/form-note.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render, screen } from '@testing-library/react';
+
+import { FormNote } from '#core/components/form/layout/form-note/form-note';
+import { buildWrapper } from '#core/test/wrappers/build-wrapper';
+import { getBaseProviderWrapper } from '#core/test/wrappers/get-base-provider-wrapper';
+
+describe('FormNote', () => {
+  it('renders markdown content as HTML', () => {
```
donovanlopez, line 8: do we need to sanitize the content if we're rendering it as HTML?

donovanlopez (approved):

javascript/packages/core/components/form/fields/number/number-field.tsx
```diff
@@ -0,0 +1,58 @@
...
+      />
+    </FormControl>
+  );
+};
+
+function parseNumber(s: string): number | undefined {
+  if (s === '') return undefined;
+  const n = Number(s);
+  return isNaN(n) ? undefined : n;
+}
```
konukselim, line 58: how about moving parseNumber method to newly created `number-utils.ts` file under `core/utils` and import it from there.

javascript/packages/core/components/form/layout/form-note/__tests__/form-note.test.tsx
craigmarker: I actually haven't considered if we should, since technically whoever deploys the app will also be responsible for configuring what props are passed to FormNote (and other users of Markdown).

I'll create an issue to track, since this larger than this specific pull request. Thanks!

javascript/packages/core/components/form/fields/number/number-field.tsx
craigmarker: Good instinct—the concern is real. parseNumber has some form-field-specific assumptions baked in (treating '' and NaN as undefined) that a general utility would need to abstract away, and the abstraction is not obvious to me without an additional use case to design against.

Happy to revisit if another number parsing use case appears!

craigmarker: We can also consider how this relates to the codebase convention and claude skill for "starting local, promoting when needed"

https://github.com/michelangelo-ai/michelangelo/blob/main/javascript/.claude/skills/file-directory-structure/SKILL.md#code-placement-strategy

PR #901, opened 2026-03-16
Support markdown in form-control captions

konukselim:
- Introduces support for Markdown content in form captions.
- Handle caption content overflow.

In order to truncate Markdown content in form-field captions, `Caption` container needs to render it's content as block(by default its a flex container). That's why `Caption` styles are overridden for this case. 

https://github.com/user-attachments/assets/ea917a7c-9608-4e14-9a9e-a53efac71d88

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
N/A

**Why?**
N/A

**How did you test it?**
Manually tested by adding a caption in one of the sandbox example form fields.

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/javascript/packages/core/components/form/components/form-control.tsx
+++ b/javascript/packages/core/components/form/components/form-control.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import { useStyletron } from 'baseui';
 import { FormControl as BaseFormControl } from 'baseui/form-control';
 
+import { Markdown } from '#core/components/markdown/markdown';
+import { TruncatedText } from '#core/components/truncated-text/truncated-text';
 import { Label } from './label/label';
 
 import type { FormControlProps } from './types';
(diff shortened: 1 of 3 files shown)


javascript/packages/core/components/form/components/form-control.tsx
```diff
@@ -21,7 +23,13 @@ export const FormControl: React.FC<FormControlProps> = ({
     <div className={css({ width: '100%' })}>
       <BaseFormControl
         label={label && <Label label={label} required={required} description={description} />}
-        caption={caption}
+        caption={
+          caption && (
+            <TruncatedText>
```
craigmarker, line 28: We should consider if it makes sense to truncate a caption, which should theoretically be important enough to always be visible.

If a caption is eligible for truncation (eg, long), it may be better to 

1. Shorten the caption
2. Use the description property instead, which will render as a tooltip.

craigmarker (approved):

javascript/packages/core/components/form/components/form-control.tsx
craigmarker: #903

konukselim: Thanks for creating the issue. We can consider `TruncatedText` as a safe-guard to prevent caption overflow. I agree with what you explained but there is nothing wrong to be cautious against "not ideal" or "bad" usages.

PR #905, opened 2026-03-17
Support parse and format in all form fields

konukselim:
- Add format and parse props to BaseFieldProps and forward them through useField to React Final Form in all field components.
- Widen input.onChange to accept unknown values so fields using parse can receive raw input values of a different type than the stored form state.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/fields/types.ts
+++ b/javascript/packages/core/components/form/fields/types.ts
@@ -1,6 +1,6 @@
 import type { FieldValidator } from '#core/components/form/validation/types';
 
-export interface BaseFieldProps<T = unknown> {
+export interface BaseFieldProps<T = unknown, InputValue = T> {
   /** Unique ID of the field,
    *
    *  - Identifies the field input in the form
@@ -67,15 +67,13 @@ export interface BaseFieldProps<T = unknown> {
    * Transforms the input value before storing it in form state.
    * Called with the input value.
    */
-  /* eslint-disable @typescript-eslint/no-explicit-any */
-  parse?: (value: any) => T;
+  parse?: (value: InputValue) => T;
 
   /**
    * Transforms the field value for display in the input.
    * Called with the field value from form state.
    */
-  /* eslint-disable @typescript-eslint/no-explicit-any */
-  format?: (value: T) => any;
+  format?: (value: T) => InputValue;
 
   /**
    * Validation function called on each value change after the field is touched.
--- a/javascript/packages/core/components/form/fields/types.ts
+++ b/javascript/packages/core/components/form/fields/types.ts
@@ -64,6 +64,20 @@ export interface BaseFieldProps<T = unknown> {
   caption?: string;
 
   /**
+   * Transforms the input value before storing it in form state.
+   * Called with the input value.
+   */
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  parse?: (value: any) => T;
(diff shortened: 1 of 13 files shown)


javascript/packages/core/components/form/fields/types.ts
```diff
@@ -63,6 +63,20 @@ export interface BaseFieldProps<T = unknown> {
    */
   caption?: string;
 
+  /**
+   * Transforms the input value before storing it in form state.
+   * Called with the input value.
+   */
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  parse?: (value: any) => T;
```
craigmarker: I'm wondering how you determined to use value: any versus value: unknown

javascript/packages/core/components/form/types.ts
```diff
@@ -91,7 +91,7 @@ export interface FieldState {
 export interface FieldInput<T = unknown> {
   value: T;
   name: string;
-  onChange: (value: T) => void;
+  onChange: (value: unknown) => void;
```
craigmarker: would add comment to onChange about the case for value: unknown being tied to the use of parse/format.

craigmarker: better if we could somehow tie onChange expectations to the presence of parse/format 😬

craigmarker (changes requested):

javascript/packages/core/components/form/fields/types.ts
konukselim: React Final Form's [parse type](https://final-form.org/docs/react-final-form/types/FieldProps#parse) is (value: any, name: string) => any. Using unknown here would create a type mismatch when forwarding to React Final Form, requiring casts at the boundary anyway. I am not sure if we should force stricter types at field level.

konukselim:
@craigmarker thanks for challenging me on thinking about these types once again. I believe I made a significant improvement on the generic type definitions. To summarize these changes:

Introduced a second generic parameter `InputValue` which defaults to `T` across three layers:

**1. FieldInput<T, InputValue = T>**

Previously value and onChange were both typed as T. But after parse and format are defined, value started holding the formatted result, not the stored type. Similarly, onChange receives the raw input from the UI, which then flows through parse to become T. The second generic makes this explicit: T is what's stored, InputValue is what the UI works with.

**2. useField<T, InputValue = T>**

parse and format now have fully typed signatures. parse takes InputValue and returns T, format takes T and returns InputValue. `any` usage is eliminated completely with this change. The returned input is FieldInput<T, InputValue>, so input.value is InputValue and input.onChange accepts InputValue.

Internally, `InputValue` is passed to React Final Form's useField hook as the third generic(since RFF's hook accepts input value as third generic). This change resulted with eliminated all the typecasts required to bridge our hook with RFF's hook.

**3. BaseFieldProps<T, InputValue = T>**

This allows field component props to carry the same type information when needed. This change requires no code change on fields that are already using BaseFieldProps with only a single generic since InputValue always defaults to passed generic if it's not provided. We can safely say this change is backward compatible.

craigmarker (approved):
LGTM, thanks for the change!

PR #906, opened 2026-03-17
Add DateField component

konukselim:
Implement DateField which supports taking dates either in ISO 8601 or epoch seconds. Overrode the parse and format of `useField` hook to be able to convert provided string dates into BaseUI DatePicker's expected Date.

For now this override will block users to provide their own parse or format methods. Later the logic can be changed to allow users provide their own parse & format methods.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/fields/date/__tests__/date-field.test.tsx
+++ b/javascript/packages/core/components/form/fields/date/__tests__/date-field.test.tsx
@@ -1,0 +1,96 @@
+import { render, screen } from '@testing-library/react';
+
+import { DateField } from '#core/components/form/fields/date/date-field';
+import { DATE_FORMAT } from '#core/components/form/fields/date/types';
+import { buildWrapper } from '#core/test/wrappers/build-wrapper';
+import { getBaseProviderWrapper } from '#core/test/wrappers/get-base-provider-wrapper';
+import { getFormProviderWrapper } from '#core/test/wrappers/get-form-provider-wrapper';
+import { getIconProviderWrapper } from '#core/test/wrappers/get-icon-provider-wrapper';
+
+const wrapper = buildWrapper([
+  getBaseProviderWrapper(),
+  getIconProviderWrapper(),
+  getFormProviderWrapper({}),
+]);
+
+describe('DateField', () => {
+  it('renders with label', () => {
+    render(<DateField name="date" label="Start Date" />, wrapper);
+
+    expect(screen.getByText('Start Date')).toBeInTheDocument();
+  });
+
+  it('shows required indicator when required', () => {
+    render(<DateField name="date" label="Start Date" required />, wrapper);
+
+    expect(
+      screen.getAllByText((_, element) => element?.textContent === 'Start Date*').length
+    ).toBeGreaterThan(0);
+  });
+
+  it('displays caption text', () => {
+    render(<DateField name="date" label="Start Date" caption="Select a start date" />, wrapper);
+
+    expect(screen.getByText('Select a start date')).toBeInTheDocument();
+  });
+
+  it('displays help tooltip when description is provided', () => {
+    render(
+      <DateField name="date" label="Start Date" description="When the pipeline should start" />,
+      wrapper
+    );
+
+    expect(screen.getByRole('img', { name: 'help' })).toBeInTheDocument();
+  });
+
+  it('renders with placeholder', () => {
+    render(<DateField name="date" label="Start Date" />, wrapper);
+
+    expect(screen.getByPlaceholderText('MM/dd/yyyy')).toBeInTheDocument();
+  });
+
+  it('hides placeholder when disabled', () => {
+    render(<DateField name="date" label="Start Date" disabled />, wrapper);
+
+    expect(screen.queryByPlaceholderText('MM/dd/yyyy')).not.toBeInTheDocument();
+  });
+
+  it('hides placeholder when readOnly', () => {
+    render(<DateField name="date" label="Start Date" readOnly />, wrapper);
+
+    expect(screen.queryByPlaceholderText('MM/dd/yyyy')).not.toBeInTheDocument();
+  });
+
+  it('displays initial value with ISO format', () => {
+    render(
+      <DateField name="date" label="Start Date" dateFormat={DATE_FORMAT.ISO_DATE_STRING} />,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({ initialValues: { date: '2024-01-15T00:00:00.000Z' } }),
+      ])
+    );
+
+    expect(screen.getByDisplayValue('01/15/2024')).toBeInTheDocument();
+  });
+
+  it('displays initial value with epoch format', () => {
+    // 1705276800 = 2024-01-15T00:00:00.000Z
+    render(
+      <DateField name="date" label="Start Date" dateFormat={DATE_FORMAT.EPOCH_SECONDS} />,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getFormProviderWrapper({ initialValues: { date: '1705276800' } }),
+      ])
+    );
+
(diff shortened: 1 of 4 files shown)


javascript/packages/core/components/form/fields/date/use-date-formatters.ts
```diff
@@ -0,0 +1,60 @@
...
+      : (value: string | null) => (value ? new Date(value) : null);
+
+  const fromDate =
+    dateFormat === DATE_FORMAT.EPOCH_SECONDS
+      ? (date: Date | null) => (date ? String(getEpochSecondsFromDate(date)) : null)
+      : (date: Date | null) => (date ? date.toISOString() : null);
+
+  return {
+    format: (value: string | null) => translateUTCDateToUserTimezone(toDate(value)),
+    parse: (value: Date | null) => fromDate(translateUserDateToUTC(value))!,
```
craigmarker: I don't think this non-null assertion is correct, seeing that fromDate definitely can return null. What am I i missing?

javascript/packages/core/components/form/fields/date/date-field.tsx
```diff
@@ -0,0 +1,99 @@
...
+      },
+    },
+  },
+};
+
+const isEmptyInputValue = (value: unknown): boolean => {
+  return isNil(value) || value === '' || (isArray(value) && !value.length);
+};
+
+export const DateField: React.FC<DateFieldProps> = ({
```
craigmarker, line 13: nit: prefer placing primary export as the first definition after imports.

javascript/packages/core/components/form/fields/date/use-date-formatters.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * The format pipeline converts stored values to display values:
+ *   Date format string → Date → timezone-adjusted Date
+ *
+ * The parse pipeline converts display values back to stored values:
+ *   timezone-adjusted Date → UTC Date → Date format string
+ *
+ * @param dateFormat - Controls the persisted date format (epoch or ISO).
+ */
+export function useDateFormatters(
```
craigmarker, line 21: nit: prefer placing primary export as the first definition after imports.

javascript/packages/core/components/form/fields/date/__tests__/date-field.test.tsx
```diff
@@ -0,0 +1,96 @@
+import { render, screen } from '@testing-library/react';
+
+import { DateField } from '#core/components/form/fields/date/date-field';
+import { DATE_FORMAT } from '#core/components/form/fields/date/types';
+import { buildWrapper } from '#core/test/wrappers/build-wrapper';
+import { getBaseProviderWrapper } from '#core/test/wrappers/get-base-provider-wrapper';
+import { getFormProviderWrapper } from '#core/test/wrappers/get-form-provider-wrapper';
+import { getIconProviderWrapper } from '#core/test/wrappers/get-icon-provider-wrapper';
+
+const wrapper = buildWrapper([
```
craigmarker, line 10: nit: consider inlining the wrapper definition. in case any new test cases emerge that would change this default wrapper instantiation in a way that obscures each tests' requirements.

javascript/packages/core/components/form/fields/date/date-field.tsx
```diff
@@ -0,0 +1,99 @@
...
+  placeholder,
+  disabled,
+  description,
+  caption,
+  noFutureDate,
+  dateFormat = DATE_FORMAT.ISO_DATE_STRING,
+}) => {
+  const { format, parse } = useDateFormatters(dateFormat);
+
+  // Date format can be epoch seconds string or ISO string
```
craigmarker:
```suggestion
```

IMO, should use props definition to get this info

javascript/packages/core/components/form/fields/date/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { BaseFieldProps } from '../types';
+
+export interface DateFieldProps extends BaseFieldProps<string> {
+  /** @default DATE_FORMAT.ISO_DATE_STRING */
+  dateFormat?: DATE_FORMAT;
+  /** Restricts selectable dates to today or earlier */
```
craigmarker:
```suggestion
  /** 
     * Restricts selectable dates to today or earlier 
     * @default false
     */
```

craigmarker (changes requested):

javascript/packages/core/components/form/fields/date/use-date-formatters.ts
konukselim: The changes were causing type mismatch when I use parse with `useField` hook since parse method passed to `useField<string>` hook was expecting the return type to be `string`. So I ended up putting a non-null assertion since I couldn't find another way. That's clearly wrong approach.

However, the stored type in form state is always string. The null only belongs on the `InputValue` side because DatePicker displays null when there's no date. Having a separate `InputValue` type simplified many implementation details. Please have another look at this hook implementation.

javascript/packages/core/components/form/fields/date/date-field.tsx
konukselim: Correct me if I'm wrong but I guess the rationale behind this request is for readability concerns so that if someone else look at the code they could easily start from the top to read without any distraction like in [newspaper code structure](https://kentcdodds.com/blog/newspaper-code-structure)

However, here `READ_ONLY_OVERRIDES` and `isEmptyInputValue` are const bindings and DateField implementation references them. I can convert them to function for hoisting reasons but it feels an unnecessary change to me.

I can import overrides from `styled-components.ts` and helper function from a `util.ts` but creating these files just for a simple override and a util method feels unnecessary and a bad trade-off to me. I'd love to hear your opinion on what could have been done to move the primary export to the top.

javascript/packages/core/components/form/fields/date/__tests__/date-field.test.tsx
konukselim: I am sorry but I disagree with you on this - The default wrapper represents the "no special state" baseline. Tests that need specific setup(like initial values in my case) already construct their own inline. Inlining the default would add repetition across 7 tests without improving clarity, and actually make it harder to spot which tests have custom setup. This also follows the same pattern used in [form.test.tsx](https://github.com/michelangelo-ai/michelangelo/blob/main/javascript/packages/core/components/form/__tests__/form.test.tsx#L567)

javascript/packages/core/components/form/fields/date/date-field.tsx
craigmarker: My understanding is similar to yours: newspaper code structure (TIL about that terminology) can improve readability.

For the `isEmptyInputValue`, I would switch to using function and move below DateField. Another TIL was that Google style guide actually [recommends function expressions for named functions](https://google.github.io/styleguide/tsguide.html#function-declarations).

For `READ_ONLY_OVERRIDES`, I'd consider inlining and extracting to a [style-components.ts function like buildSelectOverrides](https://github.com/michelangelo-ai/michelangelo/blob/c54c776840b7f946ec8f4c19c65ddafe54327ecb/javascript/packages/core/components/form/fields/select/build-select-overrides.ts#L6) if the styles ever become more complex.

Happy to chat more about either of these and ensure we're aligned 👍🏻

javascript/packages/core/components/form/fields/date/__tests__/date-field.test.tsx
craigmarker: Thanks for the opinion! Let's sync offline on this—no need to block on this. We'll document the approach moving forward for clarity.

craigmarker (approved):

javascript/packages/core/components/form/fields/date/date-field.tsx
konukselim: I also didn't know Google's recommendation. Thanks for sharing that. Turned named function into function expression and moved it to the bottom and inlined overrides 👍

PR #922, opened 2026-03-20
fix(ingester): use scheme.ObjectKinds() to resolve GVK, fixing controllermgr CrashLoopBackOff

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

<!-- Describe what has changed in this PR -->
**What changed?**
Description:                                                                                                                                                                  
                                                                                                                                                                                
  Problem                                                                                                                                                                                                                                                                                                                                                       
  michelangelo-controllermgr was in CrashLoopBackOff with the following error:                                                                                                  
                                                                                                                                                                                
  Failed: failed to setup ingester for : controller with name ingester_ already exists.                                                                                         
  Controller names must be unique to avoid multiple controllers reporting to the same metric                                                                                    
                                                                                                                                                                                
  The ingester registers one controller per CRD type in AllCRDObjects. Each controller is named ingester_<Kind>. However, every controller was being named ingester_ (empty     
  kind), so the second registration failed.                                                                                                                                     
                                                                                                                                                                                
  Root Cause                                                                                                                                                                    
   
  GetObjectKind().GroupVersionKind() was used to get the Kind in two places:                                                                                                    
                  
  1. SetupWithManager — to build the controller name (ingester_<Kind>)                                                                                                          
  2. handleDeletion / handleDeletionAnnotation — to build the TypeMeta passed to storage.Delete
                                                                                                                                                                                
  GetObjectKind() returns the struct's embedded TypeMeta field directly. This is empty in two cases:                                                                            
  - Bare Go structs (e.g. &Model{} in AllCRDObjects) — TypeMeta is never populated                                                                                              
  - Objects fetched via r.Get() — the k8s.io/apimachinery JSON deserializer intentionally does not populate TypeMeta on typed objects                                           
  (https://github.com/kubernetes/client-go/issues/308)
                                                                                                                                                                                
  Fix             
                                                                                                                                                                                
  Replace all GetObjectKind().GroupVersionKind() calls with scheme.ObjectKinds(obj), which looks up the Go type → GVK mapping from the scheme registry and is always            
  authoritative.
                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                            

<!-- Tell your future self why have you made these changes -->
**Why?**


<!-- How have you verified this change? Tested locally? Added a unit test? Checked in staging env? -->
**How did you test it?**


<!-- Assuming the worst case, what can be broken when deploying this change to production? -->
**Potential risks**

<!-- Is it notable for release? e.g. schema updates, configuration or data migration required? If so, please mention it, and also update CHANGELOG.md -->
**Release notes**

<!-- Does this PR introduce a user-facing or API change? Is user document updated? Is the change backward compatible? If so please update in wiki https://github.com/michelangelo-ai/michelangelo/wiki? -->
**Documentation Changes**


go/components/ingester/controller_test.go
```diff
@@ -512,3 +512,119 @@ func TestHelperFunctions(t *testing.T) {
 		assert.Equal(t, defaultRequeuePeriod, r2.getRequeuePeriod())
 	})
 }
+
+// TestSchemeGVKResolution verifies that all CRD objects in AllCRDObjects resolve to
+// unique, non-empty kinds via the scheme. This is the root cause of the CrashLoopBackOff:
+// GetObjectKind().GroupVersionKind() returns empty on bare structs, causing every
+// controller to be named "ingester_" and the second registration to fail.
```
aritra0227: I don't think the following has any importance to an external contributor. So best remove the line about crashbacklooping:
```suggestion
// TestSchemeGVKResolution verifies that all CRD objects in AllCRDObjects resolve to
// unique, non-empty kinds via the scheme.
```

go/components/ingester/controller.go
```diff
@@ -224,7 +230,11 @@ func (r *Reconciler) handleImmutableObject(ctx context.Context, log logr.Logger,
 
 // SetupWithManager sets up the controller with the Manager
 func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
-	kind := r.TargetKind.GetObjectKind().GroupVersionKind().Kind
+	gvks, _, err := r.Scheme.ObjectKinds(r.TargetKind)
+	if err != nil || len(gvks) == 0 {
+		return fmt.Errorf("failed to get GVK for %T: %w", r.TargetKind, err)
+	}
+	kind := gvks[0].Kind
```
aritra0227, line 244: In the entire implementation, we are assuming that each object will have exactly 1 GVK, hence we are taking `gvks[0]`. However, an object can have multiple gvks. For example:
1. {Group: "michelangelo", Version: "v1", Kind: "Model"},
2. {Group: "michelangelo", Version: "v1beta1", Kind: "Model"},

I have the following questions:
1. Do we only plan to support 1 GVK for the objects?
2. If not, then can we add a TODO with a github issue for at least adding support later?

aritra0227 (approved):
Generally looks good so approving. Left some comments.

go/components/ingester/controller.go
badcount: updated with todo

PR #953, opened 2026-03-20
Fix pipeline create tar upload failure

hwaminkim:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Fix critical bug where `mactl apply` command failed to upload uniflow tar for new pipeline instances, causing registered pipelines to be non-executable.

**Why?**

The pipeline plugin's `apply_plugin_command()` only set the metadata converter for `create` command, but not for `apply` command:

  ```python
  # Before (buggy code)
  if target_command == "create":
      crd.func_crd_metadata_converter = convert_crd_metadata_pipeline_create
```

  Impact:
  - When mactl apply -f pipeline.yaml is used on a non-existent pipeline, the CRD framework automatically calls create() internally (crd.py:400-404)
  - However, without func_crd_metadata_converter set, the create operation skips critical registration steps:
    a. Running subprocess registration to build the uniflow tar
    b. Extracting tar path and workflow inputs
    c. Adding uniflow tar path to spec.manifest.uniflowTar
  - Result: Pipeline is registered in metadata store without executable code
  - When triggered, pipeline fails because uniflow tar is missing
  - This is a silent failure - the apply command succeeds, but creates a broken pipeline

  Solution

  Include apply command in the condition to ensure identical behavior:

  # After (fixed code)
  if target_command in ("apply", "create"):
      crd.func_crd_metadata_converter = convert_crd_metadata_pipeline_create

  Test Plan

  Added test_apply_command_sets_metadata_converter() which:
  - Verifies apply command sets the metadata converter
  - Confirms apply and create use the same converter function
  - Includes detailed docstring documenting the bug impact and why this test is critical
  - ✅ Test passes with fix
  - ❌ Test fails without fix (converter is Mock instead of actual function)

  Additional Context

  The apply command is designed to be idempotent - it auto-creates if the instance doesn't exist, or updates if it does. This fix ensures the auto-create path for pipelines properly executes the
  registration workflow, making apply and create functionally equivalent for new pipelines.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/entity/pipeline/main.py
+++ b/python/michelangelo/cli/mactl/plugins/entity/pipeline/main.py
@@ -52,7 +52,7 @@ def apply_plugin_command(
     _LOG.info("Applying plugins to crd: %r / %r", crd, target_command)
     _LOG.debug("Available CRDs: %r", crds)
     _LOG.debug("gRPC Channel: %r", channel)
-    if target_command == "create":
+    if target_command in ("apply", "create"):
         crd.func_crd_metadata_converter = convert_crd_metadata_pipeline_create
     if target_command == "run":
         crd.func_crd_metadata_converter = convert_crd_metadata_pipeline_run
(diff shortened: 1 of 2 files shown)


npuppala111 (approved):
LGTM!

badcount (approved):

PR #983, opened 2026-03-24
Introduce label addon

konukselim:
Adds labelAddon prop on all form fields that renders arbitrary content(text, buttons, components) at the far right of the label row, alongside the existing counter when present.

**What changed**

- BaseFieldProps gains a `labelAddon?: ReactNode` prop, making it available to every field component.
- FormControl no longer delegates counter rendering to baseui's native counter prop. Instead, it builds a custom labelEndEnhancer via a LabelEndEnhancerContent component that renders the counter text and labelAddon side by side in a flex row. Tried to keep both counter and labelAddon, however the only way to keep them was to turn labelEndEnhancer into a function which returns ReactNode. With this change, [sharedProps are passed to the function](https://github.com/uber/baseweb/blob/main/src/form-control/form-control.tsx#L188) for it to retrieve `$length` and `$maxLength` of the counter. However, there was a bug(or feature) in form-control implementation where [maxLength is incorrectly set to length instead of maxLength](https://github.com/uber/baseweb/blob/main/src/form-control/form-control.tsx#L162) that breaks the behavior we want to have.
 - All field components destructure and forward labelAddon to FormControl.
- Also changed TextareaField component to render a resizer which could only resize the input vertically within the changes in this PR.

https://github.com/user-attachments/assets/6a8e8f84-1fea-4f07-82c9-78175fa2a1e8

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
+++ b/javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
@@ -17,7 +17,7 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
   placeholder,
   description,
   caption,
-  labelAddon,
+  labelEndEnhancer,
   format,
   parse,
   rows,
@@ -39,7 +39,7 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
       label={label}
       required={required}
       description={description}
-      labelAddon={labelAddon}
+      labelEndEnhancer={labelEndEnhancer}
       caption={caption}
       error={meta.touched && meta.error ? meta.error : undefined}
       counter={maxLength ? { length: currentLength, maxLength } : undefined}
--- a/javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
+++ b/javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
@@ -17,6 +17,7 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
   placeholder,
   description,
   caption,
+  labelAddon,
   format,
   parse,
   rows,
@@ -38,6 +39,7 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
       label={label}
       required={required}
       description={description}
+      labelAddon={labelAddon}
       caption={caption}
       error={meta.touched && meta.error ? meta.error : undefined}
       counter={maxLength ? { length: currentLength, maxLength } : undefined}
@@ -54,6 +56,13 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
         disabled={disabled}
         rows={rows}
         maxLength={maxLength}
+        overrides={{
+          Input: {
+            style: {
+              resize: 'vertical',
+            },
+          },
+        }}
       />
     </FormControl>
   );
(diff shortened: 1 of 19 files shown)


javascript/packages/core/components/form/fields/types.ts
```diff
@@ -63,6 +63,12 @@ export interface BaseFieldProps<T = unknown, InputValue = T> {
    */
   caption?: string;
 
+  /**
+   * Arbitrary content rendered at the far right of the label row.
+   * Can be a React component, text, or an action button.
+   */
+  labelAddon?: React.ReactNode;
```
craigmarker: I'm concerned that `labelAddon` isn't super explicit about its behavior, eg it could easily add on to the label name itself, or on a secondary row.

I believe `labelEndEnhancer` is more semantic and aligns with the design system. What do you think?

javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
```diff
@@ -54,6 +56,13 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
         disabled={disabled}
         rows={rows}
         maxLength={maxLength}
+        overrides={{
+          Input: {
+            style: {
+              resize: 'vertical',
```
craigmarker, line 62: I believe resize is a [first-class, though undocumented, property](https://github.com/uber/baseweb/blob/main/src/textarea/types.ts#L25) of `Textarea`.

javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
```diff
@@ -17,6 +17,7 @@ export const TextareaField: React.FC<TextareaFieldProps> = ({
   placeholder,
   description,
   caption,
+  labelAddon,
```
craigmarker: Consider adding documentation to the Textarea field props types to clarify maxLength and labelAddon from basefieldprops are appended when they are both provided.

craigmarker (changes requested):

javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
konukselim: https://github.com/michelangelo-ai/michelangelo/pull/985#discussion_r2986950942

javascript/packages/core/components/form/fields/types.ts
konukselim: Textarea case led me thinking of another name than `labelEndEnhancer`. But you are right about the design system and consistency. I'll update the name and will explain Textarea behavior in JSDoc.

javascript/packages/core/components/form/fields/textarea/textarea-field.tsx
konukselim: I explained the behavior in `TextAreaFieldProps` interface for `maxLength` prop.

craigmarker (approved):

PR #1004, opened 2026-03-25
Enforce TypeScript import and naming conventions via ESLint

craigmarker:
The codebase already followed these patterns informally but had no lint enforcement, allowing drift. Type-only imports were written inconsistently as value imports, @ts-ignore was used without requiring justification, and type naming had no automated PascalCase check.

Add four rules to codify existing conventions:
- consistent-type-imports: separate `import type` statements
- ban-ts-comment: disallow @ts-ignore, require description on @ts-expect-error
- naming-convention: PascalCase for types/interfaces, no T-suffix on type names (with separate typeParameter selector to allow single-letter generics like <T>)
- reportUnusedDisableDirectives: flag stale eslint-disable comments

Rename DATE_FORMAT enum to DateFormat — the only naming-convention violation. Autofix ~50 files for consistent-type-imports.

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**
Consumers of the library leveraging DATE_FORMAT will need to update to use DateFormat instead.

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/index.tsx
+++ b/javascript/packages/core/index.tsx
@@ -181,7 +181,7 @@ export { UrlField } from '#core/components/form/fields/url/url-field';
 export { NumberField } from '#core/components/form/fields/number/number-field';
 export { CheckboxField } from '#core/components/form/fields/checkbox/checkbox-field';
 export { DateField } from '#core/components/form/fields/date/date-field';
-export { DATE_FORMAT } from '#core/components/form/fields/date/types';
+export { DateFormat } from '#core/components/form/fields/date/types';
 
 // Form Layout
 export { FormGroup } from '#core/components/form/layout/form-group/form-group';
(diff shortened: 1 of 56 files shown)


javascript/packages/core/index.tsx
```diff
@@ -181,7 +181,7 @@ export { UrlField } from '#core/components/form/fields/url/url-field';
 export { NumberField } from '#core/components/form/fields/number/number-field';
 export { CheckboxField } from '#core/components/form/fields/checkbox/checkbox-field';
 export { DateField } from '#core/components/form/fields/date/date-field';
-export { DATE_FORMAT } from '#core/components/form/fields/date/types';
```
craigmarker, line 184: I don't believe that DATE_FORMAT is leveraged yet, but @konukselim can verify.

donovanlopez (approved):

javascript/packages/core/index.tsx
konukselim: it's not leveraged yet - we're safe here 👍

PR #1020, opened 2026-03-30
Fix: implement kill/pause/resume for trigger

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix: implement kill/pause/resume for trigger

**Why?**
Need to support pause/resume/kill for trigger

**How did you test it?**
In sandbox

**Potential risks**

**Release notes**

**Documentation Changes**


javascript/package.json
```diff
@@ -59,5 +59,9 @@
   "resolutionsComments": {
     "tslib": "Required to fix 'Cannot destructure property __extends of import_tslib.default' error that occurs with BaseUI 15 and its dependencies",
     "react-uid/tslib": "react-uid@2.3.0 (used by BaseUI) requires tslib v1.x API while CodeMirror packages require v2.x - force v1 for react-uid specifically"
+  },
+  "dependencies": {
+    "final-form-arrays": "^4.0.0",
+    "react-final-form-arrays": "^4.0.0"
```
craigmarker: We should revert this—only shared packages should live at javascript root. These are specific to javascript/packages/core

javascript/packages/core/config/entities/trigger/types.ts
```diff
@@ -10,3 +10,52 @@ export type Trigger = {
...
+      name: string;
+      namespace: string;
+    };
+    actor: {
+      name: string;
+    };
+    // Action field for trigger operations (replaces deprecated boolean fields)
+    action?: TriggerRunAction;
+    // DEPRECATED: Use 'action' field instead. Boolean field for kill operations
+    kill?: boolean;
```
craigmarker: nit: if it's deprecated, UI doesn't need to know about the type

javascript/packages/core/config/entities/trigger/trigger-run-action-dialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  // Create the trigger run spec with the appropriate action
+  const createTriggerRunWithAction = (baseRecord: TriggerRun): TriggerRun => {
+    return {
+      ...baseRecord,
+      spec: {
+        ...baseRecord.spec,
+        // Use only the action field (clean approach, deprecated boolean fields not set)
+        action: action === 'kill' ? TriggerRunAction.KILL :
```
craigmarker: nit: stale comment

javascript/packages/core/components/cell/renderers/trigger-action/trigger-action-cell.tsx
```diff
@@ -0,0 +1,192 @@
...
+    description: 'This will resume the paused trigger to allow new executions.',
+    confirmText: 'Are you sure you want to resume this trigger?',
+    buttonKind: KIND.secondary,
+    submitLabel: 'Resume',
+    // Show resume button for paused triggers
+    shouldShow: (state: TriggerRunState) => state === TriggerRunState.PAUSED,
+  },
+} as const;
+
+export const TriggerActionCell = (props: TriggerActionCellProps) => {
```
craigmarker: Can you share a screenshot? I think, in practice, the TriggerActionCell and TriggerRunActions components duplicate functionality. You should only need the TriggerRunActions

badcount:
Screenshot 2026-03-30 at 3 07 37 PM
Screenshot 2026-03-30 at 3 07 59 PM
Screenshot 2026-03-30 at 3 07 52 PM
Screenshot 2026-03-30 at 3 07 43 PM

go/base/workflowclient/cadenceclient/cadenceclient.go
```diff
@@ -274,3 +274,20 @@ func (c *CadenceClient) GetActivityTaskCompletedEventType() string {
...
+
+// UnpauseTrigger is not supported by Cadence (schedules are a Temporal feature)
+func (c *CadenceClient) UnpauseTrigger(_ context.Context, workflowID string) error {
+	return fmt.Errorf("UnpauseTrigger not supported by Cadence provider (workflowID: %s)", workflowID)
+}
+
+// DeleteTrigger terminates the cron workflow for the given workflow ID.
+// In Cadence, recurring triggers are implemented as long-running cron workflows,
+// so terminating the workflow stops all future executions.
+func (c *CadenceClient) DeleteTrigger(ctx context.Context, workflowID string) error {
```
yangfan0328: Why we need a separate delete? We have CancelWorkflow name above, can we rename as TerminateWorkflow?

go/components/triggerrun/controller.go
```diff
@@ -214,8 +225,26 @@ StateMachine:
 			}
 			log.Info("trigger run killed")
 			triggerRun.Status = status
+			// Clear action and deprecated fields after processing
+			triggerRun.Spec.Action = v2pb.TRIGGER_RUN_ACTION_NO_ACTION
```
yangfan0328, line 229: Why we need to deprecate field? After killing trigger run CR will goes to terminate state and immutable(not going to next reconciliation cycle), so no need to reset kill and action spec?

go/components/triggerrun/controller.go
```diff
@@ -224,6 +253,48 @@ StateMachine:
...
 		triggerRun.Status.State = status.State
+
+	case v2pb.TRIGGER_RUN_STATE_PAUSED:
+		log.Info("TRIGGER_RUN_STATE_PAUSED")
+
+		// Handle actions using the new action field
+		actionToPerform := triggerRun.Spec.Action
+
+		// Backward compat: if Spec.Kill is set and no explicit action, treat as KILL
+		if actionToPerform == v2pb.TRIGGER_RUN_ACTION_NO_ACTION && triggerRun.Spec.Kill {
```
yangfan0328, line 264: Can we move these 3 lines inside switch case? To make code cleaner.

go/base/workflowclient/cadenceclient/cadenceclient.go
badcount: because cancel workflow is to cancel workflow, temporal has schedule which is cancel schedule, so i think it's better to separate it make it clear. Workflow != schedule, but for implementation, you can see i call cancelWorkflow for cadence

go/components/triggerrun/controller.go
badcount: temporal support pause/resume/kill, we can't just use kill for all this actions, so i added new action field, so we can migrate it later on. but for now, we still support kill

go/components/triggerrun/controller.go
badcount: this is for backward compatibility will be removed it later

go/components/triggerrun/cron_trigger.go
```diff
@@ -118,19 +118,23 @@ func (r *cronTrigger) Run(ctx context.Context, triggerRun *v2pb.TriggerRun) (v2p
...
 	log := r.Log.WithValues("triggerRun", k8stypes.NamespacedName{
 		Namespace: triggerRun.Namespace,
 		Name:      triggerRun.Name,
 	})
+	wid := generateWorkflowID(triggerRun)
+	if err := r.WorkflowClient.DeleteTrigger(ctx, wid); err != nil {
+		log.Error(err, "failed to delete trigger (may already be deleted)", "workflowID", wid)
+	}
 	domain := r.WorkflowClient.GetDomain()
 	return killWorkflow(ctx, triggerRun, log, r.WorkflowClient, domain)
```
yangfan0328: are we able to move line 134-137 to same killWorkflow util funtion?

proto-go/api/v2/pipeline.pb.yarpc.go
yangfan0328, line 1: why updates on pipeline protobuf file? no changes this PR with pipeline and pipeline run?

badcount: This because they use trigger_run.proto

yangfan0328 (approved):

yangfan0328 (approved):
lgtm!

PR #1024, opened 2026-03-31
Extend action buttons to entity detail page headers

craigmarker:
The per-row actions system rendered actions only in table views. Detail pages had no way to surface entity-level actions, requiring users to navigate back to the list to trigger operations like Run.

DetailViewHeader now accepts actions from the entity config and renders them using ActionsButtons with the same hierarchy system (primary/secondary/tertiary). EntityDetailRoute wires entity config actions through to the detail view automatically.

Pipeline detail page serves as the first consumer: the Run button opens CreatePipelineRunForm from the header. Shared pipeline cell configs (type, state) were extracted to avoid duplication between list and detail views.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
visual-verify-target-default

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/components/actions/actions-buttons/actions-buttons.tsx
+++ b/javascript/packages/core/components/actions/actions-buttons/actions-buttons.tsx
@@ -16,7 +16,7 @@ type ActionsButtonsProps<T extends Data = Data> = {
 };
 
 /**
- * Renders action buttons in the detail page header, partitioned by hierarchy.
+ * Renders action buttons partitioned by hierarchy level and manages the selected action.
  *
  * Primary actions render as a fixed-width filled button, secondary as pill-shaped
  * buttons, and tertiary actions collapse into an overflow popover.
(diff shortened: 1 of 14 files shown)


javascript/packages/core/components/actions/actions-buttons/actions-buttons.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+export function ActionsButtons<T extends Data>({
+  actions,
+  record,
+  loading,
+}: ActionsButtonsProps<T>) {
+  const [css, theme] = useStyletron();
+  const [selectedAction, setSelectedAction] = useState<SelectedAction | null>(null);
+
+  if (!actions || actions.length === 0) return null;
```
donovanlopez: since `actions` is not optional, do we need to check `!actions`? we can do `actions?.length === 0` instead if it can be undefined. (though this is super nitpicky that it's not even worth changing unless you're making other changes as well)

donovanlopez (approved):

javascript/packages/core/components/actions/actions-buttons/actions-buttons.tsx
craigmarker: I'm all for nits :) Updated!

PR #1029, opened 2026-04-01
Add integration tests

badcount:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Adds integration tests covering end-to-end pipeline execution. Tests spin up real pipeline components and verify behavior across key scenarios rather than relying solely on unit level mocks. 

**Why?**
Unit tests verify individual components in isolation but can't catch regressions in how those components interact. Integration tests give confidence that the full execution path from pipeline submission through task scheduling and result collection — works correctly together.                                                                                  

**How did you test it?**
- Ran the new integration test suite locally: `pytest python/tests/integration/`
- Verified all tests pass against a local environment                                                                                                                               
- Confirmed tests fail as expected when intentionally breaking the behavior under test    

**Potential risks**
- Integration tests require a running environment; CI must have the necessary services available                                                                                    
- Tests may be slower than unit tests — consider running them in a separate CI stage if they impact build time     

**Release notes**
No user-facing changes. Test infrastructure only. 

**Documentation Changes**
No API or user-facing changes. No documentation update required.


python/michelangelo/cli/sandbox/resources/mysql.yaml
```diff
@@ -28,11 +32,11 @@ spec:
             - -u
             - root
             - -proot
-        initialDelaySeconds: 30
+        initialDelaySeconds: 10
```
sallycr: @badcount shorter is good?

sallycr (approved):
Thanks for adding integration tests

integration_test.md
```diff
@@ -0,0 +1,510 @@
+# Michelangelo End-to-End Integration Test Infrastructure
```
craigmarker, line 1: Should this be checked in? If so, we should scrape it for internal references (uPlan, ERD, uOwn)

.github/scripts/integration-test.sh
```diff
@@ -0,0 +1,143 @@
...
+AWS_ENDPOINT_URL="${MINIO_ENDPOINT}" \
+  ma pipeline dev-run --file="${PYTHON_DIR}/examples/bert_cola/pipeline.yaml"
+log "Pipeline dev-run triggered."
+
+# ---------------------------------------------------------------------------
+# Step 6: find the run name and wait for it
+# ---------------------------------------------------------------------------
+
+log "Waiting for pipelinerun to appear..."
+for i in $(seq 1 10); do
```
craigmarker, line 125: Is it possible for this test to look at the Michelangelo UI app instead: localhost:8090/<namespace>/train/runs?

.github/scripts/integration-test.sh
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
```
craigmarker, line 1: I wonder how much we should be relying on `kubectl` for our integration tests (instead of `ma` CLI or Michelangelo UI).

craigmarker (changes requested):
I'm having trouble understanding how sandbox changes interact with onboarding integration tests.

What do you think about breaking down changes into different commits/PRs to keep a clean git history?

PR #1054, opened 2026-04-03
    Add kill action to trigger runs

craigmarker:
Wires kill lifecycle action into the trigger run list view
using the existing ActionsPopover/ActionConfig system. Actions are always
visible but disabled with a hover tooltip when not applicable — only
RUNNING runs can be killed.

Each action opens a confirmation dialog backed by UpdateTriggerRun via
useStudioMutation. Exposes UpdateTriggerRun in the RPC handler registry.

Add stopCircle icon for Kill.

* Note that it takes a few seconds for the state to change. This is because we invalidate the `ListTriggerRun` and `GetTriggerRun` queries after 2 seconds so that we wait for the actions to process the state properly.

https://github.com/user-attachments/assets/c5758133-0902-4e71-b9b0-9f77de0400de

https://github.com/user-attachments/assets/3b95e211-d32f-4b06-acf4-5bf0b203544b

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
We added the "Kill" action for trigger runs.

**Why?**
We want to support actions for trigger runs (more actions to come in the future).

**How did you test it?**
I tested locally.

**Potential risks**
Worst case would be that the "Kill" action doesn't work and we end up with invalid or inconsistent state with trigger runs.

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
+++ b/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
@@ -35,6 +35,11 @@ function FormWrapper({
       pipeline: { name: 'test-pipeline', namespace: 'test-ns' },
       revision: { name: 'test-revision', namespace: 'test-ns' },
       actor: { name: 'test-user' },
+      sourceTriggerName: '',
+      autoFlip: false,
+      notifications: [],
+      kill: false,
+      action: TriggerRunAction.NO_ACTION,
     },
     status: { state: TriggerRunState.RUNNING },
   };
@@ -86,7 +91,6 @@ it.each([
       expect(mockRequest).toHaveBeenCalledWith(
         'UpdateTriggerRun',
         expect.objectContaining({
-          // objectContaining doesn't preserve generics — cast required by Vitest's type inference
           triggerRun: expect.objectContaining({
             spec: expect.objectContaining({ action }) as Record<string, unknown>,
           }) as Record<string, unknown>,
--- a/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
+++ b/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
@@ -11,18 +11,13 @@ import {
   createQueryMockRouter,
   getServiceProviderWrapper,
 } from '#core/test/wrappers/get-service-provider-wrapper';
-import {
-  KillTriggerRunForm,
-  PauseTriggerRunForm,
-  ResumeTriggerRunForm,
-} from '../trigger-run-action-form';
+import { KillTriggerRunForm } from '../trigger-run-action-form';
 import { TriggerRunAction, TriggerRunState } from '../types';
 
 import type { ActionComponentProps } from '#core/components/actions/types';
 import type { TriggerRun } from '../types';
 
 // Provides real isOpen state so FormDialog can close on success.
-// Record is co-located here since all three forms test the same trigger run.
 function FormWrapper({
   Form,
 }: {
@@ -52,18 +47,6 @@ it.each([
     dialogName: 'Kill Trigger Run',
     buttonLabel: 'Kill',
     action: TriggerRunAction.KILL,
-  },
-  {
-    Form: PauseTriggerRunForm,
-    dialogName: 'Pause Trigger Run',
-    buttonLabel: 'Pause',
-    action: TriggerRunAction.PAUSE,
-  },
-  {
-    Form: ResumeTriggerRunForm,
-    dialogName: 'Resume Trigger Run',
-    buttonLabel: 'Resume',
-    action: TriggerRunAction.RESUME,
   },
 ])(
   '$dialogName: submits UpdateTriggerRun with correct action and closes dialog',
--- a/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
+++ b/javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
@@ -106,7 +106,8 @@ describe('KillTriggerRunForm', () => {
     const dialog = await screen.findByRole('dialog');
     await user.click(within(dialog).getByRole('button', { name: 'Kill' }));
 
-    await screen.findByText(/Test error/);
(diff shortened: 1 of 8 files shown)


javascript/packages/core/config/entities/trigger/trigger-run-action-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  const initialValues: TriggerRun = {
+    ...record,
+    spec: { ...record.spec, action: ACTION_TO_ENUM[action] },
+  };
+
+  const handleSubmit = async (values: TriggerRun) => {
+    if (mutation.isPending) return;
+    await mutation.mutateAsync({ triggerRun: values });
```
austingreco, line 41: After this resolves, cached `GetTriggerRun`/`ListTriggerRun` data still shows the old RUNNING state, so the user has to refresh to see anything change (the PR description calls this out). Worth invalidating both keys here since Kill is reachable from list and detail views:

```ts
const queryClient = useQueryClient();

const handleSubmit = async (values: TriggerRun) => {
  await mutation.mutateAsync({ triggerRun: values });
  await queryClient.invalidateQueries({
    queryKey: ['GetTriggerRun', { name: record.metadata.name, namespace: record.metadata.namespace }],
  });
  await queryClient.invalidateQueries({ queryKey: ['ListTriggerRun'] });
};
```

`retry-cell.tsx` only invalidates the detail query because retry is detail-only — Kill needs both.

javascript/packages/core/config/entities/trigger/trigger-run-action-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+    mutationName: 'UpdateTriggerRun',
+  });
+
+  const initialValues: TriggerRun = {
+    ...record,
+    spec: { ...record.spec, action: ACTION_TO_ENUM[action] },
+  };
+
+  const handleSubmit = async (values: TriggerRun) => {
+    if (mutation.isPending) return;
```
austingreco: I think this guard is dead — `SubmitButton` already disables the button via react-final-form's `submitting` flag for the entire `await` window, so `mutation.isPending` can never be true while the button is clickable.

Worse, if it ever *did* fire: the silent `return` resolves `handleSubmit` successfully → `FormDialog` then unconditionally calls `onDismiss()` → the dialog closes without firing the mutation and without surfacing an error. Safer to just delete the line.

javascript/packages/core/config/entities/trigger/trigger.ts
```diff
@@ -1,12 +1,35 @@
...
+import { interpolate } from '#core/interpolation/interpolate';
 import { TRIGGER_DETAIL_CONFIG } from './detail';
 import { TRIGGER_LIST_CONFIG } from './list';
+import { KillTriggerRunForm } from './trigger-run-action-form';
+import { TriggerRunState } from './types';
 
 import type { PhaseEntityConfig } from '#core/types/common/studio-types';
+import type { TriggerRun } from './types';
+
+const isRunning = (record: unknown) =>
```
austingreco: Heads up — `controller.go:268-281` accepts `KILL` for both RUNNING and PAUSED runs, but this predicate disables Kill on PAUSED. Now that `PAUSED` just landed in the proto, this is a real regression: paused runs can't be killed from the UI even though the backend would happily kill them.

```ts
const isKillable = (record: unknown) => {
  const state = (record as TriggerRun).status?.state;
  return state === TriggerRunState.RUNNING || state === TriggerRunState.PAUSED;
};
```

And probably update the message to "Only running or paused trigger runs can be killed".

javascript/packages/core/config/entities/trigger/__tests__/trigger-run-action-form.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+        getErrorProviderWrapper(),
+        getRouterWrapper({ location: '/test-ns/triggers' }),
+        getServiceProviderWrapper({ request: mockRequest }),
+      ])
+    );
+
+    const dialog = await screen.findByRole('dialog');
+    await user.click(within(dialog).getByRole('button', { name: 'Kill' }));
+
+    await screen.findByText(/Test error/);
```
austingreco: I think this passes for the wrong reason. The mock throws `new Error('test')` but `getErrorProviderWrapper` normalizes that to a hardcoded `ApplicationError('Test error', ...)` regardless of input — so the assertion is matching the wrapper's literal, not anything the form renders. A regression that swallowed the real error entirely would still pass.

Suggested:
```ts
await within(dialog).findByText(/Test error/);
expect(mockRequest).toHaveBeenCalledTimes(1);
```

Scopes to the dialog and proves the mutation actually fired.

javascript/packages/core/config/entities/trigger/trigger-run-action-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <FormDialog<TriggerRun>
+      isOpen={isOpen}
+      onDismiss={onClose}
+      heading={config.heading}
+      onSubmit={handleSubmit}
+      submitLabel={config.submitLabel}
+      initialValues={initialValues}
+    >
+      <StringField name="metadata.name" label="Trigger Run Name" readOnly />
```
austingreco, line 68: Since Kill is destructive and irreversible, worth adding a sentence of context above the read-only field. Right now the dialog is just a name with a "Kill" button — easy to misclick on the wrong row when triggers generate similarly-named runs:

```tsx
<p>
  Kill run <strong>{record.metadata.name}</strong> in pipeline{' '}
  <strong>{record.spec.pipeline.name}</strong>? This action cannot be undone.
</p>
<StringField name="metadata.name" label="Trigger Run Name" readOnly />
```

(Or swap the whole thing for `ConfirmDialog` with `destructive`, but that's a bigger change — adding prose here is the smaller win.)

austingreco (approved):
We should follow up on the setTimeout. I think we need to change the backend ideally.
UpdateTriggerRun should synchronously set status.state = PENDING_KILL when spec.action = KILL

PR #1059, opened 2026-04-06
Ayao/implement pipeline apply update path

yaoyueyue:
What type of PR is this? (check all applicable)                                                                                                                                                                                                            
  - Feature                                                                                                                             
  - Bug Fix                                                                                                                                                                                                                                    
                                                                                                                                        
  What changed?                                                                                                                         
                                                                                                                                        
  - Added pipeline_apply_func_impl in plugins/entity/pipeline/apply.py — a pipeline-specific apply that performs a silent get before deciding create vs update, and copies resourceVersion for optimistic concurrency on the update path                                   
  - Added convert_crd_metadata_pipeline_apply for enriching the update request with fresh git/commit info, owner, and uniflow artifacts  (same enrichment as create)                                                                                                           
  - Wired the override in main.py: sets crd._apply_func_impl = partial(pipeline_apply_func_impl, get_method_info) and assigns separate converters for create vs update paths                                                                                                 
  - Added _apply_func_impl hook in crd.py's generate_apply so plugins can override without modifying core logic
  - Added optional yaml_dict param to get_crd_namespace_and_name_from_yaml and read_yaml_to_crd_request to avoid double file reads      
  - Consolidated apply/create/get logic out of mactl.py into the pipeline plugin;                                                             
  - E2E tested mactl pipeline apply against ma-dev-test namespace with bert-cola pipeline (both create and update paths)                
                                                                                                                                        
  Why?                                                                                                                                  
                                                                                                                                        
  The default apply_func_impl calls self.get → get_func_impl which always prints the full protobuf response — undesirable during apply. 
  The pipeline plugin now owns its apply flow end-to-end, calling gRPC directly for the existence check and handling create/update branching with the correct metadata converter for each path.                                                                          
                                                                       
  How did you test it?                                                                                                                  
  
  - E2E tested against ma-dev-test/bert-cola — confirmed both create path and update path work correctly with no spurious output        
  - 100% unit test coverage on crd.py and plugins/entity/pipeline/apply.py 
  - All ruff lint checks pass across all modified Python files                                                                          
                                                                                                                                        
  Potential risks                                                                                                                       
                                                                                                                                        
 Pipeline apply update path is new behavior — previously apply always fell through to create for existing pipelines. Other entity types are unaffected.                                                     
                                                                                                                                        
  Release notes                                                        

mactl pipeline apply now correctly updates an existing pipeline: copies resourceVersion for optimistic concurrency, re-enriches the  spec with current git/commit info and uniflow artifacts, and uses a full-replace (not deep-merge) update.
                                                                                                                                        
  Documentation Changes                                                                                                                 
   
  None


python/michelangelo/cli/mactl/crd.py
```diff
@@ -387,27 +372,65 @@ def apply_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Me
 
     _file = get_single_arg(bound_args.arguments, "file")
 
-    _namespace, _name = get_crd_namespace_and_name_from_yaml(_file)
+    yaml_dict = yaml_to_dict(_file)
```
hwaminkim: apply function supposed to be simple as possible and reuse internal `get` or `create` function. what's the reason of this change?

python/michelangelo/cli/mactl/crd.py
```diff
@@ -440,6 +463,7 @@ def __init__(self, name: str, full_name: str, metadata: list):
         self.name = name
         self.full_name = full_name
         self.func_crd_metadata_converter = convert_crd_metadata
+        self._get_method_info: Optional[CrdMethodInfo] = None
```
hwaminkim: what's the purpose of this attribute? if you'd like to call a custom get method, you can override it from the plugin level.

python/michelangelo/cli/mactl/crd.py
yaoyueyue: the original code read the YAML file twice — once in get_crd_namespace_and_name_from_yaml and again later in read_yaml_to_crd_request. The current code reads it once and passes yaml_dict down.

python/michelangelo/cli/mactl/crd.py
yaoyueyue: the purpose is to avoid self.get printing out the res during apply

python/michelangelo/cli/mactl/crd.py
yaoyueyue: I can slightly change the helper funciton and seperate this part out

python/michelangelo/cli/mactl/crd.py
yaoyueyue: Discueed off line, moved current changes in apply_func_impl to pipeline level. Will improve the apply func in crd.py in next PR

python/michelangelo/cli/mactl/crd.py
```diff
@@ -687,7 +687,9 @@ def generate_apply(self, channel: Channel, parser: Optional[ArgumentParser] = No
         self.configure_parser("apply", parser)
         func_signature = self._read_signatures("apply")
 
-        bound_func = partial(apply_func_impl, method_info)
+        bound_func = partial(
+            getattr(self, "_apply_func_impl", apply_func_impl), method_info
```
yaoyueyue: fyi: apply_plugin_command runs before generate_apply, so plugins can only inject behavior by setting attributes on crd ahead of time. Without this getattr hook, we need to reimplement and overwrite the generate_apply method.

hwaminkim: Would you try this (or similar) way? I'm still hesitating to changing CRD, unless we have a clear design principal. Also I don't prefer to put too many changes into single PR. (all PR should be atomic)

```py
     if target_command == "apply":
         crd.func_crd_metadata_converter = convert_crd_metadata_pipeline_apply
         crd.func_crd_metadata_converter_for_create = (
             convert_crd_metadata_pipeline_create
         )
         # put import clause here for easier understand
         import michelangelo.cli.mactl.crd as crd_module
         crd_module.apply_func_impl = pipeline_apply_func_impl
```

yaoyueyue: Updated wit this approach and it worked well. The potential risk is if mactl ever becomes a long-lived process handling multiple entity types, the patch would affect all CRDs. For now as a CLI tool, each invocation is a fresh process, this is safe.

hwaminkim (approved):
LGTM

PR #1068, opened 2026-04-08
Add ESLint rule: types must live in types.ts files

craigmarker:
Enforces that type/interface declarations belong in types.ts (or *-types.ts, or files in a types/ directory). Only *Props interfaces used as component function parameters or forwardRef generics are exempt.

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/javascript/eslint-local-rules/__tests__/types-in-types-file.test.js
+++ b/javascript/eslint-local-rules/__tests__/types-in-types-file.test.js
@@ -139,7 +139,7 @@ tester.run('types-in-types-file', rule, {
         interface Config { label: string; }
         export function MyComponent(props: Config) { return null; }
       `,
-      errors: [{ messageId: 'typesInTypesFile', data: { name: 'Config' } }],
+      errors: [{ messageId: 'inlineType', data: { name: 'Config' } }],
     },
 
     // ── mixed: one valid Props + one invalid non-Props ──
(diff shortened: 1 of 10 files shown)


javascript/packages/rpc/runtime-config.ts
donovanlopez, line 1: I'm not a huge fan of having to declare the interface inline in three different spots. My main concern would be  if there was any inconsistencies between them or we have to update one, we have to find them all and adjust it accordingly. Though this file isn't that bad of an example to really try to push for a change.

donovanlopez (approved):

javascript/packages/rpc/runtime-config.ts
craigmarker: When I returned to my branch, I discovered I had actually relocated the inline interface into a sibling types.ts file but not committed 😅 We're aligned!

PTAL at most recent commit

javascript/eslint-local-rules/__tests__/types-in-types-file.test.js
```diff
@@ -0,0 +1,213 @@
...
+        const Item = forwardRef<HTMLDivElement, ItemProps>((props, ref) => null);
+      `,
+    },
+
+    // ── re-exports are allowed (no declaration) ──
+    {
+      name: 're-export type from types file',
+      filename: 'component.tsx',
+      code: `export type { Foo } from './types';`,
+    },
```
konukselim, line 84: could you help me understand why this would be needed in the case where barrel files aren't allowed?

javascript/eslint-local-rules/__tests__/types-in-types-file.test.js
```diff
@@ -0,0 +1,213 @@
...
+
+    // ── mixed: one valid Props + one invalid non-Props ──
+    {
+      name: 'mixed: Props used as param (valid) + helper type (invalid)',
+      filename: 'component.tsx',
+      code: `
+        interface MyComponentProps { label: string; }
+        interface HelperType { key: string; }
+        export function MyComponent(props: MyComponentProps) { return null; }
+      `,
```
konukselim, line 153:
```
interface HelperType { key: string; }
interface MyComponentProps { helper: HelperType; }
export function MyComponent(props: MyComponentProps) { return null; }
```

Should this be allowed 🤔

Assume there are many `HelperType` attributes and they are only used within the component and the type isn't needed anywhere else.

craigmarker: Surprisingly, there were only three instances of this violation. Both of the packages/rpc types (`Services` and `RuntimeConfig`) feel core enough to be relocated to a sibling types.ts file. And the single type in packages/core fits nicely inline.

What do you think?

javascript/eslint-local-rules/__tests__/types-in-types-file.test.js
craigmarker: I _think_ barrel files are only disallowed if they are named `index.tsx`. I don't know of another lint rule in our repo that disallows exporting multiple consts/functions from a single file.

PR #1026, opened 2026-03-31
docs: add compliance guide for SOC 2, GDPR, and HIPAA

zhoward-1:
## Summary

This PR makes several documentation changes across contributing, operator, and user guides:

- **New compliance guide** (`docs/operator-guides/compliance.md`) — operator-facing guide covering SOC 2, GDPR, and HIPAA compliance for Michelangelo deployments
- **Ingester docs reorganized** — `ingester-design.md` renamed to `ingester-internals.md` (contributing guide, trimmed to developer-focused content) and split into a new `ingester-configuration.md` (operator guide covering deployment, config, and operations)
- **Sandbox validation report removed** (`docs/operator-guides/ingester-sandbox-validation.md`) — deleted as it was a one-time validation artifact, not reference documentation
- **New notifications user guide** (`docs/user-guides/notifications.md`) — covers email and Slack notification configuration for pipeline runs, trigger runs, and pipelines
- **Overview FAQ updated** (`docs/getting-started/overview.md`) — "Is my data secure?" now links to the new compliance guide
- **User guides index updated** (`docs/user-guides/index.md`) — notifications guide added to the index

## What's in the compliance guide

- **SOC 2**: RBAC config, TLS/IAM encryption, network isolation, HA setup, audit log retention
- **GDPR**: EU data residency config, data minimization, right to access/erasure/portability workflows, DPA requirements
- **HIPAA**: Encryption table (transit/rest/etcd), 6-year audit log retention, S3 Object Lock (WORM), administrative safeguards, BAA requirements, and a pre-launch PHI readiness checklist
- **General**: Kubernetes hardening, monitoring/alerting guidance, incident response runbooks with breach notification timelines

## Test plan

- [x] Verify `docs/operator-guides/compliance.md` renders correctly in the docs site
- [x] Verify the link from `overview.md` resolves correctly
- [x] Run `cd website && bun run build` to confirm no broken links

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1074, opened 2026-04-08
docs: address post-merge review feedback from PR #1026

zhoward-1:
## Summary

Addresses the four post-merge review comments left by @austingreco on michelangelo-ai/michelangelo#1026 after it was already merged.

- **`ingester-internals.md`**: Added an overview paragraph before "Finalizer Implementation" explaining what the ingester is, restoring context that existed in the old `ingester-design.md`
- **`ingester-configuration.md`**: Restored the full Pre-Existing Objects trade-off — explains orphan rows left behind when `kubectl delete` is used on objects without the finalizer, and documents both mitigations (backfill controller + API Server path), not just the workaround
- **`notifications.md`**: Added a missing `## Enabling Notification Delivery` section that fixes three broken anchor links across the file (in the `:::caution` block, the `:::tip`, and Troubleshooting)
- **`notifications.md`**: Removed the "Reference Files" section pointing users to internal Go source files — moved that context into the new operator-facing section above

## Test plan

- [ ] Verify anchor links resolve: `#enabling-notification-delivery` appears three times in `notifications.md` and should now point to the new section
- [ ] Confirm `ingester-internals.md` opens with an overview before jumping into finalizer details
- [ ] Confirm `ingester-configuration.md` Pre-Existing Objects section explains the orphan row consequence and both mitigations


austingreco (commented):
## Rebase needed before merge

The ingester doc changes are solid, but `notifications.md` has a **rebase issue** that will cause silent data loss during merge.

**What happened:** After this PR branched, commit `a291f548` (PR #1078) landed on main and replaced the three `[Enabling Notification Delivery](#enabling-notification-delivery)` anchor links in `notifications.md` with plain text (as a stopgap because the target heading didn't exist yet on main). This PR adds that heading back — but since it doesn't touch the lines with the anchor links, git's 3-way merge will silently keep main's plain-text versions.

**After merge without rebase, the result will be:**
- Line 12 (caution block): `See the reference files below for implementation details.` — but the Reference Files section no longer exists
- Lines 60 and ~198: plain text instead of the intended anchor links to `#enabling-notification-delivery`

**Fix:** Rebase onto current main, then re-add the three anchor links so they point to the new `## Enabling Notification Delivery` heading that this PR creates. This restores the PR's stated goal of fixing those three broken anchor links.

austingreco (commented):
## 3 lines need updating (outside current diff)

After the rebase, these three lines still have plain text from main that should now link to the new `## Enabling Notification Delivery` section this PR adds. They're outside the diff so I can't post inline suggestions — here are the exact changes:

---

**Line 12** (caution block) — change:
```
See the reference files below for implementation details.
```
to:
```
See [Enabling Notification Delivery](#enabling-notification-delivery) below for details.
```

---

**Line 60** (tip block) — change:
```
Messages will be delivered once your operator has enabled notification delivery.
```
to:
```
Messages will be delivered once your operator has [enabled notification delivery](#enabling-notification-delivery).
```

---

**Line 198** (troubleshooting) — change:
```
check with your platform administrator.
```
to:
```
check with your platform administrator — see [Enabling Notification Delivery](#enabling-notification-delivery) below.
```

austingreco (approved):

PR #1083, opened 2026-04-09
Update example docker image to use ghcr

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Update example docker image to use ghcr

**Why?**
we have generated example docker image so we don't have to do k3d anymore in the sandbox

**How did you test it?**
local remote run/dev-run

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/docs/user-guides/ml-pipelines/getting-started.md
+++ b/docs/user-guides/ml-pipelines/getting-started.md
@@ -198,7 +198,7 @@ metadata:
   namespace: my-project  # Your project name
   name: boston-housing-xgb
   annotations:
-    michelangelo/uniflow-image: docker.io/library/my-workflow:latest
+    michelangelo/uniflow-image: ghcr.io/michelangelo-ai/examples:main
 spec:
   type: PIPELINE_TYPE_TRAIN
   manifest:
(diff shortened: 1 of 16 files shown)


aritra0227:
According to [this doc](https://github.com/michelangelo-ai/michelangelo/blob/main/python/examples/README.md?plain=1), the current steps to build an image for a pipeline are:

```
1. docker build -t examples:latest -f ./examples/Dockerfile .
2. k3d image import examples:latest -c michelangelo-sandbox
```
Then you can run the pipeline through:
```
3.   $ PYTHONPATH="." poetry run python ./examples/bert_cola/bert_cola.py remote-run --image docker.io/library/examples:latest --storage-url s3://default --yes
```

With this change, how does the user experience change? Would you skip step 1 and 2 completely?

Maybe we should also update the doc?

aritra0227 (approved):
Resolved offline. Sorry I forgot that you don't even need the pipeline.yaml when following [this readme](https://github.com/michelangelo-ai/michelangelo/blob/main/python/examples/README.md?plain=1) since it shows steps for how to directly invoke the worklow.

PR #371, opened 2025-09-09
Cron Trigger Support with Dynamic Parameters

yangfan0328:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR implements a K8s controller that manages TriggerRun CR with dual Cadence/Temporal orchestration support, workflow/activity implementations, and comprehensive unit tests.

Here's the complete structure:
Workflow Client Abstraction
go/base/workflowclient/
├── interface/workflowclient.go              # Added ListOpenWorkflow() and TerminateWorkflow() methods
├── interface/interface_mock/workflowclient.go # Generated mocks for new interface methods
├── cadenceclient/cadenceclient.go           # Implemented ListOpenWorkflow() and TerminateWorkflow() for Cadence
├── cadenceclient/cadenceclient_test.go      # Unit tests for new Cadence client methods
├── temporalclient/temporalclient.go         # Implemented ListOpenWorkflow() and TerminateWorkflow() for Temporal
├── temporalclient/temporalclient_test.go    # Unit tests for new Temporal client methods with QueryWorkflow tests
└── temporalclient/BUILD.bazel               # Added temporal package dependency for test mocks

TriggerRun Controller
go/components/triggerrun/
├── controller.go           # K8s reconciliation controller for TriggerRun CRD lifecycle
├── runner.go              # Runner interface definition (Run, Kill, GetStatus)
├── cron_trigger.go        # Cron trigger implementation with schedule validation
├── util.go                # Utility functions for workflow execution and status management
├── module.go              # FX dependency injection module
├── controller_test.go     # Unit tests for controller reconciliation logic (gomock)
├── cron_trigger_test.go   # Unit tests for cron trigger with schedule
├── util_test.go           # Unit tests for utility functions and workflow lifecycle
└── BUILD.bazel            # Bazel build configuration for controller

Worker Activities
go/worker/activities/trigger/
├── trigger_activities.go                    # Activity implementations (GenerateBatchRunParams, GenerateConcurrentRunParams, CreatePipelineRun, GetPipelineRunSensor)
├── trigger_activities_test.go               # Unit tests for all activities with mocked logger/clients
├── module.go                                # FX module for activity registration
├── BUILD.bazel                              # Activity build configuration
└── parameter/
    ├── parameter_generator.go               # Parameter generator interface
    ├── cron_trigger.go                      # Cron parameter generation with batch policy or max concurrency support
[...]

PR #1084, opened 2026-04-10
Simplify PR template and add write-pr-description skill

craigmarker:
## Summary

Two weeks ago we changed our squash & merge configuration so that the full PR description body becomes the squash commit message. The 33 commits on main since this change suggest the existing template generates more noise than signal.

This PR simplifies the template to two sections — Summary and Test plan — and adds a Claude Code skill (/write-pull-request-description) to help authors fill them well. The two-section format intends to eliminate the redundant and routinely-dismissed fields.

## Test plan
- Ran the skill against 3 real commits. Output included in **Before / After Examples**
- Eval suite with 9 test cases checked into `.claude/skills/write-pull-request-description/evals/`
- Audited commits on main for adherence to existing template.

### Commit audit

The commit audit covering 83d66c1f^..main (33 commits). **Only 13 PRs use the existing PR template**. The other 20 commits used a custom template.

**Documentation Changes**: 12 of 13 PRs dismissed this section with "None", "N/A", or a circular answer ("This PR is the documentation change").

**Release notes**: Duplicated "What changed?" in 5 of 13 PRs (#1086, #1090, #1097, #1107, #1122) and duplicated "Why?" in 3 more. No PR used the section to add information not already present elsewhere

**What changed? / Why?**: Overlaps in the majority of PRs where both fields were filled. Authors either wrote the same sentence in both fields or left one blank.

**Type checkboxes**: The existing template instruction says, "check all applicable", but 12 of 13 PRs checked one box. Several of these PRs spanned multiple concern areas (#1097, #1122)

**Potential risks**: Many PRs include "None" or omit potential risk. Those that include potential risks contain useful information. This can be included on an as-needed basis if **Summary** and **Test plan** are insufficient.

### Before / After Examples

These were generated by running the skill against real commits from our history.

**Bug fix (3 files) — original commit message:**
> fix(ingester): use scheme.ObjectKinds() to resolve GVK, fixing controllermgr CrashLoopBackOff (#922)
> * Add unit test
> * Add AllCRDObjects to proto-go module
> * update todo
> * revert

**What the skill produces:**
> Use scheme-based GVK resolution in ingester controller
>
> ## Summary
> Replace `GetObjectKind().GroupVersionKind()` with `Scheme.ObjectKinds()` for GVK resolution in the ingester controller. `GetObjectKind()` returns an empty GVK when the object has not been through serialization, causing silent failures in metadata storage operations. `Scheme.ObjectKinds()` resolves the GVK from the scheme registry, which is always populated after `AddToScheme`.
>
> ## Test plan
> Added `TestSchemeGVKResolution` — iterates all CRD objects in the v2 scheme and asserts each resolves to a non-empty, unique Kind.

---

**Large feature (48 files) — original commit message:**
> Cron Trigger Support with Dynamic Parameters (#371)

**What the skill produces:**
> Add cron trigger support with dynamic parameters
>
> ## Summary
> Add scheduled workflow execution via cron triggers with support for dynamic parameters resolved at runtime. Previously, workflows could only be triggered manually or via external schedulers with static configurations. This introduces a cron trigger system with a new TriggerRun controller, dedicated Cadence/Temporal activities and workflows, and workflow client extensions so both backends can run cron-triggered workflows through a unified API.
>
> ## Test plan
> Added `cron_trigger_test.go` covering trigger creation, parameter resolution, and execution lifecycle. Existing workflow client tests pass with the expanded interface.

## Related
- Issue: #1138


donovanlopez:
@craigmarker I noticed that the commit titles will go from:
`fix(ingester): use scheme.ObjectKinds() to resolve GVK, fixing controllermgr CrashLoopBackOff (#922)` 
to 
`Use scheme-based GVK resolution in ingester controller`

I thought we were following the [Conventional Commit convention](https://www.conventionalcommits.org/en/v1.0.0/) before but it looks like we're moving away from that. Looking at our commit history, we were somewhat doing this but wanted to check if this is something we wanted to follow.

craigmarker:
> I thought we were following the [Conventional Commit convention](https://www.conventionalcommits.org/en/v1.0.0/) before but it looks like we're moving away from that. Looking at our commit history, we were somewhat doing this but wanted to check if this is something we wanted to follow.

I’m open to following the type prefixing (feat/fix/docs/etc). One concern I have about the type prefixing approach is that when we’re squashing and merging, a single PR of multiple commits becomes a single git history entry. It’s easy for several commits to satisfy multiple types.

If we supported rebase & merge, I wouldn’t be concerned about this problem.

What do you think?

donovanlopez:
> > I thought we were following the [Conventional Commit convention](https://www.conventionalcommits.org/en/v1.0.0/) before but it looks like we're moving away from that. Looking at our commit history, we were somewhat doing this but wanted to check if this is something we wanted to follow.
> 
> I’m open to following the type prefixing (feat/fix/docs/etc). One concern I have about the type prefixing approach is that when we’re squashing and merging, a single PR of multiple commits becomes a single git history entry. It’s easy for several commits to satisfy multiple types.
> 
> If we supported rebase & merge, I wouldn’t be concerned about this problem.
> 
> What do you think?

Ah I see. In my last company, we actively tried to keep our PRs as small as possible to have it more focused and easier to review so the type prefixing made more sense since PRs covered only one of those. If we're not enforcing anything like that, it makes sense to not follow it.

cm-howard (approved):

.claude/skills/write-pull-request-description/evals/evals.json
```diff
@@ -0,0 +1,131 @@
...
...]bool{}\n+\tfor _, obj := range v2.AllCRDObjects {\n+\t\tgvks, _, err := scheme.ObjectKinds(obj)\n+\t\trequire.NoError(t, err)\n+\t\trequire.NotEmpty(t, gvks)\n+\t\tkind := gvks[0].Kind\n+\t\tassert.NotEmpty(t, kind)\n+\t\tassert.False(t, seen[kind])\n+\t\tseen[kind] = true\n+\t}\n+}\n+\n+func TestHandleDeletion_CorrectTypeMeta(t *testing.T) {\n+\t// ... tests that handleDeletion passes correct Kind/APIVersion from scheme\n+}\n+\n+func TestHandleDeletionAnnotation_CorrectTypeMeta(t *testing.T) {\n+\t// ... tests annotation deletion path uses scheme-based GVK\n+}\ndiff --git a/go/components/ingester/module.go b/go/components/ingester/module.go\nindex 44960404..bebd5a41 100644\n--- a/go/components/ingester/module.go\n+++ b/go/components/ingester/module.go\n@@ -40,7 +40,11 @@ func register(p registerParams) error {\n \tcrdObjects := v2.AllCRDObjects\n \n \tfor _, obj := range crdObjects {\n-\t\tgvk := obj.GetObjectKind().GroupVersionKind()\n+\t\tgvks, _, err := p.Scheme.ObjectKinds(obj)\n+\t\tif err != nil || len(gvks) == 0 {\n+\t\t\treturn fmt.Errorf(\"failed to get GVK for %T: %w\", obj, err)\n+\t\t}\n+\t\tgvk := gvks[0]\n \t\tlog := p.Logger.With(zap.String(\"kind\", gvk.Kind))\n",
```
sallycr: what is this prompt?

.claude/skills/write-pull-request-description/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Summary Core Principles
+
+- Explain what problem the code solves, not how the code works
+- Explain what was wrong before, how it works now, and why this solution was chosen
+- When renaming or restructuring, explain what motivated the change
+- Summarize the approach at the decision level, not step-by-step
+- Be concise, not persuasive
+- Wrap body lines at 72 characters
+- If the branch name or commit messages reference a ticket (e.g., PROJ-123,
```
sallycr: we don't use ticket? should it be changed to github issue?

.claude/skills/write-pull-request-description/evals/evals.json
sallycr: how does evals.json works?

sallycr (approved):

PR #1092, opened 2026-04-15
[mactl] load metadata with lazy way in pipeline dev-run plugin

hwaminkim:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

dev_run no longer imports METADATA_STUB from crd.py at module load time. The gRPC stub_method is now called with [*_self.metadata, ("ttl", "600")], resolved from the CRD instance at call time.

**Why?**

  METADATA_STUB in crd.py starts as [] and is rebound to a new list in CRD.__init__. Because Python's from module import name captures the binding at import time, dev_run.py always held a reference to the original empty list — causing every dev-run gRPC call to be sent with no
  metadata. The server rejected it with "missing service name, caller name, encoding".

  This was introduced in #537 when mactl.py was split into crd.py, changing from a one-time module-level assignment to a lazy init pattern.

**How did you test it?**

  - ma pipeline dev-run -f ./examples/bert_cola/pipeline.yaml runs without the gRPC error
  - Added regression test test_dev_run_grpc_call_uses_crd_metadata that asserts stub_method is called with the full CRD metadata

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
+++ b/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
@@ -14,7 +14,6 @@ from grpc import Channel
 
 from michelangelo.cli.mactl.crd import (
     CRD,
-    METADATA_STUB,
     bind_signature,
     get_single_arg,
     inject_func_signature,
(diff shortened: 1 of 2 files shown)


yaoyueyue (approved):

badcount (approved):

hwaminkim:
Integration test is passed: https://github.com/michelangelo-ai/michelangelo/actions/runs/24475784347/job/71527461374

PR #1028, opened 2026-04-01
[MaCTL] Change plugin import timing earlier and preparation for module level override

hwaminkim:
**What type of PR is this? (check all applicable)**
- [X] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

1) Search plugins at the earliest timing of ma command.
2) Refactor plugin loading command to prepare the module level plugin override.

**Why?**

**How did you test it?**

Running ma plugin command in local
 - checked pipeline plugin commands
 - checked pipeline registration with tar file upload

**Potential risks**

**Release notes**

**Documentation Changes**

PR #1093, opened 2026-04-15
fix: use instance metadata in dev_run to fix missing YARPC headers

badcount:
## Summary

- `METADATA_STUB` in `crd.py` is initialized as `[]` at module level and only set during `CRD.__init__`
- PR #1028 moved plugin imports earlier (before CRD initialization), so `from crd import METADATA_STUB` in `dev_run.py` captured the empty list at import time
- Fix: use `_self.metadata` from the CRD instance directly inside `dev_run_func` instead of the stale module-level import

## Test plan
- [ ] `ma pipeline dev_run -f ./examples/bert_cola/pipeline.yaml` no longer fails with `missing service name, caller name, encoding`
- [ ] CI e2e pipeline run completes successfully

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
+++ b/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
@@ -14,7 +14,6 @@ from grpc import Channel
 
 from michelangelo.cli.mactl.crd import (
     CRD,
-    METADATA_STUB,
     bind_signature,
     get_single_arg,
     inject_func_signature,
@@ -218,7 +217,7 @@ def generate_dev_run(
         )
         response = stub_method(
             request_input,
-            metadata=METADATA_STUB,
+            metadata=[*_self.metadata, ("ttl", "600")],
             timeout=30,
         )
         _LOG.info("Stub method completed (%r): %r", type(response), response)


paulzim (approved):
## Review: #1093 — fix: use instance metadata in dev_run to fix missing YARPC headers

**Verdict: Approve with one follow-up note**

Correct, minimal fix with a good regression test. One sibling-file pattern worth tracking down separately.

### Root cause is accurate

`METADATA_STUB = []` in `crd.py` line 34 is the module-level initialization — an empty list. `CRD.__init__` (line 700-701) later does `global METADATA_STUB; METADATA_STUB = [*metadata, ("ttl", "600")]`, replacing the name in the `crd` module's namespace.

The problem: `dev_run.py` used `from michelangelo.cli.mactl.crd import METADATA_STUB`, which binds to the *object* at import time — the empty `[]`. When #1028 moved plugin imports earlier (before `CRD.__init__` runs), the `dev_run.py` binding was frozen to the empty list and was never updated when `CRD.__init__` later reassigned the global.

The fix correctly uses `[*_self.metadata, ("ttl", "600")]` where `_self` is the CRD instance, reading the live value at call time. The `("ttl", "600")` addition mirrors exactly what `CRD.__init__` adds to the global — confirmed by comparing `crd.py` line 701 with the new `dev_run.py` line 220.

### Regression test is good

`test_dev_run_grpc_call_uses_crd_metadata` (line 448 of `dev_run_test.py`) sets `mock_crd.metadata = [("rpc-caller", "test")]` and asserts the gRPC call receives `[*crd_metadata, ("ttl", "600")]`. This directly exercises the previously-broken path and would have caught the original bug.

### One follow-up note: `kill.py` and `delete.py` have the same stale-import pattern

Both files import `METADATA_STUB` the same way this PR fixed:

- `pipeline/kill.py` line 19: `from michelangelo.cli.mactl.crd import METADATA_STUB`, used at line 153
- `pipeline/delete.py` line 24: `from michelangelo.cli.mactl.crd import METADATA_STUB`, used at line 161

If #1028's import-ordering change affects these commands too, they'd silently send empty metadata. The safe files (`run.py`, `trigger_run/kill.py`, `trigger_run/create.py`) access `crd_module.METADATA_STUB` as a module attribute at call time — that's the correct pattern. Worth a follow-up PR applying the same instance-metadata fix to `kill.py` and `delete.py`.

Not a blocker for this PR since the scope is `dev_run` and the fix here is correct.

### CI

All checks pass. `license/cla` pending — Eric has signed on recent PRs, likely a lag.

PR #1115, opened 2026-04-23
fix: use k8s Secret for MinIO credentials to survive sandbox sync

badcount:
## Summary

- **Root cause of CI failure**: `ma sandbox sync` deleted and re-applied the `aws-credentials` Secret every run, and always applied `michelangelo-config.yaml` with hardcoded `minioadmin` values — wiping any manual credential updates on the GCP VM.
- **New `minio-credentials` k8s Secret** (`minioadmin` defaults in git) stores credentials as env vars (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`).
- **`sandbox sync` is now create-only** for both credential secrets: if a secret already exists on the VM, it is left untouched. The GCP CI runner pre-creates this secret with real credentials once; they survive across all subsequent CI runs.
- **`michelangelo-config` ConfigMap is patched at sync time** from the live secret, so Ray pods (which consume the ConfigMap via `envFrom`) also see the correct credentials.
- **Worker/controllermgr** gain `envFrom: secretRef: minio-credentials` and use `${AWS_ACCESS_KEY_ID::minioadmin}` substitution in `base.yaml`, removing the hardcoded value.
- **`integration-test.sh`** reads `MINIO_ACCESS_KEY`/`MINIO_SECRET_KEY` from the k8s Secret at runtime — no credentials in the script or workflow YAML.

## Test plan

- [ ] Verify local dev: `ma sandbox create` creates both secrets with `minioadmin` defaults, worker/controllermgr start correctly
- [ ] Verify GCP VM: pre-create `minio-credentials` secret with `ma-minio`; run `ma sandbox sync`; confirm secret is NOT overwritten and ConfigMap is patched to `ma-minio`
- [ ] Confirm integration test CI run passes

### DIFF
--- a/.github/scripts/integration-test.sh
+++ b/.github/scripts/integration-test.sh
@@ -18,8 +18,11 @@ set -euo pipefail
 
 NAMESPACE="${MA_NAMESPACE:-ma-dev-test}"
 MINIO_ENDPOINT="${MINIO_ENDPOINT:-http://localhost:9091}"
-MINIO_ACCESS_KEY="${MINIO_ACCESS_KEY:-minioadmin}"
-MINIO_SECRET_KEY="${MINIO_SECRET_KEY:-minioadmin}"
+# Read credentials from the minio-credentials k8s Secret so the script
+# automatically picks up whatever the sandbox VM is configured with.
+# Falls back to minioadmin for local dev where the Secret has defaults.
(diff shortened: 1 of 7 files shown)


sallycr (approved):
LGTM

PR #1117, opened 2026-04-24
feat: Support pipeline run ttl in config to delay the immutable set when reaching desired state

badcount:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Adds TTL-based ETCD eviction for terminal PipelineRuns and wires up the ingester finalizer correctly.

  PipelineRun TTL controller (go/components/pipelinerun/):
  - After a run reaches a terminal state (SUCCEEDED/FAILED/KILLED), the reconciler checks whether EndTime + ttlDays has elapsed
  - If elapsed: sets michelangelo/Immutable=true so the ingester evicts it from ETCD to MySQL-only storage
  - If not elapsed: requeues exactly at EndTime + ttlDays (not the default period), so eviction is timely without busy-polling
  - ttlDays=0 (default) disables TTL eviction; set >0 to enable
  - Config is read from the pipelineRun key via config.Provider, consistent with other controllers
  - Removed unused env.Context dependency from register()
  
  Ingester controller (go/components/ingester/):
  - Removed redundant AddFinalizer block from handleSync — the API handler already adds michelangelo/Ingester finalizer synchronously during Create() before writing to ETCD, so
   the controller adding it again was both redundant and racy (an object could be created and deleted before the controller's first reconcile)
  - Updated tests to reflect that objects always arrive with the finalizer pre-set
  
  Config (go/cmd/controllermgr/config/base.yaml):
  - Added ingester section with defaults (concurrentReconciles: 1, requeuePeriod: 30s)
  - Added pipelineRun section with commented ttlDays for operator visibility

**Why?**
Support pipeline run ttl in config to delay the immutable set when reaching desired state

**How did you test it?**
  - go test ./go/components/ingester/... passes
  - go test ./go/components/pipelinerun/... passes
  - Deploy to sandbox, create a PipelineRun, verify it runs to completion with finalizer present from the start
  - Set ttlDays: 0 (default) — verify no eviction occurs
  - Set ttlDays: 1 and back-date a terminal run's EndTime — verify michelangelo/Immutable=true is set and ingester removes it from ETCD

**Potential risks**

**Release notes**

**Documentation Changes**


go/components/ingester/controller.go
```diff
@@ -195,9 +227,13 @@ func (r *Reconciler) handleDeletion(ctx context.Context, log logr.Logger, object
...
 
-	// Remove our finalizer
+	// Remove our finalizer using Patch to avoid optimistic-concurrency conflicts.
+	base := object.DeepCopyObject().(client.Object)
 	ctrlutil.RemoveFinalizer(object, api.IngesterFinalizer)
-	if err := r.Update(ctx, object); err != nil {
+	if err := r.Patch(ctx, object, client.MergeFrom(base)); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			return ctrl.Result{}, nil
+		}
```
aritra0227: what does `handleDeletion` do? 

Shouldn't this function also delete the object from etcd like [how `handleDeletionAnnotation` handles it](https://github.com/michelangelo-ai/michelangelo/pull/1117/changes#diff-70d5359754db23830b2ab925611227b83d850001e0dd70bd80cf28c8a69f2371R278-R280)?

Currently this function does not delete from etcd.

go/components/ingester/controller.go
aritra0227, line 131: How do we handle the case where both `deletionTimestamp=0` and `deletionAnnotation` is set?

In the current code, both go through `r.handleDeletion`. Does that make sense?

aritra0227, line 197: Is there no logic to use `r.config.DeletionDelay` in this path?

aritra0227, line 136: Is there a reason to have a separate function for handling the case where deletion is marked via annotation?

go/components/ingester/controller_test.go
```diff
@@ -84,9 +84,10 @@ func TestReconciler_HandleSync(t *testing.T) {
 			Kind:       "Deployment",
```
aritra0227, line 84: what about test cases for the following scenarios:
1. setting deletiontimestamp=0 
2. marking object for deletion through annotations

aritra0227 (commented):
I have a few concerns. Please see comments.

go/components/ingester/controller.go
```diff
@@ -148,6 +148,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 // handleSync syncs the object to metadata storage
 func (r *Reconciler) handleSync(ctx context.Context, log logr.Logger, object client.Object) (ctrl.Result, error) {
+	// Add finalizer if not present to ensure MySQL sync happens before ETCD removal on deletion.
+	if !ctrlutil.ContainsFinalizer(object, api.IngesterFinalizer) {
```
ying-zheng: ingester finalizer is already added in api handler when the object is created
https://github.com/michelangelo-ai/michelangelo/blob/main/go/api/handler/handler.go#L547

it's too late to add finalizer in controller. An object could be created and deleted before this controller code is executed.

go/components/ingester/controller.go
```diff
@@ -176,26 +178,7 @@ func (r *Reconciler) handleDeletion(ctx context.Context, log logr.Logger, object
 		return ctrl.Result{}, nil
 	}
 
-	log.Info("Deleting from metadata storage")
```
badcount, line 179: This is duplicated logic which we have this already in line 205

go/components/ingester/controller.go
```diff
@@ -227,14 +214,9 @@ func (r *Reconciler) handleDeletionAnnotation(ctx context.Context, log logr.Logg
 		return ctrl.Result{RequeueAfter: r.getRequeuePeriod()}, err
 	}
 
-	// Remove finalizer
-	ctrlutil.RemoveFinalizer(object, api.IngesterFinalizer)
```
badcount, line 231: remove this duplicated logic which is in line 245 already

ying-zheng (approved):

PR #1123, opened 2026-04-28
Add list deployment page

donovanlopez:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

We added a list page for deployments. 

image

**Why?**
As we expand MA Studio, we want to add more features.

**How did you test it?**
I created sample deployments and created them locally to see them in the UI. Also created some unit tests for checking if the List Deployments page shows the correct columns and text. I assume the list component is already tested so no need to re-test the list's actual functionality.

**Potential risks**
None. If there's an issue, we can make this page not active. Users wouldn't be able to access this page unless they go to the specific URL.


python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -1077,6 +1079,8 @@ def _create_demo_crs(ns: argparse.Namespace):
         _create_pipeline_demo_crs()
     elif ns.demo_action == "inference":
         _create_inference_demo_crs()
+    elif ns.demo_action == "deployment":
+        _create_deployment_demo_crs()
```
aritra0227: I think we can/should remove this. So far, the `ma sandbox demo`  is _trying_ to display e2e features. I believe the `ma sandbox demo deployment` would only create a dummy CR. It would show up on the UI, but not trigger any actual model deployments. So its not really demo-ing the entire deployment feature.

craigmarker: I'm not sure this is _entirely_ true. When I originally created the sandbox demo in https://github.com/michelangelo-ai/michelangelo/commit/eca7b4ebed728cf6e6034896256a9622c3141a69, it was just to demonstrate how data would display in the UI.

It may be time for reconsidering the purpose of the sandbox demo functionality, also in the context of @sallycr work on the helm chart!

python/michelangelo/cli/sandbox/demo/deployment/mobile-deployment.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: michelangelo.api/v2
```
aritra0227: Are these manifests still required? If not, then maybe we should remove. We don't currently support any mobile/offline deployments at this moment so I think its confusing to keep these around.

donovanlopez: If that's the case, i'll remove the `mobile-deployment.yaml` and `offline-deployment.yaml`.

python/michelangelo/cli/sandbox/demo/deployment/online-deployment.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: michelangelo.api/v2
```
aritra0227: Sorry, earlier I meant that we should remove all manifests. From my understanding, this is not being used either so there's no point of keeping it.

aritra0227 (approved):
Left one comment but, otherwise lgtm. Would defer the stamp on Craig's approval for UI/Frontend. Thanks!

craigmarker (approved):

PR #1124, opened 2026-04-28
Add Ray cluster log persistence (proto + collector sidecar + log_url)

dragod812:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Wires the KubeRay History Server log-collection path end-to-end on the
controllermgr side. KubeRay History Server reference:
https://github.com/ray-project/kuberay/tree/master/historyserver

- **Proto**: add `log_url` to `RayClusterStatus` so callers can navigate
  to persisted logs after a cluster is gone.
- **Mapper** (`go/components/jobs/client/k8sengine`):
  - Load `LogPersistenceConfig` via FX from `jobs.k8sengine.mapper.logPersistence`.
  - `injectCollectorSidecar()` — inject the KubeRay collector sidecar into
    head + worker pod templates (env-var S3 config, no ConfigMap),
    reuse an existing `/tmp/ray` volume when present, set
    `ImagePullPolicy: IfNotPresent`, add the postStart raylet-node-id hook.
  - Wire injection into `mapRayCluster()` when `enabled: true`.
  - Compute `log_url` in `MapLocalClusterStatusToGlobal()` — the mapper
    is the single source of truth (it has the config, the local cluster
    name, and knows the compute-cluster Ray namespace).
    Format: `{LogUrlBaseUrl}/{Bucket}/{PathPrefix}/{cluster}_{rayLocalNamespace}/`
- **Controller** (`go/components/ray/cluster`): `applyRayClusterStatus()`
  copies `clusterStatus.Ray.LogUrl` through to `RayCluster.Status.LogUrl`.
  Controller intentionally owns no log-persistence config or formula.
- **Defaults** (`go/cmd/controllermgr/config/base.yaml`): sandbox MinIO
  defaults under `jobs.k8sengine.mapper.logPersistence`. Production
  overlays flip these.

**Why?**
The collector sidecar streams Ray cluster logs into MinIO/S3 so the
KubeRay History Server can replay the Ray Dashboard for clusters that
are already torn down. `log_url` lets uniflow / Pipeline CR expose a
direct browse link.

**How did you test it?**
Local testing in the OSS k3d sandbox (next PR in the stack brings up
History Server + MinIO + sandbox wiring). Unit tests cover sidecar
injection (env vars, lifecycle preservation, config knobs) and the
controller `log_url` pass-through.

**Potential risks**
- Requires `kuberay-collector` image to be published to GHCR before
  rollout beyond the sandbox.
- `enabled: false` in config is the safe rollback — no sidecar gets
  injected and no `log_url` is written.

**Documentation Changes**
Follow-up PR: expose `log_url` through uniflow onto the Pipeline CR.


go/components/jobs/client/k8sengine/ray.go
```diff
@@ -94,6 +115,124 @@ func getWorkerGroupSpecs(clusterName string, workers []*v2pb.RayWorkerSpec) []ra
...
+	}
+
+	eventExportEnvVars := []corev1.EnvVar{
+		{
+			Name:  "RAY_enable_core_worker_ray_event_to_aggregator",
+			Value: "1",
+		},
+		{
+			Name:  "RAY_DASHBOARD_AGGREGATOR_AGENT_EVENTS_EXPORT_ADDR",
+			Value: fmt.Sprintf("http://localhost:%d/events", collectorPort),
```
badcount: Can we put the localhost into a configurable prefix?

badcount (approved):

go/components/jobs/client/k8sengine/ray.go
dragod812: I think we don't need to make it configurable, this is inside the sidecar and the Ray container and the collector share the network, so localhost is the only correct address. I can extract the string into a constant though.

go/components/jobs/client/k8sengine/mapper.go
```diff
@@ -22,10 +26,23 @@ type MapperResult struct {
...
 const _mapperName = "k8sengineMapper"
 
+const logPersistenceConfigKey = "jobs.k8sengine.mapper.logPersistence"
+
+// NewLogPersistenceConfig loads LogPersistenceConfig from YAML config provider.
+func NewLogPersistenceConfig(provider config.Provider) (LogPersistenceConfig, error) {
+	conf := LogPersistenceConfig{}
+	err := provider.Get(logPersistenceConfigKey).Populate(&conf)
+	if err != nil {
+		// Config is optional — return zero-value (disabled) if not present
```
dragod812, line 38: fyi @badcount config is optional due to this.

go/components/jobs/common/types/mapper.go
```diff
@@ -0,0 +1,32 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	v2pb "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
+)
+
+// Mapper abstracts mapping global job objects (Michelangelo representations)
+// to their local (Kubernetes) counterparts and back. Implementations live
+// alongside specific compute backends (e.g. k8sengine for the in-cluster
+// KubeRay path); consumers depend on this interface, never on a concrete impl.
+type Mapper interface {
```
dragod812, line 14: @badcount moved the interface out of k8sengine mapper.

go/components/jobs/client/client_test.go
```diff
@@ -188,7 +188,7 @@ func TestDeleteJob(t *testing.T) {
 			k8sc := Client{
 				factory: f,
 				helper:  NewHelper(),
-				mapper:  k8sengine.NewMapper().Mapper,
+				mapper:  k8sengine.NewMapper(k8sengine.LogPersistenceConfig{}).Mapper,
```
dragod812: Move this to a mock for mapper interface.

badcount (approved):

PR #1126, opened 2026-04-29
Introduce controlled expanded state to form-group

konukselim:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Summary
Updates `FormGroup` component to accept two new props:

- `expanded?: boolean` — To allow consumers of the component to control expanded state of the group inside their own logic
- `onToggle?: (expanded: boolean) => void` — To notify consumers of the component about the changes of the expanded state. This prop is especially needed for consumers controlling the expanded state from their own logic

## Test plan
Two new test cases which tests the expected behaviors of the newly defined props

### DIFF
--- a/javascript/packages/core/components/form/layout/form-group/types.ts
+++ b/javascript/packages/core/components/form/layout/form-group/types.ts
@@ -1,7 +1,7 @@
 import type { ReactNode } from 'react';
 import type { BoxOverrides } from '#core/components/box/types';
 
-export interface FormGroupProps {
+interface FormGroupBaseProps {
   title?: string;
 
   /** Text displayed below title, **Markdown supported** */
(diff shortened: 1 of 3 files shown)


javascript/packages/core/components/form/layout/form-group/types.ts
```diff
@@ -17,6 +17,12 @@ export interface FormGroupProps {
    */
   collapsible?: boolean;
 
+  /** Controlled expanded state — requires `onToggle` to respond to user interaction */
+  expanded?: boolean;
+
+  /** Called when the user toggles the collapsible group */
+  onToggle?: (expanded: boolean) => void;
+
```
craigmarker: Thinking out loud ... `expanded` and `onToggle` are only relevant when `collapsible=true`. Does it make sense to have the `FormGroupProps` type enforce this?

craigmarker (commented):
LGTM, left one non-blockin

craigmarker (approved):

PR #1129, opened 2026-04-30
Integration Kueue into compute cluster scheduler for ray job

badcount:
What type of PR is this? (check all applicable)
  - Refactor
  x Feature
  - Bug Fix
  - Optimization
  - Documentation Update
  
  What changed?

  Added Kueue integration for Ray job scheduling on compute clusters. Changes are scoped to Kueue only — no unrelated code was modified.

  - go/components/jobs/client/k8sengine/ray.go: propagate labels from the Michelangelo RayCluster proto to the k8s RayCluster object so the kueue.x-k8s.io/queue-name label
  reaches the actual workload
  - python/michelangelo/uniflow/plugins/ray/task.star: read KUEUE_QUEUE_NAME env var (default "") and add kueue.x-k8s.io/queue-name label to every RayCluster spec when set —
  empty default makes Kueue opt-in
  - helm/michelangelo/values.yaml: add worker.kueue.queueName: ""
  - helm/michelangelo/templates/core/worker-deployment.yaml: inject KUEUE_QUEUE_NAME env var into the worker container when worker.kueue.queueName is non-empty
  - python/michelangelo/cli/sandbox/sandbox.py: add --install-kueue flag to sandbox create/sync; wire worker.kueue.queueName=user-queue into _build_helm_set_args; add ma 
  sandbox demo job and ma sandbox demo job kueue subcommands; add _install_kueue / _install_kueue_on_cluster using Helm with integrations.frameworks={ray.io/v1/RayCluster}
  - python/michelangelo/cli/sandbox/demo/job/kueue/compute-cluster-{0,1}.yaml: Kueue ClusterQueue + LocalQueue config (v1beta2) for each compute cluster

  Why?

  Kueue provides quota enforcement and fair scheduling for Ray workloads on compute clusters. Without it, jobs run immediately regardless of available resources. With Kueue,
  jobs are queued and admitted only when quota permits, enabling multi-tenant resource management on shared clusters.

  The design uses local Kueue on each compute cluster independently — no MultiKueue cross-cluster fan-out — keeping the setup simple for the POC.

  How did you test it?

  Tested end-to-end locally with two k3d compute clusters:

  1. ma sandbox demo job — created michelangelo-compute-0 and michelangelo-compute-1 with KubeRay
  2. ma sandbox demo job kueue — installed Kueue v0.9.1 on both clusters with ray.io/raycluster integration enabled
  3. Submitted a RayCluster with kueue.x-k8s.io/queue-name=user-queue label to michelangelo-compute-0
  4. Verified Kueue created a Workload, admitted it against the ClusterQueue, set spec.suspend=false, and head + worker pods reached 1/1 Running
  5. Confirmed ClusterQueue shows admittedWorkloads: 1 and quota is tracked correctly

  Potential risks

  - KUEUE_QUEUE_NAME defaults to "" so existing deployments are unaffected — no label is added to RayClusters unless explicitly configured
  - --install-kueue installs Kueue on the sandbox cluster via helm upgrade --install; idempotent if Kueue is already present
  - manageJobsWithoutQueueName=false means Kueue only manages labelled workloads — unlabelled RayClusters are unaffected

  Release notes

  No schema updates or data migration required. New opt-in feature only.
  
  To enable: ma sandbox demo job kueue then ma sandbox sync --install-kueue, or via helm --set worker.kueue.queueName=user-queue.

  Documentation Changes

  No user-facing API changes. Backward compatible — Kueue is disabled by default at every layer.

### DIFF
--- a/go/components/jobs/client/k8sengine/ray.go
+++ b/go/components/jobs/client/k8sengine/ray.go
@@ -2,14 +2,36 @@ package k8sengine
 
 import (
 	"fmt"
+	"strconv"
 
 	v2pb "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
 	rayv1 "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	k8sptr "k8s.io/utils/ptr"
 )
+
+// LogPersistenceConfig holds platform-level configuration for log persistence.
+// Loaded from YAML under jobs.k8sengine.mapper.logPersistence.
+// See: https://github.com/ray-project/kuberay/tree/master/historyserver/config
+type LogPersistenceConfig struct {
+	Enabled           bool   `yaml:"enabled"`
+	StorageEndpoint   string `yaml:"storageEndpoint"`   // S3-compatible endpoint (e.g. "minio:9091")
+	Bucket            string `yaml:"bucket"`            // S3 bucket name (e.g. "ray-history")
+	PathPrefix        string `yaml:"pathPrefix"`        // Key prefix under the bucket (e.g. "log")
+	Region            string `yaml:"region"`            // S3 region — required by AWS SDK for SigV4 signing even with custom endpoints (e.g. "us-east-1", "us-ashburn-1")
+	CredentialsSecret string `yaml:"credentialsSecret"` // K8s Secret with AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY
+	CollectorImage    string `yaml:"collectorImage"`    // KubeRay collector sidecar image
+	S3DisableSSL      bool   `yaml:"s3DisableSSL"`      // Set S3DISABLE_SSL on the collector (true for in-cluster MinIO; false for OCI/S3)
+
+	// LogURLFormat is a Go text/template applied during local→global cluster
+	// status translation to produce the human-browsable log URL surfaced on
+	// v2 RayClusterStatus. Available template variables: Bucket, PathPrefix,
+	// ClusterName, RayLocalNamespace. Empty string disables log_url emission.
+	LogURLFormat string `yaml:"logURLFormat"`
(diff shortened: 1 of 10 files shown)


go/components/jobs/client/k8sengine/ray.go
```diff
@@ -90,6 +90,7 @@ func (m Mapper) mapRayCluster(rayCluster *v2pb.RayCluster) (runtime.Object, erro
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      rayCluster.Name,
 			Namespace: RayLocalNamespace,
+			Labels:    rayCluster.GetLabels(),
```
dragod812, line 93: We can't pass all the labels, for example our michelangelo/cluster-affinity labels also exist under raycluster labels the kuberay raycluster doesn't care about that. We should have a mapLabels function to pick the necessary labels to pass to Compute clusters.

helm/michelangelo/values.yaml
```diff
@@ -315,6 +315,14 @@ worker:
 
   resources: {}
 
+  # Kueue integration. When queueName is non-empty, the worker sets
+  # KUEUE_QUEUE_NAME on every RayCluster it creates so Kueue can
+  # intercept and queue the workload on the compute cluster.
+  # Set via --set worker.kueue.queueName=user-queue or install Kueue
+  # with `ma sandbox sync --install-kueue`.
+  kueue:
+    queueName: ""
```
dragod812, line 324: What queue is this ? Cluster Queue or Local Queue ? 
Each project in michelangelo should have a local queue. 
Each compute cluster should have a cluster queue.

apurvapatkeshwar:
Is this still moving? I was considering drafting an rfc proposing a general kueue backed job queue behind the scheduler extension point, plus gang admission for ray/spark by mapping them to kueue workloads with pod-set counts. This PR is the closest existing artifact and I'd much rather build on it than duplicate it

As for the 2 open review comments - I'd go with the mapLabels allowlist direction, and with the ClusterQueue per compute cluster / LocalQueue per project topology @dragod812 described, with the helm value naming made explicit about which level it configures

@badcount would you want to revive this once fork CI (#1609) is sorted? Happy to help with the rebase/review. If you'd rather fold it into a broader scheduler backend design, I'll cite this PR as prior art in the RFC. Either way, better not to have two parallel kueue implementations in flight

PR #1133, opened 2026-04-30
 KubeRay History Server setup in the local sandbox

dragod812:
**What type of PR is this? (check all applicable)**                                                                                                                                                                            
- [ ] Refactor                                                                                                                                                                                                                 
- [x] Feature                                                                                                                                                                                                                  
- [ ] Bug Fix                                                                                                                                                                                                                  
- [ ] Optimization                                                                                                                                                                                                             
- [ ] Documentation Update                                                                                                                                                                                                     
                                                                                                                                                                                                                               
**What changed?**                                                                                                                                                                                                              
                                                                                                                                                                                                                               
Manifests (`python/michelangelo/cli/sandbox/resources/`):                                                                                                                                                                      
- Add `history-server.yaml` — Deployment + Service (NodePort 30016) +
  ServiceAccount + ClusterRole/Binding for the KubeRay History Server,                                                                                                                                                         
  using `kuberay-historyserver:v0.1.0` with env-var S3 config (no ConfigMap).                                                                                                                                                  
- Add `jobs.k8sengine.mapper.logPersistence` block to both                                                                                                                                                                     
  `michelangelo-controllermgr.yaml` and the temporal flavor, pointing the                                                                                                                                                      
  collector sidecar at in-cluster MinIO via the k3d agent NodePort, with                                                                                                                                                       
  `collectorImage: kuberay-collector:v0.1.0`.             
  the collector sidecar pattern.                                                                                                                                                                                               
                                                          
Sandbox wiring (`python/michelangelo/cli/sandbox/sandbox.py`):                                                                                                                                                                 
- Add `3001:30016` port mapping (next to Grafana on 3000 in the
  observability range; avoids the 8080 collision with Temporal Web UI).                                                                                                                                                        
- Append `history-server.yaml` to the core resource list and add a links                                                                                                                                                       
  entry so `ma sandbox create` prints `http://localhost:3001`.                                                                                                                                                                 
- Add `ray-history` to the MinIO bucket auto-creation list.                                                                                                                                                                    
- Remove the now-dead fluent-bit branch.                                                                                                                                                                                       
                                                                                                                                                                                                                               
Image build + publish:                                    
- `scripts/kuberay/build-kuberay-images.sh` — clones kuberay, patches the                                                                                                                                                      
  upstream Dockerfile to honor `TARGETARCH`, builds both images for the  
  host arch, imports them into the sandbox + compute clusters.                                                                                                                                                                 
- `.github/workflows/build-kuberay-images.yaml` — multi-arch  
[...]

PR #1134, opened 2026-04-30
Wire History Server into sandbox setup script

dragod812:
- [ ] Refactor
- [X] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Add `"8080:30016"` to `_kube_ports` so the History Server's NodePort
  (defined in PR #3) is reachable on the host at `http://localhost:8080`
- Append `history-server.yaml` to the core sandbox resource list and add
  a `links` entry so `ma sandbox create` prints the History Server URL
  alongside MinIO, Grafana, etc.
- Add `ray-history` to the MinIO bucket auto-creation list so the
  collector sidecar and History Server have a bucket on first boot
- Remove the entire `if "fluent-bit" in ns.include_experimental:` block
  (ServiceAccount provisioning + appending `fluent-bit.yaml` /
  `fluent-bit-config.yaml`) — those manifests no longer exist after PR #3

**Why?**
Wire up `sandbox.py` to actually deploy and surface the History Server
manifests landed in PR #3. Without this PR, the YAMLs exist on disk but
`ma sandbox create` never applies them, never exposes the port, never
creates the bucket, and never prints the URL — so the feature isn't
usable end to end.

**How did you test it?**
- `ma sandbox delete && ma sandbox create`: confirmed `history-server`
  pod becomes `1/1 Running`, `http://localhost:8080/readz` returns 200,
  the `ray-history` bucket appears in the MinIO console, and the
  History Server URL is printed in the creation summary

**Potential risks**
- Anyone scripting `--include-experimental fluent-bit` against an
  internal fork sees their flag silently no-op. Mitigated because the
  collector sidecar (PR #1) provides equivalent log shipping via the
  same MinIO bucket.

**Release notes**
- New host-port mapping `8080 → NodePort 30016` in `ma sandbox create`
- New `ray-history` bucket auto-created during sandbox setup

**Documentation Changes**
NA — sandbox usage flow is unchanged for end users; the new History
Server URL is self-documenting via the printed links list.


dragod812:
Not needed, covered in #1133

PR #1140, opened 2026-05-04
feat(inferenceserver): make services in target clusters discoverable within control plane cluster

aritra0227:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
For each InferenceServer, the controller now publishes a single Kubernetes Service in the control plane that fans traffic across every cluster the Inference Server is serving in. The Service has no pod selector. Instead, the controller maintains one EndpointSlice per ClusterTarget, each pointing at that cluster's ingress address.

New interfaces in `endpoints/`:
1. `EndpointSource`: resolves the ingress address of one cluster. Abstracts the per-environment differences in how a cluster's ingress is discovered (e.g., on GKE could read the gateway LoadBalancer's external IP).
**Default impl (source/k3d.go)**: reads the gateway Service's NodePort and a Node InternalIP via the per-cluster ClientFactory.

2. `EndpointPublisher`: owns the per-Inference-Server service-discovery surface in the control plane. Abstracts where that surface lives so it isn't pinned to native k8s objects (e.g., a future implementation could register each cluster's ingress as a backend in a cloud load balancer like an AWS ALB or GCP).
**Default impl:** writes a ClusterIP Service + per-cluster EndpointSlices into the control plane via controller-runtime, owner-referenced back to the Inference Server so kube GC handles teardown.

Summary of changes:
1. `base/config` gains InferenceServerConfig so the source knows where to find the gateway service
2. sandbox controllermgr configs + `base.yaml` adds the matching `inferenceServer.gateway` block
3. `rbac-inferenceserver.yaml` grants `endpointslices` verbs
4. Unit tests

**Why?**
In order for a inference-server in some remote-cluster to be in a `serving` state, it must be discoverable. At this time, we want to ensure that the target inference-servers can all be discovered from the control plane cluster. By doing this, we are ensuring that we can have an integration later on where the inference-server controller _**registers**_ the target inference-servers is some common load-balancer service. 
With the changes in this PR, we ensure that a deployment route (i.e. HTTPRoute) can discover the inference servers so that inference requests are fanned out to the target clusters.  

**How did you test it?**
1. Unit tests
2. Tested changes on sandbox

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**
Will be in a separate PR when the entire implementation stack has been merged.


go/components/inferenceserver/endpoints/source/k3d.go
```diff
@@ -0,0 +1,108 @@
+// Package source provides EndpointSource implementations for specific
```
badcount: rename to kubernetes

badcount: it should be called provider

badcount (approved):

go/components/inferenceserver/endpoints/source/k3d.go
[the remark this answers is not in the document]
aritra0227: done.

PR #1142, opened 2026-05-04
feat(inferenceserver): inference-server controller owns and provisions routing artifacts 

aritra0227:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
This PR consolidates inference routing under a new shared interface and shifts ownership of the routing resources from the Deployment controller to the Inference Server controller.

1. **New interface `routing.Manager`** (`go/components/common/routing/interface.go`). It owns the lifecycle of managing a networking route.
2. **Default implementation lives in `gatewayapi/manager.go`:** writes to Gateway API HTTPRoutes via dynamic.Interface, serialized per route by an in-process keyed mutex, with retry.RetryOnConflict wrapping rule update calls. **All business logic to orchestrate the above routes live in the deployment/serving controllers.** Route Interface has been removed from Deployment Controller.
4. **Each Inference Server now has two routes**. A **discovery route** in the control plane fans inbound requests across the hosting clusters. A **traffic route** in each target cluster forwards requests to the local inference Service. Both routes carry a default rule at Inference Server creation time, so the Inference Server is reachable at `/{inferenceServerName}` immediately, before any Deployment exists.
5. Updated InferenceServer Actors to use above interface to provision the httproutes. Updated Deployment actors to use above interface to upsert/remove deployment specific routes.
6. Unit tests.

**Why?**
InferenceServers should be discoverable without any deployments. This PR makes that possible. Earlier, we required a deployment to a target inference server for us to be able to make calls against the inference-server.
Additionally, this PR also enables multicluster discovery. This is required so that inference requests can be fanned out to all the target clusters.

**How did you test it?**
1. Deployed and tested changes against sandbox
2. Unit tests

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**
Will be included in upcoming PR once all implementation diffs are merged.


go/components/inferenceserver/routes/interface.go
```diff
@@ -0,0 +1,79 @@
...
+	// exposes {deploymentName}'s {modelName}.
+	UpsertDiscoveryRule(ctx context.Context, dynamicClient dynamic.Interface, inferenceServerName string, namespace string, deploymentName string, modelName string) error
+
+	// RemoveDiscoveryRule removes the rule for {deploymentName} from the
+	// discovery route. Tolerates a missing rule.
+	RemoveDiscoveryRule(ctx context.Context, dynamicClient dynamic.Interface, inferenceServerName string, namespace string, deploymentName string) error
+
+	// DeploymentDiscoveryRouteExists reports whether the discovery routing for
+	// {deploymentName} is configured.
+	DeploymentDiscoveryRouteExists(ctx context.Context, dynamicClient dynamic.Interface, inferenceServerName string, namespace string, deploymentName string) (bool, error)
```
badcount: why we need this?

aritra0227: This function specifically checks whether a deployment route (i.e. url for inference against that particular deployment) exists within the httproute. This function specifically enables discovery.
You can check [here](https://github.com/michelangelo-ai/michelangelo/pull/1142/changes/BASE..8d5f73c48d00917404371658a0290dbedf12283c#diff-653999b6c1487ae660971606517c8d01d1b5d9d7c574a24e0288ce32efe2229aR37) for how its used.


Similarly, the `DeploymentTrafficRouteExists` enables the actual traffic routing to individual clusters to the appropriate model. As discussed before this one is a second httproute specifically for traffic routing.

They are both required to ensure that deployment controller can properly set up an endpoint so that users can run inference against the deployed model.

go/components/inferenceserver/routes/interface.go
```diff
@@ -0,0 +1,79 @@
+//go:generate mamockgen RouteProvider
```
badcount: I suggest the route interface should be generic like in the gateway or client folder. which is used to talk to a route service

aritra0227: done

go/components/common/routing/gatewayapi/manager.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+		if setErr := unstructured.SetNestedSlice(obj.Object, desired, "spec", "rules"); setErr != nil {
+			return setErr
+		}
+		_, err = client.Resource(httprouteGVR).Namespace(namespace).Update(ctx, obj, metav1.UpdateOptions{})
+		return err
+	})
+}
+
+func (m *httpRouteManager) RuleExists(ctx context.Context, client dynamic.Interface, name, namespace string, rule routing.Rule) (bool, error) {
```
badcount, line 117: can we change to routeExists rather than "Rule" here?

go/components/common/routing/gatewayapi/manager.go
```diff
@@ -0,0 +1,331 @@
...
+			Name:               ref.Name,
+			UID:                ref.UID,
+			Controller:         &t,
+			BlockOwnerDeletion: &t,
+		}})
+	}
+	return route
+}
+
+func buildHTTPRule(rule routing.Rule) map[string]interface{} {
```
badcount, line 177: can we use httpRoute?

go/components/common/routing/gatewayapi/manager.go
```diff
@@ -0,0 +1,331 @@
...
+
+func (m *httpRouteManager) Delete(ctx context.Context, client dynamic.Interface, name, namespace string) error {
+	err := client.Resource(httprouteGVR).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+func (m *httpRouteManager) AddRules(ctx context.Context, client dynamic.Interface, name, namespace string, rules ...routing.Rule) error {
```
badcount, line 63: Can we use "Route" instead of Rule in the routemanager?

go/components/common/routing/interface.go
```diff
@@ -0,0 +1,69 @@
...
+type OwnerRef struct {
+	APIVersion string
+	Kind       string
+	Name       string
+	UID        types.UID
+}
+
+// Rule is a routing rule: match a path, optionally rewrite it, forward to a
+// named backend Service. BackendPort defaults to 80 when zero.
+type Rule struct {
```
badcount, line 36: Why we use Rule instead of Route?

badcount (approved):

go/components/common/routing/interface.go
aritra0227: resolved offline, updated comments

PR #1145, opened 2026-05-05
feat(revision): version-agnostic revision.Manager

craigmarker:
- [x] Refactor
- [x] Feature

**What changed?**

The `revision.Manager` interface and its no-op backing implementation have been replaced with a version-agnostic interface and a real implementation. The new interface includes a single function for revision upserting.

It follows the controller-runtime caller-owns-type pattern: callers build a fully-populated Revision in whatever API version they use and pass it as a `client.Object`.

The deployment controller's no-op `revision.Manager` field is removed. It existed as a placeholder but was never called.

**Why?**

The previous interface was scoped to deployments and backed by a no-op that silently did nothing. Making it version-agnostic unblocks any controller from producing revisions without the Manager needing to know about each resource type. The single-method design keeps the interface minimal: the state machine in `UpsertRevision` is the only behavior that isn't a thin wrapper over the API handler.

**How did you test it?**

Validated the end-to-end revision flow against a [sandbox cluster with wiring connected to pipeline controller](https://github.com/michelangelo-ai/michelangelo/pull/1145/changes/077b816f858ad1c36269fa75b2f85d496d60d6c0) (removed so it can land behind a config flag): applied a Pipeline CR with a commit field, and confirmed a Revision CR was created with the expected spec fields.

**Potential risks**

None in this PR — the Manager is defined and tested but not wired into any reconcile loop.


go/components/pipeline/controller.go
```diff
@@ -118,19 +110,6 @@ func (r *Reconciler) updatePipelineStatus(ctx context.Context, pipeline *v2pb.Pi
 	return result, nil
 }
 
-// formatRevisionName generates a standardized revision name for a pipeline.
-//
-// The name format is: "pipeline-{lowercase-pipeline-name}-{git-ref-prefix}"
-// where git-ref-prefix is the first 12 characters (or less) of the git reference.
-//
-// For example: "pipeline-my-model-a1b2c3d4e5f6"
-func formatRevisionName(pipeline *v2pb.Pipeline) string {
```
badcount, line 141: Why we're removing this?

craigmarker: I'm happy to add it back. Technically, revision isn't hooked up to pipeline controller at all today. So there is no "latest revision" of a pipeline. Just the pipeline CR itself.

I'll add this functionality back when we support configuring revisioning per CRD #1268

craigmarker: what do you think?

badcount (approved):

aritra0227 (approved):

PR #1153, opened 2026-05-06
fix: wire MetadataStorage into apiserver so enableMetadataStorage works

badcount:
## Summary

- `apiserver/main.go` registered `MetadataStorageConfig` (the flag) but never provided a `storage.MetadataStorage` implementation into the fx graph
- `Params.MetadataStorage` is `optional:"true"` so fx silently injected `nil`, making the `EnableMetadataStorage` check in `NewAPIServerHandler` always false
- `ListOptionsExt` was effectively unreachable in the apiserver regardless of config — reported by dev as "enableMetadataStorage doesn't work"

## Changes

- `go/cmd/apiserver/providers.go` (new) — `provideMetadataStorage` + `mysqlConfigEnabled`, mirrors `controllermgr/ingester_providers.go`
- `go/cmd/apiserver/main.go` — register `GetMySQLConfig` and `provideMetadataStorage` into the fx graph
- `go/cmd/apiserver/BUILD.bazel` — add `providers.go` to srcs, add `//go/storage` and `//go/storage/mysql` deps


go/storage/mysql/mysql.go
```diff
@@ -279,6 +300,316 @@ func (m *mysqlMetadataStorage) List(ctx context.Context, typeMeta *metav1.TypeMe
...
+// fieldName must already be the bare column name (CRD prefix stripped).
+//
+// Output format matches the internal storage/pkg/mysql/lineage_util.go exactly:
+//   - All fragments start with " " (a leading space) so they can be concatenated
+//     with " AND" / " OR" suffixes via the suffix-trim pattern in buildCriterionSQL.
+//   - IS NULL / IS NOT NULL fragments end with a trailing space (legacy from the
+//     internal map values "IS NULL "/"IS NOT NULL ").
+//   - IN / NOT IN list has no spaces between placeholders (e.g. "(?,?,?)").
+func convertCriterionOperator(fieldName string, op apipb.CriterionOperator, value string) (string, []interface{}, error) {
+	qf := " `" + fieldName + "` "
```
ishtoo1, line 565: Can we validate fieldName against an allowlist (or ^[a-zA-Z_][a-zA-Z0-9_]*$) inside processFieldName? Without it, things like a backtick in the request field name breaks out of the identifier here and lets a caller alter the WHERE clause.

go/storage/mysql/mysql.go
```diff
@@ -279,6 +300,316 @@ func (m *mysqlMetadataStorage) List(ctx context.Context, typeMeta *metav1.TypeMe
...
+				colName = col
+			} else {
+				colName = remainder
+			}
+		}
+		dir := "ASC"
+		if order.Dir == apipb.SORT_ORDER_DESC {
+			dir = "DESC"
+		}
+		clauses = append(clauses, fmt.Sprintf("`%s` %s", colName, dir))
```
ishtoo1: Can we run through the same allowlist here too and return an error for unknown fields? Right now any string survives the fallback at line 585 and gets interpolated raw into the ORDER BY identifier.

go/storage/mysql/mysql_test.go
```diff
@@ -46,3 +48,314 @@ func TestGetTableName_GVKEmpty_UnknownToScheme(t *testing.T) {
...
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, c.wantSQL, sql)
+			require.Equal(t, c.wantParams, params)
+		})
+	}
+}
+
+func TestBuildLabelCriterionSQL(t *testing.T) {
```
ishtoo1, line 135: Can we fold these into a table-driven test with a setupFixtures(t) helper and move the repeated "pipelinerun.*" / "RUNNING" / "prod" strings, and any other hardcoded constants into named constants?

go/storage/mysql/mysql.go
badcount: Thanks for suggestion, add to-do and will do it in following PR

badcount:
Integration test passed: https://github.com/michelangelo-ai/michelangelo/actions/runs/25685411795/job/75407735837

go/cmd/controllermgr/config/base.yaml
```diff
@@ -17,7 +17,7 @@ k8s:
   qps: 300
   burst: 600
 metadataStorage:
-  enableMetadataStorage: false
+  enableMetadataStorage: true
```
ishtoo1: we're flipping the default for enableMetadataStorage from false to true. if mysql isn't configured, the apiserver will crashloop. is this acceptable?

go/cmd/apiserver/config/base.yaml
```diff
@@ -15,4 +15,4 @@ k8s:
   qps: 300
   burst: 600
 metadataStorage:
-  enableMetadataStorage: false
+  enableMetadataStorage: true
```
ishtoo1: same as the below comment

ishtoo1 (approved):
approving. looks good to land after PR description is updated, and the default metadata storage flag is changed to 'false'. we can address the 2 TODOs as follow-up PRs.

go/cmd/controllermgr/config/base.yaml
ying-zheng: I think it's better to set the default to be false

where is mysql configuration?  In uber internal, there is a separate configuration for mysql. For OSS, I think  we have to put the mysql configuration in API server's and controller manager's configuration files.

scripts/ingester/ingester_schema.sql
ying-zheng, line 2: What are the differences between those 2 sql files?
compelete_ingester_schema.sql and ingester_schema.sql

badcount: i don't know, i think they are for different purpose. i will clean it up in separate PR

PR #1192, opened 2026-05-14
Fix navigation when on detail page

donovanlopez:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Previously, when navigating to a detail page (such as the Pipeline run detail page), when clicking "Back" on the browser, the page stays on the detail page.

This is because if we land on a detail page with no tab specified, we would append the first tab to the URL path using React Router's `navigate`. However this adds to the history stack so if we try to click "Back", it would go to the detail page path but we would again add the first tab to the URL which would result in a loop. The history stack would look something like this:

1. `/ma-dev-test/train/runs/`
2. `/ma-dev-test/train/runs/run-123`
3. `/ma-dev-test/train/runs/run-123/<first-tab>`

Now we will add `{ replace: true }` instead of a regular push which will instead remove the `/ma-dev-test/train/runs/run-123` entry with `/ma-dev-test/train/runs/run-123/<first-tab>`. This prevents the tab-less URL from being added to browser history so pressing back from the detail page.

**Why?**
I expect to be able to use the Back and Forward buttons as expected.

## Before
https://github.com/user-attachments/assets/8afcd9ea-303d-4cdc-abe9-b5319915b1e0

## After

https://github.com/user-attachments/assets/e943935e-c8a0-43a4-b3c3-04050b445fbe

**How did you test it?**
With unit tests and manually tested

**Potential risks**
This could mess with navigation of the pages even more but it's already messed up.

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/javascript/packages/core/router/__tests__/entity-detail-route.test.tsx
+++ b/javascript/packages/core/router/__tests__/entity-detail-route.test.tsx
@@ -951,7 +951,7 @@ describe('EntityDetailRoute', () => {
       ).toBeInTheDocument();
     });
 
-    test('pressing forward after back returns to the detail page, not to the tab-less URL', async () => {
+    test('pressing forward after back returns to the detail page', async () => {
       const user = userEvent.setup();
       const mockRequest = vi.fn().mockResolvedValue({
         pipelineRun: {
--- a/javascript/packages/core/router/__tests__/entity-detail-route.test.tsx
+++ b/javascript/packages/core/router/__tests__/entity-detail-route.test.tsx
@@ -837,7 +837,7 @@ describe('EntityDetailRoute', () => {
     expect(await screen.findByRole('button', { name: 'Resume' })).toBeInTheDocument();
   });
 
-  describe('back navigation', () => {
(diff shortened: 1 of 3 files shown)


javascript/packages/core/router/__tests__/entity-detail-route.test.tsx
```diff
@@ -835,4 +836,193 @@ describe('EntityDetailRoute', () => {
...
+                  } as CustomDetailPageConfig,
+                ],
+              },
+            ],
+          }),
+        ],
+      }),
+    };
+
+    function TestApp({ phases }: { phases: typeof detailPhases }) {
```
craigmarker: My instinct is to use the router wrapper instead of a custom function like this—to match the rest of the codebase.

If the router wrapper doesn't work for this use case, is it possible to update it?

What do you think?

craigmarker (approved):
Change LGTM -- one question about test setup.

PR #1197, opened 2026-05-15
Fix/helm dependency build

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix/helm dependency build

**Why?**
For integration test

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


helm/michelangelo/values.yaml
```diff
@@ -342,7 +342,7 @@ controllermgr:
 
   # Metadata storage knobs (ingester). Defaults match the source sandbox YAML.
   metadataStorage:
-    enable: true
```
sallycr, line 345: We should enable metadatastorage by default. isn't it?

badcount: from discussion right now, it should be false, we have other changes need to be out first

helm/michelangelo/templates/core/controllermgr-deployment.yaml
```diff
@@ -13,6 +13,8 @@ spec:
       {{- include "michelangelo.componentSelectorLabels" (dict "context" . "component" "controllermgr") | nindent 6 }}
   template:
     metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/core/controllermgr-configmap.yaml") . | sha256sum }}
```
badcount, line 17: When a ConfigMap changes (e.g., controllermgr-configmap.yaml gets updated), Kubernetes does not automatically restart the pods that consume it. The
  Deployment spec itself hasn't changed, so Kubernetes sees no reason to roll it. By computing a SHA256 checksum of the rendered ConfigMap and embedding it as a pod annotation, any change to the ConfigMap changes the annotation value → the pod template spec changes → Kubernetes triggers a rolling update of the controllermgr pods automatically.
This is a standard Helm pattern (checksum/config) used to force pod restarts when dependent ConfigMaps change. Without it, config changes would silently have no effect until pods were manually restarted.

sallycr (approved):

PR #1199, opened 2026-05-15
refactor: restructure ActionConfig as action + modal union

craigmarker:
- [x] Refactor

**What changed?**

`ActionConfig` is restructured into a union that separates what happens (action) from how it's confirmed (modal). One arm handles config-driven mutations and route navigations, with an optional confirm dialog. The other handles component-driven actions where the form owns its submit flow.

Dispatch is now handled by `ActionDispatcher`, which routes to the right sub-dispatcher based on config shape. Existing component-driven callers migrate to the new `CustomModalConfig` arm. No behavioral difference for  custom modals.

**Why?**

The previous `ActionConfig` only supported component-driven actions. The upcoming configuration-driven layer needs to declare mutations as configuration so that the framework can dispatch them without the consumer writing a component. Splitting confirm dialog logic into its own composable arm means any action type can optionally require a confirm step without duplicating that logic.

**How did you test it?**

New integration-style tests for mutation & route config
Ran locally and configured existing custom modals render correctly

**Potential risks**

This repo is the only consumer of these components. All call sites are updated in this PR, so the breaking changes to `ActionConfig` and `ActionComponentProps` are fully contained here.


javascript/packages/core/components/actions/action-dispatcher.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function ActionDispatcher<T extends Data>({ action, record, onClose }: Props<T>) {
+  if (action.modal?.type === 'custom') {
+    const Component = action.modal.component;
+    return <Component record={record} onClose={onClose} />;
+  }
+  if (action.modal?.type === 'confirm' && action.action) {
+    return (
+      <ConfirmDispatcher
+        action={{ ...action, action: action.action, modal: action.modal }}
```
donovanlopez: Why do we have to do `action: action.action, modal: action.modal` even though we're unpacking `...action`?

javascript/packages/core/components/actions/types.ts
```diff
@@ -1,19 +1,27 @@
...
+ * - **Action + optional confirm modal** — dispatches a mutation or route, optionally
+ *   gated by a confirm dialog.
+ * - **Custom modal** — opens a custom React component (a form, a wizard, etc.).
+ *   Cannot pair with `action`; the component owns its own submit flow.
  */
+export type ActionConfig<T = Data> = ActionConfigBase &
+  (
+    | { action: MutationActionConfig | RouteActionConfig; modal?: ConfirmModalConfig }
+    | { modal: CustomModalConfig<T>; action?: never }
+  );
```
donovanlopez, line 24: nit: Seeing `action.action` seems confusing. Perhaps we can name it `actionConfig` so it'd be `action.actionConfig`? Would `action` ever be something else other than a config?

javascript/packages/core/components/actions/types.ts
```diff
@@ -39,50 +47,89 @@ export type ActionConfigBase = {
...
+/**
+ * Item shape consumed by action renderers (see `useResolvedActionItems`).
+ * Only display + `onClick` — no action-type knowledge. The `onClick`
+ * delegates to whatever `onSelect` the renderer provided when the item
+ * was resolved.
+ */
 export type ResolvedActionItem = {
   display: ActionTriggerDisplay;
   hierarchy?: ActionHierarchy;
+  /** True if any of the source action's disabled rules matched. */
```
donovanlopez: nit: I don't think we need a comment explaining what `disabled` does. Plus it seems to be prescribing how it works.

javascript/packages/core/components/actions/types.ts
```diff
@@ -39,50 +47,89 @@ export type ActionConfigBase = {
...
+  PRIMARY = 'primary',
+  SECONDARY = 'secondary',
+  TERTIARY = 'tertiary',
+}
+
+/**
+ * Item shape consumed by action renderers (see `useResolvedActionItems`).
+ * Only display + `onClick` — no action-type knowledge. The `onClick`
+ * delegates to whatever `onSelect` the renderer provided when the item
+ * was resolved.
```
donovanlopez: nit: This comment seems specific to `useResolvedActionItems`. It might be better to make it more generic if this will be used elsewhere.

donovanlopez (approved):
LGTM, mainly left non-blocking comments

javascript/packages/core/components/actions/types.ts
craigmarker: yeah, that is confusing ... updated to be `operation`

javascript/packages/core/components/actions/types.ts
craigmarker: removed 👍🏻

javascript/packages/core/components/actions/types.ts
craigmarker: updated, thanks!

javascript/packages/core/components/actions/action-dispatcher.tsx
craigmarker: yeah, there is some typescript trickiness here ... action.modal?.type === 'confirm' doesn't refine such that that `action` is guaranteed to have a modal; it only refines `action.modal`.

Updated to use type guard with comment

PR #1201, opened 2026-05-15
feat: successOperations for mutation actions

craigmarker:
- [x] Feature

**What changed?**

Mutation actions can now declare `successOperations` that run after a successful submission.

This PR adds two explicit post-success operations:
- `invalidate`, which can target queries broadly by name or narrowly by `name + serviceOptions`
- `toast`, which can show a success toast with an optional action

`ConfirmDispatcher` now runs these operations after mutation success. `SnackbarProvider` was added to `CoreApp`, plus a test wrapper for components/hooks that enqueue snackbars.

This PR does **not** add automatic cache invalidation in `useStudioMutation`. The retry pipeline run cell keeps its existing explicit scoped invalidation of the current `GetPipelineRun` query.

For tests, `use-success-operations` now has dedicated integration-style coverage via a small harness fixture, and the service-provider test wrapper exposes provider test handles so tests can access the underlying `QueryClient` without capture components. Some duplicated toast-behavior assertions were removed from `actions-popover` now that the hook owns that coverage directly.

**Why?**

The main goal is to make post-success behavior declarative for mutation actions.

That gives config authors a way to express user-visible success behavior like toasts, and explicit cache refresh behavior when needed, without pushing broad invalidation rules into `useStudioMutation`.

**How did you test it?**

- Integration-style coverage in `actions-popover.test.tsx` for mutation wiring into success toasts and failure behavior
- Dedicated hook coverage in `use-success-operations.test.tsx` for explicit invalidation behavior, toast rendering, navigation, dismissal, interpolation, icons, and mixed invalidate+toast runs
- Wrapper coverage in `get-service-provider-wrapper.tests.ts`
- `yarn test:core`

**Potential risks**

Success-side effects are explicit. Any mutation flow that needs a refresh must declare it via `successOperations.invalidate` or keep its existing manual invalidation.

**Release notes**

Mutation actions now support explicit `successOperations` for post-success invalidation and toast behavior. `SnackbarProvider` was added to `CoreApp`.

**Documentation Changes**

Inline type/docs updates for `successOperations` and invalidation target behavior.


javascript/packages/core/hooks/use-studio-mutation.ts
```diff
@@ -29,5 +30,21 @@ export const useStudioMutation = <TData, TVariables extends Record<string, unkno
     onError: config?.clientOptions?.onError
       ? (error) => config.clientOptions!.onError!(error)
       : undefined,
+    // The API server names handlers as {Verb}{Kind} (Kubernetes convention);
+    // strip the verb to derive the kind and invalidate Get{Kind}+List{Kind}
+    // on every settle so reads stay consistent without per-call declarations.
+    onSettled: () => {
+      const entity = k8sEntityFromMutationName(config?.mutationName ?? '');
+      if (!entity) return;
+      void queryClient.invalidateQueries({ queryKey: [`Get${entity}`] });
```
donovanlopez: I think we need to pass in the id of the entity or else it would invalidate all ``Get${entity}`` queries

javascript/packages/core/components/actions/__tests__/use-success-operations.test.tsx
donovanlopez, line 1: We should add tests to cover the toast operation as well.

javascript/packages/core/components/actions/use-success-operations.tsx
```diff
@@ -0,0 +1,66 @@
...
+  );
+}
+
+function buildToastPayload(
+  op: ToastOperation,
+  navigate: ReturnType<typeof useNavigate>,
+  dequeue: () => void
+) {
+  const StartEnhancer = ({ size }: { size: number }) => (
+    <Icon name={op.icon ?? DEFAULT_TOAST_ICON} size={`${size}px`} />
```
donovanlopez, line 51: Can we move this `StateEnhancer` outside of the function? Since it's in a function, every time `buildToastPayload` is called, the `StartEnhancer` will remount since a new `StartEnhancer` component is created within the function.

craigmarker: hmmm... startEnhancer has to be a component that receives size, and the icon comes from the toast operation, so some inline wrapper is needed somewhere. In the current code that wrapper is created once when the toast is enqueued, not on every re-render of the same toast, so it shouldn’t cause repeated remounts in practice.

If we wanted to move it out, the alternative would be something like a hoisted default enhancer plus a small wrapper/factory for custom icons, which doesn't feel justified yet.

What am I missing?

javascript/packages/core/components/actions/__tests__/use-success-operations.test.tsx
craigmarker: For toast behavior, I kept the main coverage in actions-popover since that exercises the user-visible outcome end to end: mutation succeeds, toast renders, route actions navigate, non-route actions dismiss, and failure does not show the toast. 

We could also test useSuccessOperations with a real SnackbarProvider and assert the rendered toast behavior directly. I didn’t go that route here because actions-popover already gives us that end-to-end coverage, so adding the same behavior again at the hook level felt redundant.

What do you think?

donovanlopez: Hmmm yeah it's best to reduce any unnecessary tests and also for something like toasts, testing what the user would actually see is more valuable. The only reason to add toast coverage here is for completeness and documenting what this is supposed to do through the tests.

javascript/packages/core/components/actions/use-success-operations.tsx
donovanlopez: Ah you're right that makes sense.

donovanlopez (approved):

javascript/packages/core/components/actions/__tests__/use-success-operations.test.tsx
```diff
@@ -0,0 +1,303 @@
...
+        <UseSuccessOperationsTestHarness
+          operations={[
+            {
+              type: 'invalidate',
+              targets: [
+                { name: 'GetPipelineRun', serviceOptions: { name: 'run-1', namespace: 'ns' } },
+              ],
+            },
+            { type: 'toast', message: 'Pipeline updated' },
+          ]}
```
donovanlopez, line 143: more of a curious question than an actual feedback but what would happen if we pass in more than one toast?

donovanlopez (approved):

javascript/packages/core/components/actions/__tests__/use-success-operations.test.tsx
craigmarker: not sure lol. I can test and then we can add some guardrails if that ends up being an issue

PR #1205, opened 2026-05-18
fix: honor --exclude in sandbox k3d port forwards

craigmarker:
- [x] Bug Fix

**What changed?**

`ma sandbox create --exclude {grafana|prometheus}` now also drops the corresponding host port from the k3d cluster's port-forward list, so the cluster spec matches the services that are actually being deployed.

**Why?**

`--exclude grafana` and `--exclude prometheus` skipped deploying those services but left their host port mappings in the k3d cluster spec. The cluster still tried to bind ports 3000 and 9092 on the host at start time, and if anything else on the developer's machine was already listening on either port, `sandbox create` would fail with a bind error during cluster startup.

The port mapping and the deploy step are two halves of the same decision; they should be driven from the same source of truth.

**How did you test it?**

With port 3000 held by local dev server, ran the patched `ma sandbox create --exclude grafana`. k3d cluster came up cleanly — port 3000 was no longer in the `-p` list passed to `k3d cluster create`, so no bind attempt, no conflict. The create proceeded all the way through helm install michelangelo and ended with `🚀 Sandbox created successfully`.

**Potential risks**

Local-developer-only change — the sandbox CLI isn't part of any deployed artifact. Mapping is keyed on the exact `host:nodeport` string, so any future change to those port strings in `_infra_ports` needs a matching update to `_infra_port_owner`.

**Release notes**

N/A — local developer tooling.

**Documentation Changes**

N/A.

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -45,6 +45,14 @@ _infra_ports = [
     "5001:30013",  # MLflow Tracking Server
 ]
 
+# Infra ports owned by optionally-excluded services. When the user passes
+# --exclude {svc}, the corresponding host port is dropped from k3d's port
+# forwards so it doesn't conflict with other processes on the host.
+_infra_port_owner = {
+    "3000:30012": "grafana",
+    "9092:30015": "prometheus",
+}
+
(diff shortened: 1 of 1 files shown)


sallycr (approved):
thx

PR #1208, opened 2026-05-19
fix(pipelinerun): TTLDays=0 should disable eviction, not trigger immediately

badcount:
## Summary

- `TTLDays=0` (the default, unset) was being treated as "expire immediately" because `expireAt = lastUpdate + 0` is already in the past
- This caused every terminal PipelineRun to be marked `michelangelo/Immutable=true` on the same reconcile it became terminal — the opposite of the documented behavior
- Added an explicit early-return guard when `TTLDays == 0`

## Test plan
- [ ] `go test ./go/components/pipelinerun/...` passes
- [ ] Integration test CLI e2e passes (this bug was causing FAILED pipeline runs to be evicted from ETCD immediately, interfering with test assertions)


badcount:
Closing — not confirmed to be related to the integration test hang.

PR #1223, opened 2026-05-21
Add missing tls config for temporal

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

 What changed?

  - Fixed newTemporalWorker function to accept and apply TLS configuration via a new tlsConfig *tls.Config parameter
  - Updated provide function to pass in.TLSConfig to newTemporalWorker (this was missing, causing TLS config to be ignored for Temporal workers)
  - Added TLS connection options to Temporal client when conf.UseTLS is enabled
  - Added comprehensive unit tests covering TLS scenarios:
    - TestNewTemporalWorker_TLS: Tests worker creation with TLS enabled/disabled, custom/nil TLS config, and multiple workers
    - TestProvide_TemporalWithTLS: Tests the integration through the provide function with and without TLS config

  Why?

  The Temporal worker was not respecting TLS configuration even though:
  - The In struct had a TLSConfig field for user-provided custom TLS config
  - The Config struct had a UseTLS flag 
  - The Cadence worker implementation already supported TLS configuration
  
  The bug was in the provide function at line 102, which called newTemporalWorker without passing the tlsConfig parameter. This meant Temporal workers could not connect to
  TLS-enabled Temporal servers, breaking parity with the Cadence implementation.

  How did you test it?

  - Added unit tests that verify TLS configuration is properly applied in various scenarios
  - Ran all tests in the module: bazel test //go/worker/workflowfx:go_default_test - all tests pass
  - Tested with TLS enabled/disabled, custom TLS config, and nil TLS config (defaults to &tls.Config{})

  Potential risks

  - Low risk: This is a bug fix that makes Temporal workers work correctly with TLS configuration
  - The change is backward compatible - if UseTLS is false, behavior is unchanged
  - If UseTLS is true and TLSConfig is nil, a default &tls.Config{} is used (same as Cadence implementation)
  - Existing deployments without TLS are unaffected

  Release notes

  Bug fix: Temporal workers now properly respect TLS configuration when useTLS: true is set in the workflow-engine config.

  Documentation Changes

  No documentation changes required. The configuration format remains the same - this fix makes the existing useTLS configuration option work correctly for Temporal workers.

### DIFF
--- a/go/worker/workflowfx/module_test.go
+++ b/go/worker/workflowfx/module_test.go
@@ -25,9 +25,11 @@ func (m *MockTemporalClient) Close() {
 
 type MockTemporalClientFactory struct {
 	mock.Mock
+	CapturedOpts *tempclient.Options
 }
 
 func (m *MockTemporalClientFactory) NewTemporalClient(opts tempclient.Options) (tempclient.Client, error) {
+	m.CapturedOpts = &opts
 	return tempclient.NewLazyClient(opts)
 }
 
@@ -314,6 +316,9 @@ func TestNewTemporalWorker_TLS(t *testing.T) {
 
 		assert.NoError(t, err)
 		assert.Len(t, workers, 1)
+		assert.NotNil(t, mockFactory.CapturedOpts)
+		assert.NotNil(t, mockFactory.CapturedOpts.ConnectionOptions.TLS)
+		assert.Equal(t, customTLSConfig, mockFactory.CapturedOpts.ConnectionOptions.TLS)
 	})
 
 	t.Run("creates temporal worker with TLS enabled and nil config", func(t *testing.T) {
@@ -333,6 +338,9 @@ func TestNewTemporalWorker_TLS(t *testing.T) {
 
 		assert.NoError(t, err)
 		assert.Len(t, workers, 1)
+		assert.NotNil(t, mockFactory.CapturedOpts)
+		assert.NotNil(t, mockFactory.CapturedOpts.ConnectionOptions.TLS)
+		// When nil TLS config is passed with UseTLS=true, a default &tls.Config{} should be used
 	})
 
 	t.Run("creates temporal worker with TLS disabled", func(t *testing.T) {
@@ -355,6 +363,9 @@ func TestNewTemporalWorker_TLS(t *testing.T) {
 
 		assert.NoError(t, err)
 		assert.Len(t, workers, 1)
+		assert.NotNil(t, mockFactory.CapturedOpts)
+		assert.Nil(t, mockFactory.CapturedOpts.ConnectionOptions.TLS)
+		// When UseTLS=false, TLS config should not be set even if provided
 	})
 
 	t.Run("creates temporal worker with multiple workers", func(t *testing.T) {
@@ -381,6 +392,9 @@ func TestNewTemporalWorker_TLS(t *testing.T) {
 
 		assert.NoError(t, err)
 		assert.Len(t, workers, 3)
+		assert.NotNil(t, mockFactory.CapturedOpts)
+		assert.NotNil(t, mockFactory.CapturedOpts.ConnectionOptions.TLS)
+		assert.Equal(t, customTLSConfig, mockFactory.CapturedOpts.ConnectionOptions.TLS)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+		logger := zap.NewNop()
+		mockFactory := &MockTemporalClientFactory{}
+		customTLSConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+
+		workers, err := newTemporalWorker(mockFactory, conf, customTLSConfig, logger)
+
+		assert.NoError(t, err)
+		assert.Len(t, workers, 1)
```
aritra0227, line 318: over here, if possible, I think we should verify that the TLS config is actually getting populated. 
Something like:
```
assert.Equal(t, workers.TLSConfig, customTLSConfig) 
```

go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+			Workers: []WorkerConfig{{TaskList: "test-tasklist"}},
+		}
+
+		logger := zap.NewNop()
+		mockFactory := &MockTemporalClientFactory{}
+
+		workers, err := newTemporalWorker(mockFactory, conf, nil, logger)
+
+		assert.NoError(t, err)
+		assert.Len(t, workers, 1)
```
aritra0227, line 340: same here

go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+		logger := zap.NewNop()
+		mockFactory := &MockTemporalClientFactory{}
+		customTLSConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+
+		workers, err := newTemporalWorker(mockFactory, conf, customTLSConfig, logger)
+
+		assert.NoError(t, err)
+		assert.Len(t, workers, 1)
```
aritra0227, line 365: over here, maybe we should check that the TLS config is **not populated** for the workers.

go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+		logger := zap.NewNop()
+		mockFactory := &MockTemporalClientFactory{}
+		customTLSConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+
+		workers, err := newTemporalWorker(mockFactory, conf, customTLSConfig, logger)
+
+		assert.NoError(t, err)
+		assert.Len(t, workers, 3)
```
aritra0227, line 394: same here, we should check that the tls config for each worker is populated

go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+			Logger:          logger,
+			TemporalFactory: mockTemporalFactory,
+			TLSConfig:       customTLSConfig,
+		}
+
+		out, err := provide(in)
+
+		assert.NoError(t, err)
+		assert.Equal(t, service.BackendType("temporal"), out.Backend)
+		assert.Len(t, out.Workers, 1)
```
aritra0227, line 430: same here, should check whether the tls config is populated.

go/worker/workflowfx/module_test.go
```diff
@@ -292,3 +292,155 @@ func TestNewCadenceClient_TLSFlow(t *testing.T) {
...
+			Logger:          logger,
+			TemporalFactory: mockTemporalFactory,
+			TLSConfig:       nil,
+		}
+
+		out, err := provide(in)
+
+		assert.NoError(t, err)
+		assert.Equal(t, service.BackendType("temporal"), out.Backend)
+		assert.Len(t, out.Workers, 1)
```
aritra0227, line 458: over here, should check that the tls config is **not populated**

aritra0227 (approved):
Some concerns regarding the tests. In particular, we are not verifying whether the TLS config actually gets populated.
Otherwise, stamping since the implementation looks fine to me.

PR #1224, opened 2026-05-21
feat: [CanvasFlex] Pusher Phase 1 PR3 — DatasetVariable + DataSink + PandasIO

sallycr:
## Summary

Implements the dataset pusher plugin as a complete, extensible unit. Ships together so provider layers (Uber's \`UberHiveSink\`, community \`S3Sink\`) have the full extension contract before any other plugin is built.

**Closes #1217**

#### \`workflow/variables/_private/dataset.py\` — \`DatasetVariable(Variable)\`
- Subclasses \`Variable\` from \`_private/base.py\` — **matches internal repo pattern exactly**
- Single constructor: \`DatasetVariable(value=..., path=...)\` or \`DatasetVariable.create(value)\`
- \`backend\` property: isinstance dispatch → \`"pandas"\` | \`"spark"\` | \`"ray"\` | \`"unknown"\`
- \`save()\` dispatches on runtime type → \`save_pandas/spark/ray_*()\`
- Each save/load method delegates to \`Variable._save/_load_value_using_io(IO)\` — inherits \`_saved\` guard, \`_io_metadata\` round-trip, and logging for free
- \`_load()\` runtime detection: active SparkSession → Ray initialized → pandas fallback
- All pyspark/ray/pandas imports lazy (inside try/except ImportError)
- path auto-generated from \`UF_STORAGE_URL\` env var (default \`memory://storage\`)

#### \`workflow/schema/data_sink.py\` — \`DataSink\` ABC + built-in sinks
- \`DataSink.write(variable: DatasetVariable) -> SinkResult\` — each sink accesses \`variable.value\` in its native format
- \`LocalFileSink\` — pandas only; single CSV/Parquet/JSON file; raises \`TypeError\` for Spark/Ray
- \`InMemorySink\` — zero I/O; \`.records\` list for test assertions
- \`HiveSink\` — Spark only; \`artifact.value\` natively (no \`toPandas()\` collect); count() before saveAsTable(); mode validated in \`__init__\`
- \`SinkResult(uri, num_records, extra)\` — frozen dataclass; \`extra\` dict spread into plugin result

#### \`workflow/schema/pusher.py\` — \`DatasetPluginConfig\` updated
- \`sinks: list[DataSink] | None = None\` sentinel — \`None\` = not provided; \`[]\` = explicitly empty; never overridden after construction
- \`destination_path\` shorthand auto-creates \`LocalFileSink\` when \`sinks is None\`

[...]

PR #1233, opened 2026-05-22
feat: [CanvasFlex] Pusher Phase 1 PR3 — DatasetVariable + DataSink + PandasIO

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Implements the CanvasFlex Pusher Phase 1 dataset plugin as a complete, extensible unit. Supersedes #1224 (clean rebase onto main).

**`DatasetVariable`** (`workflow/variables/_private/dataset.py`):
- Subclasses `Variable` from `_private/base.py` — matches internal repo pattern exactly
- Single constructor: `DatasetVariable(value=..., path=...)` or `DatasetVariable.create(value)`
- `backend` property: isinstance dispatch → `"pandas"` | `"spark"` | `"ray"` | `"unknown"`
- `save()` / `_load()` dispatch to `_save/_load_value_using_io` (inherited from `Variable`)
- `_load()` runtime detection: active SparkSession → Ray initialized → pandas fallback
- All pyspark/ray/pandas imports lazy; `ValueError` when saving with no value set
- `__repr__` shows path and backend; path auto-generated from `UF_STORAGE_URL` env var

**`DataSink` ABC + built-in sinks** (`workflow/schema/data_sink.py`):
- `LocalFileSink` — pandas only; single CSV/Parquet/JSON file
- `InMemorySink` — zero-I/O; `.records` list for test assertions
- `HiveSink` — Spark only; accesses `variable.value` natively (no `toPandas()`); count() before saveAsTable(); mode validated in `__init__`
- `SinkResult(uri, num_records, extra)` — frozen dataclass

**`DatasetPluginConfig`** (`workflow/schema/pusher.py`):
- `sinks: list[DataSink] | None = None` sentinel — `None` = not provided; `[]` = explicitly empty
- `destination_path` shorthand auto-creates `LocalFileSink`

**`DatasetPusherPlugin`** (`workflow/tasks/pusher/plugins/dataset_plugin.py`):
- Dispatches via `sink.write(variable)` — no format logic at plugin level
- Returns `{"sinks": [...], "num_records": N, "destination_path": "..."}`
- `ConfigurationError` on `artifact=None` or empty sinks

**`PandasIO`** (`uniflow/plugins/pandas/io.py`) — new:
- OS `uniflow/plugins/` had `SparkIO` + `RayDatasetIO` but was missing pandas equivalent
- Writes directory of `part-*.parquet` files via PyArrow + fsspec; zstd compression
- Lazy pyarrow/fsspec imports (circular import fix); added missing `uniflow/plugins/__init__.py`

**Why?**

Phase 1 of the CanvasFlex migration — ports the dataset pusher plugin from the internal Michelangelo monorepo to the open source package. `DatasetVariable` matches the internal `DatasetVariable(Variable)` pattern exactly so provider layers (Uber) can subclass with minimal code.

**How did you test it?**

```bash
cd python && source .venv/bin/activate
pytest michelangelo/workflow/ michelangelo/uniflow/ -v
# 153 passed
ruff check michelangelo/workflow/ michelangelo/uniflow/plugins/pandas/
# All checks passed
```

**Potential risks**

None — all new files, no changes to existing functionality.

**Release notes**

N/A

**Documentation Changes**

Docstrings on all new classes and methods.


python/michelangelo/workflow/schema/data_sink.py
```diff
@@ -0,0 +1,108 @@
...
+        >>> result.num_records
+        3
+    """
+
+    uri: str
+    num_records: int
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+class DataSink(ABC):
```
kenns29: this abstract class is also an implementation detail and should be moved under workflow/sinks

python/michelangelo/workflow/tasks/functions/sinks/__init__.py
```diff
@@ -0,0 +1,25 @@
+"""Built-in DataSink implementations for DatasetPusherPlugin.
```
kenns29, line 1: we can make the path  `workflow/common/sinks`

We will have a lot of these shared components for tasks it's better to group them under one folder `common`

python/michelangelo/workflow/schema/sinks/hive.py
```diff
@@ -0,0 +1,40 @@
+"""Config dataclass for HiveSink."""
```
sallycr, line 1: @kenns29 note:   
  - schema/sink.py (single file): simpler, easier to read at a glance, all sink contracts in one place. Makes sense now since there are only 4 small dataclasses + 1 result
  type.
  - schema/sinks/ (folder): each config in its own file, easier to extend — providers add schema/sinks/terrablob.py without touching the core file.

  Given that providers will add their own sink configs (e.g. TerrablobSinkConfig), the folder approach is more extensible.

kenns29 (approved):

PR #1162, opened 2026-05-08
feat: Helm integration in sandbox, Cadence & Temporal subcharts, port sync

sallycr:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Phase 4 of the Helm chart integration for the Michelangelo sandbox:

- **Control plane via Helm**: Replaced 7 raw `kubectl apply` resource YAMLs with a single `helm install/upgrade michelangelo` call using `values-k3d.yaml`
- **Cadence as Helm subchart**: Migrated from a bare Pod to a subchart bundled inside the michelangelo release; schema init and domain registration happen automatically
- **Temporal as Helm subchart**: Migrated from a separate `helm install` to a subchart; enabled/disabled dynamically via `--set temporal.enabled=true/false`
- **`ma sandbox sync/delete`**: Now use `helm upgrade --reuse-values` / `helm uninstall`
- **Port management**: Single source of truth via `_helm_chart_ports()` reading NodePorts from `values-k3d.yaml` dynamically
- **ClusterRole**: Fixed `resources: ["*"]` wildcard for `michelangelo.api` group to cover dynamically-registered CRDs
- **logPersistence config**: Wires the KubeRay collector sidecar config (from PR #1124) into the Helm chart under `controllermgr.jobs.k8sengine.mapper.logPersistence` (mirrors the Go config path). The collector image is managed in the top-level `images.kuberayCollector` block alongside other images (`ghcr.io/michelangelo-ai/kuberay-collector:main`). `enabled` is a proper pass-through value (was previously hardcoded `true` in the template). Enabled by default in both `values.yaml` and `values-k3d.yaml`
- **Bug fixes from live sandbox validation**:
  - Excluded `history-server` from `kubectl wait` (requires custom-built image not on Docker Hub)
  - Fixed secret name `minio-credentials` → `object-storage-credentials` in `history-server.yaml`
  - Fixed MinIO endpoint `minio:9000` → `minio:9091` in `values-k3d.yaml`
  - Moved `_create_cadence_domain()` after `helm install`
  - Fixed Cadence subchart disabled in Temporal CI step (both-enabled guard was triggering)
  - Fixed Temporal Helm repo URL and removed obsolete top-level persistence keys

**Why?**

[...]

PR #1234, opened 2026-05-22
fix(sandbox): enable Cadence strongly-consistent queries in helm sandbox

craigmarker:
## What

The michelangelo controller hardcodes `QueryConsistencyLevelStrong` when querying
the `task_progress` workflow handler on every reconcile loop. Cadence requires
this to be explicitly enabled at the cluster level.

The helm migration in #1162 switched the sandbox from `ubercadence/server:auto-setup`
(which bundled a `development.yaml` with this flag on by default) to the Cadence
helm subchart (which ships only a minimal dynamicconfig). Every `task_progress`
query started failing with:

```
BadRequestError: cluster or domain does not enable strongly consistent
query but strongly consistent query was requested
```

Since `constructPipelineRunStepInfo` is the only path that populates substep
state and advances the workflow-level state machine, pipeline runs in the helm
sandbox no longer made it past `Running` — even after the underlying Cadence
workflow completed.

### Before (on `craig.marker/kafka-helm-bitnami-legacy-fix`)

I tried to run a pipeline; it never reached a terminal state. `controllermgr`
logs showed the BadRequestError on every reconcile:

```bash
kubectl logs deployment/michelangelo-controllermgr | grep "strongly consistent query"
```

### After (on this branch)

I recreated the sandbox on this branch:

```bash
ma sandbox delete && ma sandbox create --exclude grafana
```

Then ran a pipeline to completion and verified:

1. No `strongly consistent query` errors:
   ```bash
   kubectl logs deployment/michelangelo-controllermgr | grep "strongly consistent query"
   # (no output)
   ```

2. Step `activityId` fields are populated:
   ```bash
   kubectl get pipelinerun -n ma-dev-test <run-name> \
     -o jsonpath='{.status.steps[*].subSteps[*].activityId}'
   # 1 10
   ```

### DIFF
--- a/helm/michelangelo/values-k3d.yaml
+++ b/helm/michelangelo/values-k3d.yaml
@@ -131,6 +131,40 @@ cadence:
       type: NodePort
       nodePort: 30004
 
+  # Dynamic config overrides.
+  #
+  # The upstream Cadence helm subchart ships a minimal dynamicconfig that only
+  # sets retention + visibility store. It does NOT enable strongly-consistent
+  # queries, which Cadence's default is "off".
+  #
+  # The michelangelo controller hardcodes QueryConsistencyLevelStrong when
+  # querying the `task_progress` workflow handler on every reconcile loop
+  # (cadenceclient.go). Without this flag, every query fails with:
+  #   BadRequestError: cluster or domain does not enable strongly consistent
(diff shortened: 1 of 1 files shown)


sallycr (approved):
thx for fixing the sandbox problem.

PR #1237, opened 2026-05-23
feat(pusher): pluggable EvalReportSink interface + gRPC/LocalFile built-ins

### LINKED ISSUE #1218: feat: [CanvasFlex Pusher PR4] EvalReportPusherPlugin
## PR4 — EvalReportPusherPlugin

Standalone evaluation report plugin. No storage backend or registry client needed — the simplest plugin, ideal for independent review.

### Scope
- `pusher/plugins/eval_report_plugin.py` — `EvalReportPusherPlugin`
- 8 tests

### Key design
- Artifact is plain `dict` — no protobuf, no Spark
- `num_keys` counted before `extra_fields` merge
- Writes to `tempfile.mkdtemp(prefix='michelangelo_reports_')`

**Depends on:** #1217 (PR3)
**Part of:** CanvasFlex Pusher Phase 1

sallycr:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

The Pusher task — the final stage of a Michelangelo training pipeline, responsible for publishing a trained model and its evaluation results — now exposes a pluggable `EvalReportSink` interface so contributors can route evaluation reports to MLflow, Weights & Biases, Comet, S3, or any custom backend without modifying the plugin itself.

**New components (`tasks/functions/eval_report_sinks/`):**

- **`EvalReportSink`** — abstract base class defining the sink contract (`write(report, extra_fields)`). Documents fail-fast sequential dispatch, synchronous call contract, read-only `extra_fields` semantics, and idempotency guidance for retried pipeline runs. Includes a complete `S3EvalReportSink` example for community contributors.
- **`flatten_report_to_metrics(report) -> dict[str, float]`** — converts a nested `EvaluationReport` proto into a flat metric dict for sinks targeting MLflow `log_metrics`, W&B `wandb.log`, Comet, etc. Reduces a custom sink to a handful of lines.
- **`LocalFileEvalReportSink`** — default sink; serializes the `EvaluationReport` proto as JSON to a local directory. Zero configuration; auto-creates a temp directory.
- **`GRPCEvalReportSink`** — pushes reports to any gRPC service implementing the `EvaluationReportService` proto. Stores the channel as an instance attribute (`self._channel`) with `close()` and context-manager support (`with GRPCEvalReportSink(cfg) as sink`). Catches `grpc.RpcError` specifically, preserving status code and details in the wrapped `OSError`.

**`EvalReportPusherPlugin` refactored:**

- Now accepts proto-only artifacts (`EvaluationReport`); plain-dict path removed.
- Dispatches to an ordered list of sinks via `EvalReportPluginConfig.sinks`.
- Resolves `metadata.name` (config override → proto field → auto-generated UUID).
- Passes a defensive copy of `extra_fields` to each sink so multi-sink loops cannot interfere with each other.
- Top-level `output_path` / `namespace` in the result reflect the **first** sink; full per-sink results are in `result["sinks"]`.

**`EvalReportPluginConfig` sentinel semantics:**

- `sinks=None` (default) — auto-creates a `LocalFileEvalReportSink` at execute time (construction deferred to `execute()`, not `__post_init__`, keeping the schema layer free of implementation imports).
- `sinks=[]` — explicit dry-run; the plugin runs but writes nothing.
- `sinks=[sink1, sink2, ...]` — dispatches in order, fail-fast.

**Other changes:**

- `schema/eval_report_sinks/__init__.py` re-exports sink config classes consistent with `schema/sinks/`.
- `LocalFileEvalReportSink` writes with explicit `utf-8` encoding.
- `python/HOW_TO_WRITE_A_TASK.md` — new contributor guide for writing workflow task plugins, using the sink pattern as a worked example. Internal project naming stripped.

Closes #1218

**Why?**

The Pusher task's evaluation-report path previously hard-coded local JSON writes inside the plugin with no extension seam — community contributors couldn't add MLflow, W&B, cloud-storage, or gRPC destinations without forking. The proto-only refactor (which dropped the legacy dict path) made this an even bigger gap, since users now have a richly-typed `EvaluationReport` they want to ship to existing experiment-tracking systems.

`EvalReportSink` is the strategy-pattern interface that closes this gap, modeled on Great Expectations' `Action` and OpenTelemetry's `SpanExporter`: a single method to implement, an ordered list to configure, and a built-in helper (`flatten_report_to_metrics`) to bridge the proto schema to flat-metric platforms.

**How did you test it?**

- 37 unit tests covering the sink ABC, both built-in sinks, the plugin's dispatch logic, name resolution, multi-sink dispatch, `extra_fields` mutation guard, and the dry-run path (`sinks=[]`).
- `ruff check` clean on all changed files.

**Potential risks**

- **`APISink` → `GRPCEvalReportSink` rename** (and `APISinkConfig` → `GRPCEvalReportSinkConfig`, `LocalFileEvalSinkConfig` → `LocalFileEvalReportSinkConfig`). Backwards-incompatible for any caller importing these by name. No existing callers in this repo.
- **`EvalReportPluginConfig(sinks=[])` now means explicit dry-run.** Previously `__post_init__` auto-created a `LocalFileEvalReportSink` even for empty lists. Now `[]` is honored as "write nothing" and `None` triggers the default. Callers passing `sinks=[]` expecting writes will produce no output.
- **Plugin no longer accepts plain dicts.** Callers must construct an `EvaluationReport` proto. Migration: `EvaluationReport(spec=EvaluationReportSpec(...))`.
- `EvalReportPluginConfig.__post_init__` no longer eagerly constructs the default sink — deferred to `execute()`. Observable behavior is identical for all documented usage.

**Release notes**

**New extension point: `EvalReportSink`.** Write your own evaluation-report destination by subclassing `EvalReportSink` and passing instances via `EvalReportPluginConfig(sinks=[...])`. Two built-ins ship in the box: `LocalFileEvalReportSink` (zero-config JSON to tempdir) and `GRPCEvalReportSink` (pushes to any `EvaluationReportService` gRPC endpoint;
requires `grpcio`).

**New helper: `flatten_report_to_metrics(report)`.** Converts an `EvaluationReport` proto into a flat `{name: float}` dict — drop-in for `mlflow.log_metrics()`, `wandb.log()`, or `comet.log_metrics()`. Reduces a custom sink to a handful of lines.

**Dispatch semantics:** sinks are called synchronously in order; the first failure aborts the loop (fail-fast). Sinks that create server-side resources should be idempotent — retried pipeline runs call `write()` again with the same `metadata.name`. `extra_fields` is read-only for sink implementations.

> `EvaluationReport` uses Michelangelo's own proto schema
> (`proto/api/v2/evaluation_report.proto`), not OpenMetrics or OTLP.
> Community sinks targeting flat-metric platforms should use
> `flatten_report_to_metrics()` to bridge.

**Design notes**

- `flatten_report_to_metrics()` is intentionally co-located with the `EvalReportSink` ABC in `base.py` so custom-sink authors who import the ABC see the helper immediately. If additional proto-conversion utilities are added later (e.g. proto→OTLP), they should be split into a `utils.py` at that point.

**Documentation Changes**

- New: `python/HOW_TO_WRITE_A_TASK.md` — contributor guide for workflow task plugins
- Updated: `EvalReportSink` ABC docstring — fail-fast contract, sync semantics, idempotency, read-only `extra_fields`, `S3EvalReportSink` example
- Updated: `GRPCEvalReportSink` docstring — context-manager usage, TLS configuration, channel lifecycle
- Updated: `EvalReportPusherPlugin.execute()` docstring — fail-fast semantics, first-sink shortcut fields


python/michelangelo/workflow/tasks/functions/eval_report_sinks/api.py
```diff
@@ -0,0 +1,174 @@
...
+            from michelangelo.gen.api.v2.evaluation_report_svc_pb2_grpc import (
+                EvaluationReportServiceStub,
+            )
+        except ImportError as exc:
+            raise ImportError(
+                "GRPCEvalReportSink requires the 'grpcio' package. "
+                "Install it with: pip install grpcio"
+            ) from exc
+
+        self._channel = (
```
kenns29, line 142: why don't we use APIClient https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/api/v2/client.py#L12

kenns29 (approved):

python/michelangelo/workflow/tasks/functions/eval_report_sinks/api.py
sallycr: APIClient needs to be changed to support the features e.g. secure. will make a separate PR for that. thx

PR #1240, opened 2026-05-27
cli: print dev_run response so it is visible at default log level

yaoyueyue:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
Currently \`ma pipeline dev_run\` only logs the response via \`_LOG.info(...)\`. The default log level (set by \`mactl.py\` to WARNING) suppresses it, so \`ma pipeline dev_run -f <yaml>\` returns exit 0 with **zero output** even on success — looks like a silent no-op.

This adds a one-line \`print(response)\` after the existing log call, matching the behavior of \`mactl pipelinerun\` and \`ma pipeline {apply, create, get, run}\` which all print results regardless of log level.

  ## Test plan
  - [x] All 16 existing \`dev_run_test.py\` cases still pass

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
+++ b/python/michelangelo/cli/mactl/plugins/entity/pipeline/dev_run.py
@@ -221,6 +221,10 @@ def generate_dev_run(
             timeout=30,
         )
         _LOG.info("Stub method completed (%r): %r", type(response), response)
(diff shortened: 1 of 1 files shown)


hwaminkim (approved):
LGTM

PR #1247, opened 2026-05-27
docs: address sandbox setup feedback — timing, sync, missing prereqs, troubleshooting

zhoward-1:
## Summary
Intent:
- Apply 10 pieces of real-user feedback from sandbox testing to reduce the gap between what the doc promises and what first-time users actually experience

Changes:
- Time estimate: updated from "~20 min" to "30–60 min on first run, 5–10 min on subsequent recreates"
- Prerequisites: changed Docker verify from `docker --version` to `docker info` (requires running daemon); narrowed Python to "3.11 or 3.12" with a note about newer versions lacking pre-built ML wheels
- Quick start: added step 2 to build local-only kuberay images via scripts/kuberay/build-kuberay-images.sh before ma sandbox create; added note explaining that ContainerCreating during image pulls is normal
- Verifying success: replaced "all Running in 2–3 minutes" with realistic guidance (5–10 min ContainerCreating is expected; act only on ImagePullBackOff/CrashLoopBackOff)
- Sandbox commands: added ma sandbox sync with description; updated lifecycle diagram to show sync as recovery path from a failed create
- New section: "Recovering from a failed create" explaining when to use sync vs delete+create
- Troubleshooting: added six new entries — command not found: ma, Failed to create cluster already exists, history-server ImagePullBackOff, Progress deadline exceeded, Grafana CrashLoopBackOff from plugin fetch, updated ImagePullBackOff entry to reference local-only image issue

## Test Plan
Doc-only change. Verified all internal cross-links and anchor references are consistent.

## Revert Plan
Revert this PR via `git revert <commit_sha>`.


paulzim:
Thanks for picking this up so quickly! I just confirmed end-to-end on a fresh external install — the sync recovery path, the kuberay-historyserver build step, and the "Progress deadline exceeded" guidance all match what I would have written as new issues. Two small notes:
- The "Progress deadline exceeded" troubleshooting entry might want a line clarifying it's specifically a *first-run* phenomenon — `kubectl get pods -A` will show the operator deployments as `Running` within a few minutes of the helm error.
- Related but separate: I just filed [#1252](https://github.com/michelangelo-ai/michelangelo/issues/1252) about a code bug in _sync(): it skips the compute-cluster setup that _create() does, so after a sync-based recovery ma sandbox demo inference fails at BackendProvision until the compute-cluster Secrets are back-filled. Until that's fixed, it might be worth adding a warning callout in your new "Recovering from a failed create" section pointing at the issue, or a separate troubleshooting entry for BackendProvisioningFailed. I can send you a short troubleshooting bit if you think it should be added.

paulzim (approved):
Changes look good. Thanks, Zach!

PR #1254, opened 2026-05-29
feat(eval-report): EvaluationReportService + APIClientEvalReportSink

sallycr:
- [x] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

- **`APIClientEvalReportSink(svc=None)`** — replaces `GRPCEvalReportSink` as the single network sink. Delegates `write()` to `APIClient.EvaluationReportService`, reusing the shared singleton channel. Accepts an optional `svc` param for DI or custom endpoints (`APIClientEvalReportSink(svc=APIClient(endpoint=...).EvaluationReportService)`). Emits `UserWarning` when `extra_fields` is passed (unsupported). Raises `RuntimeError` with a clear `MA_API_SERVER` hint when the service is not initialized.
- **`GRPCEvalReportSink` and `GRPCEvalReportSinkConfig` removed** — the `svc=` DI pattern via per-instance `APIClient` replaces all use cases, matching the single-client convention of MLflow, W&B, and Comet.
- **`LocalFileEvalReportSink`** unchanged — writes JSON to a local directory with `extra_fields` merge support.
- **`flatten_report_to_metrics`** — skip-path logs at `WARNING`; missing `value` key skips instead of recording `0.0`; `TODO(v2)` markers reference GitHub issue [#1258](https://github.com/michelangelo-ai/michelangelo/issues/1258).
- **`_raise_as_oserror`** helper — centralises gRPC→OSError wrapping; annotated `NoReturn` for correct static analysis.
- **`EvalReportPluginConfig` docstring** — examples updated from `GRPCEvalReportSink` to `APIClientEvalReportSink`.
- **README** — new "Publishing Evaluation Reports" section: concept mapping table, `APIClientEvalReportSink` example, custom-endpoint snippet, `flatten_report_to_metrics` → `mlflow.log_metrics` snippet.
- **Tests** — 32 tests: `APIClientEvalReportSink` (9 incl. DI, None guard, no-warning path, non-RpcError propagation), `LocalFileEvalReportSink` (8), `flatten_report_to_metrics` (7), ABC (1), config (1), others.

**Why?**

- `GRPCEvalReportSink` removal eliminates the dual-mode footgun (namespace mutation, dual-timeout, cross-layer singleton import) flagged in OSS review.
- A single `APIClientEvalReportSink` matches how MLflow (`MlflowClient`), W&B (`wandb.Api`), and Comet (`API()`) work — one client, env-var-driven default, explicit injection for overrides — rather than exposing raw gRPC channel management to callers.

**How did you test it?**

```
cd python
.venv/bin/pytest \
  michelangelo/workflow/tasks/functions/eval_report_sinks/tests/ \
  michelangelo/workflow/schema/eval_report_sinks/tests/ -q
# 32 passed

.venv/bin/pytest michelangelo/api/v2/tests/test_client.py -q
# 22 passed (singleton + instance isolation, unchanged)

ruff check michelangelo/workflow/tasks/functions/eval_report_sinks/
# All checks passed
```

Four rounds of 4-agent OSS review (Architect, Researcher, Dev Advocate, Coder) with converging APPROVE after each round's findings were addressed.

**Potential risks**

- `APIClientEvalReportSink()` raises `RuntimeError` (not `AttributeError`) when `MA_API_SERVER` is not set — callers catching `AttributeError` will need to update exception handling.
- `extra_fields` on `APIClientEvalReportSink` now emits `UserWarning` instead of silently dropping — visible in logs but not an exception.
- `flatten_report_to_metrics` now logs at `WARNING` for skipped multi-series/multi-point charts — existing pipelines with time-series charts will see new warning log lines.
- `GRPCEvalReportSink` is removed — any direct usage must migrate to `APIClientEvalReportSink(svc=APIClient(endpoint=...).EvaluationReportService)`.

**Release notes**

New: `APIClientEvalReportSink(svc=None)`. Removed: `GRPCEvalReportSink`, `GRPCEvalReportSinkConfig`. All new API is additive; removal is a breaking change for direct `GRPCEvalReportSink` users (migration: use `APIClientEvalReportSink(svc=)`). `MA_API_SERVER` is the correct env var name (README previously listed wrong name). Note: `APIClient.EvaluationReportService` landed separately in PR #1263.

**Documentation Changes**

README: new "Publishing Evaluation Reports" section with MLflow/W&B concept mapping table, `APIClientEvalReportSink` example, custom-endpoint pattern, `flatten_report_to_metrics` → MLflow snippet. All public methods documented with Google-style docstrings and copy-pasteable examples. GitHub issue [#1258](https://github.com/michelangelo-ai/michelangelo/issues/1258) tracks `flatten_report_to_metrics` multi-series/step support.


sallycr:
**Design feedback — consider splitting `GRPCEvalReportSink` before merge**

The current `GRPCEvalReportSink` serves two distinct transports via a single `config=None` branch:

| Mode | Transport | Lifecycle | Auth |
|---|---|---|---|
| `config` provided | Own gRPC channel | Sink owns open/close | Configurable |
| `config=None` | `APIClient` shared channel | APIClient owns lifecycle | Inherits APIClient |

This dual-mode design was flagged in the OSS review (plan item 2B). Every major metrics SDK separates these (MLflow has `MlflowClient` vs `mlflow.log_metric`, W&B has `wandb.Run` vs `wandb.Api`). The union type in `write()` and the branching throughout `__init__`/`close()` are symptoms of two classes trying to share one.

**Proposed split:**

```python
# Self-contained — opens/owns its own gRPC channel
class GRPCEvalReportSink(EvalReportSink):
    def __init__(self, config: GRPCEvalReportSinkConfig): ...

# APIClient convenience — delegates to APIClient.EvaluationReportService
class APIClientEvalReportSink(EvalReportSink):
    def __init__(self): ...
```

Both would live in the same module; `GRPCEvalReportSink` is the unchanged existing class, so no breaking change.

**Asking the team:** is this the right call before merging, or should we defer to a follow-up PR and keep the current dual-mode design with a TODO? Happy to do the split here if the team agrees — it's a contained change.

/cc @sallycr

python/michelangelo/api/v2/services/gen/__init__.py
```diff
@@ -28,6 +28,7 @@ def _wire(target, context):
 
 class ServicesGen(object):
     CachedOutputService = None
+    EvaluationReportService = None
```
kenns29, line 31: can we split the API Client update from the pusher update in two separate PRs, API code contains a lot of auto generated code we want to make sure they are easy to revert

kenns29 (approved):

python/michelangelo/api/v2/services/gen/__init__.py
sallycr: ack. separate PR created https://github.com/michelangelo-ai/michelangelo/pull/1263

PR #1265, opened 2026-06-02
fix: Pipeline deletion prereqs — propagation threading, reconcile fixes, SparkJob termination

ishtoo1:
## Summary

Three independent bug fixes that are prerequisites for the Pipeline cascade-delete feature (follow-up PR #1266). Each is safe to land independently.

- **Propagation threading:** The metadata-storage delete path dropped the caller's `propagationPolicy` — the apiserver never recorded it and the ingester issued a bare `Delete` with no options. Thread the policy end-to-end via a new `michelangelo/DeletePropagation` annotation so foreground/cascade deletion works on the metadata-storage path.
- **Pipeline reconcile fixes:** When a Pipeline has `deletionTimestamp` set (foreground deletion), the controller kept reconciling — stamping Status and requeueing every 10s, racing with GC. Short-circuit immediately on `deletionTimestamp`. Also fix `client.IgnoreNotFound` → `utils.IsNotFoundError` for gRPC-aware not-found detection.
- **SparkJob termination:** The SparkJob controller ignored `Spec.Termination` — when a workflow was cancelled, the underlying SparkApplication was never torn down. Gate termination on the shared `jobsutils.IsTerminationInfoSet`, add `handleTermination` that deletes the SparkApplication via a new `DeleteJob` on `SparkClient`, and drive the SparkJob to a terminal state through the shared condition helpers — setting `Killed` and the `Succeeded` condition from the termination type (matching the Ray job controller) — then mark it immutable and stop requeueing.

## Test plan

- [x] `./tools/bazel test //go/components/pipeline/...` — reconcile + not-found tests pass
- [x] `./tools/bazel test //go/components/spark/...` — SparkJob termination tests pass (Killed + Succeeded set per termination type)
- [x] `./tools/bazel test //go/components/jobs/common/utils/...` — termination-spec helper tests pass
- [x] `./tools/bazel test //go/components/ingester/...` — propagation annotation tests pass
- [x] `./tools/bazel build //go/cmd/controllermgr/...` — clean build


go/components/spark/job/controller.go
```diff
@@ -112,6 +112,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	original := sparkJob.DeepCopy()
 
+	// Honor a termination request placed on the SparkJob's spec. The worker's
+	// TerminateSparkJob activity sets Spec.Termination (e.g. when a workflow or
+	// pipeline is torn down). When set, terminate the underlying SparkApplication
+	// and drive the SparkJob to a terminal, immutable state so a cascade delete is
+	// never blocked on an in-flight Spark workload.
```
craigmarker: nit: comment is a little self-aware about its context. The code is relatively simple: if termination is requested, terminate. I get a little confused by the additional comment

go/components/spark/job/controller.go
```diff
@@ -241,3 +250,67 @@ func (r *Reconciler) createJob(ctx context.Context, log logr.Logger, job *v2pb.S
 func (r *Reconciler) getJobStatus(ctx context.Context, logger logr.Logger, job *v2pb.SparkJob) (*string, string, string, error) {
 	return r.sparkClient.GetJobStatus(ctx, logger, job)
 }
+
+// isTerminationRequested reports whether a caller has requested termination of
```
craigmarker: same nit about comment

go/components/spark/job/controller.go
```diff
@@ -241,3 +250,67 @@ func (r *Reconciler) createJob(ctx context.Context, log logr.Logger, job *v2pb.S
...
+
+	message := "Spark job killed"
+	if job.Spec.Termination.Reason != "" {
+		message = job.Spec.Termination.Reason
+	}
+	setCondition(&job.Status.StatusConditions, constants.KilledCondition, apipb.CONDITION_STATUS_TRUE, message, constants.SparkAppKilled)
+
+	// Mark the SparkJob immutable so it cannot transition further once killed.
+	utils.MarkImmutable(job)
+	if err := r.Status().Update(ctx, job); err != nil {
```
craigmarker, line 329: Should this update the entire resource instead of just Status so the annotation gets persisted as well?

ishtoo1: Makes sense. This same pattern exists for RayJob as well. Should we fix that as well then?

go/components/spark/job/controller.go
ishtoo1: yes, i missed carefully reading the generated comments. updated

go/components/jobs/common/utils/utils.go
```diff
@@ -24,7 +24,7 @@ func IsTerminationInfoSet(
 	case *v2pb.RayCluster:
 		return cluster.(*v2pb.RayCluster).Spec.GetTermination().GetType() != v2pb.TERMINATION_TYPE_INVALID, nil
 	case *v2pb.SparkJob:
-		panic("not implemented")
+		return cluster.(*v2pb.SparkJob).Spec.GetTermination().GetType() != v2pb.TERMINATION_TYPE_INVALID, nil
```
dragod812: You don't need to use this right now. 
The spark job controller will change a lot when we support scheduling on spark jobs. 
I would keep a separate function in 
go/components/spark/job/client/client.go for IsTerminationInfoSet . 
But eventually we will replace go/components/spark/job/client/client.go with go/components/jobs/client/client.go
Then we will implement the SparkJob handling under Job components.

go/components/spark/job/client/client.go
```diff
@@ -135,6 +136,31 @@ func (r SparkClient) GetJobStatus(ctx context.Context, logger logr.Logger, job *
...
 }
 
+// DeleteJob terminates a running Spark job by deleting its SparkApplication.
+//
+// Deleting the SparkApplication custom resource instructs the Spark Operator to
+// tear down the driver and executor pods, terminating the underlying workload.
+//
+// Returns an error if the deletion fails. Callers should treat a not-found error
+// as success, since it means the SparkApplication has already been removed.
+func (r SparkClient) DeleteJob(ctx context.Context, log logr.Logger, job *v2pb.SparkJob) error {
```
dragod812, line 146: Just FYI, for now this is okay, but eventually we will delete this file, when Job components support spark jobs.

go/components/spark/job/controller.go
dragod812: Yeah, i think this is a bug in RayJob, we should fix that in RayJob too, you don't need to do that as part of this PR.

go/components/spark/job/controller_test.go
```diff
@@ -30,10 +33,12 @@ const (
 )
 
 type mockSparkClient struct {
```
dragod812: @ishtoo1 can we move to mocks docs/contributing/use-go-mocks-in-unit-test.md

dragod812 (approved):

badcount (approved):

PR #1266, opened 2026-06-02
feat(cascadedelete): owner-ref stamping, drain, and retention for cascade delete

ishtoo1:
The `go/cascadedelete` package — the shared cascade-delete mechanics for a parent's child runs. Consumers that wire this into Pipeline / PipelineRun / TriggerRun are in the stacked PR #1316.

- `EnsureControllerRef` / `StampOwnerRefOnCreate` — stamp the owning parent as the controller ownerReference so Kubernetes foreground GC removes children with it.
- `RunDrainStep` + `DrainTarget` — a generic drain driver that cancels a child's in-flight work before GC removes it (the driver holds no client; each kind adapts via thin `DrainTarget` methods).
- `RetainPolicy` — per-kind opt-in for retaining a child's final state in MySQL; the kind set is injected at the composition root.
- drain accounting, a 24h drain safety timeout, and Prometheus metrics.

Unit-tested in isolation — `//go/cascadedelete:go_default_test` passing. No consumer wiring lives here.

### DIFF
--- a/go/cascadedelete/BUILD.bazel
+++ b/go/cascadedelete/BUILD.bazel
@@ -1,0 +1,47 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "apihook.go",
+        "drain.go",
+        "metrics.go",
+        "ownerref.go",
+        "retain.go",
+    ],
+    importpath = "github.com/michelangelo-ai/michelangelo/go/cascadedelete",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@com_github_prometheus_client_golang//prometheus:go_default_library",
+        "@io_k8s_apimachinery//pkg/apis/meta/v1:go_default_library",
+        "@io_k8s_apimachinery//pkg/runtime:go_default_library",
+        "@io_k8s_sigs_controller_runtime//:go_default_library",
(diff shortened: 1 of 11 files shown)


go/cascade/ownerref.go
```diff
@@ -0,0 +1,45 @@
...
+// If child already has a controller ownerReference whose UID matches the
+// Pipeline's UID, nothing changes and it returns false. Otherwise it appends a
+// controller ownerReference built like the gatewayapi pattern (APIVersion =
+// v2pb.GroupVersion ("michelangelo.api/v2"), Kind = "Pipeline", with both
+// Controller and BlockOwnerDeletion set to true) and returns true.
+//
+// The change is in-memory only; the caller persists it with a metadata Update.
+// Resolving and Get-ing the Pipeline (needed for its UID) is the caller's job —
+// see the per-controller wiring.
+func EnsurePipelineOwnerRef(child client.Object, pipeline *v2pb.Pipeline) (changed bool) {
```
badcount: If this is only used for pipeline, why we don't put this into the pipeline folder?

ishtoo1: The name might be misleading but it's not used by the Pipeline component. Both the PipelineRun and TriggerRun controllers call it to stamp the parent Pipeline as their ownerRef (child=run, pipeline=owner), so it can't live in either folder without coupling them.

badcount (approved):

PR #1268, opened 2026-06-03
feat(pipeline): make revision management configurable

craigmarker:
- [x] Feature

**What changed?**

Wires pipeline revision snapshotting behind a runtime config flag and makes `revision.Manager` optional in the Fx graph.

`Config.RevisioningEnabled` (YAML key `pipeline.revisioningEnabled`, defaults to `false`) gives operators a runtime knob to enable revisioning without recompiling. `LatestRevision` is only written to the pipeline status and the Revision CR is only created when this flag is true.

`revision.Manager` is now optional in Fx, so OSS deployments get a working manager automatically, and callers that need a custom implementation can still inject one via Fx.

The Helm configmap template now surfaces `pipeline.revisioningEnabled` (default `false`) so operators can toggle it without rebuilding the image.

**Why?**

The pipeline controller revision integration was built in the previous PR but held back because any deployer using the full controllermgr image needed an opt-out before it could be turned on by default. This PR adds that opt-out and fixes the wiring so the real manager is used by default rather than a no-op.

**How did you test it?**

I deployed the controllermgr locally to a k3d sandbox and exercised both flag states against real pipelines.

**Enabled (`pipeline.revisioningEnabled: true`)**

I patched the configmap to set `revisioningEnabled: true` and restarted the deployment. After patching `training-pipeline` with a new gitRef:

```
ts=1781287524.281  "Reconciling pipeline starts"   namespace-name=ma-dev-test/training-pipeline
ts=1781287524.287  "created revision"               revision_name=pipeline-training-pipeline-helm-fix-178  namespace=ma-dev-test
```

`kubectl get revisions -n ma-dev-test` returned four Revision CRs, one per pipeline:

```
NAME                                           AGE
pipeline-eval-pipeline-main                    29s
pipeline-training-pipeline-dbg-17812871        29s
pipeline-training-pipeline-helm-fix-178        8s
pipeline-training-pipeline-with-trigger-main   29s
```

**Disabled (`pipeline.revisioningEnabled: false`)**

I patched the configmap back to `revisioningEnabled: false` and restarted. After patching the same pipeline:

```
ts=1781287583.078  "Reconciling pipeline starts"   namespace-name=ma-dev-test/training-pipeline
```

No `"created revision"` log followed. `kubectl get revisions -n ma-dev-test` returned the same four CRs from the previous run.

**Potential risks**

None — revisioning defaults to off. Existing behaviour is unchanged unless an operator explicitly sets `pipeline.revisioningEnabled: true`.

**Release notes**

`pipeline.revisioningEnabled` (Helm value `controllermgr.pipeline.revisioningEnabled`, default `false`) controls whether the controller snapshots a Revision CR on each successful pipeline reconcile.

**Documentation Changes**

None.


go/base/revision/noop.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"go.uber.org/yarpc"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type noopManager struct{}
+
+// NewNoOpManager returns a Manager that does nothing.
+// Used as the default when revision storage is not configured.
+func NewNoOpManager() Manager {
```
badcount: Where this is used?

craigmarker: nowhere. artifact of coding agent not understanding about optional Fx params. sorry about that noise

badcount (approved):

PR #1279, opened 2026-06-05
fix: CRD update mutations return 400 on JSON-encoded gRPC path

craigmarker:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

CRD mutation requests sent over the JSON-encoded gRPC path were failing
with 400 errors because gogo's jsonpb encodes `int64`/`uint64` fields as
quoted strings (`"42"`) per the proto3 JSON spec, but `encoding/json` —
which handles final deserialization — cannot decode quoted strings into
integer fields.

`UnmarshalJSONPB` is now generated on every top-level CRD type. It
intercepts gogo's JSON dispatch, unquotes any int64/uint64 values in the
raw bytes, flattens Kubernetes inline fields (TypeMeta), then delegates
to `encoding/json`. Map fields with int64 values are also handled.
Errors from the unquoting and inline-field passes include the offending
field path.

**Why?**

gogo jsonpb dispatches through the `JSONPBUnmarshaler` interface when
present. Without it, CRD types fell through to a default path that could
not reconcile proto3-encoded int64 strings with Go's `encoding/json`.

**How did you test it?**

**Potential risks**

Low. `UnmarshalJSONPB` is only invoked by gogo's jsonpb dispatcher —
standard `json.Unmarshal` callers are unaffected. The unquoting is a
no-op when values are already bare numbers, so existing correctly-encoded
payloads pass through unchanged.

**Release notes**

N/A

**Documentation Changes**

N/A


sallycr:
## Code Review — 4-Agent OSS Review Team

**Verdict: Approve with comments.** Solid tactical fix — no active bugs, tests pass, correctly handles the motivating fields. A few latent gaps and polish items worth addressing before or shortly after merge.

---

### How It Works (Verified)
gogo jsonpb dispatches to `UnmarshalJSONPB` on CRD messages → unquotes int64 fields → flattens inline TypeMeta → hands the subtree to `encoding/json` via the `type Alias` trick. `CollectInt64Fields` reflect-walks proto descriptors and caches paths in a `sync.Map`. All 16 generated CRD types gain the same template-driven dispatcher block.

### No Active Bugs
All current CRD types unmarshal correctly. `metadata.generation` and `metadata.deletionGracePeriodSeconds` are collected through embedded TypeMeta + nested ObjectMeta. Defensive guard (`isQuotedInteger`) prevents corruption of legitimate string fields.

---

### P1 — Should Fix

**1. `reflect.Map` not traversed in `CollectInt64Fields` (`util.go:328-393`)**
A `map<string,int64>` field collects `[]` — quoted int64 map values would fail to decode. No CRD uses this today, but it's a trap for the next person who adds one. Either handle `reflect.Map` (walk into elem) or add a comment documenting maps as unsupported.

**2. `reflect.Interface` (oneof) not traversed**
int64 fields inside oneof variants are silently skipped. Pre-existing limitation (not a regression), but worth a tracking comment since any future bare int64 in a oneof variant → zero-value field, hard to debug.

**3. No package doc on `go/kubeproto/util`**
The core "why" — proto3 encodes int64 as quoted strings, K8s JSON uses bare numbers, and the `type Alias` trick avoids infinite recursion — is only explained in commit messages. A contributor opening `util.go` has no top-level context. Suggest adding a `// Package util ...` doc comment capturing the proto3-int64/K8s-JSON mismatch and design rationale.

**4. No back-pointer from generated code to template**
Generated files carry the standard `// Code generated by protoc-gen-gogo. DO NOT EDIT.` header, but the `UnmarshalJSONPB` block comes from `go/kubeproto/templates/crd.tmpl`. A developer hitting a bug here will try to hand-edit the generated file. Suggest adding `// generated from go/kubeproto/templates/crd.tmpl` in the template block.

**5. Runtime errors lack field-path context**
`UnquoteInt64Fields` / `ApplyInlineFields` propagate raw sjson/json errors with no breadcrumb to which field or CRD. Suggest wrapping: `fmt.Errorf("unquoting int64 field %q: %w", resolvedPath, err)`.

---

### P2 — Nice to Have

**6. Test coverage gaps**
Only plain `int64` is exercised. No `uint64`, `sint64`, `fixed64`, `sfixed64`, max-int64 boundary, or `map<string,int64>` cases. Adding a `uint64` and a max-int64 case would harden it cheaply. Direct unit tests on `CollectInt64Fields` / `UnquoteInt64Fields` (vs only transitive coverage through codegen) would also help.

**7. Marshal asymmetry**
PR adds `UnmarshalJSONPB` but no `MarshalJSONPB`. gogo's interface doc says implement both. Works today via default gogo marshal, but deserves a comment explaining why this is intentional.

**8. Commit history cleanup**
Commits tell a good story and messages are excellent, but there's a self-cancelling pair (break in `51f759c7` → restore in `f59b0842`). Consider squashing to ~4 logical commits (proto+template, util library, generated regen, tests) before merge.

**9. Template duplication**
The inline-flattening dance is duplicated 3x in `crd.tmpl` (UnmarshalJSONPB + Spec/Status UnmarshalJSON). Drift risk — could extract a shared helper.

**10. Pre-existing: Spec/Status `UnmarshalJSON` doesn't check `ApplyInlineFields` error**
The NEW `UnmarshalJSONPB` this PR adds correctly checks the error, but the existing Spec/Status path at `crd.tmpl ~line 326-330` does not. Out of scope but worth a follow-up since it's right next door.

---

### Strategic Note — gogo-protobuf Deprecation

gogo-protobuf is officially deprecated and unmaintained. Kubernetes is actively removing it ([KEP-5589](https://github.com/kubernetes/enhancements/issues/5589)), and it's already causing Go build breaks ([kubernetes#135786](https://github.com/kubernetes/kubernetes/issues/135786)). Two lighter-weight alternatives exist for this specific problem:

- **`json:",string"` struct tag** on int64 fields — far less code, no gjson dependency. Only downside: doesn't accept bare numbers (this PR's approach is more lenient).
- **Istio's protoc-gen-jsonshim pattern** — generate marshalers delegating to proto JSON marshaler instead of `encoding/json`.

**Recommendation:** This PR is a good tactical fix — merge it. But consider opening a tracking issue for migrating to `google.golang.org/protobuf` + `protojson`. Once migrated, `protojson` handles int64 correctly and `util.go` can likely be deleted.

---

*Review performed by a 4-agent team: OSS Architect, OSS Researcher, OSS Dev Advocate, OSS Coder.*

craigmarker:
> ### P1 — Should Fix
> **1. `reflect.Map` not traversed in `CollectInt64Fields` (`util.go:328-393`)** A `map<string,int64>` field collects `[]` — quoted int64 map values would fail to decode. No CRD uses this today, but it's a trap for the next person who adds one. Either handle `reflect.Map` (walk into elem) or add a comment documenting maps as unsupported.

Done.

> **2. `reflect.Interface` (oneof) not traversed** int64 fields inside oneof variants are silently skipped. Pre-existing limitation (not a regression), but worth a tracking comment since any future bare int64 in a oneof variant → zero-value field, hard to debug.

I couldn't find a good solution to this. There is nothing in the reflect.Inferface API that can be recursed into. The NumField function used to enumerate fields for Struct recursion [panics if called on a non-Struct kind](https://pkg.go.dev/reflect#Value.NumField).
 
> **3. No package doc on `go/kubeproto/util`** The core "why" — proto3 encodes int64 as quoted strings, K8s JSON uses bare numbers, and the `type Alias` trick avoids infinite recursion — is only explained in commit messages. A contributor opening `util.go` has no top-level context. Suggest adding a `// Package util ...` doc comment capturing the proto3-int64/K8s-JSON mismatch and design rationale.

Happy to do this as a separate effort—these files are pre-existing.
 
> **4. No back-pointer from generated code to template** Generated files carry the standard `// Code generated by protoc-gen-gogo. DO NOT EDIT.` header, but the `UnmarshalJSONPB` block comes from `go/kubeproto/templates/crd.tmpl`. A developer hitting a bug here will try to hand-edit the generated file. Suggest adding `// generated from go/kubeproto/templates/crd.tmpl` in the template block.

Done.

> **5. Runtime errors lack field-path context** `UnquoteInt64Fields` / `ApplyInlineFields` propagate raw sjson/json errors with no breadcrumb to which field or CRD. Suggest wrapping: `fmt.Errorf("unquoting int64 field %q: %w", resolvedPath, err)`.

Done.

> ### P2 — Nice to Have
> **6. Test coverage gaps** Only plain `int64` is exercised. No `uint64`, `sint64`, `fixed64`, `sfixed64`, max-int64 boundary, or `map<string,int64>` cases. Adding a `uint64` and a max-int64 case would harden it cheaply. Direct unit tests on `CollectInt64Fields` / `UnquoteInt64Fields` (vs only transitive coverage through codegen) would also help.

Done.

> **7. Marshal asymmetry** PR adds `UnmarshalJSONPB` but no `MarshalJSONPB`. gogo's interface doc says implement both. Works today via default gogo marshal, but deserves a comment explaining why this is intentional.

Added documentation—Marshal works correctly so we don't need to override.

> **8. Commit history cleanup** Commits tell a good story and messages are excellent, but there's a self-cancelling pair (break in `51f759c7` → restore in `f59b0842`). Consider squashing to ~4 logical commits (proto+template, util library, generated regen, tests) before merge.

Squash & merge
 
> **9. Template duplication** The inline-flattening dance is duplicated 3x in `crd.tmpl` (UnmarshalJSONPB + Spec/Status UnmarshalJSON). Drift risk — could extract a shared helper.

Happy to do this as a separate refactor.

> **10. Pre-existing: Spec/Status `UnmarshalJSON` doesn't check `ApplyInlineFields` error** The NEW `UnmarshalJSONPB` this PR adds correctly checks the error, but the existing Spec/Status path at `crd.tmpl ~line 326-330` does not. Out of scope but worth a follow-up since it's right next door.

Sure, I can do this in a follow up.

go/kubeproto/templates/crd.tmpl
```diff
@@ -371,6 +371,36 @@ func (m *{{.Name}}Status) UnmarshalJSON(b []byte) error {
     return json.Unmarshal(b, aux)
 }
 
+// UnmarshalJSONPB implements jsonpb.JSONPBUnmarshaler.
+// generated from go/kubeproto/templates/crd.tmpl — edit the template, not this file.
```
badcount: why this file not editable?

craigmarker:
Superceded by #1468

PR #1281, opened 2026-06-06
fix(sandbox): adopt orphaned Helm resources via annotate instead of delete

badcount:
## Problem

The GCP integration test was failing with:

```
Error: UPGRADE FAILED: Unable to continue with update: Service "michelangelo-cadence-frontend"
in namespace "default" exists and cannot be imported into the current release: invalid ownership
metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to
"Helm"; annotation validation error: missing key "meta.helm.sh/release-name"...
```

The existing cluster has pre-existing resources (e.g. `michelangelo-cadence-frontend`) that lack Helm ownership labels/annotations. `helm upgrade --install` refuses to manage them.

## Fix

The `_helm_adopt_orphaned_resources` function previously **deleted** orphaned resources so Helm could recreate them. This approach is fragile (causes service restarts, and if `helm template` doesn't render a resource for any reason, the resource is missed).

Switch to **annotating and labeling** orphaned resources with the Helm ownership metadata in-place:
- `meta.helm.sh/release-name=michelangelo`
- `meta.helm.sh/release-namespace=default`
- `app.kubernetes.io/managed-by=Helm`

This lets Helm adopt the existing resources without recreating them, which is less disruptive and more reliable.

## Test plan

- [ ] Re-run the GCP integration test (`sandbox-e2e`)
- [ ] Verify `helm upgrade --install` succeeds on a cluster with pre-existing resources

Fixes: https://github.com/michelangelo-ai/michelangelo/actions/runs/27042506073/job/79821472928

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -346,15 +346,25 @@ def _sync(ns: argparse.Namespace):
     if release_healthy:
         # Healthy release: upgrade in-place, keeping infra running.
 
-        # If a previous run dropped the Cadence databases, the schema Job is
-        # still marked "Completed" but the databases are empty. Helm won't
-        # recreate a Completed Job, so we must delete it first so the upgrade
-        # recreates and reapplies the schema.
+        # If a previous run dropped the Cadence databases, or the schema is
+        # version-mismatched (Cadence refuses to bind its gRPC port on start),
+        # we must drop and recreate the databases. The schema Job is still
+        # "Completed" in k8s, so Helm won't recreate it — delete it first so
+        # the upgrade creates a fresh Job that applies the correct schema.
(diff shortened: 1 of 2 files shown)


python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -463,6 +505,70 @@ def _refresh_mysql_schema():
...
+        [
+            "kubectl",
+            "delete",
+            "job",
+            "-l",
+            "app.kubernetes.io/name=cadence,app.kubernetes.io/component=schema",
+            "--ignore-not-found=true",
+        ],
+        check=False,
+    )
```
aritra0227, line 544: I don't think this function is used anywhere. Remove?

aritra0227 (approved):
Approved with a comment.

PR #1283, opened 2026-06-08
fix(notification): OSS quality — extensible sinks, shared worker, configurable types (#1259 #1260)

### LINKED ISSUE #1259: feat(notification): task_list config, pipeline-run-started event, pluggable worker
## Background

During Grab's review of the forked notification support code, three follow-up blockers were identified that need to be addressed before the notification feature is production-ready for external adopters.

## Blocker 1 — `task_list` hardcoded as `"pipeline_run"`

**File:** `go/components/pipelinerun/notification/notifier.go:61`

```go
options := clientInterfaces.StartWorkflowOptions{
    ID:       workflowID,
    TaskList: "pipeline_run",  // hardcoded — should come from config
    ...
}
```

`PipelineRunNotifier` only receives `workflowClient` and `logger` — there is no config struct injected that could carry a task list value. The worker config at `go/cmd/worker/config/base.yaml` defines the three task lists (`default`, `trigger_run`, `pipeline_run`) but the notifier never reads from it.

**Proposed fix:** Inject a config struct into `PipelineRunNotifier` at construction and read `TaskList` from it — matching the pattern the worker already uses for its own task list registration.

---

## Blocker 2 — Missing "pipeline run started" notification event

**Files:**
- `proto/api/v2/notification.proto:43-65`
- `go/base/notification/types/types.go:28-40`

The state→event map in `types.go` only covers terminal states:

```go
var stateMap = map[v2pb.PipelineRunState]v2pb.Notification_EventType{
    PIPELINE_RUN_STATE_FAILED:    EVENT_TYPE_PIPELINE_RUN_STATE_FAILED,
    PIPELINE_RUN_STATE_SUCCEEDED: EVENT_TYPE_PIPELINE_RUN_STATE_SUCCEEDED,
    PIPELINE_RUN_STATE_KILLED:    EVENT_TYPE_PIPELINE_RUN_STATE_KILLED,
    PIPELINE_RUN_STATE_SKIPPED:   EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED,
    // RUNNING / PENDING missing
}
```

`PIPELINE_RUN_STATE_PENDING=1` and `PIPELINE_RUN_STATE_RUNNING=2` exist in `pipeline_run.proto` but have no corresponding notification event type — so users can never be notified when a run starts.

**Proposed fix:**
1. Add `EVENT_TYPE_PIPELINE_RUN_STATE_STARTED = 10` to `proto/api/v2/notification.proto`
2. Regenerate Go proto stubs
3. Add the mapping entry in `go/base/notification/types/types.go`

---

## Blocker 3 — Notification not pluggable (embedded in shared worker binary)

**Files:**
- `go/worker/workflows/module.go:13`
- `go/cmd/worker/`

The notification module is always registered unconditionally in the shared worker binary:

```go
var Module = fx.Options(
    ray.Module,
    trigger.Module,
    notification.Module,  // always on, no opt-out
)
```

There is no way to deploy or configure the notification worker independently. External adopters (e.g. Grab) that want to customise the Slack integration or disable email must fork the entire worker binary.

**Proposed fix:** Move the notification worker to its own binary at `go/cmd/notification-worker/main.go` with its own config file. This lets operators:
- Deploy it independently
- Configure the Cadence/Temporal task list without touching the shared worker
- Extend or replace the notification implementation without forking

```
go/cmd/
├── apiserver/
├── controllermgr/
├── notification-worker/   ← new
└── worker/
```

---

## Work items

- [ ] **Blocker 1:** Inject `TaskList` config into `PipelineRunNotifier` — remove hardcoded `"pipeline_run"`
- [ ] **Blocker 2:** Add `EVENT_TYPE_PIPELINE_RUN_STATE_STARTED` proto enum value + state map entry + regenerate
- [ ] **Blocker 3:** Create `go/cmd/notification-worker/main.go` as a standalone binary with its own config

## References

- Grab fork review discussion (Slack thread, 2026-05-29)

### LINKED ISSUE #1260: fix(notification): OSS quality — internal constants, naming, stubs, error handling
## Background

Follows up on #1259 (task_list config, STARTED event, pluggable worker). The OSS architecture review identified additional issues that block external adopters from using the notification system without forking. These are independent of the 3 blockers in #1259 and can be addressed in parallel.

Review file: `go/components/pipelinerun/notification/OSS_REVIEW.md`

---

## Issue 1 — Uber-internal URLs and email hardcoded as constants 

**Files:** `go/base/notification/types/types.go:16-17`, `go/worker/workflows/notification/workflows.go:76`

```go
_defaultMaURL   = "https://michelangelo-studio.uberinternal.com/ma/"  // in every notification body
_defaultMaEmail = "michelangelo@uber.com"                               // unused, leaks
SendAs:           "michelangelo@uber.com"                               // in every outgoing email
```

Every notification message body contains a broken Michelangelo Studio link for external adopters. Every outgoing email shows `michelangelo@uber.com` as sender.

**Fix:** Move `StudioBaseURL` and `SenderEmail` into the `Config` struct introduced in #1259. Remove the unused `_defaultMaEmail` constant.

---

## Issue 2 — `getPipelinePhase` hardcodes Uber pipeline taxonomy 

**File:** `go/base/notification/types/types.go:71-95`

A switch over 15+ `PIPELINE_TYPE_*` enum values maps to Michelangelo Studio URL path segments (`/train`, `/deploy`, `/genai-finetune`, etc.). This is Uber's internal ML product structure. External adopters with different pipeline types get `"unknown"` in all notification links.

**Fix (option A):** Accept a `PhaseResolver func(pipelineType string) string` parameter on `GenerateText`, defaulting to the existing switch.

**Fix (option B):** Move `getPipelinePhase` out of `go/base/notification/types/` into an Uber-specific package, leaving `types/` with generic helpers only.

---

## Issue 3 — Activity stubs are no-ops but godoc says "sends a message" 

**File:** `go/worker/activities/notification/activities.go:51, 90`

`SendMessageToSlackActivity` and `SendMessageToEmailActivity` log and return `nil`. There is no actual delivery. The godoc says "sends a message" — that is false and misleads adopters into thinking notifications work out of the box.

**Fix:** Rewrite godoc to be explicit:
```go
// SendMessageToSlackActivity is a stub. The default implementation logs the
// request and returns nil. Replace the body with your Slack transport
// (webhook POST, slack-go SDK, etc.) before relying on notifications in production.
```

Long-term: introduce a `Sink` interface (see #1259 Blocker 3 Step 3) so adopters plug in their transport without forking the activity.

---

## Issue 4 — `SendPRNotification` reads as "Pull Request" 

**File:** `go/worker/workflows/notification/workflows.go:15, 96`

`PRNotificationWorkflowName` and `SendPRNotification` — every external developer's first read is "Pull Request". This name appears in the Cadence/Temporal UI as the registered workflow name and in all workflow history records.

**Fix:**
- `PRNotificationWorkflowName` → `PipelineRunNotificationWorkflowName`
- `SendPRNotification` → `SendPipelineRunNotification`

**Migration note:** Renaming a registered Cadence/Temporal workflow name affects existing in-flight workflow histories. Add a CHANGELOG entry and deprecate the old name for one release before removing.

---

## Issue 5 — Error swallowing is doubled — controller warning never fires 

**Files:** `go/components/pipelinerun/notification/notifier.go:73-77`, `go/components/pipelinerun/controller.go:133-138`

`NotifyOnStateChange` always returns `nil` (even when `StartWorkflow` fails), so `controller.go:134`'s `logger.Warn("Failed to send notifications")` is dead code — it can never fire.

**Fix:** Bubble the error from `NotifyOnStateChange` and let the controller log it. Keep the controller's `Warn`-only (not `Error`) behavior to avoid blocking reconciliation — notification failures must not stall pipeline run processing.

---

## Issue 6 — Notification fan-out returns only the last error 

**File:** `go/worker/workflows/notification/workflows.go:99-119`

The loop that fans out notifications reassigns `err` on each iteration. If 3 of 4 notifications fail and the 4th succeeds, the workflow reports success. If 3 succeed and the 4th fails, only the 4th error is returned — the 3 failures are lost.

**Fix:** Accumulate with `errors.Join` (Go 1.20+) or `go.uber.org/multierr`:
```go
var errs error
for _, notif := range notifications {
    if err := sendOne(ctx, notif); err != nil {
        logger.Error("notification failed", zap.Error(err))
        errs = errors.Join(errs, err)
    }
}
return errs
```

---

## Issue 7 — `TRIGGER_RUN` and `PIPELINE_STATE` event types are dead code 

**File:** `proto/api/v2/notification.proto:54-63`

`EVENT_TYPE_TRIGGER_RUN_STATE_*` (values 5-7) and `EVENT_TYPE_PIPELINE_STATE_*` (values 8-9) are defined in the proto but `ContainsEventType` in `types.go:28-40` only maps `PIPELINE_RUN` states. Adopters who configure `EVENT_TYPE_TRIGGER_RUN_STATE_FAILED` receive no notifications silently.

**Fix:** Add a proto comment on each unimplemented value: `// Reserved. Not yet implemented — configuring this event type has no effect.`

---

## Issue 8 — Layering import: controller imports worker package 

**File:** `go/components/pipelinerun/notification/notifier.go:11`

The controller-side `notifier.go` imports `go/worker/workflows/notification` just to get `PRNotificationWorkflowName` (a string constant). A control-plane component should not import a worker-plane package.

**Fix:** Move workflow name constants into `go/base/notification/types` (or a new sibling `go/base/notification/names` package) so neither side imports the other.

---

## Work items

- [ ] **Issue 1:** Move `StudioBaseURL` and `SenderEmail` to `Config` struct; remove `_defaultMaEmail`
- [ ] **Issue 2:** Make `getPipelinePhase` injectable via `PhaseResolver` func param, or move to Uber-specific package
- [ ] **Issue 3:** Rewrite activity stub godoc to be explicit about no-op status
- [ ] **Issue 4:** Rename `SendPRNotification` → `SendPipelineRunNotification` (+ CHANGELOG migration note)
- [ ] **Issue 5:** Bubble `StartWorkflow` error from `NotifyOnStateChange`; remove double-swallow
- [ ] **Issue 6:** Accumulate fan-out errors with `errors.Join`
- [ ] **Issue 7:** Add proto comments documenting which event types are implemented vs. reserved
- [ ] **Issue 8:** Move workflow name constants to `go/base/notification/` to fix layering import

## References

- Parent issue: #1259 (task_list config, STARTED event, pluggable worker)

sallycr:
- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

- Consolidated notification into a shared worker (removed standalone notification-worker binary)
- Replaced string-based notification type constants with proto enum (`Notification_NotificationType`)
- Introduced extensible `Sink` interface for fan-out delivery (Email, Slack, or custom transports like PagerDuty/Teams)
- Generic `Message` struct with `Body` (plain-text fallback) + `FormattedBodies map[string]string` (MIME-style keys: `text/html`, `text/slack`) — sinks pick their preferred format and fall back to Body
- Added `PhaseResolver` callback for custom pipeline-type-to-UI-path mapping
- Made `SourcePipelineTypeLabelName` and `SourcePipelineManifestTypeLabelName` private — zero cross-package consumers
- Renamed `_pipelineManifestTypeASL` to `pipelineManifestTypeASL` (idiomatic Go unexported naming)
- Moved notification activity AND workflow registration out of shared packages into `cmd/worker/main.go`
- Added `EVENT_TYPE_PIPELINE_RUN_STATE_RUNNING` (10, deprecated) for backward compatibility; `EVENT_TYPE_PIPELINE_RUN_STATE_STARTED` shifted to 11
- Registered deprecated `PRNotificationWorkflow` alias for rolling deploy backward compatibility
- Dispatched notification after status persist to prevent duplicate STARTED events
- Added state-scoped workflow IDs and YAML tags on Config
- Added Helm values for notification configuration (taskList, studioBaseURL, senderEmail)
- Fixed Bazel BUILD files for notification activities and workflows dependencies

**Why?**

Addresses OSS quality issues #1259 and #1260:
1. The notification subsystem had Uber-internal patterns (hardcoded webhook sink, standalone binary) that don't fit open-source deployment
2. Notification type constants used raw strings instead of proto enums, making the API fragile
3. No extension point existed for operators to add custom notification channels
4. Notification activities and workflows were bundled in shared modules, preventing downstream forks from registering custom transports

**How did you test it?**

- Unit tests for all notification types, workflows, activities, and sinks
- End-to-end sandbox validation on Temporal (FAILED and SUCCEEDED pipeline runs)
- `go build` / `go vet` on all changed packages
- Verified zero remaining references to old exported label constant names via grep
- Verified backward compat: ContainsEventType matches both RUNNING and STARTED

**Potential risks**

- `DeprecatedPRNotificationWorkflowName` alias must stay registered for one release cycle to drain in-flight workflows
- Proto enum renumbering: STARTED moved from 10→11 — systems persisting the numeric value 10 will now resolve to RUNNING (deprecated)
- Downstream forks that imported `notificationActivities.Module` or `notificationWorkflows.Module` must update their wiring

**Breaking Changes**

- [ ] No breaking changes
- [x] Proto changes (enum value renumbering, field number changes, field removal, service removal)
- [x] Helm changes (new required values, renamed or removed keys, changed value semantics)
- [x] Config/deployment changes (new required env vars, renamed container args, changed ports or mount paths)

**Migration guide**

### 1. Proto enum: `EVENT_TYPE_PIPELINE_RUN_STATE_RUNNING` deprecated

`EVENT_TYPE_PIPELINE_RUN_STATE_RUNNING = 10` is now `[deprecated = true]`. Replacement: `EVENT_TYPE_PIPELINE_RUN_STATE_STARTED = 11`.

**Action:** Update any `spec.notifications[].eventTypes` referencing `RUNNING` to use `STARTED`. The worker accepts both for one release cycle.

### 2. Workflow rename: `PRNotificationWorkflow` → `io.michelangelo.notification.PipelineRunFanout`

**Action:** If your fork registers or dispatches the notification workflow by string name, update to `io.michelangelo.notification.PipelineRunFanout`. The old name is aliased for one release to drain in-flight executions.

### 3. Module move: `notificationActivities.Module` and `notificationWorkflows.Module`

Registration moved from shared packages to `go/cmd/worker/main.go`.

**Action:** If your fork imported these modules:
- Remove `fx.Options(notificationActivities.Module)` calls
- Add `notificationWorkflows.Module` to your fx options
- Add `fx.Invoke(RegisterNotificationActivities)` to register activity functions

See `go/cmd/worker/main.go` for the reference implementation.

**Release notes**

- Notification workflow renamed (`PRNotificationWorkflow` → `io.michelangelo.notification.PipelineRunFanout`, alias registered for one release)
- New Helm values: `notification.taskList`, `notification.studioBaseURL`, `notification.senderEmail`
- Notification types now use proto enum instead of strings
- `EVENT_TYPE_PIPELINE_RUN_STATE_RUNNING` deprecated — use `STARTED`
- Notification registration moved to `cmd/worker/main.go` for fork-friendly customization

**Documentation Changes**

Updated docstrings on all notification types, constants, and functions. Sink interface documented with extension examples.

Closes #1259, closes #1260


sallycr:
tested in sandbox temporal: 
-----
Email activity:
to:      ["test@example.com"]
subject: "Pipeline Run (sandbox-notif-temporal-1) state: SUCCEEDED"
send_as: ""   ← (no sender email configured in sandbox)

Slack activity:
channel: "#test-alerts"
text:
  Pipeline Run (sandbox-notif-temporal-1) state: SUCCEEDED:
  - Name: sandbox-notif-temporal-1
  - Project: default
  - State: SUCCEEDED
  - Pipeline Type: TRAIN
  - <http://localhost:3000/studio/default/train/runs/sandbox-notif-temporal-1|Studio URL>
  - <https://workflow.example.com/run/abc123|Workflow Log URL>
-----

sallycr:
verified it in cadence sandbox : 
--- 
Email activity (00:52:01.204):
to:      ["test@example.com"]
subject: "Pipeline Run (sandbox-notif-run-3) state: SUCCEEDED"
send_as: ""

Slack activity (00:52:01.230):
channel: "#test-alerts"
text:
  Pipeline Run (sandbox-notif-run-3) state: SUCCEEDED:
  - Name: sandbox-notif-run-3
  - Project: default
  - State: SUCCEEDED
  - Pipeline Type: TRAIN
  - <http://localhost:3000/studio/default/train/runs/sandbox-notif-run-3|Studio URL>
  - <https://workflow.example.com/run/abc123|Workflow Log URL>
---

go/worker/activities/notification/activities.go
```diff
@@ -1,123 +1,122 @@
...
 	if logger == nil {
-		// For testing contexts where activity logger is not available
 		logger = zap.NewNop()
 	}
-	logger.Info("Sending slack notification",
+	logger.Info("SendMessageToSlackActivity called (no-op: no transport configured)",
 		zap.String("channel", req.Channel),
 		zap.String("text", req.Text))
-
-	// Placeholder implementation - replace with actual CAG integration
```
badcount, line 63: I think those comments has value to show this is a place holder, don't suggest to remove

go/worker/activities/notification/activities.go
```diff
@@ -1,123 +1,122 @@
...
+// message. Replace the body of this function with your transport integration
+// (SMTP, SendGrid, etc.) before relying on email notifications in production.
 func SendMessageToEmailActivity(ctx context.Context, req *SendMessageToEmailActivityRequest) error {
-	// TODO(#701): Implement email sending logic
-	// This would typically integrate with internal CAG (Communication API Gateway) service
+	if req == nil {
+		return errors.New("SendMessageToEmailActivityRequest cannot be nil")
+	}
 	logger := activity.GetLogger(ctx)
 	if logger == nil {
```
badcount: this should be refactored we don't usually check logger nil in other activities

go/worker/activities/notification/module.go
```diff
@@ -21,9 +21,11 @@ var Module = fx.Options(
 // Registered Components:
 // - SendMessageToEmailActivity: Sends email notifications
 // - SendMessageToSlackActivity: Sends Slack notifications
+// - SendMessageToWebhookActivity: Sends webhook notifications
 func register(workers []worker.Worker) {
 	for _, w := range workers {
 		w.RegisterActivity(SendMessageToEmailActivity)
 		w.RegisterActivity(SendMessageToSlackActivity)
+		w.RegisterActivity(SendMessageToWebhookActivity)
```
badcount: why do we need to add this? let's separate out refactor with adding new feature

go/worker/workflows/notification/workflows.go
```diff
@@ -1,120 +1,104 @@
...
+// only matching notifications are delivered. Delivery failures are accumulated
+// with errors.Join so that a failure on one sink does not suppress others.
+func (wf *Workflow) SendPipelineRunNotification(ctx workflow.Context, req *types.PipelineRunNotificationRequest) error {
+	// Attach the workflow backend so that workflow.ExecuteActivity can resolve
+	// activities. This mirrors the pattern used by other Go workflows in this
+	// codebase (e.g. trigger.CronTrigger).
+	if wf.backend != nil {
+		ctx = workflow.WithBackend(ctx, wf.backend)
+	}
 	ctx = workflow.WithActivityOptions(ctx, workflowActivityOpts)
```
badcount: This code should be refactored to remove any activity from this method, only passing logger

go/cmd/notification-worker/config/base.yaml
```diff
@@ -0,0 +1,52 @@
+worker:
```
badcount: do we really need a separated worker for this?

go/base/notification/types/types.go
```diff
@@ -13,95 +14,169 @@ import (
...
+	// PipelineRunNotificationWorkflowName is the registered name of the pipeline run
+	// notification workflow in Cadence/Temporal. Both the notifier (which starts the
+	// workflow) and the worker (which registers it) must use this constant.
+	//
+	// The name follows the reverse-DNS convention recommended by Temporal to avoid
+	// collisions in multi-tenant namespaces.
+	PipelineRunNotificationWorkflowName = "io.michelangelo.notification.PipelineRunFanout"
+
+	// NotificationTypeEmail selects the plain-text email body in GenerateText.
+	NotificationTypeEmail = "email"
```
badcount: why we wan't just use     NOTIFICATION_TYPE_EMAIL = 1;

go/worker/activities/notification/activities.go
sallycr: CAG is the uber internal system name. no good for open source.

go/worker/activities/notification/activities.go
sallycr: ah. good catch. thx

sallycr:
Thanks for flagging this — I dug into it and the picture is a bit nuanced, so leaving a full write-up.

**You're right that the original `if logger == nil { logger = zap.NewNop() }` was wrong** — in a real Cadence/Temporal worker context `activity.GetLogger` will never return nil, so the fallback was dead code. That's now removed.

**What I found tracing the call chain:**

The starlark-worker `activity.GetLogger` has its own safety wrapper:

```go
// starlark-worker activity/activity.go
func GetLogger(ctx context.Context) *zap.Logger {
    if b, ok := workflow.GetBackend(ctx); ok {   // comma-ok, never panics
        return b.GetActivityLogger(ctx)
    }
    return nil   // returns nil when no backend in context
}
```

The underlying SDK calls it wraps are **not** safe — both Cadence and Temporal panic with `"Not an activity context"` if you call their native `activity.GetLogger` on a bare `context.Background()`. The starlark-worker wrapper absorbs that by using a comma-ok type assertion on the context key, so nil is the defined return value for non-worker contexts.

The starlark-worker's own production activities (`plugin/request/activities.go`) call `logger.Info(...)` directly without a nil check — safe because those activities only run under a real worker where the backend is always present.

**What we settled on:**

```go
if logger := activity.GetLogger(ctx); logger != nil {
    logger.Info("SendMessageToSlackActivity called (no-op: no transport configured)", ...)
}
```

This is the right call for our case: the nil path only occurs in unit tests using `context.Background()`, where silently skipping the log line is fine. In production the backend is always set so the branch is never taken. Removed the `zap.NewNop()` fallback per your feedback; tests now pass `context.Background()` directly.

Happy to revisit if you'd prefer a different approach (e.g. requiring callers to pass a logger explicitly), but this seemed like the minimal correct fix.

go/worker/activities/notification/module.go
sallycr: This is the identified that common way implemented other mlops solutions with the benefits of : 

WebhookSink is a reference implementation, not a user-facing feature. The Sink interface is the actual extension point — any operator who wants to add PagerDuty, OpsGenie, or a custom HTTP endpoint can drop in a new Sink without touching the workflow. But an interface with only two examples (email and Slack) tells an integrator "here are two things that happen to implement this." A third example — especially one using a different configuration pattern (URLs at construction time rather than from the proto) — makes the contract obvious: the workflow doesn't need to know about your channel, and you don't need a proto change to add one.

Concretely, WebhookSink demonstrates two things neither EmailSink nor SlackSink shows:
1. A sink that is configured at startup (operator-supplied URLs) rather than driven by user proto fields
2. errors.Join accumulation across multiple endpoints in one sink

Without it, an operator looking at the code might assume every new channel requires a new proto field and a workflow change. With it, the pattern is self-evident.

The cost is low — it's a stub activity that logs and returns nil, same as the others. The proto not having a webhook_urls field is intentional: operators are expected to configure it via the worker's dependency injection (FX), not through the user API.

go/worker/workflows/notification/workflows.go
sallycr: ack. thx

go/components/pipelinerun/notification/notifier.go
```diff
@@ -8,29 +8,61 @@ import (
...
 	"github.com/michelangelo-ai/michelangelo/go/base/notification/types"
 	clientInterfaces "github.com/michelangelo-ai/michelangelo/go/base/workflowclient/interface"
-	"github.com/michelangelo-ai/michelangelo/go/worker/workflows/notification"
 	v2pb "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
 	"go.uber.org/zap"
 )
 
-// PipelineRunNotifier handles notification logic for pipeline run state changes.
+// Config holds operator-supplied configuration for PipelineRunNotifier.
+type Config struct {
```
andriika, line 17: ### 2. `Config` struct has no yaml tags — camelCase YAML keys won't populate fields

`yaml.v2` lowercases unannotated field names (`gopkg.in/yaml.v2/yaml.go:395` — `info.Key = strings.ToLower(field.Name)`), and looks up YAML keys case-sensitively (`decode.go:753` — `sinfo.FieldsMap[name.String()]`). So:

- struct field `TaskList` → fieldsMap key `tasklist`
- YAML key `taskList` (camelCase, as in the documented example and in `notification-worker/config/base.yaml:31`) → not found in fieldsMap
- `Populate` silently leaves `cfg.TaskList = ""`

**Failure scenario:** Even after fixing finding #1 by adding `notification: { taskList: foo }` to `controllermgr/config/base.yaml`, the field stays empty and `NewPipelineRunNotifier` still errors. The sibling `Config` at `controller.go:39` uses `yaml:"ttlDays"` — that's the established convention; the new struct breaks it.

**Fix:**
```go
type Config struct {
    TaskList      string `yaml:"taskList"`
    StudioBaseURL string `yaml:"studioBaseURL"`
    SenderEmail   string `yaml:"senderEmail"`
}
```

go/worker/workflows/notification/sinks.go
```diff
@@ -0,0 +1,139 @@
...
+// Errors from individual endpoints are accumulated with errors.Join so that a
+// failure on one endpoint does not suppress delivery to others.
+func (s *WebhookSink) Notify(ctx workflow.Context, notif *v2pb.Notification, msg Message) error {
+	if len(s.URLs) == 0 {
+		return nil
+	}
+
+	bodyBytes, err := json.Marshal(map[string]any{
+		"subject":  msg.Subject,
+		"text":     msg.SlackText,
```
andriika: ### 4. WebhookSink JSON body uses `msg.SlackText` — Slack-formatted markup leaks to all webhook receivers

`msg.SlackText` contains Slack's `<https://...|Studio URL>` markup. Webhook receivers expecting plain text or HTML get literal angle-bracket markup they cannot reliably parse.

**Failure scenario:** WebhookSink wired for PagerDuty/Teams/a generic HTTP endpoint. Payload contains `"text": "Pipeline Run (X) state: SUCCEEDED:\n- <https://...|Studio URL>\n"`. Teams renders literal `<https://...|Studio URL>` instead of a hyperlink; PagerDuty incident title is corrupted.

**Fix:** Use `msg.EmailText` (plain text) for the webhook body, OR add a channel-neutral `Message.Body` field, OR have each sink render its own body from the underlying `PipelineRun` (would eliminate `EmailText`/`SlackText` duplication on `Message` as well).

go/worker/workflows/notification/sinks.go
```diff
@@ -0,0 +1,139 @@
...
+	URLs []string
+	// Headers is an optional map of HTTP headers sent with every request.
+	Headers map[string]string
+}
+
+// Notify POSTs a JSON payload to every URL configured on the sink.
+// Returns nil immediately when URLs is empty.
+// Errors from individual endpoints are accumulated with errors.Join so that a
+// failure on one endpoint does not suppress delivery to others.
+func (s *WebhookSink) Notify(ctx workflow.Context, notif *v2pb.Notification, msg Message) error {
```
andriika: ### 5. `WebhookSink.Notify` ignores the `notif` argument — fires every URL on every matched notification

Unlike `EmailSink` (gates on `len(notif.Emails)`) and `SlackSink` (gates on `len(notif.SlackDestinations)`), `WebhookSink` only checks its own configured `s.URLs` and ignores per-notification intent.

**Failure scenario:** Operator wires `WebhookSink` with a global audit URL alongside `EmailSink`. A user configures `Notification{NotificationType: NOTIFICATION_TYPE_EMAIL, Emails: ["private@example.com"]}` for a confidential team. The webhook still receives a POST containing that team's pipeline event — confidential data leaks to an unintended endpoint.

**Fix:** Either gate on a per-notification webhook destination (requires proto change), or treat the per-sink URL list as deliberate fan-out to all matched events and document that explicitly (the doc comment currently calls this design choice out for *proto schema avoidance* but not the leakage implication).

go/components/pipelinerun/notification/notifier.go
```diff
@@ -44,86 +76,78 @@ func (n *PipelineRunNotifier) NotifyOnStateChange(
...
-	// Crop pipeline run to reduce payload size for workflow
-	croppedPipelineRun := types.CropPipelineRun(newPipelineRun)
+	req := &types.PipelineRunNotificationRequest{
+		PipelineRun:   types.CropPipelineRun(newPipelineRun),
+		StudioBaseURL: n.cfg.StudioBaseURL,
+		SenderEmail:   n.cfg.SenderEmail,
+	}
 
-	// Start notification workflow
 	workflowID := fmt.Sprintf("%s.%s.notification", newPipelineRun.Namespace, newPipelineRun.Name)
```
andriika: ### 6. workflowID collision drops terminal-state notifications when STARTED workflow is in-flight

The workflowID `<ns>.<name>.notification` is constant per PipelineRun. When both STARTED + a terminal state are subscribed for the same run, and the STARTED workflow is still active, Cadence/Temporal rejects the second `StartWorkflow` with `WorkflowExecutionAlreadyStartedError`. The `AllowDuplicate` policy at `cadenceclient/cadenceclient.go:53` only permits reuse *after* the previous execution closes — it does not allow two concurrent executions with the same ID.

**Failure scenario:** User subscribes to both STARTED and SUCCEEDED. Pipeline transitions PENDING→RUNNING → fires workflow id `ns.foo.notification` for STARTED. A real transport activity retries on transient failure for a few minutes. Pipeline finishes RUNNING→SUCCEEDED while STARTED workflow is still in flight. Notifier tries to start SUCCEEDED workflow with the same ID → Cadence rejects → notifier logs Warn, returns nil, **terminal notification is silently dropped** (the more important one of the two).

**Fix:** Include the state name or a timestamp in the workflow ID, e.g. `fmt.Sprintf("%s.%s.notification.%s", ns, name, strings.ToLower(state.String()))`.

andriika (commented):
Code review findings (top 8, severity-ranked). Findings 2 / 4 / 5 / 6 / 7 / 8 are posted as inline comments on the relevant lines. Findings 1 and 3 are below because they're about absent files/templates and don't anchor to a specific line in the diff.

---

### 1. controllermgr crashloops — `go/cmd/controllermgr/config/base.yaml` has no `notification:` block

`provideConfig` in `notification/module.go` returns `Config{TaskList:""}`; `NewPipelineRunNotifier` then returns the error `notification.Config.TaskList is required`; FX aborts startup. The helm template at `helm/michelangelo/templates/core/controllermgr-configmap.yaml` and the sandbox manifest under `python/.tox/.../sandbox/resources/michelangelo-controllermgr.yaml` also have no `notification:` block, so every default deployment crashloops.

**Failure scenario:** `helm install` (or `mactl sandbox`) → controllermgr pod fails `fx.New` → every CRD reconciler (PipelineRun, RayCluster, Spark, Trigger, Deployment, InferenceServer, Ingester) is down.

**Fix options:** (a) add a `notification: { taskList: ... }` block to controllermgr's base.yaml and the helm template; (b) make `NewPipelineRunNotifier` return a `NoopNotifier` when `TaskList==""` so the controller stays up but skips dispatch.

---

### 3. No helm template / sandbox manifest deploys the new `notification-worker` binary

`grep -rin notification helm/` returns zero hits for a `notification-worker` deployment template; `helm/michelangelo/values.yaml` `workflow.taskLists` defaults to `[default, trigger_run, pipeline_run]` with no `notification_worker` entry; the sandbox resources under `python/.tox/.../sandbox/resources/` likewise don't include it. Combined with `go/worker/workflows/module.go:10-11` ("Notification workflows are registered separately by the notification-worker binary") removing notification registration from the shared worker, nothing polls the `notification_worker` task list out of the box.

**Failure scenario:** Even if #1 is fixed and controllermgr dispatches to task list `notification_worker`, no listener exists. Workflows wait through `ScheduleToStartTimeout` (1 min) inside a 60h `ExecutionStartToCloseTimeout` shell, then time out per attempt. Cadence/Temporal namespace fills with stuck workflows. No notifications delivered.

**Fix:** Ship a `helm/michelangelo/templates/core/notification-worker-deployment.yaml` + configmap + value gate, and document the new task list in the upgrade notes.

---

### Notable non-findings (verified during the review)

- **STARTED storm on every reconcile** — refuted. `originalPipelineRun` is a deep-copy of the persisted state at `controller.go:115`, so `shouldNotify` short-circuits on `oldState==newState`. The narrower restart/status-update-failure re-fire race is captured as finding 7.
- **`CropPipelineRun` shared-map race with concurrent reconciler mutation** — refuted. `CadenceClient.StartWorkflow` serializes the workflow args synchronously before returning; no concurrent reads on the shared maps.
- **Nil entries in `Spec.Notifications` panicking the workflow** — possible from CRD JSON deserialization but pre-existing in the older code, not a regression of this PR.

go/components/pipelinerun/notification/notifier.go
```diff
@@ -44,86 +76,78 @@ func (n *PipelineRunNotifier) NotifyOnStateChange(
...
 		ExecutionStartToCloseTimeout:    60 * time.Hour,
 		DecisionTaskStartToCloseTimeout: 30 * time.Second,
 	}
 
 	execution, err := n.workflowClient.StartWorkflow(
 		ctx,
 		options,
-		notification.PRNotificationWorkflowName,
-		croppedPipelineRun,
+		types.PipelineRunNotificationWorkflowName,
```
andriika, line 119: ### 8. Rolling deploy breaks in-flight workflows — name AND input schema BOTH changed

This PR renames the workflow (`PRNotificationWorkflow` → `io.michelangelo.notification.PipelineRunFanout`) AND changes the input type (`*v2pb.PipelineRun` → `*types.PipelineRunNotificationRequest`).

**Failure scenario:** During the rollout window:
- Pre-PR controllermgr dispatches `PRNotificationWorkflow` with `*v2pb.PipelineRun` arg → lands in the Cadence/Temporal queue.
- New worker binary rolls out and registers only `io.michelangelo.notification.PipelineRunFanout`. Old-name workflows hang as "WorkflowType not registered" until the 60h `ExecutionStartToCloseTimeout` expires. No notifications delivered for the events queued during the rollout window.

Even if only the schema had changed (name kept), the old payload's `PipelineRun` bytes would decode into the `PipelineRun` field of the new request struct only by coincidence of protobuf field ordering — silent data loss for `StudioBaseURL`/`SenderEmail`.

**Fix:** Register both the old and new workflow names on the worker for one release, or drain the queue before deploy.

go/components/pipelinerun/notification/notifier.go
```diff
@@ -44,86 +76,78 @@ func (n *PipelineRunNotifier) NotifyOnStateChange(
...
 func (n *PipelineRunNotifier) shouldNotify(
 	oldPipelineRun, newPipelineRun *v2pb.PipelineRun,
 	logger *zap.Logger,
 ) bool {
-	if newPipelineRun == nil {
-		return false
-	}
-
-	// Extract effective states for comparison
 	oldState := getEffectiveState(oldPipelineRun)
```
andriika, line 138: ### 7. STARTED has no persistent idempotency marker — restart between dispatch and status-persist re-fires the notification

`shouldNotify` compares the engine's computed `newState` against the persisted `originalPipelineRun.Status.State`. If `updatePipelineRunStatus` at `controller.go:156` fails (network blip, conflict) or the controller restarts between the line-141 dispatch and the line-156 persist, the next reconcile sees the same `original=PENDING` and re-fires the STARTED workflow.

**Failure scenario:**
1. Reconcile N: `original=PENDING`, engine sets `new=RUNNING`, line 141 dispatches STARTED workflow.
2. Line 156 `updatePipelineRunStatus` fails. Reconciler returns; nothing is persisted.
3. Reconcile N+1: `original=PENDING` again (no state was written), engine again sets `new=RUNNING`, line 141 dispatches a **second** STARTED workflow.

Same pattern on a controller restart between the two calls. Combined with finding #6, the duplicate STARTED workflow can also block a legitimate terminal-state workflow via the workflowID collision.

**Fix:** Persist a `michelangelo/started-notified` annotation on successful dispatch and check it in `shouldNotify` before re-dispatching STARTED. Or, push the dispatch to AFTER `updatePipelineRunStatus` succeeds so a write failure doesn't leave the notification ahead of the persisted state.

go/cmd/notification-worker/config/base.yaml
sallycr: The notification workflow could run on the shared worker by adding notification_worker to its task list config and including the module in go/cmd/worker/main.go. 

This way lets deployers provide transport credentials (SMTP, Slack tokens) and scale/restart notification delivery independently of the Ray/Spark worker fleet.

Do you want to combine worker with ray/ spark activities? @badcount @andriika

go/base/notification/types/types.go
sallycr: thx. good idea.

go/worker/workflows/notification/sinks.go
sallycr, line 1: Stale finding — WebhookSink was removed from this PR in commit f925fd83. The Sink interface remains as an extension point; operators can add a webhook sink via fx.Decorate when they have a real transport to wire.

sallycr, line 1: Stale finding — WebhookSink was removed from this PR in commit f925fd83. The notif-scoping concern is noted for any future WebhookSink implementation.

sallycr:
**Re: findings 1, 2, 3, 4, 5, 6 — addressed in latest push.**

**Finding 1 (controllermgr crashloop):** Already fixed — `go/cmd/controllermgr/config/base.yaml` has a `notification:` block with env-var-overridable defaults; `NewPipelineRunNotifier` returns `nil, nil` on empty `taskList` so the controller starts cleanly with notifications disabled.

**Finding 2 (missing yaml tags):** Fixed in commit `1a33ee3d` — added `yaml:"taskList"`, `yaml:"studioBaseURL"`, `yaml:"senderEmail"` to all three `Config` fields.

**Finding 3 (no Helm deployment for notification-worker):** Stale — the standalone `notification-worker` binary has been removed entirely (commit `c2f59c58`). Notification activities and workflows are now registered in the shared worker via the `notification_worker` task list, which is included in `workflow.taskLists` in `values.yaml` and in `go/cmd/worker/config/base.yaml`.

**Finding 4 (WebhookSink uses SlackText):** Stale — `WebhookSink` was removed in commit `f925fd83`. The `Sink` interface and `Message.Metadata` map remain as extension points for operators to add their own transports via `fx.Decorate`.

**Finding 5 (WebhookSink ignores notif arg):** Stale — same as finding 4; `WebhookSink` is gone. Noted for any future implementation.

**Finding 6 (workflowID collision):** Fixed in commit `1a33ee3d` — workflow ID now includes the state name: `<ns>.<name>.notification.<state>`, giving each state transition its own execution slot.

go/base/notification/types/types.go
```diff
@@ -13,95 +14,164 @@ import (
...
+
+// ContainsEventType reports whether any of eventTypes corresponds to prState.
 func ContainsEventType(eventTypes []v2pb.Notification_EventType, prState v2pb.PipelineRunState) bool {
-	var stateMap = map[v2pb.PipelineRunState]v2pb.Notification_EventType{
+	stateMap := map[v2pb.PipelineRunState]v2pb.Notification_EventType{
 		v2pb.PIPELINE_RUN_STATE_FAILED:    v2pb.EVENT_TYPE_PIPELINE_RUN_STATE_FAILED,
 		v2pb.PIPELINE_RUN_STATE_SUCCEEDED: v2pb.EVENT_TYPE_PIPELINE_RUN_STATE_SUCCEEDED,
 		v2pb.PIPELINE_RUN_STATE_KILLED:    v2pb.EVENT_TYPE_PIPELINE_RUN_STATE_KILLED,
 		v2pb.PIPELINE_RUN_STATE_SKIPPED:   v2pb.EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED,
+		v2pb.PIPELINE_RUN_STATE_RUNNING:   v2pb.EVENT_TYPE_PIPELINE_RUN_STATE_STARTED,
```
zhouxianzx, line 108: nit: Would it be clearer to align the state/event naming here? `PIPELINE_RUN_STATE_RUNNING` maps to `EVENT_TYPE_PIPELINE_RUN_STATE_STARTED`, which is understandable, but slightly surprising for config/API users. Maybe either name the event `EVENT_TYPE_PIPELINE_RUN_STATE_RUNNING`, or document explicitly that `STARTED` means “transitioned into RUNNING”.

badcount (changes requested):

go/base/notification/types/types.go
```diff
@@ -13,95 +14,164 @@ import (
...
 
+// GenerateSubject returns the email subject line for a pipeline run notification.
 func GenerateSubject(pipelineRun *v2pb.PipelineRun) string {
 	state := strings.TrimPrefix(pipelineRun.Status.State.String(), "PIPELINE_RUN_STATE_")
-	return fmt.Sprintf("Pipeline Run (%s) has completed with state %s", pipelineRun.Name, state)
+	return fmt.Sprintf("Pipeline Run (%s) state: %s", pipelineRun.Name, state)
 }
 
-func GenerateText(pipelineRun *v2pb.PipelineRun, textType string) string {
-	pipelineType := pipelineRun.Labels[_sourcePipelineTypeLabelName]
```
badcount, line 49: Why we remove this?

helm/michelangelo/values.yaml
```diff
@@ -97,6 +97,7 @@ workflow:
     - default
     - trigger_run
     - pipeline_run
+    - notification_worker
```
badcount, line 101: why do we need this?

proto/api/v2/notification.proto
```diff
@@ -51,17 +51,18 @@ message Notification {
...
     EVENT_TYPE_TRIGGER_RUN_STATE_SUCCEEDED = 7;
-    // Notification is generated by pipeline state ready
+    // Reserved. Not yet implemented — configuring this event type has no effect.
     EVENT_TYPE_PIPELINE_STATE_READY = 8;
-    // Notification is generated by pipeline state error
+    // Reserved. Not yet implemented — configuring this event type has no effect.
     EVENT_TYPE_PIPELINE_STATE_ERROR = 9;
-
+    // Notification is generated when a pipeline run transitions to the running state.
+    EVENT_TYPE_PIPELINE_RUN_STATE_STARTED = 10;
```
badcount: I remember RUNNING is needed not started

go/cmd/controllermgr/config/base.yaml
```diff
@@ -26,6 +26,16 @@ ingester:
...
 
+# Notification configuration for PipelineRun state-change workflows.
+# taskList must match the task list registered by the worker
+# (default: notification_worker — registered by the shared worker binary).
+# Leave taskList empty to disable notifications — the controller starts normally
+# but no notification workflows are dispatched.
+notification:
+  taskList: ${NOTIFICATION_TASK_LIST:""}
+  studioBaseURL: ${NOTIFICATION_STUDIO_BASE_URL:""}
+  senderEmail: ${NOTIFICATION_SENDER_EMAIL:""}
```
badcount, line 39: Why this is needed?

go/cmd/worker/config/base.yaml
```diff
@@ -24,6 +24,8 @@ workflow-engine:
       taskList: trigger_run
     - domain: default
       taskList: pipeline_run
```
badcount: "pipeline_run" is for notification task list

go/components/pipelinerun/notification/notifier.go
sallycr: 1. Workflow name: PRNotificationWorkflow → io.michelangelo.notification.PipelineRunFanout

PRNotificationWorkflow is a private shorthand — it means nothing to an external operator looking at a Cadence/Temporal workflow list. The new name has two benefits:

- Reverse-domain namespacing (io.michelangelo.*): Cadence/Temporal shares a namespace across all task lists. Without namespacing, workflow names collide if an operator runs Michelangelo alongside other systems that also have a workflow called PRNotificationWorkflow. The reverse-domain prefix is the standard convention for OSS workflow engines (same pattern as Java package names, Android intents, etc.).
- Self-describing (PipelineRunFanout): the name tells you exactly what the workflow does in the Cadence/Temporal UI — it fans out a pipeline run state change to all configured sinks. PR was ambiguous (pull request? pipeline run? production?).

---
2. Input type: *v2pb.PipelineRun → *types.PipelineRunNotificationRequest

On main, the workflow took the raw PipelineRun proto. Two problems with that:

- No way to pass operator config into the workflow: StudioBaseURL (for deep links in messages) and SenderEmail (the From address) are runtime config values that vary per deployment. The workflow had them hardcoded to internal URLs. With a bare PipelineRun argument there's nowhere to carry them — the only options were hardcoding or a global config read inside the workflow, both of which break OSS deployability.
- Unnecessary data on the wire: The full PipelineRun proto includes large fields (spec, labels, annotations) that the notification workflow doesn't need. CropPipelineRun already existed to trim it down, but the cropped value was still typed as *v2pb.PipelineRun, making the trimming invisible at the call site.

PipelineRunNotificationRequest bundles the cropped run with the two operator-supplied config values in one explicit struct, making the workflow's actual inputs clear and keeping internal config out of the proto.

I like this finding and we should fix for long run. For Uber, we're in migrating so won't break. how about @zhouxianzx @josephsahayaraj What do you think?

go/components/pipelinerun/notification/notifier.go
sallycr: thx.

go/components/pipelinerun/notification/notifier.go
sallycr: thx. change the code to "push the dispatch to AFTER updatePipelineRunStatus succeeds so a write failure doesn't leave the notification ahead of the persisted state"

sallycr:
**Finding #7 — fixed in commit `084f6723`.**

Moved `NotifyOnStateChange` to after `updatePipelineRunStatus` succeeds, gating on `returnErr == nil`. The notification now only fires when both the engine step and the status write have completed successfully, so a write failure or controller restart no longer re-dispatches the event on the next reconcile.

Went with Option B (dispatch after persist) over Option A (annotation) — it's the minimal change and avoids an extra Kubernetes write per STARTED event.

sallycr:
**Finding #8 — fixed in commit `37705a08`.**

Registered `PRNotificationWorkflow` as an alias on the same handler alongside the new `io.michelangelo.notification.PipelineRunFanout` name. In-flight executions dispatched by a pre-upgrade controllermgr will now be picked up by the new worker and drain cleanly instead of hanging until the 60h `ExecutionStartToCloseTimeout`.

Added `DeprecatedPRNotificationWorkflowName` constant in `types` with a removal note — the alias registration in `module.go` should be dropped after all operators have rolled past this release.

go/worker/activities/notification/activities.go
sallycr: Fixed in 53559ebb — restored a generic extension comment on SendMessageToEmailActivityRequest indicating that attachments, categories, and headers can be added when integrating a real transport. The CAG-specific reference was removed as it named an internal system.

go/worker/activities/notification/activities.go
sallycr: activity.GetLogger(ctx) returns nil when there is no backend in the workflow context — this happens in unit tests that do not run through a real Cadence/Temporal worker. The nil guard prevents a panic in those cases. Other activities do not need it because they only run in live workflow contexts.

go/worker/activities/notification/module.go
sallycr: The verbose multi-line docstring on the unexported register() function was removed as cleanup — it restated what the function body and the registered activity names already made obvious. The module-level doc and the exported type comments cover the public contract.

go/worker/workflows/notification/workflows.go
sallycr: Addressed in commit 8baf48e9 — activity options are now set inside each Sink.Notify call (workflow.WithActivityOptions wrapping ExecuteActivity), and logger is extracted once at the workflow level and passed down to sinks. The workflow method no longer touches activities directly.

go/base/notification/types/types.go
sallycr: Addressed in commit ed438db3 — replaced the string constants NotificationTypeEmail / NotificationTypeSlack with v2pb.NOTIFICATION_TYPE_EMAIL / v2pb.NOTIFICATION_TYPE_SLACK proto enums. Using the proto enum gives type safety and keeps the source of truth in the proto definition.

go/base/notification/types/types.go
sallycr: _defaultMaURL (uberinternal.com) and _defaultMaEmail (uber.com) were hardcoded internal values that cannot ship in an OSS codebase. They are now operator-supplied via Config.StudioBaseURL and Config.SenderEmail.

helm/michelangelo/values.yaml
sallycr: notification_worker is added so the shared worker polls that task list and picks up notification workflows dispatched by controllermgr. Without this entry, StartWorkflow succeeds on the client side but no worker processes the execution — workflows hang until the 60h timeout expires.

proto/api/v2/notification.proto
sallycr: EVENT_TYPE_PIPELINE_RUN_STATE_STARTED fires when the pipeline run transitions to RUNNING state. STARTED is the user-facing milestone name, not the internal state constant. Added a comment in ContainsEventType (commit 53559ebb) explaining the mapping explicitly.

go/cmd/controllermgr/config/base.yaml
sallycr: controllermgr reads this block via go.uber.org/config to populate notification.Config in FX, which constructs PipelineRunNotifier. Without it, the config provider returns empty strings, NewPipelineRunNotifier sees an empty TaskList, and notifications are silently disabled.

go/cmd/worker/config/base.yaml
sallycr: pipeline_run is the task list for pipeline run execution workflows (engine steps, Ray cluster management). notification_worker is separate so notification delivery can be scaled or restarted independently without affecting the main execution pipeline.

go/base/notification/types/types.go
sallycr: Good catch — fixed in commit 53559ebb. Added a comment in ContainsEventType explaining that PIPELINE_RUN_STATE_RUNNING maps to EVENT_TYPE_PIPELINE_RUN_STATE_STARTED because STARTED is the user-visible lifecycle milestone, not the internal state constant name.

go/components/pipelinerun/notification/notifier.go
zhouxianzx: I think the rollout compatibility risk is low. We always release the MA controller and worker together, so the mixed-version window should be short. The notification feature is also not officially released or widely adopted internally yet.

go/base/notification/types/types.go
```diff
@@ -13,95 +14,175 @@ import (
 )
 
 const (
-	_defaultMaURL                        = "https://michelangelo-studio.uberinternal.com/ma/"
-	_defaultMaEmail                      = "michelangelo@uber.com"
-	_sourcePipelineTypeLabelName         = "michelangelo/SourcePipelineType"
-	_sourcePipelineManifestTypeLabelName = "pipeline.michelangelo/PipelineManifestType"
```
badcount, line 19: can you make this same as private before

go/base/notification/types/types.go
```diff
@@ -13,95 +14,175 @@ import (
...
+	// package was open-sourced. The worker registers this alias for one release so
+	// that in-flight executions dispatched by a pre-upgrade controllermgr can drain
+	// without hanging until their 60h ExecutionStartToCloseTimeout. Remove once all
+	// operators have rolled past this release.
+	DeprecatedPRNotificationWorkflowName = "PRNotificationWorkflow"
+
+	// SourcePipelineTypeLabelName is the Kubernetes label key that identifies the
+	// pipeline type (e.g. PIPELINE_TYPE_TRAIN). External integrators can read this
+	// label to drive custom routing logic without hardcoding the key string.
+	SourcePipelineTypeLabelName = "michelangelo/SourcePipelineType"
```
sishilong: could u make it private as before

go/cmd/worker/config/base.yaml
```diff
@@ -24,6 +24,8 @@ workflow-engine:
       taskList: trigger_run
     - domain: default
```
sishilong, line 25: just a note: pipelineRun internally is using conditionengine and also trigger a cadence workflow

internal:
cadence:
  clients:
    - domain: michelangelo-controllermgr
      service: cadence
  workers:
    - domain: michelangelo-controllermgr
      task_list: pipeline_run

go/worker/workflows/module.go
```diff
@@ -1,14 +1,16 @@
...
 	"github.com/michelangelo-ai/michelangelo/go/worker/workflows/trigger"
-	"go.uber.org/fx"
 )
 
+// Module provides workflow registrations for the shared worker binary.
 var Module = fx.Options(
 	ray.Module,
 	trigger.Module,
-	notification.Module,
+	notificationWorkflows.Module,
```
badcount: Let's move this to cmd/module.go

proto/api/v2/notification.proto
badcount: Change the number to 11

yangfan0328: Will make this backward compatible, grab has some trigger run with notification setup, need to make a script to do backfill.

go/base/notification/types/types.go
```diff
@@ -13,95 +14,175 @@ import (
...
-	_defaultMaEmail                      = "michelangelo@uber.com"
-	_sourcePipelineTypeLabelName         = "michelangelo/SourcePipelineType"
-	_sourcePipelineManifestTypeLabelName = "pipeline.michelangelo/PipelineManifestType"
+	// PipelineRunNotificationWorkflowName is the registered name of the pipeline run
+	// notification workflow in Cadence/Temporal. Both the notifier (which starts the
+	// workflow) and the worker (which registers it) must use this constant.
+	//
+	// The name follows the reverse-DNS convention recommended by Temporal to avoid
+	// collisions in multi-tenant namespaces.
+	PipelineRunNotificationWorkflowName = "io.michelangelo.notification.PipelineRunFanout"
```
sallycr, line 25: this and DeprecatedPRNotificationWorkflowname is used outside of module. 

- PipelineRunNotificationWorkflowName — used in 3 packages:
  - types.go (definition)
  - notifier.go:115 (controller starts workflow)
  - module.go:49 (worker registers workflow)
- DeprecatedPRNotificationWorkflowName — used in module.go:50 (worker registers alias)

go/cmd/controllermgr/config/base.yaml
sallycr: https://github.com/michelangelo-ai/michelangelo/pull/1283/changes#r3410235886 default values are replaced by the config.

go/cmd/worker/config/base.yaml
sallycr: add future deprecated message.

go/cmd/worker/main.go
```diff
@@ -41,3 +44,13 @@ func RegisterRayPlugin(registry map[string]service.IPlugin) {
 func RegisterSparkPlugin(registry map[string]service.IPlugin) {
 	registry[sparkplugin.Plugin.ID()] = sparkplugin.Plugin
 }
+
+// RegisterNotificationActivities registers the default (no-op) email and Slack
+// notification activities. Downstream forks should replace this function with
+// their own transport implementations (SMTP, Slack API, PagerDuty, etc.).
+func RegisterNotificationActivities(workers []worker.Worker) {
+	for _, w := range workers {
+		w.RegisterActivity(notification.SendMessageToEmailActivity)
```
sallycr, line 60: @josephsahayaraj worker is moved to be controller here.

go/cmd/worker/config/base.yaml
sallycr: this is needed for the backward compatibility. added the comment for the deprecation.

helm/michelangelo/values.yaml
```diff
@@ -97,6 +97,7 @@ workflow:
     - default
     - trigger_run
     - pipeline_run
```
badcount, line 100: as discussed, we need to remove pipeline_run

go/cmd/controllermgr/config/base.yaml
```diff
@@ -26,6 +26,16 @@ ingester:
 pipelineRun:
   # ttlDays: 0  # 0 disables TTL eviction; set e.g. 30 to evict runs older than 30 days
 
+# Notification configuration for PipelineRun state-change workflows.
+# taskList must match the task list registered by the worker
+# (default: notification_worker — registered by the shared worker binary).
+# Leave taskList empty to disable notifications — the controller starts normally
+# but no notification workflows are dispatched.
+notification:
+  taskList: ${NOTIFICATION_TASK_LIST:""}
```
badcount: should remove this

sallycr: why you don't think this should be remove this?

go/components/pipelinerun/notification/notifier.go
sallycr: thanks @zhouxianzx

helm/michelangelo/values.yaml
sallycr: this is needed for the backward compatible. note above.

badcount (approved):

PR #1285, opened 2026-06-09
feat(example): California Housing XGBoost end-to-end UniFlow workflow

### LINKED ISSUE #1228: feat: [CanvasFlex Pusher PR8] California Housing end-to-end push example
## PR8 — California Housing End-to-End Example

Canonical 'how to use Phase 1 pusher' reference for external contributors. Demonstrates the full UniFlow workflow API with a clean, ethically sound dataset.

### Dataset

Uses the [California Housing dataset](https://scikit-learn.org/stable/datasets/real_world.html#california-housing-dataset) from scikit-learn (20,640 samples, 8 features). Replaces Boston Housing, which was removed by scikit-learn v1.2 over racial-bias concerns.

### Scope

`examples/california_housing_xgb/` — full UniFlow workflow split into per-step modules:

| File | Description |
|---|---|
| `california_housing_xgb.py` | Workflow entry point (`@uniflow.workflow`) |
| `feature_prep.py` | Ray task — load dataset, train/test split, Ray Datasets |
| `preprocess.py` | Spark task — cast columns to float |
| `train.py` | Ray task — distributed XGBoost training |
| `push_model.py` | Ray task — push model + eval report to storage and registry |

### Key design

- `DatasetVariable` reference-based data passing between tasks (no serialization of large datasets through the workflow engine)
- `MinioStorageBackend` for remote runs, `LocalStorageBackend` for local/CI — no external services required for local development
- `APIRegistryClient` when `REGISTRY_ENDPOINT` is set, `InMemoryRegistryClient` otherwise
- All stateful objects (storage clients, gRPC channels) constructed inside task body — required by UniFlow codec boundary
- `cache_enabled=True` on `feature_prep` and `preprocess` — skips re-execution when inputs are unchanged
- `with_overrides()` pattern demonstrated for per-run resource overrides
- `breakpoint=True` debugging pattern documented in README
- Post-push serving documented — `deployable_artifact_uri` explained with KServe/Seldon/BentoML as consumers

**Depends on:** #1220 (PR6)
**Part of:** CanvasFlex Pusher Phase 1
**PR:** #1282

sallycr:
- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Replaces the deprecated Boston Housing dataset (removed from sklearn v1.2 over racial-bias concerns) with California Housing, and adds a complete end-to-end UniFlow example with Ray and Spark tasks. Also includes several core library fixes discovered during end-to-end testing and OSS review.

**Example (`python/examples/california_housing_xgb/`):**
- New 3-step workflow: `feature_prep` (Ray) → `preprocess` (Spark) → `train` (Ray), split one task per file
- `__init__.py` is docstring-only with explanation — prevents eager SparkTask import in Ray containers
- Explicit sandbox-friendly memory limits: `SparkTask` (driver: 4G, executor: 2G) and `RayTask` (4Gi, 2 worker instances for `train` so `create_scaling_config`'s 50% CPU reserve has enough headroom)
- `entrypoint.sh` sources `/etc/environment` before `spark-submit` so `JAVA_HOME` resolves on arm64 k3d nodes
- `train.py` defers `ray`/`xgboost` imports into the function body — workflow orchestrator no longer loads them at module import time
- `preprocess.py` uses `schema.simpleString()` (public PySpark API, not private `_jdf` Py4J bridge)

**Core library fixes:**
- `spark/io.py`: `_ensure_s3a_config()` is now lazy (called at I/O time, not import) and idempotent (module-level flag); restored `fs.s3.impl` mapping so `s3://` URIs are handled by `S3AFileSystem` in Hadoop 3.x (required — without it, `s3://` raises `UnsupportedFileSystemException`); added `# type: ignore` on `_jsc` with link to follow-up issue #1286; fixed `SparkIO.read_data` None guard
- `dataset.py`: `_io_metadata` now passed through `super().__init__()` instead of double-written after the fact (issue #1287)
- `sandbox-bucket-setup.yaml`: adds `michelangelo` bucket required by `--storage-url s3://michelangelo/workflows`; fixes YAML parse error (stray `t` on `AWS_ENDPOINT_URL` key)

**Docs (`docs/`, `python/README.md`):**
- `getting-started.md`: fixed two broken code snippets (`seed`/`log` undefined in Step 1; `pr` undefined in Step 2); correct `--storage-url s3://michelangelo/workflows`; Java 17 prerequisite with `brew` install command; `k3d image import` step; troubleshooting bullet for `getSubject is not supported` (Java 21 + Hadoop 3.3 incompatibility)
- All Boston Housing references replaced across docs, sandbox YAMLs, and README
- Example `README.md` specifies Java 17 requirement

**Why?**

Boston Housing was removed from sklearn v1.2 due to racial-bias concerns. California Housing is the recommended replacement. The core fixes were discovered by end-to-end testing on a k3d sandbox and a follow-up 4-agent OSS quality review.

**How did you test it?**

- Local run: `JAVA_HOME=.../openjdk@17 PYTHONPATH=. poetry run python examples/california_housing_xgb/california_housing_xgb.py` ✅ (`validation-rmse: 0.876`)
- Remote run on k3d sandbox: `feature_prep` (Ray) → `preprocess` train+validation (Spark ×2) → `train` (Ray) all completed ✅ (`validation-rmse: 0.925`)
- Ruff lint and format pass ✅

**Potential risks**

- `DatasetVariable.__init__` change is backward-compatible (`_io_metadata` defaults to `None`; now initialized via `super().__init__()`)
- `_ensure_s3a_config()` idempotency flag is module-level; resets across process boundaries (fine — each task container is a new process)
- Memory limits in example tasks are sized for k3d sandbox; production runs use `SPARK_OVERRIDE_*` env vars or `with_overrides()` for production-appropriate values
- `fs.s3.impl` is restored intentionally — required for `s3://` URIs in Hadoop 3.x

**Release notes**

N/A

**Documentation Changes**

- `docs/user-guides/getting-started/getting-started.md` — rewritten for California Housing; fixed broken snippets; correct remote-run commands; Java 17 prereq; expanded troubleshooting
- `docs/intro.md`, `docs/user-guides/index.md`, `docs/getting-started/sandbox-setup.md`, `docs/user-guides/ml-pipelines/pipeline-management.md` — Boston Housing references replaced
- `python/README.md` — example link updated
- `python/examples/california_housing_xgb/README.md` — new example README with Java 17 requirement
- Follow-up issues for deferred items: #1286 (`_jsc` → `SparkTask.pre_run` injection), #1288 (`__all__` for `spark/io.py`)


python/michelangelo/cli/sandbox/resources/sandbox-bucket-setup.yaml
```diff
@@ -18,7 +18,7 @@ spec:
             - name: "AWS_ENDPOINT_URL"
               value: "http://minio:9091"
             - name: "BUCKET_NAMES"
-              value: "logs,default,deploy-models"
+              value: "logs,default,deploy-models,michelangelo"
```
badcount, line 21: why we're adding new bucket?

python/examples/entrypoint.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -e
 
+# Load auto-detected JAVA_HOME (written at image build time to handle arm64/amd64).
+# shellcheck disable=SC1091
+[ -f /etc/environment ] && . /etc/environment
+export JAVA_HOME PATH
+
```
badcount, line 9: what issue this is used to solve?

kenns29 (approved):

python/examples/entrypoint.sh
sallycr: java path are different for arm64 (local macOS sandbox) vs amd64.

python/michelangelo/cli/sandbox/resources/sandbox-bucket-setup.yaml
sallycr: instead of using default bucket, we can use michelangelo bucket. I changed the readme and leave default for the backward compatibility.

python/examples/entrypoint.sh
sallycr: Better answer from claude code

The ENV JAVA_HOME in the Dockerfile is hardcoded as /usr/lib/jvm/java-11-openjdk-${TARGETARCH}, where TARGETARCH defaults to amd64. On arm64 k3d nodes (Apple Silicon), this path doesn't exist, so spark-submit fails immediately with:

/usr/lib/jvm/java-11-openjdk-amd64/bin/java: No such file or directory

The Dockerfile already auto-detects the correct Java path at build time and writes it to /etc/environment:

RUN JAVA_HOME=$(dirname $(dirname $(readlink -f $(which java)))) && \
    echo "JAVA_HOME=${JAVA_HOME}" >> /etc/environment

But non-login shells (which is what spark-submit runs in) don't source /etc/environment automatically. The entrypoint line loads it explicitly so spark-submit uses the architecture-correct path.

Root cause in one line: ENV JAVA_HOME is a build-time constant that gets the wrong arch; /etc/environment has the runtime truth.

sallycr: adding TODO with github issue.

PR #1297, opened 2026-06-09
feat(pipelinerun): add status reconstruction from workflow engine

badcount:
Add automatic status recovery for PipelineRuns when status is lost due to etcd corruption, TTL eviction, or controller restarts. The reconciler now queries Cadence/Temporal to rebuild WorkflowId, WorkflowRunId, and State when status fields are empty.

Key changes:
- Add reconstructStatusFromWorkflowEngine() to query workflow engine
- Integrate reconstruction early in reconcile loop before condition engine
- Map workflow execution status to PipelineRun states
- Handle retry scenarios correctly by detecting run ID changes
- Comprehensive test coverage including retry edge cases

This ensures PipelineRuns can self-heal from status data loss without requiring manual intervention or workflow restarts.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


go/components/pipelinerun/controller.go
```diff
@@ -145,6 +146,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			pipelineRun.Status.WorkflowRunId = exec.Execution.RunID
+			pipelineRun.Status.State = mapWorkflowStatusToPipelineRunState(exec.Status)
+
+			// Set stable metadata storage primary key annotation if not already set
+			// This enables consistent identity across cluster migrations where K8s UIDs change
+			annotations := pipelineRun.GetAnnotations()
+			if annotations == nil {
+				annotations = make(map[string]string)
+			}
+			if _, exists := annotations[api.MetadataStoragePrimaryKeyAnnotation]; !exists {
```
aritra0227: why do we have to set this on the pipelinerun controller? Is this set by the ingester controller?

go/components/pipelinerun/controller.go
```diff
@@ -145,6 +146,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// MySQL primary key mismatches (MySQL uses UID as PK, workflow uses name as ID).
+	// Status reconstruction will work, but the MySQL metadata storage will have orphaned
+	// records from the old UID. Avoid recreating PipelineRuns with the same name.
+	if r.workflowClient != nil && (pipelineRun.Status.WorkflowId == "" || pipelineRun.Status.WorkflowRunId == "") {
+		workflowID := pipelineRun.Name
+		if exec, err := r.workflowClient.GetWorkflowExecutionInfo(ctx, workflowID, ""); err == nil && exec.Execution != nil {
+			// Restore workflow IDs and derive state from workflow engine
+			pipelineRun.Status.WorkflowId = workflowID
+			pipelineRun.Status.WorkflowRunId = exec.Execution.RunID
+			pipelineRun.Status.State = mapWorkflowStatusToPipelineRunState(exec.Status)
```
aritra0227, line 164: is this necessary? How do we currently populate the state?

aritra0227 (commented):
Left some comments.

Also from the PR summary:
> Handle retry scenarios correctly by detecting run ID changes

I'm not sure if the current code is handling the retries.

> Add reconstructStatusFromWorkflowEngine() to query workflow engine

I also do not see this function used anywhere

go/components/pipelinerun/controller.go
aritra0227: Nvm, I misunderstood this line. I'm assuming this is strictly in the scenario that the workflowID is nill but a workflow does exist in cadence/temporal

go/components/pipelinerun/controller.go
badcount: Good catch - you're right, this is redundant. The ingester controller already does this generically for every CRD kind registered in `v2pb.CrdObjects` (which includes `PipelineRun`): `ensureMetadataStoragePrimaryKey` in `go/components/ingester/controller.go` stamps `api.MetadataStoragePrimaryKeyAnnotation` with the object's UID on first reconcile before it's ever upserted into MySQL.

Removed the duplicate annotation-setting block from this controller - just left the workflow ID/state reconstruction, which is the actual purpose of this block.

aritra0227: I don't think the changes were pushed

aritra0227 (approved):

PR #1304, opened 2026-06-10
fix: register TypedStruct in ConnectRPC type registry

craigmarker:
- [x] Bug Fix

**What changed?**

The RPC transport now includes a type registry so the client can decode pipeline manifests that embed custom types inside protobuf `Any` fields. End-to-end test of `request` exercises the failure more. This required the unit test github action to generate RPC files. Exclude these from coverage metrics.

**Why?**

When the Pipeline Runs table loaded, the browser threw a parse error decoding the manifest content of any run whose pipeline had a structured config. The root cause: the transport's JSON decoder had no knowledge of custom types, so it failed the moment it encountered one embedded in an `Any` field. The table rendered an error state instead of data.

**How did you test it?**

**Reproduction & fix**

https://github.com/user-attachments/assets/9e86d01c-0f18-4b59-96c6-bb122085f79f

Unit test added to confirm the decoder throws without the registry and succeeds with it.


javascript/packages/core/components/actions/__tests__/actions-popover.test.tsx
```diff
@@ -384,7 +384,9 @@ describe('ActionsPopover', () => {
       expect(await screen.findByText('Cannot archive')).toBeInTheDocument();
       await user.hover(screen.getByRole('option', { name: 'Delete' }));
       expect(await screen.findByText('Cannot delete')).toBeInTheDocument();
-      expect(screen.queryByText('Cannot archive')).not.toBeInTheDocument();
+      await waitFor(() => {
```
craigmarker, line 387: just a flaky test fix

donovanlopez (approved):

PR #1309, opened 2026-06-12
refactor: migrate CreatePipelineRun to declarative MutationActionConfig

craigmarker:
- [x] Refactor

**What changed?**

The pipeline Run action is now expressed as declarative config rather than a custom form component. The form and its test file are deleted; the action is described directly in the pipeline entity config using a mutation operation with middleware that builds the run payload from the pipeline record and routing context.

Two supporting changes in the middleware layer make this possible. First, a new `startEmpty` option lets the middleware produce a fresh payload object instead of starting from a clone of the action record — necessary here because the action record (a Pipeline) and the mutation payload (a PipelineRun) are different types, and without it the pipeline's own fields would leak into the request. Second, a source operation without a transformation now performs a plain copy to the destination; previously it was a silent no-op, which was a gap rather than intended behavior.

**Why?**

Like the kill trigger run migration, the form existed only because the framework couldn't express "fire a mutation behind a confirm dialog with a derived payload" declaratively. The middleware layer now covers everything the form was doing — routing context via function defaults, record field copying via source operations, and generated names via inline functions. Keeping the form would leave two parallel ways to write the same kind of action.

The `startEmpty` option generalizes cleanly to any future CREATE action where the record that triggers the action is a different type from what the mutation expects.

**How did you test it?**

The new `pipeline.test.tsx` covers the happy path end-to-end: opens the popover, confirms, asserts the mutation was called with the correct run name pattern, namespace, actor, and pipeline reference. Also covers the error path. The middleware test for the plain-copy behavior was updated to assert the new (correct) behavior. Full test suite passes.

**Potential risks**

None expected. The `startEmpty` and plain-copy changes only affect callers that opt in — no existing middleware schemas use `startEmpty`, and no existing schemas rely on the silent no-op behavior of source-without-transformation (the one test that documented it was testing the gap, not a feature).

**Release notes**

Pipeline Run action migrated from a custom form component to declarative config. Middleware gains `startEmpty` for CREATE-style actions and plain-copy support for source operations.

**Documentation Changes**

N/A.


craigmarker:
Reverting — pipeline run migration requires FormModalConfig support first. Will revisit once that's in place.

PR #761, opened 2026-01-30
Add notebook support using Uniflow

badcount:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Add notebook support using Uniflow

**Why?**
For supporting databrick notebook  solution

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

PR #1315, opened 2026-06-13
Fix/resilient cache sync

badcount:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

What changed?
Added a resilient cache wrapper (go/controllermgr/resilientcache/) that enforces a 2-minute timeout on WaitForCacheSync. If the timeout expires, the controller manager proceeds in degraded mode instead of blocking forever. A timedOut flag ensures subsequent sync checks from per-controller source watchers return immediately. Wired into the manager via the NewCache option — one line in module.go.

Why?
A single malformed CRD object (e.g. a Deployment with shadowSpec.duration: "99999999999999s") causes the Deployment informer's List to fail on JSON unmarshal (time.ParseDuration overflows for values > 292 years). Since all controllers share one cache and WaitForCacheSync blocks until every informer syncs, this one bad object prevents every controller (PipelineRun, InferenceServer, RayJob, TriggerRun, etc.) from starting — not just Deployment. This was the root cause of the incident on 2026-06-12.
                                                                                                                                                                                
  How did you test it?                                                                                                                                                          
  1. Applied a poisoned Deployment with shadowSpec.duration: "99999999999999s" to the sandbox cluster                                                                           
  2. Ran the controller manager locally (bazel run //go/cmd/controllermgr)                                                                                                      
  3. Before fix: verified all controllers blocked forever — no Starting workers log, no reconciliation for any resource type                                                    
  4. After fix: verified the WARNING is logged after 2 minutes, all controllers start their workers, and a newly created PipelineRun is reconciled successfully while the Deployment informer continues retrying in the background                                                                                                                      
  5. Verified recovery: after deleting the bad Deployment, the Deployment informer syncs and reconciliation resumes                                                             

Potential risks                                                                                                                                                               
  - Controllers whose informer failed to sync will operate in degraded mode: individual Get()/List() calls may return stale or empty results until the informer recovers. This  
  is strictly better than the current behavior where ALL controllers are permanently dead.                                                                                      
  - The 2-minute timeout is hardcoded. In production, this means a ~2 minute delay at startup before unaffected controllers begin processing if any CRD has a bad object.       
                                                                                                                                                                                
  Release notes                                                                                                                                                                 
  No schema, configuration, or data migration changes. The controller manager now tolerates informer sync failures at startup instead of blocking indefinitely.                 
                                                                                                                                                                                
  Documentation Changes                                                                                                                                                         
  None. No user-facing or API changes. The fix is internal to controller manager startup behavior.

### DIFF
--- a/go/controllermgr/BUILD.bazel
+++ b/go/controllermgr/BUILD.bazel
@@ -11,6 +11,7 @@ go_library(
     deps = [
         "//go/base/blobstore:go_default_library",
         "//go/base/blobstore/minio:go_default_library",
+        "//go/controllermgr/resilientcache:go_default_library",
         "//go/kubeproto/metrics:go_default_library",
         "@com_github_uber_go_tally//:go_default_library",
         "@io_k8s_apimachinery//pkg/runtime:go_default_library",
--- a/go/controllermgr/module.go
+++ b/go/controllermgr/module.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/michelangelo-ai/michelangelo/go/base/blobstore"
 	"github.com/michelangelo-ai/michelangelo/go/base/blobstore/minio"
+	"github.com/michelangelo-ai/michelangelo/go/controllermgr/resilientcache"
 	"github.com/michelangelo-ai/michelangelo/go/kubeproto/metrics"
 )
 
@@ -77,6 +78,7 @@ func create(p params) (result, error) {
 		HealthProbeBindAddress: p.Config.HealthProbeBindAddress,
 		LeaderElection:         p.Config.LeaderElection,
 		LeaderElectionID:       p.Config.LeaderElectionID,
+		NewCache:               resilientcache.NewCacheFunc(),
 	})
 	if err != nil {
 		return result{}, err
--- a/go/controllermgr/module.go
+++ b/go/controllermgr/module.go
@@ -78,7 +78,7 @@ func create(p params) (result, error) {
 		HealthProbeBindAddress: p.Config.HealthProbeBindAddress,
 		LeaderElection:         p.Config.LeaderElection,
 		LeaderElectionID:       p.Config.LeaderElectionID,
-		NewCache:               resilientcache.NewCacheFunc(),
+		NewCache:               resilientcache.NewCacheFunc(ctrl.Log.WithName("cache")),
 	})
 	if err != nil {
 		return result{}, err
--- a/go/controllermgr/resilientcache/BUILD.bazel
+++ b/go/controllermgr/resilientcache/BUILD.bazel
@@ -1,0 +1,12 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["resilientcache.go"],
+    importpath = "github.com/michelangelo-ai/michelangelo/go/controllermgr/resilientcache",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@io_k8s_client_go//rest:go_default_library",
+        "@io_k8s_sigs_controller_runtime//pkg/cache:go_default_library",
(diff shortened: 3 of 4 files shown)


paulzim (changes requested):
## Review: PR #1315 — Fix/resilient cache sync

**Verdict: Request Changes (two correctness issues; pending CLA + rebase)**

---

### Context

This fixes the June 12 incident where a single malformed Deployment (`shadowSpec.duration: "99999999999999s"` — `time.ParseDuration` overflows for values > ~292 years) caused the Deployment informer's List to fail on unmarshal, which in turn blocked every controller in the shared manager forever. The fix is well-scoped and the design is sound.

### Net change

Branch was cut in mid-June; the diff against `upstream/main` looks enormous (stale-merge-base noise) but the actual change from the branch point is 4 files, 67 insertions, 0 deletions.

### Design review

The structure is correct:

- `resilientCache` embeds `cache.Cache`, overriding only `WaitForCacheSync`.
- A 2-minute child context gates the blocking call; on timeout, a `sync.Once` logs the warning once and sets a `timedOut` flag.
- Subsequent calls from per-controller informer watches check `timedOut` first and return immediately — this is the right pattern to avoid re-blocking each controller individually after the initial timeout.
- Wiring in `module.go` is a single line (`NewCache` option), no structural changes.
- BUILD.bazel deps are correct.

### Fix 1 (required): `timedOut` has a data race — use `atomic.Bool`

`timedOut` is written inside `once.Do()` and read at the top of `WaitForCacheSync` without any synchronization. These are unsynchronized accesses from potentially different goroutines — `go test -race` would flag this. Go 1.24 has `sync/atomic.Bool` which makes this a one-line fix:

```go
type resilientCache struct {
    cache.Cache
    logger   logr.Logger
    once     sync.Once
    timedOut atomic.Bool
}

// Read: c.timedOut.Load()
// Write: c.timedOut.Store(true)
```

### Fix 2 (required): Parent context cancellation triggers a false "timed out" log and wrong return value

If the manager shuts down during the 2-minute window, the parent `ctx` is cancelled, which cancels `syncCtx` too. `c.Cache.WaitForCacheSync(syncCtx)` returns false — not because of a bad CRD, but because of normal shutdown. Two problems follow: the error log fires spuriously, and the function returns `true`, telling the manager "cache is synced, proceed" when it should be propagating the cancellation so the manager can stop cleanly. The fix is to distinguish deadline expiry from parent cancellation:

```go
if !c.Cache.WaitForCacheSync(syncCtx) {
    if syncCtx.Err() == context.DeadlineExceeded {
        c.once.Do(func() {
            c.timedOut.Store(true)
            c.logger.Error(nil, "cache sync timed out ...", "timeout", syncTimeout)
        })
    }
    // Parent context cancelled (normal shutdown) — return false so the manager
    // can propagate the shutdown correctly rather than masking it with true.
    return syncCtx.Err() == context.DeadlineExceeded
}
```

Note: returning `false` on parent cancellation is also more correct — the manager interprets a false return during shutdown as a clean stop, not an error.

### No tests

The new package ships with no `_test.go`. A table-driven test with a mock `cache.Cache` whose `WaitForCacheSync` blocks until cancelled would cover: (1) timeout triggers degraded mode, (2) subsequent calls return immediately, (3) healthy sync still works. Given this fixes a production incident, test coverage would be valuable. Not a hard blocker given the controllermgr package has no existing test files to establish a pattern, but worth adding.

### CI and CLA

All CI checks pass (bazel-build, dirty-check, Go Test & Coverage, python-build, Go Lint). CLA is unsigned — same situation as #761; Eric will need to sign before this can merge.

---

The fix is the right approach — degraded mode with a bounded timeout is strictly better than indefinite blocking. The two required fixes are small and clearly specified; after those, the rebase, and the CLA, this is ready.

PR #1316, opened 2026-06-14
feat(cascadedelete): Pipeline cascade-delete consumers, wiring, CLI, docs

ishtoo1:
Consumers + wiring for cascade delete, built on the `go/cascadedelete` package (#1266, merged to `main`).

- **PipelineRun / TriggerRun controllers** drain via `cascadedelete.RunDrainStep` + thin per-kind adapters; ownerRefs stamped at create via API hooks; TriggerRun timeout teardown via `ForceKillWorkflow`.
- **ingester** routes retain-vs-delete via an injected `RetainPolicy`, scoped to opted-in kinds (retained children keep their final state in MySQL instead of being soft-deleted).
- **controllermgr** provides the `RetainPolicy` + registers metrics; **apiserver** registers the create hooks.
- **`ma pipeline delete`**: foreground propagation + unconditional confirm.
- docs + CHANGELOG updated; invariant + ingester-branch unit tests added.

### Validation
- **Unit**: `ingester` / `pipelinerun` / `triggerrun` / `apiserver` / `controllermgr` `go_default_test` — 5/5 pass; CLI `delete_test.py` + `main_test.py` — 13/13 pass.
- **Live (k3d, metadata storage on)**: cascade RETAIN verified for PipelineRun + TriggerRun (gettable from MySQL, gone from etcd, `delete_time IS NULL`); live Spark kill chain verified high-fidelity — SparkApplication RUNNING → foreground delete → `SparkJob Killed=TRUE` → SparkApplication deleted → run retained terminal.
- **Integration tests**: michelangelo-ai/integration-test#5.

### Follow-up
- #1337 — remove the reconciler `ensureOwnerRef` migration backfill once CRs predating the `BeforeCreate` apihook have aged out.


go/components/pipelinerun/controller.go
```diff
@@ -181,6 +237,160 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return result, returnErr
 }
 
+// backfillOwnerRef stamps the owning Pipeline as the run's controller
+// ownerReference so GC cascades Pipeline deletes to it. Idempotent; a no-op when
+// the Pipeline or its reference is absent.
+func (r *Reconciler) backfillOwnerRef(ctx context.Context, logger *zap.Logger, pipelineRun *v2pb.PipelineRun) error {
```
badcount: Should we have todo here to deprecate backfill later on?

ishtoo1: The create hook only stamps runs made through the apiserver; backfill is the steady-state catch-all for runs created out-of-band (kubectl apply/direct k8s API) that bypass it. I can update the comment and the method name here to avoid any confusions. Does that sound good?

badcount (approved):

PR #1323, opened 2026-06-16
Add eslint-plugin-react with structural component rules

craigmarker:
- [x] Refactor

**What changed?**

Adds `eslint-plugin-react` and enables three rules that enforce how components are structured: one component per file, no components defined inline in JSX, and no inline object literals passed to Context Providers.

Relocated FormControl Label private component to dedicated file.

**Why?**

These patterns were already followed by convention but had no enforcement. The goal is to reduce surface area required to be annotated in agent skills/agent md files.

FormControl Label relocated rather than eslint ignored because logic is substantial and there is already a dedicated label/ directory within the form-control component.

**How did you test it?**

Lint passes

**Potential risks**

Low. The two `useMemo` additions are correctness fixes with no functional side effects.

### DIFF
--- a/javascript/packages/core/components/form/components/label/label-end-enhancer-content.tsx
+++ b/javascript/packages/core/components/form/components/label/label-end-enhancer-content.tsx
@@ -1,0 +1,32 @@
+import { useStyletron } from 'baseui';
+
+import type { FormControlProps } from '#core/components/form/components/types';
+
+/**
+ * Renders content intended for FormControl's label end enhancer slot. Includes counter
+ * display and/or arbitrary enhancer content, side-by-side.
+ *
+ * **Relies on parent FormControl's LabelEndEnhancer override for flex layout**
+ */
+export const LabelEndEnhancerContent = ({
+  counter,
+  labelEndEnhancer,
+}: Pick<FormControlProps, 'counter' | 'labelEndEnhancer'>) => {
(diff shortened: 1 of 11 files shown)


badcount (approved):

javascript/packages/core/components/form/components/label/label-end-enhancer-content.tsx
donovanlopez, line 1: Regarding file organization, since we put the `LabelEndEnhancerContent` component into its own file, should it have its own folder? I noticed it's in the `/label` folder but the other components have dedicated folders. 

For example, [/sticky-footer](https://github.com/michelangelo-ai/michelangelo/tree/f2e9917ea7fb4e863a3af002166270c452b769ff/javascript/packages/core/components/form/components/sticky-footer) mainly houses the `StickyFooter` component as its main export.

donovanlopez (approved):
LGTM. Left a comment that's not necessarily blocking this PR.

javascript/packages/core/components/form/components/label/label-end-enhancer-content.tsx
craigmarker: it's a good question ... I don't have a great answer. My instinct suggests the issue is more that form-control should be its own directory that contains form-control and the label-end-enhancer-content. It's a little tricky, since Label is used outside of form control for rows.

let's keep an eye on it

PR #1325, opened 2026-06-16
Add require-handler-prefix; enforce handle* prefix on locally-defined handlers

craigmarker:
- [x] Refactor

**What changed?**

`no-handler-mirror` landed in #1326 as part of a stacked PR. This PR picks up where that left off, completing the two-rule convention for event handler naming.

Adds a new custom rule, `require-handler-prefix`, which requires any locally-defined function passed to an `on*` JSX prop to begin with `handle`. Passthrough props — where a callback received from outside is forwarded directly to a child — are exempt, because the `handle*` naming already applies at the call site where the handler was defined. Test files are also exempt; mocks are better named after the prop they stand in for.

`react/jsx-handler-names` was considered but fires on both passthrough props and test mocks, which forces aliases and renames that add noise without improving clarity.

Together `no-handler-mirror` and `require-handler-prefix` rules enforce a complete convention:

- `onChange={handleChange}` — rejected by `no-handler-mirror` (mirrors the prop, adds no context)
- `onChange={persistSelection}` — rejected by `require-handler-prefix` (no `handle` prefix)
- `onChange={handleSelectionChange}` — ✅

**Why?**

Requiring `handle` as a prefix gives a uniform, greppable signal that an identifier is an event handler. Exempting passthroughs avoids aliases that communicate nothing.

**How did you test it?**

I ran `yarn lint` — 0 errors in `packages/core`. 1321 tests pass.

**Potential risks**

Low. The rule applies only to source files, not tests. Passthrough props are exempt. Existing violations were addressed in this PR.

### DIFF
--- a/javascript/eslint-local-rules/no-handler-mirror.md
+++ b/javascript/eslint-local-rules/no-handler-mirror.md
@@ -36,13 +36,13 @@ This rule works in tandem with `react/jsx-handler-names`, which requires all han
 
 Name the **effect**, not the trigger. The `handle` prefix is required (enforced by `react/jsx-handler-names`), and the suffix should describe what the handler does to application state.
 
-| Context | Instead of | Use |
-|---------|-----------|-----|
+| Context                                 | Instead of     | Use                     |
+| --------------------------------------- | -------------- | ----------------------- |
 | Radio coerces DOM string to typed value | `handleChange` | `handleSelectionChange` |
 | Select maps option objects to typed IDs | `handleChange` | `handleCommitSelection` |
-| Checkbox stops propagation + toggles | `handleChange` | `handleToggleSelection` |
-| Form submit that also closes the dialog | `handleSubmit` | `handleFormSubmit` |
-| Row delete on click | `handleClick` | `handleRowDelete` |
+| Checkbox stops propagation + toggles    | `handleChange` | `handleToggleSelection` |
+| Form submit that also closes the dialog | `handleSubmit` | `handleFormSubmit`      |
+| Row delete on click                     | `handleClick`  | `handleRowDelete`       |
 
 The name should answer "what does this function do to the application state?" — not "what event is it responding to?"
 
--- a/javascript/eslint-local-rules/no-handler-mirror.md
+++ b/javascript/eslint-local-rules/no-handler-mirror.md
@@ -32,19 +32,7 @@ This rule works in tandem with `require-handler-prefix`, which requires all loca
 <Child onClick={props.onClick} />
 ```
 
-## Naming guidance
-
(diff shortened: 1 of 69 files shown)


badcount (approved):

javascript/eslint-local-rules/no-handler-mirror.md
donovanlopez, line 1: I agree that names like `handleChange` and `handleOnClick` aren't all that helpful but I think the `handle` prefix is an easy way to distinguish that we're looking at an event handler. I think it's also a pretty common naming convention (but don't take this as "If everyone else is doing it, then we should too" but as someone that's used to the `handle` prefix, it's more intuitive at a glance). So doing `handleCommitSelection` rather than `commitSelection` would be a way to have a more descriptive name. I'm not really sure we can enforce that through a lint rule.

donovanlopez (commented):
I think i see file changes from other PRs. Also left a comment about the rule.

javascript/eslint-local-rules/no-handler-mirror.md
craigmarker: thanks for the critical eye and thoughtful perspective! to clarify, `handleCommitSelection` would pass this lint rule.

I see how it's valuable to denote that a function is an event handler, ie requires wiring to a DOM event. That said, I think that information is not lost by dropping the `handle` prefix—the function signature & how it is invoked still carry that.

We could modify documentation to be less critical about the `handle` prefix, even add a test case to no-event-handler-prefix.test.js that clarifies `handleCommitSelection` is valid event handler name for `onChange`.

what do you think?

javascript/eslint-local-rules/no-event-handler-prefix.js
```diff
@@ -3,7 +3,7 @@
  *
  * onClick={onClick}, onChange={handleChange}, onClick={handleOnClick} all tell
  * the reader nothing about *what* is being handled. Descriptive names like
- * onChange={handleRowChange} make the intent clear at the call site.
+ * onChange={handleRowChange} or onChange={commitSelection} make the intent clear.
```
donovanlopez, line 6: What are your thoughts on enforcing only one way (either with or without the `handle` prefix)? Whichever way we go, I think we should only allow one naming pattern so we can have uniformity across the codebase. I'm leaning toward keeping the `handle` prefix as it's a common industry convention.

craigmarker: SGTM, I actually see that there is an existing eslint rule in eslint-plugin-react that gives us this: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md

donovanlopez: Nice i love it when things are given to us for free

donovanlopez (approved):

donovanlopez (approved):

PR #1330, opened 2026-06-17
fix: atomic update to prevent Temporal schedule race condition

badcount:
When updating both a cron schedule and pause/resume action simultaneously, the controller made two separate Temporal API calls (schedule.Update for cron, then schedule.Pause/Unpause for action). These calls raced at the Temporal server level, causing ~2.5% of operations to lose the pause/resume state. This merges both into a single atomic schedule.Update() call by passing the desired paused state into the DoUpdate callback alongside the cron change.

Verified: 0 failures across 200 stress-test iterations (vs 5/200 before fix).

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

<!-- Describe what has changed in this PR -->
                                                                                                                                                                                         
  Merged the Temporal schedule.Update() (cron change) and schedule.Pause()/schedule.Unpause() (action change) into a single atomic API call when both are requested simultaneously.      
                                                                                                                                                                                         
  - Added paused *bool parameter to UpdateTrigger in the WorkflowClient interface                                                                                                        
  - The Temporal implementation now sets State.Paused inside the DoUpdate callback alongside the cron expression change                                                                  
  - Runner.Update() now accepts the pending action and returns whether it was handled atomically, so the controller skips the separate Pause()/Unpause() call when the action was already
  folded into the cron update                                                                                                                                                            
  - When only an action OR only a cron change is requested (no overlap), the existing separate code paths are used — no behavioral change                                                
                                                                                                                                                                                         
  Why?                                                                                                                                                                                   
                                                                                                                                                                                         
  When a user updates both the crontab and the run action (e.g. PAUSE) on a TriggerRun simultaneously, the controller made two sequential Temporal API calls: schedule.Update() for the  cron change, then schedule.Pause() for the action. These two calls race at t Jump to bottom (ctrl+End) ↓ hedule.Update() does a full schedule replace (including the State.Paused field), which can overwrite or conflict with the subsequent Pause()/Unpause() call. This caused ~2.5% of simultaneous cron+action updates to silently lose the pause/resume state while the cron change succeeded.                                                                                                                                                             
        
<!-- How have you verified this change? Tested locally? Added a unit test? Checked in staging env? -->
**How did you test it?**
                                                                                                                                                                                 
  1. Unit tests pass: bazel test //go/components/triggerrun:go_default_test — all existing + updated tests pass                                                                          
  2. Reproduced the race condition before the fix: ran a 200-iteration stress test alternating cron+PAUSE and cron+RESUME via kubectl apply, verified against Temporal HTTP API          
  (/api/v1/namespaces/default/schedules/...). Result: 5 failures out of 200 iterations (2.5%)                                                                                            
  3. Verified the fix: same 200-iteration stress test with the fix deployed. Result: 0 failures out of 200 iterations                                                                    
  4. Each iteration verified both the cron step value and the paused state directly from the Temporal API (not just the TriggerRun CR status)                                            
                         

<!-- Assuming the worst case, what can be broken when deploying this change to production? -->
**Potential risks**
 
  implementations of WorkflowClient will need to add the parameter.                                                                                                                      
  - When there is no cron drift (only an action change), the controller still uses the separate Pause()/Unpause() path — this is the same behavior as before and is race-free since      
  there's no competing Update() call.      

<!-- Is it notable for release? e.g. schema updates, configuration or data migration required? If so, please mention it, and also update CHANGELOG.md -->
**Release notes**

<!-- Does this PR introduce a user-facing or API change? Is user document updated? Is the change backward compatible? If so please update in wiki https://github.com/michelangelo-ai/michelangelo/wiki? -->
**Documentation Changes**

### DIFF
--- a/go/components/triggerrun/controller.go
+++ b/go/components/triggerrun/controller.go
@@ -204,14 +204,6 @@ StateMachine:
 	case v2pb.TRIGGER_RUN_STATE_RUNNING:
 		log.Info("TRIGGER_RUN_STATE_RUNNING")
 
-		// Sync TriggerRun spec changes to workflow engine
-		if status, err := runner.Update(ctx, triggerRun); err != nil {
-			log.Error(err, "failed to sync trigger spec to workflow engine")
-			triggerRun.Status.ErrorMessage = err.Error()
-			triggerRun.Status.State = status.State
-			break StateMachine
-		}
-
 		// Handle actions using the new action field (preferred) or deprecated boolean fields (backward compatibility)
 		actionToPerform := triggerRun.Spec.Action
 
@@ -222,9 +214,27 @@ StateMachine:
 			}
 		}
 
+		// Sync TriggerRun spec changes to workflow engine, passing the action so cron
+		// update and pause/resume can be applied atomically in a single API call.
+		status, actionHandled, err := runner.Update(ctx, triggerRun, actionToPerform)
+		if err != nil {
+			log.Error(err, "failed to sync trigger spec to workflow engine")
+			triggerRun.Status.ErrorMessage = err.Error()
+			triggerRun.Status.State = status.State
+			break StateMachine
+		}
+		triggerRun.Status = status
+
+		// If Update already applied the action atomically, clear and skip
+		if actionHandled {
+			triggerRun.Spec.Action = v2pb.TRIGGER_RUN_ACTION_NO_ACTION
+			triggerRun.Spec.Kill = false
+			break StateMachine
+		}
+
 		switch actionToPerform {
 		case v2pb.TRIGGER_RUN_ACTION_KILL:
-			status, err := runner.Kill(ctx, triggerRun)
+			status, err = runner.Kill(ctx, triggerRun)
 			if err != nil {
 				log.Error(err, "failed to kill scheduled workflow")
 				triggerRun.Status.ErrorMessage = err.Error()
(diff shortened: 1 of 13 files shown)


go/components/triggerrun/controller.go
```diff
@@ -222,9 +214,27 @@ StateMachine:
...
+		triggerRun.Status = status
+
+		// If Update already applied the action atomically, clear and skip
+		if actionHandled {
+			triggerRun.Spec.Action = v2pb.TRIGGER_RUN_ACTION_NO_ACTION
+			triggerRun.Spec.Kill = false
+			break StateMachine
+		}
+
 		switch actionToPerform {
```
sishilong, line 235: this (RUNNING) state's switch block has only KILL and PAUSE cases but no RESUME case. A TRIGGER_RUN_ACTION_RESUME on an already-running trigger (a user mistake, or after transient state) will:
- Call Update() with the RESUME action
- If there is no cron drift, Update() returns actionHandled=false
- The switch finds no matching case, falls through to GetStatus()

triggerRun.Spec.Action is never cleared — RESUME persists forever in the spec, causing an infinite retry on every reconcile

badcount: user can't do resume from running and pass from already pause. This is original behavior, i didn't change it. I will create a separate one to fix so we can discuss there

sishilong (approved):

PR #1331, opened 2026-06-17
cli: add [plugin.packages] discovery for import-based plugin loading

yaoyueyue:
## Summary

Adds a parallel plugin discovery mechanism that loads plugins **by importable module path** instead of by filesystem path. New TOML key:

```toml
[plugin]
packages = ["my_pkg.plugins"]   # importable package containing entity/<crd>/main.py
```

Both `[plugin.dirs]` (existing) and `[plugin.packages]` (new) are scanned by `discover_all_plugins`; results accumulate into the same registry. No behavior change for existing consumers — the new key defaults to `[]`.

## Motivation

The existing `[plugin.dirs]` loader has two structural limitations that show up the moment a downstream distribution ships plugins inside a wheel or PEX:

1. **PEX / zipimport incompatibility.** Discovery uses `Path.exists()` + `Path.iterdir()` on a real filesystem directory. Inside a PEX, bundled `plugins/` is a zip entry — `Path.exists()` returns False and the loader silently skips every bundled plugin.

2. **No `__package__` on loaded plugins.** `read_module_from_file` uses `spec_from_file_location` with no `__package__` set. Python then refuses relative imports (`from . import apply`, `from ...git_validation import X`) inside the plugin. Worse, plugins shipping in multiple namespaces (e.g. `my_pkg.plugins.entity.foo` at pip-install vs `org.team.product.my_pkg.plugins.entity.foo` inside a monorepo's bazel sandbox) have to manage that themselves with dual-namespace try/except imports.

The package-based loader fixes both:

- Uses `importlib.import_module` + `pkgutil.iter_modules` — both honor zipimporter, so PEX and wheel-bundled plugins are discoverable.
- Loaded plugin has a real `__package__`, so relative imports just work and plugin authors don't need to know their plugin's absolute name.

## Design

- `read_module_from_package(crd_name, package_name)` — imports `{package_name}.entity.{crd_name}.main` and returns it. Returns None on `ModuleNotFoundError` (different packages contain different CRDs — expected) or any other plugin-load exception (resilience — one broken plugin can't crash the CLI).
- `_iter_entity_names_in_package(package_name)` — uses `pkgutil.iter_modules` over `{package_name}.entity` to enumerate entity subpackages without touching the filesystem.
- `discover_all_plugins` — after scanning `[plugin.dirs]` as today, also iterates `[plugin.packages]` and appends discovered plugins to the same registry.

## What this unblocks (referenced consumers)

A downstream CLI that wraps `ma` and ships its plugins inside a pip wheel currently has to:

- Maintain dual-namespace try/except imports in every plugin's `main.py` and `apply.py` (e.g. `import my_pkg.git_validation` with a fallback to `from org.team.product.my_pkg import git_validation`).
- Add `# gazelle:ignore` directives in build-system tooling because the pip namespace doesn't exist as a build target.
- Keep a path-based `_load_sibling()` helper because `from . import apply` can't resolve.
- Lose the option to package itself as a PEX because the bundled plugins go undiscovered.

Switching that CLI's config from `[plugin.dirs]` to `[plugin.packages]` lets it drop all of the above. The cleanup is purely consumer-side after this lands.

## Test plan

| command | run | verified |
|---|---|---|
| `pytest michelangelo/cli/mactl/mactl_test.py -k 'ReadModuleFromPackage or IterEntityNamesInPackage or DiscoverAllPluginsFromPackages'` | yes | 13/13 pass |
| `pytest michelangelo/cli/mactl/mactl_test.py` (full suite, including existing `DiscoverAllPluginsTest`) | yes | no regressions vs main |
| `ruff check python/michelangelo/cli/mactl/` + `ruff format --check python/michelangelo/cli/mactl/` | yes | clean |
| **ETE hot-patch:** built a local-tagged wheel from this branch, `pip install --force-reinstall` over an existing `michelangelo-0.1.1.58` install, pointed a downstream wrapper CLI's `~/.ma/config.toml` at `[plugin] dirs=[] packages=["downstream_pkg.plugins"]`, ran `LOG_LEVEL=info ma --help` | yes | all 5 wrapper-CLI plugins (`pipeline`, `project`, `prompt_template`, `ray_job`, `spark_job`) registered via the new path; log shows `Registered plugin for entity 'X' from package 'downstream_pkg.plugins'`; loaded modules have real `__package__` (e.g. `downstream_pkg.plugins.entity.pipeline`); `ma --help` dispatch table includes every plugin CRD |

New coverage:
- Successful package import + module return
- Missing package returns None (not raised)
- Existing package without entity subpackage returns None
- Plugin whose top-level code raises is skipped (resilience)
- Relative imports inside a package-loaded plugin resolve correctly (the structural win)
- Entity enumeration via `pkgutil.iter_modules` filters non-package files
- Entity subpackage whose top-level code raises is skipped (covers broad-Exception branch)
- Entity resolved to a single-file module (no `__path__`) returns `[]`
- `discover_all_plugins` end-to-end with packages
- `[plugin.dirs]` and `[plugin.packages]` accumulate into one registry
- Missing `packages` config key behaves identically to `packages = []`

### DIFF
--- a/python/michelangelo/cli/mactl/mactl.py
+++ b/python/michelangelo/cli/mactl/mactl.py
@@ -5,6 +5,7 @@ A command line interface to interact with the Michelangelo API server via gRPC.
 
 import importlib
 import logging
+import pkgutil
 import re
 import sys
 from argparse import ArgumentParser, Namespace
@@ -126,6 +127,75 @@ def read_module_from_file(
         return None
     _LOG.info("Loaded plugin module: %r", plugin_module)
     return plugin_module
+
+
+def read_module_from_package(
+    crd_name: str, package_name: str
+) -> Union[object, None]:
+    """Read and load a plugin module via the import system, not by file path.
+
+    Looks for ``{package_name}.entity.{crd_name}.main`` and returns it on
+    success.
+
+    Why prefer this over :func:`read_module_from_file`:
+
+    * Works inside PEX / zipimport bundles (no ``Path.exists()`` filesystem
+      check on a synthetic zip path).
+    * The loaded module has a real ``__package__``, so relative imports
+      (``from . import apply``, ``from ...git_validation import X``) work
+      inside the plugin.
+    * Plugins do not have to know their own absolute module name —
+      consumers register the parent package once and discovery walks it.
+    """
+    module_name = f"{package_name}.entity.{crd_name}.main"
+    _LOG.info("Importing plugin module: %r", module_name)
+    try:
+        return importlib.import_module(module_name)
+    except ModuleNotFoundError as e:
+        # The whole point of multi-package discovery is that not every
+        # registered package contains every CRD. Log and move on.
+        _LOG.debug("Plugin module %r not found: %s", module_name, e)
+        return None
+    except Exception as e:
+        _LOG.error(
+            "Failed to import plugin module %r (skipped): %s",
+            module_name,
+            e,
+            exc_info=True,
+        )
+        return None
+
+
+def _iter_entity_names_in_package(package_name: str) -> list[str]:
+    """Yield entity names from ``{package_name}.entity`` via the import system.
+
+    PEX/zipimport-safe: uses :func:`pkgutil.iter_modules` rather than
+    :func:`Path.iterdir`. Returns an empty list if the entity subpackage does
+    not exist or cannot be imported.
+    """
+    entity_pkg_name = f"{package_name}.entity"
+    try:
+        entity_pkg = importlib.import_module(entity_pkg_name)
+    except ModuleNotFoundError:
+        _LOG.debug("Entity subpackage not found: %r", entity_pkg_name)
+        return []
+    except Exception as e:
+        _LOG.error(
+            "Failed to import entity subpackage %r (skipped): %s",
+            entity_pkg_name,
+            e,
+            exc_info=True,
+        )
+        return []
+
+    paths = getattr(entity_pkg, "__path__", None)
+    if not paths:
+        return []
+    return [
+        name
+        for _, name, is_pkg in pkgutil.iter_modules(paths)
+        if is_pkg and not name.startswith("__")
+    ]
 
 
 def read_plugins(crd: CRD, channel: Channel) -> None:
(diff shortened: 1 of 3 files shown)


python/michelangelo/cli/mactl/mactl.py
```diff
@@ -387,6 +462,19 @@ def discover_all_plugins() -> dict[str, list[object]]:
                 registry[entity_name].append(plugin_module)
                 _LOG.info("Registered plugin for entity '%s'", entity_name)
 
+    for package_name in plugin_packages:
+        _LOG.debug("Scanning plugin package: %r", package_name)
+        for entity_name in _iter_entity_names_in_package(package_name):
+            _LOG.debug("Found entity plugin: %s (from %s)", entity_name, package_name)
+            plugin_module = read_module_from_package(entity_name, package_name)
```
sishilong, line 469: should we let user know(logging print) that which plugin_module is skipped

yaoyueyue: added,a skipped entity prints to stderr in their terminal automatically

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -387,6 +462,25 @@ def discover_all_plugins() -> dict[str, list[object]]:
                 registry[entity_name].append(plugin_module)
                 _LOG.info("Registered plugin for entity '%s'", entity_name)
 
+    for package_name in plugin_packages:
```
hwaminkim, line 465: This may register redundant plugins in `registry`. in this case, how would we handle it?

python/michelangelo/cli/mactl/mactl.py
```diff
@@ -341,8 +409,14 @@ def apply_module_overrides() -> None:
 def discover_all_plugins() -> dict[str, list[object]]:
```
hwaminkim, line 409: nit: this function is getting longer. we may separate out by wrapping the directory based, module based, and package based search in the future.

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: We can control config.toml and will not use file path plugin anymore, because it fails in bazel environments.

python/michelangelo/cli/mactl/mactl.py
yaoyueyue: Thank you, we can do it as follow up

sishilong (approved):

PR #1332, opened 2026-06-18
feat: add MetadataStoragePrimaryKey annotation for k8s migration

badcount:
Add annotation-based stable primary key for metadata storage to solve the UID mismatch problem when migrating CRs across Kubernetes clusters.

Problem:
- K8s UID changes when recreating CRs in new cluster
- Metadata storage (MySQL/Postgres) uses UID as primary key
- Ingester syncs CRs to metadata storage on recreation
- Result: duplicate records (old UID + new UID) for same resource

Solution:
- New annotation: michelangelo/MetadataStoragePrimaryKey
- Metadata storage uses annotation value if present, else falls back to UID
- Enables stable identity across cluster migrations
- Set annotation value before migration to preserve PK

Changes:
- go/api/api.go: Define MetadataStoragePrimaryKeyAnnotation constant
- go/storage/mysql/mysql.go: Check annotation in fullUpsert, use as PK

Benefits:
- No duplicate records when recreating CRs with same name
- Stable foreign key references across migrations
- Database agnostic (works with MySQL, Postgres, etc)
- Backwards compatible (falls back to UID if annotation not present)
- Enables true disaster recovery scenarios

Usage:
Set the annotation before migrating to a new cluster:
  kubectl annotate pipelinerun my-run \
    michelangelo/MetadataStoragePrimaryKey=$(kubectl get pipelinerun my-run -o jsonpath='{.metadata.uid}')

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/go/api/api.go
+++ b/go/api/api.go
@@ -34,6 +34,13 @@ const (
 	// object can be changed in MySQL later on.
 	ImmutableAnnotation = "michelangelo/Immutable"
 
+	// MetadataStoragePrimaryKeyAnnotation stores a stable identifier to use as the primary
+	// key in metadata storage (MySQL, Postgres, etc). When set, ingester and metadata
+	// storage use this value instead of K8s UID. This enables disaster recovery across
+	// clusters where UIDs change but logical identity (namespace + name) remains the same.
+	// The value is typically the original UID from the first cluster, preserved during migration.
+	MetadataStoragePrimaryKeyAnnotation = "michelangelo/MetadataStoragePrimaryKey"
+
 	/////////////////////////// K8s Labels ///////////////////////////
 
 	// SpecUpdateTimestampLabel is used to record the last time the spec of the object is updated.
--- a/go/storage/mysql/mysql.go
+++ b/go/storage/mysql/mysql.go
@@ -13,6 +13,7 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	proto "github.com/gogo/protobuf/proto"
 	gogotypes "github.com/gogo/protobuf/types"
+	api "github.com/michelangelo-ai/michelangelo/go/api"
 	"github.com/michelangelo-ai/michelangelo/go/api/utils"
 	"github.com/michelangelo-ai/michelangelo/go/storage"
 	apipb "github.com/michelangelo-ai/michelangelo/proto-go/api"
@@ -1108,7 +1109,7 @@ func (m *mysqlMetadataStorage) fullUpsert(ctx context.Context, tx *sql.Tx, table
 	// This enables stable identity across cluster migrations
 	primaryKey := string(metaObj.GetUID())
 	if annotations := metaObj.GetAnnotations(); annotations != nil {
-		if storagePK, exists := annotations["michelangelo/MetadataStoragePrimaryKey"]; exists && storagePK != "" {
+		if storagePK, exists := annotations[api.MetadataStoragePrimaryKeyAnnotation]; exists && storagePK != "" {
 			primaryKey = storagePK
 		}
 	}
--- a/go/storage/mysql/mysql.go
+++ b/go/storage/mysql/mysql.go
@@ -148,20 +148,23 @@ func (m *mysqlMetadataStorage) Upsert(ctx context.Context, object runtime.Object
 		return m.directUpdate(ctx, tx, tableName, metaObj, object)
 	}
 
-	// Full upsert: update all fields
+	// Full upsert: update all fields. Compute the stable primary key once so the main
+	// object row and its label/annotation child rows are all keyed consistently.
+	primaryKey := metadataStoragePrimaryKey(metaObj)
+
 	err = m.fullUpsert(ctx, tx, tableName, groupVer, metaObj, protoBytes, jsonBytes, indexedFields)
 	if err != nil {
 		return err
 	}
 
 	// Upsert labels
-	err = m.upsertLabels(ctx, tx, tableName, string(metaObj.GetUID()), metaObj.GetLabels())
+	err = m.upsertLabels(ctx, tx, tableName, primaryKey, metaObj.GetLabels())
 	if err != nil {
 		return err
 	}
 
 	// Upsert annotations
-	err = m.upsertAnnotations(ctx, tx, tableName, string(metaObj.GetUID()), metaObj.GetAnnotations())
+	err = m.upsertAnnotations(ctx, tx, tableName, primaryKey, metaObj.GetAnnotations())
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


go/storage/mysql/mysql.go
```diff
@@ -1104,11 +1104,20 @@ func (m *mysqlMetadataStorage) fullUpsert(ctx context.Context, tx *sql.Tx, table
 		indexedFieldsMap[field.Key] = field.Value
 	}
 
+	// Determine primary key: use MetadataStoragePrimaryKeyAnnotation if set, else UID
+	// This enables stable identity across cluster migrations
+	primaryKey := string(metaObj.GetUID())
+	if annotations := metaObj.GetAnnotations(); annotations != nil {
+		if storagePK, exists := annotations["michelangelo/MetadataStoragePrimaryKey"]; exists && storagePK != "" {
```
aritra0227: can we use the defined variable here?
```suggestion
		if storagePK, exists := annotations[MetadataStoragePrimaryKeyAnnotation]; exists && storagePK != "" {
```

aritra0227 (commented):
@badcount curious if this was tested through deployment? I think an integration test might be good in this case.

go/api/api.go
```diff
@@ -34,6 +34,13 @@ const (
 	// object can be changed in MySQL later on.
 	ImmutableAnnotation = "michelangelo/Immutable"
 
+	// MetadataStoragePrimaryKeyAnnotation stores a stable identifier to use as the primary
+	// key in metadata storage (MySQL, Postgres, etc). When set, ingester and metadata
+	// storage use this value instead of K8s UID. This enables disaster recovery across
+	// clusters where UIDs change but logical identity (namespace + name) remains the same.
+	// The value is typically the original UID from the first cluster, preserved during migration.
+	MetadataStoragePrimaryKeyAnnotation = "michelangelo/MetadataStoragePrimaryKey"
```
aritra0227, line 42: when is this annotation actually set on the resource? How do we guarantee that every resource has this annotation set?

aritra0227 (commented):
In the summary I see
> Set the annotation before migrating to a new cluster:

However, what happens if a user does not set this before migrating?
Then they will encounter issues with duplicate resources being persisted in mysql right?

go/api/api.go
aritra0227, line 42: In terms of what value we should set for MetadataStoragePrimaryKeyAnnotation, I think some string concatenation of `{Group}:{Version}:{Kind}:{Namespace}:{Name}` might be a good candidate. I believe the above combination is unique per CR.

badcount: we can't use this as people can create same name/namespace after deletion

aritra0227: So what is the goal of the mysql storage? isn't it supposed to be a mirror of etcd?

If someone does create a new CR with the same name/namespace, then isn't the correct path to reflect that in mysql? I'm wondering what you would do with the old value in that case?

go/storage/mysql/mysql.go
badcount: Done in 7f067b8851c43bacf9f0d09626ff66124d02940e. Added import for the api package and replaced the hardcoded string with api.MetadataStoragePrimaryKeyAnnotation.

[addressed by agent]

go/api/api.go
badcount: Correcting my earlier reply here — it was wrong. The annotation is **not** manually set and is **not** opt-in; the ingester sets it automatically, which is exactly what guarantees every reconciled resource has it.

On every reconcile, `Reconciler.ensureMetadataStoragePrimaryKey` (`go/components/ingester/controller.go`) runs *before* the metadata-storage upsert. If `michelangelo/MetadataStoragePrimaryKey` is absent, it sets it to the resource's current UID, patches the object, and requeues so the subsequent reconcile upserts with the annotation already in place. It only writes when the annotation is missing, so it never overwrites an existing value.

That answers both of your questions:

- **When is it set / how is it guaranteed:** on the resource's first reconcile, before it ever reaches MySQL. Any resource that gets ingested necessarily has it.
- **Existing resources (backwards compat):** on their next reconcile they get the annotation set to their current UID — which is exactly the primary key already in use — so no data change and no migration step is required. The UID fallback in `fullUpsert` covers anything not yet annotated.
- **Cross-cluster migration:** because the value is captured at first ingestion and travels with the resource (etcd backup/restore preserves annotations), a CR recreated on a new cluster still carries the original UID in the annotation. k8s assigns a fresh UID on recreate, but `ensureMetadataStoragePrimaryKey` sees the preserved annotation and leaves it untouched, so the upsert reuses the same MySQL row instead of inserting a duplicate.

So the stable identity is established by the ingester from the very first ingestion — no operator action needed.

[addressed by agent]

badcount: Agreed with @badcount - {Group}:{Version}:{Kind}:{Namespace}:{Name} is not safe since a CR with the same name can be recreated after deletion, giving a different logical identity but the same key.

Re the MySQL mirror question: yes, MySQL is a mirror of etcd, but migration breaks this invariant. When you recreate the same CR on a new cluster, k8s assigns a fresh UID. Without this annotation, the ingester would insert a new MySQL row (keyed by the new UID) while the old row (keyed by the original UID) is left orphaned - two rows for the same logical resource, with any foreign-key references pointing at the old one.

How the annotation gets the original UID (correcting an earlier reply in the other thread): the ingester sets it automatically. On first ingestion in the source cluster, `ensureMetadataStoragePrimaryKey` writes the annotation = that resource's UID. During migration the annotation travels with the resource (etcd backup/restore preserves annotations), so on the new cluster it still holds the original UID; `ensureMetadataStoragePrimaryKey` only writes when the annotation is absent, so it leaves the preserved value untouched and the UPSERT in `fullUpsert` hits the existing row instead of creating a new one - references stay intact.

A genuine delete + recreate with the same name (not a migration) is naturally handled: the recreated object is fresh and does not carry the old annotation, so the ingester auto-assigns a new stable key = the new UID, producing a fresh row - which is the correct behavior for a new logical resource. The difference between the two cases is whether the annotation travels with the resource, not whether an operator sets it.

[addressed by agent]

aritra0227: >A genuine delete + recreate with the same name (not a migration) is naturally handled: the recreated object is fresh and does not carry the old annotation, so the ingester auto-assigns a new stable key = the new UID, producing a fresh row - which is the correct behavior for a new logical resource.

In this case, you would have 1 orphaned row in mysql which does not mirror anything in etcd. Is this fine?

What happens when a resource in deleted in kubernetes? For example, if I run 
```
kubectl delete michelangelo.api.deployments/deployment-foo
```
The etcd entry will likely remove that entry. What happens to the mysql row that corresponds to this? If the mysql row also gets invalidated/removed, then I think this solution is fine.

aritra0227 (commented):
During migration from cluster-A to cluster-B, what if there are some CRs in cluster-A which have not been synced within the Metadata storage yet? What is the solution in that case?

Do we just expect that in cluster-B, the new CRs would get synced within the metadatastore with their new UIDs?

go/api/api.go
badcount: Good catch — validated this against the actual ingester code and it's a real, pre-existing gap, not introduced by this PR: for a plain `kubectl delete` (no `michelangelo/Deleting` annotation), `handleCascadeDeletion` in `go/components/ingester/controller.go` was a no-op for any kind not opted into the cascade `RetainPolicy` — it just removed the finalizer without touching MySQL. Since the table's `PRIMARY KEY` is `uid`, not `(namespace, name)`, deleting and recreating a CR with the same name gets a new UID and inserts a *distinct* row rather than colliding with the stale one — a genuine, permanent orphan.

I reproduced it against a live MySQL and confirmed: `Upsert(cr1)` → (no delete) → `Upsert(cr2, new UID)` leaves 2 live rows for the same namespace/name.

Since this affects every CR kind's delete handling regardless of the stable-PK annotation, I'm fixing it as a separate PR: `handleCascadeDeletion` now soft-deletes the MySQL row for non-opted-in kinds before removing the finalizer. Will link the PR here once it's up.

badcount:
Re: CRs not yet synced to metadata storage before cluster migration — that's expected, and it's fine:

- `MetadataStoragePrimaryKey` is only set by `ensureMetadataStoragePrimaryKey` on the ingester's *first* reconcile of a resource, right before its first Upsert to MySQL. If a CR hasn't synced yet in cluster-A, it also hasn't gotten the annotation.
- After migration it reconciles fresh in cluster-B, gets the annotation set to its (new) UID, and upserts as a normal first-sync — no stale row to collide with, since it was never in MySQL from cluster-A.
- So un-synced CRs just become new rows in cluster-B, same as if created there natively. Stable-PK preservation only matters for CRs that were already synced (and thus already carry the annotation) before migration.

aritra0227 (approved):

PR #1333, opened 2026-06-18
Add require-cast-comment rule; annotate all existing as-assertions

craigmarker:
## Summary

Type assertions can be escape hatches for real issues. They also can
lose their context over time. This PR adds a new eslint rules that requires
comments on type assertions.

Turning the rule on meant reckoning with every assertion already in the 
codebase. Where a cast could be removed outright, it was. Where a cast
could be fixed cheaply, it was fixed rather than re-justified. Where a cast
could hide a real design gap or bug, an issue is created and referenced
in the cast for later follow up. Where a cast is believed to be a genuine,
permanent boundary (third-party library's typing gaps), the comment
clarifies.

## Test plan

Smoke test the application by clicking through impacted functionality.

## Issues

Related: #1416, #1417, #1418, #1419, #1424, #1425, #1443, #1453, #1454, #1456, #1458, #1459, #1460, #1461, #1462, #1464

### DIFF
--- a/javascript/eslint-local-rules/__tests__/require-cast-comment.test.js
+++ b/javascript/eslint-local-rules/__tests__/require-cast-comment.test.js
@@ -1,0 +1,72 @@
+import { RuleTester } from 'eslint';
+
+import rule from '../require-cast-comment.js';
+
+RuleTester.describe = describe;
+RuleTester.it = it;
+
+const tester = new RuleTester({
+  languageOptions: {
+    parser: (await import('@typescript-eslint/parser')).default,
+  },
+});
+
+tester.run('require-cast-comment', rule, {
+  valid: [
+    // as const — safe, no comment needed
+    { code: 'const x = [] as const;' },
+
+    // as unknown — widening, no comment needed
+    { code: 'const x = foo as unknown;' },
+
+    // Comment on the same line
+    { code: 'const x = foo as Bar; // cast: Bar is the only concrete type here' },
+
+    // Comment on the line above
+    {
+      code: `// cast: narrowing from union — checked by caller
+const x = foo as Bar;`,
+    },
+
+    // Double assertion: inner is unknown (safe), outer has comment
+    { code: 'const x = (foo as unknown as Bar); // cast: required for generic override' },
+
+    // as const on object literal
+    { code: 'const cfg = { a: 1 } as const;' },
+  ],
+
+  invalid: [
+    // Missing comment — simple assertion
+    {
+      code: 'const x = foo as Bar;',
+      errors: [{ messageId: 'missingCastComment', data: { type: 'Bar' } }],
+    },
+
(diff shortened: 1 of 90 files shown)


javascript/packages/core/components/form/fields/select/select-field.tsx
```diff
@@ -67,26 +67,29 @@ export function SelectField<V = string | number>({
     if (!currentValue || (Array.isArray(currentValue) && currentValue.length === 0)) return;
```
craigmarker, line 67: @konukselim and I had some trouble getting the typing clean in select-field. A lot of these casts are explained by conversation in https://github.com/michelangelo-ai/michelangelo/pull/890

javascript/eslint-local-rules/require-cast-comment.js
```diff
@@ -0,0 +1,65 @@
...
+      const lines = src.getText().split('\n');
+      const line = node.loc.start.line; // 1-indexed
+
+      // Same line
+      if (lines[line - 1]?.includes('// cast:')) return true;
+
+      // Line immediately above — blank lines break the connection
+      if (line >= 2) {
+        return lines[line - 2].trim().startsWith('// cast:');
+      }
```
donovanlopez: Question: Will we ever enforce a [max-len](https://eslint.style/rules/max-len) lint rule? I screenshotted what one of the cast comments look like below. The comment extends where if you're not looking on a big screen (which I wasn't), you have to scroll to the right to read the whole comment making it hard to read. If we have a max-len rule, it's possible the cast comment won't be immediately above the cast.This isn't really addressable in this PR but if we eventually do enforce a max-len, then we have to rework this rule. 

Image

donovanlopez (approved):
LGTM. Had a curious, non-blocking question

javascript/eslint-local-rules/__tests__/require-cast-comment.test.js
```diff
@@ -0,0 +1,109 @@
...
+      errors: [{ messageId: 'missingCastComment', data: { type: 'Array<string>' } }],
+    },
+
+    // Missing comment — inline in expression
+    {
+      code: 'doSomething(value as SpecificType);',
+      errors: [{ messageId: 'missingCastComment', data: { type: 'SpecificType' } }],
+    },
+
+    // Trailing same-line comment no longer counts — must be a leading block
```
donovanlopez: nit: remove "no longer counts" from the comment as it seems to be referring to something specifically

donovanlopez (approved):
left a small comment but otherwise LGTM

javascript/eslint-local-rules/__tests__/require-cast-comment.test.js
craigmarker: done!

PR #1336, opened 2026-06-18
feat(rayjob): add ownerReference and TTL for automatic RayJob cleanup

dragod812:
## Summary
- **OwnerReference on RayJobs**: `CreateJob()` now fetches the remote RayCluster's UID and stamps an `ownerReference` on each RayJob before creating it. When the RayCluster is deleted, Kubernetes GC automatically cascade-deletes all owned RayJobs, their submitter K8s Jobs, and pods — eliminating orphaned resources.
- **TTLSecondsAfterFinished**: RayJobs are now created with `ttlSecondsAfterFinished: 300` (5 minutes) so KubeRay can auto-cleanup finished job artifacts on the compute cluster.

## Problem
When Michelangelo terminates a Ray cluster, it deletes the `RayCluster` object. But `RayJob` CRs had no `ownerReferences`, so Kubernetes GC had nothing to cascade from. The RayJobs (and their submitter Jobs + pods) were permanently orphaned on the compute cluster.

## Changes
- `go/components/jobs/client/client.go` — `CreateJob()` fetches remote RayCluster UID, sets ownerReference on the mapped RayJob
- `go/components/jobs/client/k8sengine/ray.go` — `mapRay()` sets `TTLSecondsAfterFinished: 300` on the RayJob spec
- `go/components/jobs/client/client_test.go` — Added `TestCreateJob_OwnerReferenceSet` and `TestCreateJob_GetRemoteClusterFailure`
- `go/components/jobs/client/k8sengine/mapper_test.go` — Updated expected RayJob spec to include TTL field

## Operator-side note
To also auto-delete the RayJob CR itself after job completion, set `DELETE_RAYJOB_CR_AFTER_JOB_FINISHES=true` as an environment variable on the KubeRay operator deployment (not part of this PR — requires operator-level config).

## Test plan
- [x] `go test ./components/jobs/client/...` — all 28 tests pass
- [x] `go test ./components/ray/...` — all 18 tests pass
- [x] Integration: create RayCluster + RayJob in sandbox, verify ownerRef is set, delete cluster, confirm RayJob is cascade-deleted

### DIFF
--- a/go/components/jobs/client/client.go
+++ b/go/components/jobs/client/client.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
+	k8sptr "k8s.io/utils/ptr"
 
 	apipb "github.com/michelangelo-ai/michelangelo/proto-go/api"
 	v2pb "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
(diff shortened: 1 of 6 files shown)


go/components/jobs/client/client.go
ishtoo1, line 1: from what i understand this only ensures the ownerRef connection to newly created rayjobs. how do we ensure this for existing rayjobs?

dragod812: The existing rayjobs which have not been cleaned up will be manually cleaned up from the compute clusters. The existing RayJobs have already completed and reached a terminal state so it shouldn't be a problem to clean them up.

ishtoo1 (approved):

go/components/jobs/client/client_test.go
```diff
@@ -762,6 +764,113 @@ func TestCreateJob(t *testing.T) {
 	}
 }
 
+// spyHelper is a test stub of the Helper interface for CreateJob owner-reference tests.
+// It captures the object passed to CreateResource and lets tests inject GetResource behavior.
+type spyHelper struct {
```
ishtoo1: do you think defining this entity might be an overkill here?

dragod812: Yeah i agree, this is not needed, removing this.

PR #1341, opened 2026-06-19
Rename no-event-handler-prefix → no-handler-mirror

craigmarker:
The rule name implied it was about prefixes. The actual constraint is that handler names cannot mirror the prop name without adding context — `onChange={handleChange}` and `onClick={onClick}` are both banned regardless of prefix.

`no-handler-mirror` says exactly what the rule does, and matches the existing `noHandlerMirror` messageId.

## Changes

Renames three files (`*.js`, `*.md`, `*.test.js`) and updates `eslint.config.js` — import variable, plugin registration key, and rule activation key.

## Test plan

I ran `yarn test eslint-local-rules/__tests__/no-handler-mirror.test.js` — all 15 cases pass under the new name.


javascript/eslint-local-rules/no-handler-mirror.md
donovanlopez, line 1: The title of this page is still `no-event-handler-prefix`

donovanlopez (approved):
LGTM! I left a comment that we need to fix the title of the .md file.

PR #1364, opened 2026-06-23
feat(trainer): add TrainingObserver protocol for pluggable metrics observation

### LINKED ISSUE #1355: [Trainer] TrainingObserver protocol for pluggable metrics
The trainer currently has no metrics or observability hook. Users have no pluggable way to emit training metrics to their monitoring system (Prometheus, Datadog, custom, etc.).

- Define a `TrainingObserver` protocol with:
  - `on_result(success: bool, tags: dict)` — called after each training result
  - `on_checkpoint_upload(parallelism: int)` — called on checkpoint save
- Add optional `training_observer` parameter to `LightningTrainerParam`
- Wire into `train()` and callback hooks
- Default: no-op implementation (zero overhead when not configured)
- Users supply their own implementation via constructor injection

- Callback-only approach: rejected because metrics emission is cross-cutting, not a Lightning callback concern
- Logger subclass: rejected because it conflates experiment tracking with operational metrics

**Additional context**
- Estimated size: ~200 lines
- Dependencies: Test suite issue
- Priority: **P2**
- Pattern: constructor injection (same pattern as pusher's `StorageBackend`)

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds a `TrainingObserver` Protocol to the PyTorch Lightning trainer that provides two hooks:
- `on_result(metrics, checkpoint_path)` — called driver-side after training completes
- `on_checkpoint_saved(epoch, step, metrics, checkpoint_path)` — called worker-side after each checkpoint is saved and reported to Ray Train

The observer is an optional field on `LightningTrainerParam` (`training_observer`). It is threaded through to the Ray Train callbacks (`RayTrainReportCallback`, `RayTrainReportPerNodeCallback`) via `_resolve_callbacks`. When `None` (default), all code paths are no-ops.

**Why?**

The internal SDK hardcodes M3/Comet logging with no pluggable observation point. The OSS trainer needs an extensible protocol so users can hook custom metric sinks (MLflow, W&B, custom dashboards) without modifying trainer internals. Resolves #1355.

**How did you test it?**

- Added tests in `test_schema.py`: Protocol is `runtime_checkable`, rejects incomplete implementations
- Added tests in `test_callbacks.py`: observer called after checkpoint in both callback types, `None` observer no-ops
- Added tests in `test_lightning_trainer.py`: observer field defaults, pop/re-inject wiring in `__init__`, `on_result` called after `train()`, no-observer no-op
- All files pass Python AST syntax validation

**Potential risks**

Low risk — the observer is opt-in (`None` default). Existing code paths are unchanged when no observer is provided. The only behavioral addition is guarded `if self._training_observer is not None:` checks after existing `ray.train.report()` calls.

**Release notes**

New `TrainingObserver` protocol for pluggable training event observation. Users can implement `on_result()` and `on_checkpoint_saved()` to receive training lifecycle events.

**Documentation Changes**

- Google-style docstrings on `TrainingObserver` protocol, both methods, and the `training_observer` field on `LightningTrainerParam`
- Example usage in the Protocol docstring


kenns29 (approved):

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py
```diff
@@ -5,11 +5,17 @@
...
 import ray.train.lightning
 from ray.train import Checkpoint
 
+if TYPE_CHECKING:
+    from michelangelo.lib.trainer.torch.pytorch_lightning.schema import (
+        TrainingObserver,
+    )
+
 
 class RayTrainReportCallback(ray.train.lightning.RayTrainReportCallback):
```
TheGreatFrankie, line 20: In the current log of lightning trainer sdk, in RayTrainReportCallback & RayTrainReportPerNodeCallback, we only do model checkpointing in on_train_epoch_end(), which is explicitly on epoch_end. 

will this assumption always be true? Will there be future use case when user need mid-epoch checkpointing? or checkpoint per every N steps?

TheGreatFrankie (approved):

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py
sallycr: This is not relevant on this PR changes.

Issue #1338, opened 2026-06-19
[Release] Phase 1: Foundation — Release Infrastructure (v0.3.0)


sallycr:
## Phase 1: Foundation

Establishes the core release infrastructure needed before the first coordinated release. All 6 tasks must complete before Phase 2 begins.

**RFC:** [RFC-20260616: Coordinated Release Process](https://github.com/michelangelo-ai/enhancements/pull/9)
**Target:** v0.3.0-rc.1

### Tasks

- [ ] **001** — Adopt conventional commits — update `CONTRIBUTING.md` with commit message format guide
- [ ] **002** — Add `cliff.toml` for changelog generation — configure git-cliff with conventional commit parsers
- [ ] **003** — Add PyPI publishing to release workflow — `poetry publish` step in `release.yaml` (requires `PYPI_TOKEN` secret)
- [ ] **004** — Add Helm OCI chart publishing — `helm push` job in `release.yaml` to `oci://ghcr.io/uber/michelangelo/charts`
- [ ] **005** — Align all version numbers to 0.3.0 — `pyproject.toml`, `package.json` (×3), `Chart.yaml`
- [ ] **006** — Fix tag format to standard SemVer — document `vX.Y.Z` format, update CI references

### Prerequisites

- [ ] `PYPI_TOKEN` secret provisioned on the GitHub org (blocks task 003)

### Acceptance Criteria

- `pip install michelangelo==0.3.0rc1` works from PyPI
- `helm pull oci://ghcr.io/uber/michelangelo/charts/michelangelo --version 0.3.0-rc.1` works
- All component version files show `0.3.0`
- `git cliff` generates a changelog from recent commits
- `CONTRIBUTING.md` documents conventional commit format

sallycr:
Phase 1 complete — all 8 tasks done:

- #1335 (001) Conventional commits & contribution guidelines ✅
- #1365 (002) cliff.toml for changelog generation ✅
- #1367 (003) PyPI publishing in release workflow ✅
- #1371 (004) Helm OCI chart publishing ✅
- #1376 (005) Version alignment to 0.3.0 ✅
- #1377 (006) SemVer tag format docs ✅
- #1370 (021) Breaking change checklist in PR template ✅
- #1378 (022) Deprecation policy in CONTRIBUTING.md ✅

PR #1335, opened 2026-06-18
docs(contributing): add conventional commit message guide

sallycr:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

- Added a **Commit Messages** section to `CONTRIBUTING.md` documenting the conventional commit format: `type(scope): description`
- Defined 8 allowed types (`feat`, `fix`, `docs`, `ci`, `chore`, `refactor`, `test`, `perf`) and 6 allowed scopes (`python`, `helm`, `ci`, `npm`, `ui`, `go`)
- Documented the `BREAKING CHANGE:` footer convention for changelog generation via git-cliff
- Aligned `.claude/skills/pr-template/SKILL.md` to match — added `ci` type and optional scopes

**Why?**

Conventional commits are a prerequisite for automated changelog generation (git-cliff, task 002) and semantic versioning. Without a documented format, contributors use inconsistent commit messages that can't be parsed into release notes.

**How did you test it?**

- Verified all 8 types and 6 scopes are documented in CONTRIBUTING.md
- Verified Table of Contents anchor links correctly
- Confirmed pr-template skill types now match CONTRIBUTING.md
- No functional code changes — documentation only

**Potential risks**

None — documentation-only change.

**Release notes**

N/A

**Documentation Changes**

- `CONTRIBUTING.md`: new "Commit Messages" section with format, types, scopes, examples, and BREAKING CHANGE convention
- `.claude/skills/pr-template/SKILL.md`: added `ci` type and optional scopes to PR title format

PR #1365, opened 2026-06-23
ci(release): add cliff.toml for changelog generation

sallycr:
## Summary
- Adds `cliff.toml` at repo root for [git-cliff](https://git-cliff.org/) changelog generation
- Configures conventional commit parsers mapping to 8 groups: Features, Bug Fixes, Documentation, Performance, Refactoring, CI/CD, Testing, Miscellaneous
- Enables BREAKING CHANGE detection via commit body/footer (`protect_breaking_commits = true`)
- Skips nightly dev tags (`*.dev.*`), targets release tags (`v[0-9].*`)
- Catch-all parser routes unrecognized commit types to Miscellaneous

## Test plan
- [x] TOML validates cleanly (`python3 -c "import tomllib; tomllib.load(open('cliff.toml', 'rb'))"`)
- [ ] Run `git-cliff` against repo history to verify output format

Part of #1338

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -1,0 +1,48 @@
+[changelog]
+header = """
+# Changelog\n
+All notable changes to this project will be documented in this file.\n
+"""
+body = """
+{% if version %}\
+    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+{% else %}\
+    ## [Unreleased]
+{% endif %}\
+{% for group, commits in commits | group_by(attribute="group") %}
+    ### {{ group }}\n
+    {% for commit in commits %}
+        - {% if commit.scope %}**{{ commit.scope }}:** {% endif %}{{ commit.message | upper_first }}\
+            {% if commit.breaking %} [**BREAKING**]{% endif %}\n
+    {% endfor %}
+{% endfor %}\n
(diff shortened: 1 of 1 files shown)


sallycr:
verify git-cliff 
------
- **types:** Type the accessor input so callers don't have to cast records (#1278)
- **registry:** APIRegistryClient delegates to APIClient.ModelService (#1302)
- Migrate kill trigger run to declarative MutationActionConfig (#1203)
------

The core groups are working correctly — Features, Bug Fixes, Documentation, CI/CD, Refactoring, Testing, Miscellaneous all show up. But there's noise from old non-conventional commits whose message bodies contain markdown headers (e.g., "1. WorkflowClient Interface Extension", "Applied Consistent Terminology").

Two options:
1. Set filter_unconventional = true — only conventional commits appear in changelogs. Cleanest output, but loses pre-convention history.
2. Leave as-is — since changelogs will only be generated for new tags going forward (post PR #1335), this noise won't appear in real releases.

I'd recommend option 2 since the changelog will only cover commits after v0.3.0 tagging.

paulzim (approved):
This should provide consistent content for the changelog generation. I agree with the as- is option, since as you mentioned, changelogs generated for the 0.3.0 release will contain the proper formatting.

Issue #1339, opened 2026-06-19
[Release] Phase 2: Automation — Nightly Builds, Changelog & Release Tooling (v0.3.0)


sallycr:
## Phase 2: Automation

Adds automated nightly builds, changelog generation, release tooling, and integration validation. Builds on Phase 1 foundation.

**RFC:** [RFC-20260616: Coordinated Release Process](https://github.com/michelangelo-ai/enhancements/pull/9)
**Target:** v0.3.0
**Depends on:** Phase 1: Foundation (#1338)

### Tasks

- [ ] **007** — Create `nightly.yml` workflow — scheduled daily at 02:00 UTC, builds all artifacts with nightly version suffix
- [ ] **008** — Create `changelog.yml` workflow — triggered on `v*` tag push, uses git-cliff to auto-populate GitHub Release notes
- [ ] **009** — Add release checklist issue template — `.github/ISSUE_TEMPLATE/release-checklist.md` with pre-release, RC, and final checklists
- [ ] **010** — Add tag-based triggers to existing workflows — `dev-release.yml` and `ui-release.yml` trigger on `v*` tags
- [ ] **011** — Add nightly runtime warning to Python SDK — `warnings.warn()` when version contains `.dev`
- [ ] **012** — Add version-bump script — single command updates version in all component files
- [ ] **013** — Add nightly integration test workflow — `nightly-integration.yml` validates all 7 nightly artifacts are mutually compatible

### Acceptance Criteria

- Nightly builds run automatically and publish to all channels (PyPI pre-release, ghcr.io, npm nightly tag, Helm OCI)
- Nightly integration tests pull all 7 artifacts and run the integration test suite; failures auto-open a GitHub Issue
- `git tag v0.3.0` on a release branch auto-generates release notes on the GitHub Release
- `scripts/version-bump.sh 0.3.0` updates all 5 component version files
- Release checklist template is available when creating new issues
- Python nightly installs show a runtime warning directing users to stable releases

sallycr:
## Phase 2 Complete ✅

All 9 tasks delivered and merged:

| Task | PR | Status |
|---|---|---|
| 007 — Nightly workflow | #1379 | Merged |
| 008 — Changelog workflow | #1380 | Merged |
| 009 — Release checklist template | #1381 | Merged |
| 010 — Tag-based triggers | #1382 | Merged |
| 011 — Nightly runtime warning | #1388 | Merged |
[...]

PR #1379, opened 2026-06-25
ci: add nightly build workflow

sallycr:
## What type of PR is this?

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [x] CI/CD changes
- [ ] Other

## What changed?

Added `.github/workflows/nightly.yml` — a scheduled workflow that builds and publishes nightly artifacts across all components:
- **Python**: wheel with `.devYYYYMMDD` suffix published to PyPI
- **Go containers**: controllermgr, worker, apiserver images tagged `nightly-YYYYMMDD` pushed to ghcr.io
- **UI container**: tagged `nightly-YYYYMMDD` pushed to ghcr.io
- **Helm chart**: OCI artifact with nightly version pushed to ghcr.io
- **npm packages**: `@michelangelo/core` and `@michelangelo/rpc` published with `nightly` dist-tag

## Why?

Release task 007 (Phase 2). Nightly builds give early adopters and internal teams access to the latest changes without waiting for a formal release. The workflow includes change detection to avoid publishing when main has no new commits.

## How did you test it?

- Verified YAML syntax and structure
- Cross-referenced action versions and patterns with existing workflows (dev-release.yml, ui-release.yml, release.yaml)
- Confirmed registry, secret, and permission patterns match established conventions

## Potential risks

None — new workflow file only. Does not affect existing CI pipelines. Requires `NPM_TOKEN` secret to be configured for npm publishing.

## Release notes

Added nightly build pipeline running at 02:00 UTC daily. Install nightly Python SDK with `pip install --pre michelangelo`.

## Documentation Changes

N/A

Closes: Part of #1339

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,302 @@
+name: Nightly Build
+
+on:
+  schedule:
+    - cron: '0 2 * * *'  # 02:00 UTC daily
+  workflow_dispatch: {}
+
+env:
+  REGISTRY: ghcr.io
+  NIGHTLY_TAG: nightly-${{ github.run_id }}
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # Gate: skip the entire pipeline when main has no new commits since yesterday
+  # ---------------------------------------------------------------------------
+  check-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      has_changes: ${{ steps.check.outputs.has_changes }}
+      date_tag: ${{ steps.check.outputs.date_tag }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for new commits in the last 25 hours
+        id: check
+        run: |
+          DATE_TAG=$(date -u +%Y%m%d)
+          echo "date_tag=$DATE_TAG" >> "$GITHUB_OUTPUT"
+
+          RECENT=$(git log --since="25 hours ago" --oneline | head -1)
+          if [ -n "$RECENT" ] || [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "has_changes=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


paulzim (approved):
LGTM! Reviewed all 6 jobs and the referenced composite action.

What I verified:

- `check-changes` uses a 25-hour window (smart boundary choice to avoid missing commits at cron boundaries), and correctly bypasses the gate on `workflow_dispatch`.
- `python-nightly` dynamically reads the current version with `poetry version -s | sed 's/\.dev.*//'` before constructing the nightly version — good, no hardcoding.
- The `setup-python-poetry` composite action exists and is well-structured with venv caching, so the Python job will be fast on repeat runs.
- `container-nightly` correctly uses `ubuntu-24.04` (matching `dev-release.yml`) and reuses the same Bazel cache key pattern.
- `ui-nightly` mirrors `ui-release.yml` closely — correct Node.js version, buf CLI setup, and Dockerfile.
- `helm-nightly` correctly adds subchart repos before `helm dependency update`.
- Both `nightly-YYYYMMDD` and `nightly` rolling tags are published for containers, which is the right pattern for nightlies.

Two nits (non-blocking):

1. **Hardcoded `0.3.0` in `helm-nightly` and `npm-nightly`**: Both jobs construct versions like `0.3.0-nightly.${DATE_TAG}` with a hardcoded base. After the next version bump, these will publish artifacts with the wrong base version until someone manually updates them. `python-nightly` avoids this by reading the version dynamically — the same approach (`grep '^version:' helm/michelangelo/Chart.yaml | awk '{print $2}'` and `node -p "require('./packages/core/package.json').version"`) would make these self-maintaining.

2. **Unused `NIGHTLY_TAG` top-level env var**: `NIGHTLY_TAG: nightly-${{ github.run_id }}` is defined but never referenced — all jobs use `needs.check-changes.outputs.date_tag` for their tags instead. Dead code, safe to remove.

PR #1380, opened 2026-06-25
ci: add changelog generation workflow

sallycr:
## What type of PR is this?

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [x] CI/CD changes
- [ ] Other

## What changed?

Added `.github/workflows/changelog.yml` that automatically generates release notes when a version tag is pushed. Uses `git-cliff` with the existing `cliff.toml` configuration to produce categorized changelogs from conventional commits.

## Why?

Release task 008 (Phase 2). Automated changelog generation ensures every release has consistent, accurate release notes derived from commit history. Eliminates manual changelog authoring.

## How did you test it?

- Verified YAML syntax and workflow structure
- Confirmed `orhun/git-cliff-action@v4` and `softprops/action-gh-release@v2` match existing repo patterns
- Validated compatibility with existing `cliff.toml` commit parsers and template

## Potential risks

None — new workflow file only. Runs only on tag pushes and manual dispatch.

## Release notes

Added automated changelog generation. Release notes are now auto-generated from conventional commits when a version tag is pushed.

## Documentation Changes

N/A

Closes: Part of #1339

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,79 @@
+name: Changelog
+
+on:
+  push:
+    tags: ['v*']
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to generate changelog for (e.g. v0.3.0). Defaults to latest tag.'
+        required: false
+
+permissions:
+  contents: write
+
+jobs:
+  generate-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Determine tag
+        id: tag
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ] && [ -n "${{ inputs.tag }}" ]; then
+            TAG="${{ inputs.tag }}"
+          elif [ "${{ github.event_name }}" = "push" ]; then
+            TAG="${{ github.ref_name }}"
+          else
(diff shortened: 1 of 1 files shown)


paulzim (approved):
LGTM! Reviewed the full workflow and `cliff.toml` compatibility.

What I verified:

- `fetch-depth: 0` is correct — git-cliff needs full history to walk the commit range.
- `--strip header` on the release notes step correctly removes the `# Changelog` heading (appropriate for GitHub Release body).
- `cliff.toml` `tag_pattern = "v[0-9].*"` matches the `v*` trigger, and `skip_tags = ".*\\.dev.*"` will correctly skip RC tags like `v0.3.0-rc.1`.
- `softprops/action-gh-release@v2` with `append_body: false` will create the release if it doesn't exist or overwrite the body if it does — both correct behaviors.
- No conflict with the tag-triggered container workflows in PR #1382; they operate independently.

One nit (non-blocking): The `Determine tag` step resolves and outputs the tag correctly, and it's used for `Update GitHub Release`. However, the two `git-cliff-action` steps use `--latest` rather than referencing `steps.tag.outputs.tag`. For the primary tag-push use case this is fine — git-cliff picks up the just-pushed tag as latest. But if someone runs `workflow_dispatch` targeting an older tag (e.g., to regenerate notes for `v0.2.0` when `v0.3.0` is already the latest), git-cliff will generate notes for `v0.3.0` instead. Worth addressing if backfill is ever needed — adding `--tag ${{ steps.tag.outputs.tag }}` to the cliff args would fix it.

PR #1382, opened 2026-06-25
ci: add tag-based triggers to container and UI release workflows

sallycr:
## What type of PR is this?

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [x] CI/CD changes
- [ ] Other

## What changed?

Added `tags: ['v*']` triggers to two existing workflows:
- **`dev-release.yml`** — Go service containers (controllermgr, worker, apiserver) now build on version tag pushes
- **`ui-release.yml`** — UI container now builds on version tag pushes, skipping the version-change check for tag events

## Why?

Release task 010 (Phase 2). Tagged releases need to automatically build and push all container images. Previously, only `release.yaml` (Python/PyPI) had tag triggers. The Go service and UI container workflows only ran on main branch pushes.

## How did you test it?

- Verified YAML syntax
- Confirmed `release.yaml` already has `v*` tag trigger (no change needed)
- Verified the `startsWith(github.ref, 'refs/tags/v')` condition correctly bypasses version-change check

## Potential risks

None — additive trigger change. Existing main-branch behavior is unchanged.

## Release notes

Container release workflows now trigger automatically on version tag pushes.

## Documentation Changes

N/A

Closes: Part of #1339

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches: [ main ]
+    tags: ['v*']
     paths-ignore:
       - 'docs/**'
       - 'website/**'
--- a/.github/workflows/ui-release.yml
+++ b/.github/workflows/ui-release.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches: [ main ]
+    tags: ['v*']
     paths:
       - 'javascript/**'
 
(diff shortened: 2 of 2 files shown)


paulzim (approved):
LGTM! Reviewed both workflow files in full.

- `dev-release.yml` — `tags: ['v*']` correctly extends the existing push trigger; `paths-ignore` is safe since a standard release commit always touches `helm/michelangelo/Chart.yaml` (not ignored), so Go service containers will build.
- `ui-release.yml` — tag trigger added, and the `startsWith(github.ref, 'refs/tags/v')` bypass on `build-ui-image` correctly skips the version-change check for tag events. Both docker metadata configs already had `type=ref,event=tag` so image tagging was ready.

One nit (non-blocking): `ui-release.yml` has a `paths` allow-list (`javascript/**`). GitHub applies this filter to tag pushes too, so if a `v*` tag is ever pushed on a commit that doesn't touch `javascript/**` (e.g., a Go-only hotfix), the UI build will silently not run. For the standard version-bump workflow this is fine since `javascript/packages/*/package.json` will always be updated. Worth being aware of if the release process ever deviates.

PR #1385, opened 2026-06-25
feat(triggerrun): make notification settings updatable post-creation

badcount:
## Summary
Intent:
- Allow users to update `spec.notifications` on a TriggerRun after creation and have changes reflected in future pipeline runs spawned by that trigger
- Previously, notifications were locked to whatever was set at creation time because they were encoded in the Temporal schedule's workflow action args and never re-synced

Changes:
- Added `actual_notifications` field (proto field 6) to `TriggerRunStatus` to track the notification configuration currently active in the workflow engine
- Added `UpdateScheduleArgs` to the `WorkflowClient` interface; Temporal implementation replaces the schedule action's workflow args atomically via `handle.Update()`; Cadence implementation is a no-op
- `cronTrigger.Update()` now detects drift between `spec.notifications` and `status.actual_notifications` and calls `UpdateScheduleArgs` with the updated `CreateTriggerRequest` when they diverge
- `cronTrigger.Run()` stamps `ActualNotifications` in the initial status so drift detection has a baseline from the first reconcile
- Updated mock for `WorkflowClient` interface


go/components/triggerrun/cron_trigger.go
```diff
@@ -228,30 +231,46 @@ func (c *cronTrigger) Update(ctx context.Context, triggerRun *v2pb.TriggerRun, a
...
+	}
+
+	// cronToUpdate is empty when only notifications/paused changed — UpdateTrigger
+	// skips the cron spec update in that case.
+	cronToUpdate := ""
+	if cronDrifted {
+		cronToUpdate = desiredCron
+	}
+
+	if cronDrifted || notifDrifted || paused != nil {
```
yangfan0328: Can we remove line 238-252 logic inside this if condition?

go/components/triggerrun/cron_trigger.go
```diff
@@ -228,30 +231,46 @@ func (c *cronTrigger) Update(ctx context.Context, triggerRun *v2pb.TriggerRun, a
...
 
-		// Get workflow ID
-		wid := generateWorkflowID(triggerRun)
-
-		// Try to update the schedule in workflow engine (with optional pause state)
-		err := c.WorkflowClient.UpdateTrigger(ctx, wid, desiredCron, paused)
+		err := c.WorkflowClient.UpdateTrigger(ctx, wid, cronToUpdate, paused, args)
 		if err != nil {
 			// If schedule doesn't exist (e.g., manually deleted), recreate it
 			if strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "NotFound") {
```
yangfan0328: Not sure if we need to consider Cadence region failover case in OSS, we usually do 3-5 times retry and keep getting not found error then we create new.

yangfan0328 (approved):

Issue #1340, opened 2026-06-19
[Release] Phase 3: Polish — Security, Compatibility & Retention (v0.4.0)


sallycr:
## Phase 3: Polish

Hardens the release pipeline with security scanning, compatibility testing, artifact retention, and additional build targets.

**RFC:** [RFC-20260616: Coordinated Release Process](https://github.com/michelangelo-ai/enhancements/pull/9)
**Target:** v0.4.0
**Depends on:** Phase 2: Automation (#1339)

### Tasks

- [ ] **014** — Add Go binary builds to release — cross-compile for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64; attach to GitHub Release
- [ ] **015** — Add CVE scanning workflow — `cve-scan.yml` using trivy or grype on every container build; block release on critical/high CVEs
- [ ] **016** — Add nightly retention cleanup workflow — `cleanup-nightlies.yml` prunes container images, npm nightly tags, and PyPI dev releases older than 14 days
- [ ] **017** — Add compatibility testing matrix — `compat-matrix.yml` tests against Python 3.9–3.12, Kubernetes 1.28–1.30, Node.js 18/20, Helm 3.13+
- [ ] **018** — Evaluate and configure Release Drafter — assess as complement to git-cliff for PR-label-based release notes

### Acceptance Criteria

- Go binaries for 4 OS/arch combinations are attached to GitHub Releases
- Container images are scanned for CVEs on every build; critical/high findings block the release
- Nightly artifacts older than 14 days are automatically cleaned up
- CI matrix tests pass across all targeted Python, K8s, Node.js, and Helm versions
- Release Drafter evaluation documented with recommendation (adopt or skip)

### Notes

- PyPI does not support deleting published versions — nightly cleanup for Python is limited to not publishing new nightlies, not removing old ones
- Each task in this phase is independent and can be implemented in any order

paulzim:
All 5 sub-tasks in this checklist have landed:

- **014** Multi-arch Go binary builds — `.github/workflows/release.yaml` now has a `GOOS`/`GOARCH` build matrix attached to releases
- **015** CVE scanning — `.github/workflows/cve-scan.yml` (Trivy)
- **016** Nightly retention cleanup — `.github/workflows/cleanup-nightlies.yml`
- **017** Compatibility testing matrix — `.github/workflows/compat-matrix.yml`
- **018** Release Drafter — `.github/workflows/release-drafter.yml`

[...]

PR #1398, opened 2026-06-26
ci: add Trivy CVE scanning workflow for container images

sallycr:
## Summary

Adds a scheduled + post-release Trivy CVE scanning workflow that scans all container images for CRITICAL and HIGH severity vulnerabilities and uploads results to the GitHub Security tab.

## What this adds

**`.github/workflows/cve-scan.yml`** — new workflow:
- **Triggers:**
  - `workflow_run` on `Dev Release`, `UI Release`, and `Nightly Build` completing on `main` (scans freshly pushed images)
  - Weekly schedule: Mondays at 06:00 UTC (ongoing baseline scan)
  - `workflow_dispatch` for manual runs
- **Matrix:** `apiserver`, `worker`, `controllermgr`, `ui`
- **Severity:** `CRITICAL,HIGH` — fails the job if found
- **Output:** SARIF uploaded to GitHub Security tab + artifact per service + table summary in job log
- **`fail-fast: false`** so all services are scanned even if one has CVEs

## Validation

- `actionlint` passes
- `workflow_run` trigger names verified against actual `.github/workflows/` files:
  - `"Dev Release"` → `dev-release.yml` (Go service containers)
  - `"UI Release"` → `ui-release.yml` (UI container)
  - `"Nightly Build"` → `nightly.yml` (all nightly images)
- Trivy scan tested locally against `ghcr.io/michelangelo-ai/apiserver:nightly` — connects to GHCR and returns CVE results correctly

## Note on double scan (nit from review)

Two Trivy steps per service is intentional: SARIF mode captures structured results for the Security tab (with `exit-code: 1` to fail on CVEs); table mode provides human-readable output in the job log (with `exit-code: 0`, always runs). The extra scan time is acceptable for a weekly + post-release cadence.

Related: #1340

### DIFF
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -60,3 +60,21 @@ jobs:
           name: trivy-report-${{ matrix.service }}
           path: trivy-${{ matrix.service }}.sarif
           retention-days: 30
+
+  # ---------------------------------------------------------------------------
+  # Notify Slack when the CVE scan fails (i.e. CRITICAL/HIGH vulns found)
+  # ---------------------------------------------------------------------------
+  notify-on-failure:
+    needs: scan
+    if: failure()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Slack on failure
+        uses: slackapi/slack-github-action@v1.27.0
+        with:
+          channel-id: '#michelangelo-releases'
+          slack-message: |
+            :x: *${{ github.workflow }}* failed
+            Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -2,7 +2,7 @@ name: CVE Scan
 
 on:
   workflow_run:
-    workflows: ["Go Containers Release", "UI Release"]
+    workflows: ["Dev Release", "UI Release", "Nightly Build"]
     types: [completed]
     branches: [main]
   schedule:
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -1,0 +1,62 @@
+name: CVE Scan
+
+on:
+  workflow_run:
+    workflows: ["Go Containers Release", "UI Release"]
+    types: [completed]
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  security-events: write
+  packages: read
+
+env:
+  REGISTRY: ghcr.io
+  ORG: michelangelo-ai
+
+jobs:
+  scan:
+    runs-on: ubuntu-24.04
+    if: github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'
+    strategy:
+      fail-fast: false
+      matrix:
+        service: [apiserver, worker, controllermgr, ui]
+
+    steps:
+      - name: Run Trivy vulnerability scanner (${{ matrix.service }})
+        uses: aquasecurity/trivy-action@0.28.0
+        with:
+          image-ref: ${{ env.REGISTRY }}/${{ env.ORG }}/${{ matrix.service }}:latest
+          format: sarif
+          output: trivy-${{ matrix.service }}.sarif
+          severity: CRITICAL,HIGH
+          exit-code: '1'
+
+      - name: Upload SARIF to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        if: always()
+        with:
+          sarif_file: trivy-${{ matrix.service }}.sarif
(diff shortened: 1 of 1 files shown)


paulzim (changes requested):
Reviewed both new workflows. Two blocking issues, one nit.

---

**Blocking 1 (cve-scan.yml): `"Go Containers Release"` workflow does not exist**

The `workflow_run` trigger lists:

```yaml
workflows: ["Go Containers Release", "UI Release"]
```

Looking at every workflow in `.github/workflows/`, there is no workflow named `"Go Containers Release"`. The existing container-adjacent workflows are `"build-release"`, `"Nightly Build"`, `"Dev Release"`, and `"Build examples Docker image"` — none match. `"UI Release"` is correct (`ui-release.yml` exists).

A `workflow_run` trigger for a non-existent workflow name silently never fires. The on-build scanning half of this PR's stated goal won't work until the name is corrected. Needs to match whatever workflow actually builds and pushes the Go container images.

---

**Blocking 2 (cleanup-nightlies.yml): Scheduled runs are permanently in dry-run mode**

```yaml
DRY_RUN: ${{ inputs.dry_run != false }}
```

When triggered by the daily cron schedule, `inputs.dry_run` is unset (empty string). In GitHub Actions expression evaluation, `'' != false` is `true` — so `DRY_RUN` evaluates to `"true"` on every scheduled run. Nightly cleanup will log what it *would* delete but never actually delete anything.

The fix is to use a form that resolves to `false` when there are no inputs:

```yaml
DRY_RUN: ${{ inputs.dry_run == true }}
```

With this form:
- Schedule (no inputs): `'' == true` → `false` → deletes ✓
- Manual dispatch `dry_run: true`: `true == true` → `true` → dry-run ✓
- Manual dispatch `dry_run: false`: `false == true` → `false` → deletes ✓

---

**Nit (cve-scan.yml): Trivy runs twice per service**

Each matrix leg runs Trivy twice: once in SARIF mode (fails on HIGH/CRITICAL) and once in table mode (always, exit-code 0). This doubles scan time and doubles GHCR pulls. Reasonable tradeoff for simplicity, but worth noting.

---

**Everything else looks good:** `fail-fast: false` is correct for the CVE matrix, `if: always()` on SARIF upload ensures results are captured even when Trivy finds CVEs, `aquasecurity/trivy-action@0.28.0` is pinned (good for a security workflow), `dry_run` defaulting to `true` on manual dispatch is a good safety default, and the container pagination and npm deprecation logic are correct.

Requesting changes on the two blocking issues before approving.

sallycr (commented):
Thanks for the detailed review!

**Blocking 1 — workflow_run names:** Fixed in 989afabc. `"Go Containers Release"` → verified actual workflow names:
- `"Dev Release"` (`dev-release.yml`) — builds and pushes Go service containers
- `"UI Release"` (`ui-release.yml`) — builds and pushes UI container
- Added `"Nightly Build"` (`nightly.yml`) as well since it also pushes tagged container images

**Blocking 2 — DRY_RUN expression:** Fixed `inputs.dry_run != false` → `inputs.dry_run == true` across all three cleanup jobs. Same fix as #1400.

**Nit — double Trivy scan:** Intentional tradeoff — SARIF captures structured results for the Security tab, table gives readable output in the job log. Keeping both.

**Validation:** actionlint passes; tested `trivy image --severity CRITICAL,HIGH` locally against `ghcr.io/michelangelo-ai/apiserver:nightly` — successfully scans and returns CVE results.

paulzim (approved):
Both blocking issues from the prior review are resolved. Approved.

**Blocking 1 (wrong workflow name in trigger) — fixed ✓**

`workflow_run` now lists `["Dev Release", "UI Release", "Nightly Build"]`. All three match actual workflow `name:` fields in the repo (`dev-release.yml`, `ui-release.yml`, `nightly.yml`). The silent never-fires problem is gone.

**Blocking 2 (dry-run logic in cleanup-nightlies.yml) — N/A**

`cleanup-nightlies.yml` was removed from this PR entirely. If it's coming back in a separate PR, the `${{ inputs.dry_run != false }}` bug I flagged still needs fixing before that one merges.

**Nit (double Trivy scan) — still present**

Each matrix leg still pulls and scans the image twice: once in SARIF mode, once in table mode. Doubles scan time and GHCR traffic. Not a blocker, just worth knowing.

**Minor nit on Slack step (same as PR 1399)**

`channel-id: '#michelangelo-releases'` should be a Slack channel ID (`C0123456789`), not a `#channel-name` string — the Slack API's `chat.postMessage` endpoint requires the ID. Failure notifications may silently drop if the channel isn't found by name. Same issue I flagged on PR 1399; worth fixing in both places.

- `if: github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'` is the correct guard pattern for mixed `workflow_run` + `schedule` + `dispatch` triggers
- `fail-fast: false` correct — a CVE in one image shouldn't cancel scans of the others
- `if: always()` on SARIF upload and table steps ensures results are captured when Trivy exits 1
- `aquasecurity/trivy-action@0.28.0` pinned — good for a security workflow
- `permissions: packages: read` + `security-events: write` scoped correctly
- `retention-days: 30` on SARIF artifacts is appropriate

Approved.

PR #1399, opened 2026-06-26
ci: add cross-compiled Go binary builds to release workflow

sallycr:
## Summary
- Add cross-compiled Go binaries (3 services × 4 platforms) to GitHub Release artifacts
- Generate SHA-256 checksums file

## Test plan
- [ ] Verify workflow syntax with actionlint
- [ ] Verify Go cross-compilation works locally
- [ ] CI passes on PR

Related: #1340

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -145,3 +145,18 @@ jobs:
           helm push \
             michelangelo-${{ steps.version.outputs.VERSION }}.tgz \
             oci://ghcr.io/michelangelo-ai/michelangelo/charts
+
+  notify-on-failure:
+    needs: [build-upload, build-go-binaries, upload-binaries, helm-publish]
+    if: failure()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Slack on failure
+        uses: slackapi/slack-github-action@v1.27.0
+        with:
+          channel-id: '#michelangelo-releases'
+          slack-message: |
+            :x: *${{ github.workflow }}* failed for ${{ github.ref_name }}
+            Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,14 +3,8 @@ name: build-release
 on:
   push:
     tags: ["v*"]
-  workflow_dispatch:
-    inputs:
-      test_version:
-        description: 'Test version for workflow_dispatch runs'
-        default: '0.0.0-test.1'
 jobs:
   build-upload:
-    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     permissions:
       contents: write
@@ -88,7 +82,6 @@ jobs:
           retention-days: 1
 
   upload-binaries:
-    if: github.event_name == 'push'
     needs: build-go-binaries
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


paulzim (changes requested):
Reviewed the workflow and verified cross-compilation locally. There's one blocking bug, plus a minor nit.

---

**Blocking: `go build` will fail — wrong working directory**

The `Build binary` step runs from the GitHub Actions workspace root:

```yaml
run: go build -o "$BIN" ./go/cmd/${{ matrix.service }}
```

But there is no `go.mod` (or `go.work`) at the repo root — it lives in `go/`. Running `go build` from the root produces:

```
go: cannot find main module, but found .git/config in /home/user/michelangelo
```

I hit this exactly when testing locally. Cross-compilation works correctly once you're in the right directory.

Fix: add `working-directory: go` to the job defaults and adjust the import path:

```yaml
  build-go-binaries:
    runs-on: ubuntu-latest
    defaults:
      run:
        working-directory: go
    ...
      - name: Build binary
        env:
          ...
          BIN: ${{ matrix.service }}-${{ matrix.os }}-${{ matrix.arch }}
        run: go build -o "../$BIN" ./cmd/${{ matrix.service }}
```

Note: the output path needs `../` to write the binary outside `go/` where the upload step can find it — or use `${{ github.workspace }}/$BIN` as an absolute path.

---

**Nit: `checksums-sha256.txt` listed twice in `files:`**

`binaries/*` already includes `checksums-sha256.txt`. `softprops/action-gh-release` deduplicates so it won't double-upload, but the explicit line is redundant and can be removed.

---

**Everything else looks good:** Go version matches `go.mod`, `CGO_ENABLED=0` is correct for cross-compilation, action versions are consistent, `merge-multiple: true` is right for the flat artifact merge, permissions are scoped correctly.

Requesting changes for the working-directory bug before approving.

sallycr (commented):
Thanks for catching this and verifying locally!

Fixed in 3430ffd:
- Added `defaults.run.working-directory: go` to the `build-go-binaries` job so `go build` runs from the correct module root
- Updated import path from `./go/cmd/...` to `./cmd/...`
- Switched binary output to `${{ github.workspace }}/$BIN` (absolute path) so `upload-artifact` can find it from the workspace root
- Removed the redundant `checksums-sha256.txt` entry from `files:` — already covered by `binaries/*`

Validated: all 12 matrix legs (apiserver/worker/controllermgr × linux/darwin × amd64/arm64) passed in CI run #28677058596.

paulzim (approved):
Both issues from the prior review are fixed. Approving with one minor note on the new Slack step.

**Blocking (working directory) — fixed ✓**

`defaults.run.working-directory: go` is now on the `build-go-binaries` job, and `BIN` uses the absolute `${{ github.workspace }}/...` path. The `upload-artifact` step (a `uses:` step, unaffected by `defaults.run`) resolves its relative `path:` from the workspace root — exactly where `BIN` writes the binary. Path resolution is correct for all 12 matrix combinations.

**Nit (duplicate checksums in `files:`) — fixed ✓**

`files: binaries/*` only. Clean.

**Go version ✓**

`go.mod` declares `go 1.24.0`; `setup-go` with `'1.24'` installs the latest 1.24 patch. Compatible.

**Minor nit on the new Slack step**

```yaml
channel-id: '#michelangelo-releases'
```

`slackapi/slack-github-action` expects a channel ID (e.g. `C0123456789`) in `channel-id`, not a `#channel-name` string. Channel names can work for public channels via the Slack API, but it's not guaranteed and the action docs recommend the ID. If the notification silently fails on a release failure, this is why. Worth swapping in the actual channel ID — not a blocker for the binary build functionality.

- `CGO_ENABLED=0` correct for cross-compilation
- `merge-multiple: true` correct — 12 uniquely-named artifacts, no collisions
- `retention-days: 1` appropriate for intermediate build artifacts
- `upload-binaries` correctly has `contents: write`; `build-go-binaries` correctly does not
- `notify-on-failure` correctly lists all four jobs in `needs`; `if: failure()` is the right condition

Approved.

PR #1403, opened 2026-06-26
ci: add compatibility testing matrix for Python, Node.js, and Helm

sallycr:
## Summary

Adds a weekly scheduled workflow that runs the test suite against multiple runtime versions to catch compatibility regressions early.

## What this adds

**`.github/workflows/compat-matrix.yml`** — three parallel matrix jobs + summary:

| Job | Matrix | What it runs |
|---|---|---|
| `python-compat` | Python 3.9, 3.10, 3.11, 3.12 | `poetry run pytest` |
| `node-compat` | Node.js 18, 20 | `yarn install --frozen-lockfile && yarn test` |
| `helm-compat` | Helm 3.13.0, 3.14.0, 3.15.0 | `helm dependency build` + `helm lint` + `helm template` |
| `summary` | — | Writes a result table to the job summary |

**Triggers:** weekly Monday at 03:00 UTC + `workflow_dispatch`

**`fail-fast: false`** on all matrix jobs so all versions are tested even if one fails.

## Fix applied

Removed unnecessary `tar xzf` loop after `helm dependency build` — Helm reads `.tgz` sub-charts directly from `charts/` without extraction (addressed reviewer nit).

## Validation

- `actionlint` passes ✅
- Helm steps tested locally end-to-end **without** `tar xzf`:
  - `helm dependency build helm/michelangelo` — downloads cadence + temporal sub-charts ✅
  - `helm lint helm/michelangelo` — 1 chart linted, 0 failed ✅
  - `helm template michelangelo helm/michelangelo -f values-k3d.yaml` — renders successfully ✅
- Python: `pytest --collect-only` finds 1500 tests ✅
- Node: `yarn test` command confirmed present in `javascript/package.json` ✅

Related: #1340

### DIFF
--- a/.github/workflows/compat-matrix.yml
+++ b/.github/workflows/compat-matrix.yml
@@ -1,0 +1,122 @@
+name: Compatibility Matrix
+
+on:
+  schedule:
+    - cron: '0 3 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  python-compat:
+    name: Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.9', '3.10', '3.11', '3.12']
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Python and Poetry
+        uses: ./.github/actions/setup-python-poetry
+        with:
+          python-version: ${{ matrix.python-version }}
+          working-directory: './python'
+
+      - name: Run tests
+        working-directory: ./python
+        run: poetry run pytest
+
+  node-compat:
+    name: Node.js ${{ matrix.node-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [18, 20]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'yarn'
+          cache-dependency-path: javascript/yarn.lock
+
+      - name: Install dependencies
+        working-directory: ./javascript
+        run: yarn install --frozen-lockfile
+
+      - name: Run tests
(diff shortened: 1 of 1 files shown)


sallycr (commented):
Branch is now clean — rebased onto main with 2 commits only (`compat-matrix.yml` add + tar xzf fix).

**Nit addressed:** removed the unnecessary `tar xzf` loop. Helm reads `.tgz` sub-charts directly from `charts/` — extraction is not needed.

**Validation:**
- `actionlint` passes ✅
- All three helm steps tested locally without `tar xzf`:
  - `helm dependency build` — downloads cadence + temporal ✅
  - `helm lint` — 1 chart linted, 0 failed ✅
  - `helm template -f values-k3d.yaml` — renders successfully ✅
- Python: `pytest --collect-only` finds 1500 tests ✅

paulzim (approved):
Reviewed the compatibility matrix workflow — looks solid. One nit, everything else is correct.

**Verified:**
- `setup-python-poetry` composite action accepts `python-version` as an input and passes it to `actions/setup-python@v5` — the Python matrix is meaningful, not cosmetic
- `yarn test` script exists in `javascript/package.json` (runs vitest)
- Helm chart path (`helm/michelangelo`) and `values-k3d.yaml` are correct; cadence and temporal repos are added before `helm dependency build`
- `fail-fast: false` on all three matrix jobs — correct, a failure in one version shouldn't cancel the others
- `if: always()` on the summary job — correct, the summary should always render regardless of matrix outcomes
- `permissions: contents: read` only — minimal and appropriate

**Nit (helm-compat): `tar xzf` extraction is unnecessary**

```yaml
for f in helm/michelangelo/charts/*.tgz; do
  tar xzf "$f" -C helm/michelangelo/charts/
done
```

`helm lint` and `helm template` both read sub-chart `.tgz` files directly — no extraction needed. The loop creates extracted directories alongside the tarballs, which is harmless but redundant. Can be removed.

**Minor usability note:** The summary table reports the overall job result per component (`success`/`failure`) rather than per version. If Python 3.10 fails but 3.11 passes, the table shows `failure` without indicating which version broke. The per-version detail is available in the matrix job logs, so this isn't a bug — just something to be aware of if you want finer-grained visibility in the summary.

LGTM. Approving.

PR #1406, opened 2026-06-27
fix(ci): bypass poetry-dynamic-versioning in nightly Python build

sallycr:
## Summary
- Nightly Python build failed with PyPI 400: local version identifiers not allowed
- Root cause: `poetry-dynamic-versioning` plugin generates versions from `git describe`, producing `0.3.1.post2.dev0+9f309156` instead of clean `0.3.0.dev20260627`
- Fix: read base version from `pyproject.toml` directly (not `poetry version -s`), set `POETRY_DYNAMIC_VERSIONING_BYPASS` env var to prevent the plugin from overriding the version at build time

## Test plan
- [ ] CI passes on PR
- [ ] Next nightly run publishes to PyPI successfully

Related: #1340

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -61,14 +61,20 @@ jobs:
           working-directory: './python'
 
       - name: Set nightly version
+        env:
+          POETRY_DYNAMIC_VERSIONING_BYPASS: "0.0.0"
         run: |
           DATE_TAG="${{ needs.check-changes.outputs.date_tag }}"
-          BASE_VERSION=$(poetry version -s | sed 's/\.dev.*//')
+          BASE_VERSION=$(grep '^version' pyproject.toml | head -1 | sed 's/version = "\(.*\)"/\1/')
           NIGHTLY_VERSION="${BASE_VERSION}.dev${DATE_TAG}"
           poetry version "$NIGHTLY_VERSION"
+          echo "nightly_version=$NIGHTLY_VERSION" >> "$GITHUB_OUTPUT"
           echo "Publishing Python nightly: $NIGHTLY_VERSION"
+        id: version
 
       - name: Build wheel
+        env:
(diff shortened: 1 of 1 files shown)


sallycr:
## CI Validation ✓

All checks passed on [workflow run #28294906340](https://github.com/michelangelo-ai/michelangelo/actions/runs/28294906340):

- ✅ `bazel-build` — passed
- ✅ `python-build` — passed
- ✅ `dirty-check` — passed
- ✅ `license/cla` — passed

Note: the actual nightly publish fix can only be verified by the next nightly run (02:00 UTC). The change bypasses `poetry-dynamic-versioning` to produce clean `0.3.0.dev{DATE}` versions instead of `0.3.1.post2.dev0+hash`.

paulzim (approved):
Correct diagnosis and minimal fix. The two-step bypass strategy is the right approach here:

1. **Version extraction** — reading directly from `pyproject.toml` via `grep`/`sed` instead of `poetry version -s` sidesteps the dynamic-versioning plugin entirely, so the base version stays `0.3.0` rather than becoming `0.3.1.post2.dev0+9f309156`.

2. **Build bypass** — setting `POETRY_DYNAMIC_VERSIONING_BYPASS` to the computed nightly version during `poetry build` prevents the plugin from re-running its git-describe logic and re-introducing a local version identifier at wheel-build time.

The `grep '^version' pyproject.toml | head -1 | sed 's/version = "\(.*\)"/\1/'` pattern is correct for the current `pyproject.toml` layout (`version = "0.3.0"` under `[tool.poetry]`). The `POETRY_DYNAMIC_VERSIONING_BYPASS: "0.0.0"` on the "Set nightly version" step is defensive but harmless — the plugin doesn't intercept `poetry version <set>` writes, so it has no practical effect there, but it does make the step's intent explicit.

Looks good to merge.

PR #1408, opened 2026-06-27
fix(ci): add helm repo registration before dependency build

sallycr:
## Summary
- Adds `helm repo add` for cadence and temporal chart repositories before `helm dependency build` in the release workflow
- The v0.3.1 helm-publish job failed because the chart dependencies (cadence, temporal) require their Helm repos to be registered first
- Same fix already exists in `nightly.yml` — this aligns `release.yaml`

## Test plan
- [ ] Re-run the v0.3.1 release workflow after merge to verify helm-publish succeeds
- [ ] Verify `helm dependency build helm/michelangelo` completes without errors

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -67,6 +67,12 @@ jobs:
         id: version
         run: echo "VERSION=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
 
+      - name: Add Helm dependency repos
+        run: |
+          helm repo add cadence https://cadence-workflow.github.io/cadence-charts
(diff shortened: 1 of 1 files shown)


paulzim (approved):
Correct fix. `Chart.yaml` declares the cadence and temporal dependencies by repository URL, and `helm dependency build` resolves against those URLs — so this unblocks the v0.3.1 helm-publish job.

One minor nit: `nightly.yml` registers the cadence repo under the alias `cadence-workflow`, while this PR uses `cadence`. Since `helm dependency build` matches by URL (not local alias), both work fine and the inconsistency has no functional impact. Worth aligning them in a follow-up for tidiness, but not a blocker.

Looks good to merge.

PR #1410, opened 2026-06-28
feat(ui): add pipeline run notification settings (toggle + email/Slack recipients)

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds a **"Set Up Notifications"** section to the `CreatePipelineRunForm` dialog:

- A toggle (`Send notifications` — Enabled/Disabled) that enables/disables the section
- Email address recipients, entered as tags (multi + creatable select)
- Slack channel / user handle recipients, entered as tags
- Checkboxes for which run states trigger a notification (Succeeded, Failed, Killed, Skipped — all selected by default)
- Notifications are omitted from the `CreatePipelineRun` payload unless the toggle is on and at least one email or Slack recipient is set

The notification toggle and recipient inputs are pure UI state, not part of the `PipelineRun` entity being submitted, so they're held in local `useState` rather than modeled as virtual fields on the form. `CreatePipelineRunForm` is typed as `FormDialog<PipelineRun>` directly — no separate form-values type. `handleRunSubmit` receives a clean `PipelineRun` and merges in `spec.notifications` from local state before calling the mutation.

Files changed:

| File | Change |
|---|---|
| `packages/core/config/entities/run/types.ts` | Export `ALL_NOTIFICATION_EVENT_TYPES` const; `NotificationEventType`, `PipelineRunNotification` types; extend `PipelineRun.spec.notifications?` |
| `packages/core/config/entities/pipeline/types.ts` | `NotificationDetailsValue` type (local UI state shape) |
| `packages/core/config/entities/pipeline/create-pipeline-run-form.tsx` | `FormDialog<PipelineRun>`; notification toggle (`useState`) merged into payload in `handleRunSubmit` |
| `packages/core/config/entities/pipeline/notification-details.tsx` | Controlled component (`enabled`/`value`/`onChange` props) rendering email/Slack/event-type inputs via BaseUI `Select`/`Checkbox` primitives directly |
| `packages/core/components/form/fields/select/build-select-overrides.ts` | Minor override tweak needed for the tag-style multi/creatable select |
| `packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx` | Tests: notification fields disabled by default and enabled after toggling; notifications included in payload when toggle on + email entered |

**Why?**

Pipeline runs previously had no way to trigger completion alerts from the Studio UI. This mirrors the internal MAStudio notification UX (enable/disable toggle + tag-style recipient inputs) rather than a collapsible-optional section, per review feedback.

**How did you test it?**

- `yarn vitest run packages/core/config/entities/pipeline` — 4/4 tests pass
- `tsc --noEmit` — no new type errors
- `eslint` — 0 errors on changed files

**Potential risks**

None. The notification section is off by default — existing `CreatePipelineRun` calls are unchanged when the toggle is left off. No breaking changes to existing types (`notifications` is optional on `PipelineRun.spec`).

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Pipeline run form now supports a notification toggle with email and Slack recipients, and a choice of which run states trigger an alert.

### DIFF
--- a/javascript/packages/core/components/form/fields/select/build-select-overrides.ts
+++ b/javascript/packages/core/components/form/fields/select/build-select-overrides.ts
@@ -9,8 +9,8 @@ export function buildSelectOverrides(
   readOnly?: boolean
 ): SelectOverrides {
   // When rendered inside modals, dropdown options can overflow past the modal body into its backdrop.
-  // ignoreBoundary lets the dropdown escape the modal clip region; popperOptions ensures
-  // preventOverflow is registered before the hide modifier (Popper.js v1 requires this order).
+  // popperOptions ensures preventOverflow is registered before the hide modifier (Popper.js v1
+  // requires this order), so ignoreBoundary doesn't get its overflow-prevention undone by hide.
   // name is forwarded to the internal <input> so form-focus libraries (e.g. final-form-focus) can
   // match this control to its field error by name.
   const base = {
--- a/javascript/packages/core/components/form/fields/select/build-select-overrides.ts
+++ b/javascript/packages/core/components/form/fields/select/build-select-overrides.ts
@@ -9,10 +9,21 @@ export function buildSelectOverrides(
   readOnly?: boolean
 ): SelectOverrides {
   // When rendered inside modals, dropdown options can overflow past the modal body into its backdrop.
+  // ignoreBoundary lets the dropdown escape the modal clip region; popperOptions ensures
+  // preventOverflow is registered before the hide modifier (Popper.js v1 requires this order).
   // name is forwarded to the internal <input> so form-focus libraries (e.g. final-form-focus) can
   // match this control to its field error by name.
   const base = {
-    Popover: { props: { ignoreBoundary: true } },
+    Popover: {
+      props: {
+        ignoreBoundary: true,
+        popperOptions: {
+          modifiers: {
+            preventOverflow: { enabled: true, boundariesElement: 'window', padding: 8 },
+          },
+        },
+      },
+    },
     Input: { props: { name } },
   };
 
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -220,5 +220,12 @@ describe('CreatePipelineRunForm', () => {
     await user.keyboard('{Enter}');
 
     await screen.findByText('Enter valid email addresses, e.g. user@example.com.');
+
+    // Submitting with the invalid email must not call the API or close the dialog.
+    await user.click(screen.getByRole('button', { name: 'Run' }));
+
+    await screen.findAllByText('Enter valid email addresses, e.g. user@example.com.');
+    expect(mockRequest).not.toHaveBeenCalled();
+    expect(screen.getByRole('dialog')).toBeInTheDocument();
   });
 });
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -138,7 +138,7 @@ describe('CreatePipelineRunForm', () => {
     });
   });
 
-  it('notification fields are disabled by default and enabled after toggling on', async () => {
+  it('notification fields are disabled until the toggle is switched on, then submit includes them', async () => {
     const user = userEvent.setup();
     const mockRequest = createQueryMockRouter({ CreatePipelineRun: {} });
 
@@ -156,45 +156,24 @@ describe('CreatePipelineRunForm', () => {
 
     await screen.findByRole('dialog', { name: 'Start new pipeline run' });
 
-    // Toggle is visible; email combobox is present but disabled
+    // Toggle is visible and off by default; email combobox is present but disabled.
+    // The toggle's accessible name is its own label text ("Enabled"/"Disabled"), which
+    // starts as "Disabled" since notifications default to off.
     expect(screen.getByText('Send notifications')).toBeInTheDocument();
     expect(screen.getByRole('combobox', { name: /email addresses/i })).toBeDisabled();
 
     // Enable notifications via toggle
-    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
+    await user.click(screen.getByRole('checkbox', { name: 'Disabled' }));
 
     // Email combobox is now enabled
-    await waitFor(() => {
-      expect(screen.getByRole('combobox', { name: /email addresses/i })).not.toBeDisabled();
-    });
-  });
-
-  it('includes notifications in payload when toggle is on and email is entered', async () => {
-    const user = userEvent.setup();
-    const mockRequest = createQueryMockRouter({ CreatePipelineRun: {} });
-
-    render(
-      <FormWrapper />,
-      buildWrapper([
-        getBaseProviderWrapper(),
-        getIconProviderWrapper(),
-        getErrorProviderWrapper(),
-        getInterpolationProviderWrapper(),
-        getRouterWrapper({ location: '/ma-dev-test/train/pipelines' }),
-        getServiceProviderWrapper({ request: mockRequest }),
-      ])
-    );
-
-    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
-
-    // Enable notifications
-    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
-
-    // Type an email into the email addresses combobox and commit with Enter
     const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
+    await waitFor(() => {
+      expect(emailCombo).toBeEnabled();
+    });
+
+    // Type an email and submit — payload should include the notification
     await user.type(emailCombo, 'notify@example.com');
     await user.keyboard('{Enter}');
-
     await user.click(screen.getByRole('button', { name: 'Run' }));
 
     await waitFor(() => {
@@ -233,7 +212,7 @@ describe('CreatePipelineRunForm', () => {
     await screen.findByRole('dialog', { name: 'Start new pipeline run' });
 
     // Enable notifications
-    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
+    await user.click(screen.getByRole('checkbox', { name: 'Disabled' }));
 
     // Type an invalid email into the email addresses combobox and commit with Enter
     const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -156,23 +156,23 @@ describe('CreatePipelineRunForm', () => {
 
     await screen.findByRole('dialog', { name: 'Start new pipeline run' });
 
-    // Toggle is visible and off by default; email combobox is present but disabled.
+    // Toggle is visible and off by default; email input is present but disabled.
     // The toggle's accessible name is its own label text ("Enabled"/"Disabled"), which
     // starts as "Disabled" since notifications default to off.
     expect(screen.getByText('Send notifications')).toBeInTheDocument();
-    expect(screen.getByRole('combobox', { name: /email addresses/i })).toBeDisabled();
+    expect(screen.getByRole('textbox', { name: /email addresses/i })).toBeDisabled();
 
     // Enable notifications via toggle
     await user.click(screen.getByRole('checkbox', { name: 'Disabled' }));
 
-    // Email combobox is now enabled
-    const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
+    // Email input is now enabled
+    const emailInput = await screen.findByRole('textbox', { name: /email addresses/i });
     await waitFor(() => {
-      expect(emailCombo).toBeEnabled();
+      expect(emailInput).toBeEnabled();
     });
 
     // Type an email and submit — payload should include the notification
-    await user.type(emailCombo, 'notify@example.com');
+    await user.type(emailInput, 'notify@example.com');
     await user.keyboard('{Enter}');
     await user.click(screen.getByRole('button', { name: 'Run' }));
 
@@ -216,12 +216,12 @@ describe('CreatePipelineRunForm', () => {
     // Enable notifications
     await user.click(screen.getByRole('checkbox', { name: 'Disabled' }));
 
-    // Type an invalid email into the email addresses combobox and commit with Enter
-    const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
-    await user.type(emailCombo, 'not-an-email');
+    // Type an invalid email into the email addresses input and commit with Enter
+    const emailInput = await screen.findByRole('textbox', { name: /email addresses/i });
+    await user.type(emailInput, 'not-an-email');
     await user.keyboard('{Enter}');
 
-    await screen.findByText('Enter valid email addresses, e.g. user@example.com.');
+    await screen.findAllByText('Enter valid email addresses, e.g. user@example.com.');
 
     // Submitting with the invalid email must not call the API or close the dialog.
     await user.click(screen.getByRole('button', { name: 'Run' }));
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -186,7 +186,7 @@ describe('CreatePipelineRunForm', () => {
                 emails: ['notify@example.com'],
                 slackDestinations: [],
                 // Default selection covers every trigger condition (see NOTIFICATION_EVENT_TYPES).
-                eventTypes: [1, 2, 3, 4],
+                eventTypes: [11, 1, 2, 3, 4],
               }),
             ],
           }) as Record<string, unknown>,
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -137,4 +137,109 @@ describe('CreatePipelineRunForm', () => {
       );
     });
   });
+
+  it('notification fields are disabled by default and enabled after toggling on', async () => {
+    const user = userEvent.setup();
+    const mockRequest = createQueryMockRouter({ CreatePipelineRun: {} });
+
+    render(
+      <FormWrapper />,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getErrorProviderWrapper(),
+        getInterpolationProviderWrapper(),
+        getRouterWrapper({ location: '/ma-dev-test/train/pipelines' }),
+        getServiceProviderWrapper({ request: mockRequest }),
+      ])
+    );
+
+    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
+
+    // Toggle is visible; email combobox is present but disabled
+    expect(screen.getByText('Send notifications')).toBeInTheDocument();
+    expect(screen.getByRole('combobox', { name: /email addresses/i })).toBeDisabled();
+
+    // Enable notifications via toggle
+    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
+
+    // Email combobox is now enabled
+    await waitFor(() => {
+      expect(screen.getByRole('combobox', { name: /email addresses/i })).not.toBeDisabled();
+    });
+  });
+
+  it('includes notifications in payload when toggle is on and email is entered', async () => {
+    const user = userEvent.setup();
+    const mockRequest = createQueryMockRouter({ CreatePipelineRun: {} });
+
+    render(
+      <FormWrapper />,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getErrorProviderWrapper(),
+        getInterpolationProviderWrapper(),
+        getRouterWrapper({ location: '/ma-dev-test/train/pipelines' }),
+        getServiceProviderWrapper({ request: mockRequest }),
+      ])
+    );
+
+    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
+
+    // Enable notifications
+    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
+
+    // Type an email into the email addresses combobox and commit with Enter
+    const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
+    await user.type(emailCombo, 'notify@example.com');
+    await user.keyboard('{Enter}');
+
+    await user.click(screen.getByRole('button', { name: 'Run' }));
+
+    await waitFor(() => {
+      expect(mockRequest).toHaveBeenCalledWith(
+        'CreatePipelineRun',
+        expect.objectContaining({
+          spec: expect.objectContaining({
+            notifications: [
+              expect.objectContaining({
+                emails: ['notify@example.com'],
+                slackDestinations: [],
+              }),
+            ],
+          }) as Record<string, unknown>,
+        })
+      );
+    });
+  });
+
+  it('shows a validation error when an invalid email address is entered', async () => {
+    const user = userEvent.setup();
+    const mockRequest = createQueryMockRouter({ CreatePipelineRun: {} });
+
+    render(
+      <FormWrapper />,
+      buildWrapper([
+        getBaseProviderWrapper(),
+        getIconProviderWrapper(),
+        getErrorProviderWrapper(),
+        getInterpolationProviderWrapper(),
+        getRouterWrapper({ location: '/ma-dev-test/train/pipelines' }),
+        getServiceProviderWrapper({ request: mockRequest }),
+      ])
+    );
+
+    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
+
+    // Enable notifications
+    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
+
+    // Type an invalid email into the email addresses combobox and commit with Enter
+    const emailCombo = await screen.findByRole('combobox', { name: /email addresses/i });
+    await user.type(emailCombo, 'not-an-email');
+    await user.keyboard('{Enter}');
+
+    await screen.findByText('Enter valid email addresses, e.g. user@example.com.');
+  });
 });
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -185,6 +185,8 @@ describe('CreatePipelineRunForm', () => {
               expect.objectContaining({
                 emails: ['notify@example.com'],
                 slackDestinations: [],
+                // Default selection covers every trigger condition (see NOTIFICATION_EVENT_TYPES).
+                eventTypes: [1, 2, 3, 4],
               }),
             ],
           }) as Record<string, unknown>,
--- a/javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
+++ b/javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
@@ -1,5 +1,4 @@
 import { useState } from 'react';
-import { useStyletron } from 'baseui';
 import { Checkbox, LABEL_PLACEMENT, STYLE_TYPE } from 'baseui/checkbox';
 
 import { FormControl } from '#core/components/form/components/form-control';
@@ -8,7 +7,6 @@ import { StringField } from '#core/components/form/fields/string/string-field';
 import { TextareaField } from '#core/components/form/fields/textarea/textarea-field';
 import { FormGroup } from '#core/components/form/layout/form-group/form-group';
 import { NotificationDetails } from '#core/config/entities/pipeline/notification-details';
-import { getEmailValidationError } from '#core/config/entities/pipeline/notification-validation';
 import {
   ALL_NOTIFICATION_EVENT_TYPES,
   NOTIFICATION_EVENT_TYPES,
@@ -18,7 +16,7 @@ import { useStudioMutation } from '#core/hooks/use-studio-mutation';
 import { generateSuffix } from '#core/utils/name-utils';
 
 import type { ActionComponentProps } from '#core/components/actions/types';
-import type { NotificationDetailsValue, Pipeline } from '#core/config/entities/pipeline/types';
+import type { NotificationFormFields, Pipeline } from '#core/config/entities/pipeline/types';
 import type { NotificationEventType, PipelineRun } from '#core/config/entities/run/types';
 
 const EVENT_TYPE_TO_PROTO_VALUE: Record<NotificationEventType, number> = Object.fromEntries(
@@ -26,15 +24,12 @@ const EVENT_TYPE_TO_PROTO_VALUE: Record<NotificationEventType, number> = Object.
 ) as Record<NotificationEventType, number>;
 
 export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<Pipeline>) => {
-  const [css, theme] = useStyletron();
   const { projectId } = useStudioParams('base');
 
   const [notificationEnabled, setNotificationEnabled] = useState(false);
-  const [notificationDetails, setNotificationDetails] = useState<NotificationDetailsValue>({
-    emails: [],
-    slackChannels: [],
-    eventTypes: [...ALL_NOTIFICATION_EVENT_TYPES],
-  });
+  const [eventTypes, setEventTypes] = useState<NotificationEventType[]>([
+    ...ALL_NOTIFICATION_EVENT_TYPES,
+  ]);
 
   const createPipelineRunMutation = useStudioMutation<PipelineRun, PipelineRun>({
     mutationName: 'CreatePipelineRun',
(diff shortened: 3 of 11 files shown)


sallycr:
## OSS Review — REQUEST_CHANGES

CI is all green and the code diff is clean of internal references. One thing to fix before merge, plus a few minor polish items.

---

### 🔴 Major — fix before merge

**M1. PR body leaks internal implementation details**

The "Why?" section publicly references `MAStudio`, `shared.ts NOTIFICATIONS_ENTITIES`, "no internal URLs", and "no feature flags" — these are internal implementation details on a public GitHub PR, indexed by search engines.

The code itself is clean. Please rewrite the "Why?" section without internal references, e.g.:

> Pipeline runs previously had no way to be annotated with a description or to trigger completion alerts from the Studio UI. This PR adds both: an optional description field (visible in the run list) and a collapsible notification section where users can configure email and Slack destinations per run.

---

### 🟡 Minor — polish

**m1. `getByPlaceholderText` used where `getByRole` is preferred**

In the "includes description in payload" test:
```tsx
// current
screen.getByPlaceholderText('Enter a description for this run…')

// preferred — tests the accessible interface, not a hint string
screen.getByRole('textbox', { name: /description/i })
```

---

**m2. Fragile `.mock.calls` double-cast in the "omits notifications" test**

```tsx
// current — bypasses the type system
const calls = (mockRequest as ReturnType<typeof vi.fn>).mock.calls;
const payload = calls[0][1] as { spec: Record<string, unknown> };
expect(payload.spec).not.toHaveProperty('notifications');
```
Prefer the type-safe matcher:
```tsx
await waitFor(() => {
  expect(mockRequest).toHaveBeenCalledWith(
    'CreatePipelineRun',
    expect.not.objectContaining({
      spec: expect.objectContaining({ notifications: expect.anything() }),
    })
  );
});
```
Also removes the need for `import type { vi } from 'vitest'`.

---

**m3. `ALL_EVENT_TYPES` silently out-of-sync with `NotificationEventType`**

A new union variant won't automatically appear in the default selection. Consider deriving the type from the constant:

```ts
// run/types.ts
export const ALL_NOTIFICATION_EVENT_TYPES = [
  'EVENT_TYPE_PIPELINE_RUN_STATE_SUCCEEDED',
  'EVENT_TYPE_PIPELINE_RUN_STATE_FAILED',
  'EVENT_TYPE_PIPELINE_RUN_STATE_KILLED',
  'EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED',
] as const;

export type NotificationEventType = typeof ALL_NOTIFICATION_EVENT_TYPES[number];
```

Or at minimum add a comment noting they must stay in sync.

---

### ✅ What's good

- Zero internal references in the code diff — `user@example.com` placeholder, `notify@example.com` in tests ✓
- All form fields use existing `core` primitives — no new low-level components added ✓
- `useStudioMutation` wired correctly ✓
- `PipelineRunFormValues` extending `PipelineRun` with virtual fields is the right pre-submission transformation pattern ✓
- Notifications correctly omitted from payload when section left blank ✓
- `validateEmail` returns `undefined` for empty values — field correctly optional ✓

sallycr:
Screenshot 2026-06-28 at 4 02 28 PM
Screenshot 2026-06-28 at 4 02 53 PM

and spec 

`run-20260628-230156-4a7a4ed1 spec:

description: "final test"
notifications {
  event_types: SUCCEEDED / FAILED / KILLED / SKIPPED   ✓
  emails: "sally.lee@uber.com"                         ✓
  slack_destinations: "#core-ml-platform"              ✓
}
status: PIPELINE_RUN_STATE_FAILED  (expected — no compute)

Clean — all fields stored correctly, no extra quotes.
`

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -102,4 +102,146 @@ describe('CreatePipelineRunForm', () => {
...
+        getRouterWrapper({ location: '/ma-dev-test/train/pipelines' }),
+        getServiceProviderWrapper({ request: mockRequest }),
+      ])
+    );
+
+    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
+
+    // Section and toggle are visible; email combobox is present but disabled
+    expect(screen.getByText('Send notifications')).toBeInTheDocument();
+    expect(screen.getAllByRole('combobox')[0]).toBeDisabled();
```
craigmarker: this is not a very good matcher, since it will break whenever the form order changes

craigmarker (changes requested):

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -102,4 +102,146 @@ describe('CreatePipelineRunForm', () => {
...
+                slackDestinations: [],
+              }),
+            ],
+          }) as Record<string, unknown>,
+        })
+      );
+    });
+  });
+
+  it('omits notifications from payload when toggle is off', async () => {
```
craigmarker: this test isn't verifying anything unique, consider removing.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -1,29 +1,70 @@
...
 export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<Pipeline>) => {
+  const [css, theme] = useStyletron();
   const { projectId } = useStudioParams('base');
 
   const createPipelineRunMutation = useStudioMutation<PipelineRun, PipelineRun>({
     mutationName: 'CreatePipelineRun',
   });
 
-  const handleRunSubmit = async (values: PipelineRun) => {
+  const handleRunSubmit = async (values: PipelineRunFormValues) => {
```
craigmarker: These casts are incorrect :/ `PipelineRunFormValues` is a subset of a `PipelineRun`

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -37,10 +78,14 @@ export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<
...
     },
+    isNotificationEnabled: false,
+    notificationEmails: [],
+    notificationSlackChannels: [],
+    notificationEventTypes: [...ALL_NOTIFICATION_EVENT_TYPES],
   };
 
   return (
-    <FormDialog<PipelineRun>
+    <FormDialog<PipelineRunFormValues>
```
craigmarker: I don't think this change is correct :/ the form should submit an entire PIpelineRun, not just the notifications

javascript/packages/core/components/form/components/add-button/add-button.tsx
```diff
@@ -35,7 +35,13 @@ export function AddButton({ label = 'Add more', shape, onClick }: AddButtonProps
         <Icon name="plus" color={theme.colors.contentPrimary} size={theme.sizing.scale600} />
       }
       overrides={{
-        BaseButton: { style: { marginBottom: theme.sizing.scale600, width: '260px' } },
+        BaseButton: {
```
craigmarker: what issue were we seeing here?

javascript/packages/core/components/form/components/add-button/add-button.tsx
```diff
@@ -35,7 +35,13 @@ export function AddButton({ label = 'Add more', shape, onClick }: AddButtonProps
...
+          style: {
+            marginTop: theme.sizing.scale400,
+            marginBottom: theme.sizing.scale600,
+            width: '260px',
+          },
+        },
```
austingreco:
```suggestion
        BaseButton: { style: { marginBottom: theme.sizing.scale600, width: '260px' } },
```

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Fixed — replaced `getAllByRole('combobox')[0]` with `getByRole('combobox', { name: /email addresses/i })` so the selector is stable regardless of field order.

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Removed — you're right that the basic submit test already implicitly covers the toggle-off path (notifications absent by default). Down to 5 tests.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Fixed — switched to destructuring: `const { isNotificationEnabled, notificationEmails, notificationSlackChannels, notificationEventTypes, ...pipelineRun } = values`. The virtual fields are stripped out, the rest spreads directly as the `PipelineRun` payload. No casts needed since `PipelineRunFormValues` already satisfies `PipelineRun`.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Fixed — the submit handler now destructures virtual notification fields out of `values` and spreads the rest as the full `PipelineRun` payload. Notifications are merged into `spec` on top of the spread, so the entire run (metadata, actor, pipeline, description, etc.) is always submitted.

javascript/packages/core/components/form/components/add-button/add-button.tsx
sallycr: Reverted — the `marginTop` was added to create spacing between the form group header and the add button, but the spacing should come from the parent layout, not the button itself. Reverted to the original `marginBottom` + `width` only.

javascript/packages/core/components/form/components/add-button/add-button.tsx
sallycr: Applied — reverted to the original `{ marginBottom: theme.sizing.scale600, width: '260px' }`, removing the `marginTop`. Thanks!

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
donovanlopez: @craigmarker I think `PipelineRunFormValues` extends `PipelineRun` so it wouldn't be a subset.

https://github.com/michelangelo-ai/michelangelo/blob/ddc2d35271e06a3ce8ad3e05c765e19bbf8cd65a/javascript/packages/core/config/entities/pipeline/types.ts#L3-L25

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -1,29 +1,77 @@
...
+    const {
+      isNotificationEnabled,
+      notificationEmails,
+      notificationSlackChannels,
+      notificationEventTypes,
+      ...pipelineRun
+    } = values;
+
+    const emails = notificationEmails ?? [];
+    const slack = notificationSlackChannels ?? [];
```
donovanlopez: nit: we should rename this variable to `slackDestinations` to make it more clear what this variable holds

javascript/packages/core/config/entities/pipeline/types.ts
```diff
@@ -9,3 +11,15 @@ export interface Pipeline {
     };
   };
 }
+
+/** Extended form values type for CreatePipelineRunForm, adding virtual notification fields. */
+export type PipelineRunFormValues = PipelineRun & {
```
donovanlopez: This is what the `PipelineRunFormValues` looks like destructured:
```
type PipelineRunFormValues = {
  metadata: {
    name: string;
    namespace: string;
  };
  spec: {
    actor: {
      name: string;
    };
    pipeline: {
      name: string;
      namespace: string;
    };
    description?: string;
    notifications?: PipelineRunNotification[];
  };

  isNotificationEnabled?: boolean;
  notificationEmails?: string[];
  notificationSlackChannels?: string[];
  notificationEventTypes?: Array<
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_SUCCEEDED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_FAILED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_KILLED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED'
  >;
};
```


Should we shape this to distinguish which part is the `pipelineRun` and which part is the form state?
```
type PipelineRunFormValues = {
  pipelineRun: PipelineRun;

  isNotificationEnabled?: boolean;
  notificationEmails?: string[];
  notificationSlackChannels?: string[];
  notificationEventTypes?: Array<
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_SUCCEEDED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_FAILED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_KILLED'
    | 'EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED'
  >;
};

```

javascript/packages/core/config/entities/run/types.ts
```diff
@@ -1,3 +1,25 @@
...
+] as const;
+
+/** Trigger condition for a pipeline run notification. */
+export type NotificationEventType = (typeof ALL_NOTIFICATION_EVENT_TYPES)[number];
+
+/**
+ * A single notification channel attached to a pipeline run.
+ * Sent when any of the listed event types fires.
+ */
+export type PipelineRunNotification = {
```
donovanlopez: I noticed we use notifications for multiple CRDs. Do we want to create specific notification types for each CRD (ex. `ModelNotification`, `PipelineNotification`, `PipelineRunNotification`) or mirror the proto and have a generic shared `Notification` type?

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
donovanlopez, line 1: can we add a test that tests when we input an invalid email address?

javascript/packages/core/config/entities/run/types.ts
sallycr: let's discuss this topic with the team.

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Added in e3244609: a test that enters an invalid email and asserts the validation message appears. Also made the error message distinct from the field's default caption so the two states are distinguishable.

sallycr: Note: the branch was rebased onto main and squashed (e3244609 no longer exists) — the invalid-email test is now in 6d01418b, same file, still present and passing.

javascript/packages/core/config/entities/run/types.ts
donovanlopez: nit: can we rename from `PipelineRunNotification` to `Notification` to mirror the notification proto?

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,100 @@
...
+  { id: 'EVENT_TYPE_PIPELINE_RUN_STATE_KILLED', label: 'Killed' },
+  { id: 'EVENT_TYPE_PIPELINE_RUN_STATE_SKIPPED', label: 'Skipped' },
+];
+
+const EMAIL_REGEX = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+
+type NotificationDetailsProps = {
+  enabled: boolean;
+  value: NotificationDetailsValue;
+  onChange: (value: NotificationDetailsValue) => void;
```
donovanlopez: nit: can we change the name of this handler to `onNotificationDetailsChange`so it's more descriptive?

donovanlopez (approved):

javascript/packages/core/components/form/fields/select/build-select-overrides.ts
```diff
@@ -9,10 +9,21 @@ export function buildSelectOverrides(
   readOnly?: boolean
 ): SelectOverrides {
   // When rendered inside modals, dropdown options can overflow past the modal body into its backdrop.
+  // ignoreBoundary lets the dropdown escape the modal clip region; popperOptions ensures
```
craigmarker: there was already a comment about why ignoreBoundary is included

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -137,4 +137,109 @@ describe('CreatePipelineRunForm', () => {
...
+    );
+
+    await screen.findByRole('dialog', { name: 'Start new pipeline run' });
+
+    // Toggle is visible; email combobox is present but disabled
+    expect(screen.getByText('Send notifications')).toBeInTheDocument();
+    expect(screen.getByRole('combobox', { name: /email addresses/i })).toBeDisabled();
+
+    // Enable notifications via toggle
+    await user.click(screen.getByRole('checkbox', { name: /enabled|disabled/i }));
```
craigmarker: why is the regex matching enabled | disabled? i’ve never seen something like this before. what am i missing?

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -137,4 +137,109 @@ describe('CreatePipelineRunForm', () => {
...
+
+    // Email combobox is now enabled
+    await waitFor(() => {
+      expect(screen.getByRole('combobox', { name: /email addresses/i })).not.toBeDisabled();
```
craigmarker: not.toBeDisabled —> toBeEnabled?

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -137,4 +137,109 @@ describe('CreatePipelineRunForm', () => {
...
+    });
+  });
+
+  it('includes notifications in payload when toggle is on and email is entered', async () => {
```
craigmarker: why not include the emai submission in the previous test? there’s a lot of duplication here

craigmarker (changes requested):

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -1,27 +1,79 @@
...
+
   const createPipelineRunMutation = useStudioMutation<PipelineRun, PipelineRun>({
     mutationName: 'CreatePipelineRun',
   });
 
+  const handleNotificationEnabledChange = (event: React.FormEvent<HTMLInputElement>) => {
+    setNotificationEnabled(event.currentTarget.checked);
+  };
+
+  const handleNotificationDetailsChange = (details: NotificationDetailsValue) => {
```
craigmarker: what does this handler provide above and beyond the state setter?

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,100 @@
...
+      ...value,
+      eventTypes: value.eventTypes.includes(id)
+        ? value.eventTypes.filter((t) => t !== id)
+        : [...value.eventTypes, id],
+    });
+  };
+
+  return (
+    <>
+      <FormControl label="Email addresses" caption="Use valid email addresses." error={emailError}>
```
craigmarker: I’m not sure that the caption provides much more useful information than the field label

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Select
+          id="notificationEmails"
+          value={toSelectValue(value.emails)}
+          options={[]}
+          onChange={(params) => onChange({ ...value, emails: fromSelectParams(params) })}
+          disabled={!enabled}
+          multi
+          creatable
+          placeholder="user@example.com"
```
craigmarker: placeholders should be prefixed with “eg” to indicate it is not a default value. but probably not necessary for email address

craigmarker (changes requested):

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Already reads as `slackChannels` (destructured from `notificationDetails`) and maps to `slackDestinations` on the payload in 15cf9053 — should be clear now which is which.

javascript/packages/core/config/entities/pipeline/types.ts
sallycr: Moot now — PipelineRunFormValues was removed entirely in c50e1365/6d01418b per Craig's later feedback (FormDialog is typed directly as PipelineRun; notification UI state lives in local useState, not a virtual form-values type).

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Renamed to onNotificationDetailsChange in 15cf9053.

javascript/packages/core/components/form/fields/select/build-select-overrides.ts
sallycr: You're right, that was an unintentional restatement — fixed in 15cf9053 to only add the new preventOverflow-ordering rationale, keeping the original ignoreBoundary comment as the single source of truth for that part.

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Fair — that was a workaround for the toggle's accessible name being its own label text ("Enabled"/"Disabled"), which changes with state. Replaced with a literal match on 'Disabled' (the known initial state) in 15cf9053, which is much clearer.

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Done in 15cf9053.

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Agreed — merged into a single test ('notification fields are disabled until the toggle is switched on, then submit includes them') in 15cf9053, removing the duplicated toggle/enable setup.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Nothing beyond the state setter — that was a leftover from an earlier lint fix. Removed the wrapper; the prop is now named onNotificationDetailsChange (not onChange) so the require-handler-prefix rule doesn't force a handle-prefixed wrapper here, and a descriptive handleRecipientsChange is only needed where NotificationDetails' own onChange-shaped calls happen. See 15cf9053.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Agreed, removed in 15cf9053 — the label already says "Email addresses"; the caption added nothing beyond that. Kept the error message and the Slack caption, which does add format guidance not in the label.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Agreed it's not necessary for the email placeholder — left as-is per your note. Happy to add the "e.g." prefix to the Slack placeholder too if you'd like consistency there.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Correction to my last reply: the wrapper is still there, just renamed to handleRecipientsChange — it's required by this repo's local/require-handler-prefix ESLint rule, which flags any `on*`-shaped prop (not just `onChange`) unless the passed value starts with `handle`. The old name (handleNotificationDetailsChange) got flagged by a second rule, local/no-handler-mirror, for just repeating the prop name with the handle prefix and no added context — so it needed a name describing the effect instead, hence handleRecipientsChange. It's still a thin pass-through to the state setter; that's unavoidable here given the two lint rules, not a sign of missing logic.

javascript/packages/core/config/entities/run/types.ts
sallycr: Renamed `PipelineRunNotification` to `Notification` to mirror the proto (83811d50). It was only referenced within this one file, so this was a clean rename with no other call sites to update. Lint/format/typecheck all pass.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Added the "e.g." prefix to the Slack placeholder (816d9827) for consistency, per your note. Left the email placeholder as-is since you flagged that one as not necessary.

sallycr: Updated — added "e.g." to the email placeholder too for full consistency across both fields (d44f5848).

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
```diff
@@ -137,4 +137,88 @@ describe('CreatePipelineRunForm', () => {
...
+      expect(mockRequest).toHaveBeenCalledWith(
+        'CreatePipelineRun',
+        expect.objectContaining({
+          spec: expect.objectContaining({
+            notifications: [
+              expect.objectContaining({
+                emails: ['notify@example.com'],
+                slackDestinations: [],
+              }),
+            ],
```
craigmarker, line 191: Why does this test not assert `eventTypes` exists in the payload?

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -1,27 +1,79 @@
...
 import { generateSuffix } from '#core/utils/name-utils';
 
 import type { ActionComponentProps } from '#core/components/actions/types';
-import type { Pipeline } from '#core/config/entities/pipeline/types';
-import type { PipelineRun } from '#core/config/entities/run/types';
+import type { NotificationDetailsValue, Pipeline } from '#core/config/entities/pipeline/types';
+import type { NotificationEventType, PipelineRun } from '#core/config/entities/run/types';
+
+// Proto enum Notification.EventType numeric values (notification.proto).
+const EVENT_TYPE_TO_NUMBER: Record<NotificationEventType, number> = {
```
craigmarker: Anytime a new enum is added to `NotificationEventType`, onboarding support to the UI requires changing several different constants.

What do you think about consolidating into a single constant (probably exported from shared.ts) that has { id, label, protoValue } properties.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,108 @@
...
+export function NotificationDetails({
+  enabled,
+  value,
+  onNotificationDetailsChange,
+}: NotificationDetailsProps) {
+  const [css, theme] = useStyletron();
+
+  const emailError = value.emails.some((email) => !EMAIL_REGEX.test(email.trim()))
+    ? 'Enter valid email addresses, e.g. user@example.com.'
+    : undefined;
```
craigmarker: what prevents a user from submitting the form with an invalid email?

javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
sallycr: Now asserts eventTypes too (ab336fad): default selection covers all four trigger conditions, so the expected payload is `eventTypes: [1, 2, 3, 4]`.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Good call — consolidated into a single `NOTIFICATION_EVENT_TYPES` constant in run/types.ts with { id, label, protoValue } per entry (ab336fad). The id list, UI options, and proto-value lookup all derive from it now, so adding a trigger condition only needs one new entry.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Confirmed this was a real gap — fixed in dca594ec: handleRunSubmit now re-validates the email(s) and throws if invalid, which FormDialog catches to keep the dialog open and block the mutation (previously it submitted regardless of the inline error). Added a test asserting the request is never sent and the dialog stays open.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Following up — this whole debate is moot now: per your later feedback, PipelineRunFormValues was removed entirely and the form now types as FormDialog<PipelineRun> directly, with the notification UI state held in local useState. No more virtual-fields type, so no more casts/destructuring needed here.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -1,27 +1,86 @@
...
 
-    await createPipelineRunMutation.mutateAsync(values);
+    const { emails, slackChannels, eventTypes } = notificationDetails;
+
+    if (notificationEnabled) {
+      const emailError = getEmailValidationError(emails);
+      if (emailError) {
+        throw new Error(emailError);
+      }
+    }
```
craigmarker: Let's try not to reinvent the wheel, there is validation functionality built into the [Field API](https://github.com/michelangelo-ai/michelangelo/blob/62103e34744dc9217190134dde464ae60cd43cc2/javascript/packages/core/components/form/fields/types.ts#L84-L89)

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -57,6 +116,31 @@ export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<
...
         description="Optional. Helps identify this run in the pipeline run list."
       />
+
+      <FormGroup
+        title="Set Up Notifications"
+        description="Receive an alert when this run completes, fails, or is killed."
+      >
+        <div
+          className={css({ display: 'flex', flexDirection: 'column', gap: theme.sizing.scale600 })}
+        >
```
craigmarker: we shouldn't be adding one-off styling within the form components. layout should be entirely controlled by exports from [components/form/layout](https://github.com/michelangelo-ai/michelangelo/tree/main/javascript/packages/core/components/form/layout)

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,99 @@
...
+          value={toSelectValue(value.emails)}
+          options={[]}
+          onChange={(params) =>
+            onNotificationDetailsChange({ ...value, emails: fromSelectParams(params) })
+          }
+          disabled={!enabled}
+          multi
+          creatable
+          placeholder="e.g. user@example.com"
+        />
```
craigmarker: I don't think we should be using a select field with no options... instead we should support multi-input https://baseweb.design/components/input/#create-delete-tags

javascript/packages/core/config/entities/pipeline/notification-details.tsx
```diff
@@ -0,0 +1,99 @@
+import { useStyletron } from 'baseui';
+import { Checkbox } from 'baseui/checkbox';
+import { Select } from 'baseui/select';
+
+import { FormControl } from '#core/components/form/components/form-control';
```
craigmarker, line 4: we should be using exports from https://github.com/michelangelo-ai/michelangelo/tree/main/javascript/packages/core/components/form

craigmarker (changes requested):

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Fixed in 6073923e: notificationEmails/notificationSlackChannels are now registered as real (if virtual) fields via the new MultiInputField (built on useField), with validateEmails passed as the field's validate. handleRunSubmit no longer re-checks validity manually - Form's built-in submission gating (final-form's hasSyncErrors check) now blocks the mutation automatically, matching the Field API pattern used everywhere else in the form.

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
sallycr: Fixed in 6073923e: removed the one-off flex/gap wrapping div in create-pipeline-run-form.tsx. It was redundantly re-declaring spacing that Box (which FormGroup renders its children into) already provides via its own display:flex; flexDirection:column; gap styling - so removing the div and making NotificationDetails/the checkbox FormControl direct children of FormGroup preserves identical spacing with no duplicate CSS.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Fixed in 6073923e: added a new reusable MultiInputField component (components/form/fields/multi-input/) implementing the baseweb create-delete-tags pattern - type a value, press Enter to add it as a removable tag. Replaced both Select multi creatable options=[] inputs (emails, Slack channels) with it.

javascript/packages/core/config/entities/pipeline/notification-details.tsx
sallycr: Fixed in 6073923e: notification-details.tsx now uses the new MultiInputField from components/form/fields/multi-input instead of hand-assembling FormControl + a raw Select, consistent with how the rest of the form fields are built.

PR #1421, opened 2026-07-01
feat(trainer): tabular_trainer pure helpers module (_dataset.py) [PR 7b]

sallycr:
## Summary

- Ports 10 pure-function helpers from internal `lightning_dataset.py` to `python/michelangelo/workflow/tasks/tabular_trainer/_dataset.py`
- Replaces internal `pad_batch` (absent in OSS) with `_pad_row()` using `pad_ragged_lists` / `ast.literal_eval`
- Builds `construct_read_kwargs` by iterating `ParquetReadConfig` fields directly (no internal `parquet_read_config_to_kwargs` needed)
- 69 unit tests in `tests/dataset_test.py`, all passing
- No Uber-internal references; zero OSS-infra dependencies (no Ray cluster required)

## Functions ported

| Function | Purpose |
|---|---|
| `_map_torch_dtype_to_numpy` | dtype string → numpy type |
| `_map_torch_dtype_to_datatype` | dtype string → DataType enum |
| `get_model_schema` | builds ModelSchema from ColumnConfig dicts |
| `_pad_row` | OSS replacement for `pad_batch` |
| `collate_sample_row` | normalises a single sample row for `sample_data` |
| `get_sample_data` | extracts / type-casts features for inference testing |
| `_raise_hyperparameter_deprecation_warnings` | warns for deprecated hyperparameter keys |
| `_raise_trainer_config_deprecation_warnings` | extension point for future field deprecations |
| `raise_lightning_trainer_config_deprecation_warnings` | combines both warning helpers |
| `construct_read_kwargs` | builds `ray.data.read_parquet` kwargs |

## Related

- Closes #1359
- Builds on PR 7a (#1415) — imports `ParquetReadConfig`/`DataloadingConfig` from `ray_data_io`
- PR 7c (dispatcher + glue) will follow

## Test plan

- [x] `ruff format --check` clean
- [x] `ruff check` clean
- [x] `pytest` — 69/69 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1415, opened 2026-06-30
feat(trainer): add tabular_trainer config schema (PR 7a, issue #1359)

sallycr:
## Summary

Implements PR 7a from the [tabular_trainer implementation plan](https://vibe-mcp.uberinternal.com/v/webdocs/#/d/0918273f-7c5e-4c1b-a9d2-84be5d32f159) — the config schema layer, which the pure helpers (7b) and dispatcher (7c) will stack on top of.

**New files:**
- `python/michelangelo/workflow/schema/tabular_trainer.py` (~400 lines) — all configuration dataclasses and enums
- `python/michelangelo/workflow/schema/tests/tabular_trainer_test.py` (~250 lines) — 41 tests

**What's in the schema:**
- `CheckpointScoreOrder`, `IncrementalTrainingModeConfig` (str enums)
- `ColumnConfig`, `CometConfig`, `ScalingConfig`, `CheckpointConfig`
- `ParquetReadConfig`, `BatchIterConfig`, `DataloadingConfig`
- `LightningTrainerKwargs` — 44 passthrough Lightning `Trainer.__init__` kwargs + `__post_init__` exclusivity validator for `limit_*_batches` vs `limit_*_batches_count` pairs
- `TransferLearningSpecConfig`, `CustomTrainerConfig`
- `LightningTrainerConfig` — main config; `comet` now Optional; `data_collate_fn` dropped (use `dataloading_config.batch_iter_config.collate_fn`)
- `TabularTrainerConfig` — `__post_init__` enforces exactly one of `lightning`/`custom`; `TorchTrainerConfig` dropped (never dispatched)

**Design divergences from internal (intentional):**
- Plain `@dataclass` — no Pydantic/JSONData dependency
- `comet` is `Optional` on `LightningTrainerConfig` (was required internally)
- Top-level `data_collate_fn` dropped (use `batch_iter_config.collate_fn`)
- `TorchTrainerConfig` dropped (internal dead code, never dispatched)
- `ParquetReadConfig` simplified to single Ray version (OSS pins one Ray)
- `TabularTrainerConfig.__post_init__` fails at construction vs internal's call-time check

## Test plan

- [x] 41 unit tests — all passing locally
- [x] `ruff format --check` passes
- [x] `ruff check` passes
- [ ] CI green

Closes #1359 (part 1 of 3 — stacked with 7b and 7c)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1426, opened 2026-07-01
feat(trainer): tabular_trainer dispatcher + ModelMetadata fields (PR 7c)

### LINKED ISSUE #1359: [Trainer] tabular_trainer workflow task
The `workflow/tasks/` directory only has `pusher/`. There is no workflow task for training tabular/structured data models, which means users cannot include model training as a step in their workflow pipelines.

- Add `tabular_trainer` Canvas workflow task that:
  - Accepts training/validation datasets and optional pre-initialized model weights
  - Routes to Lightning or Custom Ray trainer backend based on config
  - Returns a `ModelVariable` with metadata (schema, sample data, training framework)
- Add utility module re-exporting dataset helper functions (`get_model_schema`, `get_sample_data`, `collate_sample_row`, `construct_read_kwargs`)
- Add comprehensive tests covering both Lightning and Custom trainer paths
- Register as a Canvas workflow task in the OSS task registry

- Requiring users to write custom training tasks: rejected because a standard tabular training task is a common need
- Deferring until full workflow module migrates: the task only needs schema + variables, not the full workflow engine

**Additional context**
- Estimated size: ~1,500 lines (including tests)
- Dependencies: Test suite issue, TrainingObserver issue, _internal utils issue
- Priority: **P2**

sallycr:
## Summary

Final PR in the tabular_trainer split (7a config schema, 7b pure helpers → **7c dispatcher + glue**).

- **`train_tabular()`** — top-level OSS entry point for tabular training. Dispatches lightning/custom; uses injected `StorageBackend`, `run_config`, `scaling_config`, `is_local_run` for full testability (no internal MAF/workflow-context dependencies)
- **`_train_lightning()`** — ports the Ray Train orchestration from internal `lightning_train.py`: dataset loading with post-load column projection, batch/collate resolution, experiment tracker wiring (`ExperimentTrackerConfig` → `CometParam`), `LightningTrainerParam` + `LightningTrainerWithStateDict`, schema/sample_data serialised to metadata blobs, model upload via `StorageBackend`
- **`ModelMetadata`** — adds `is_incremental_training: bool` and `baseline_model_identifier: str | None` fields; both serialised in `to_registry_dict()`
- **`__init__.py`** — re-exports all schema classes + `train_tabular` with `__all__`

### Gated with NotImplementedError (not yet in OSS)
- Custom trainer backend
- Mid-epoch checkpointing (`save_every_n_steps`)
- Transfer-learning spec building (`transfer_learning_spec`)

### Key divergences from internal
- Returns `ModelArtifact` (not `ModelVariable`) — OSS workflow-variable type
- `StorageBackend` injected (not `ModelVariable.save_lightning_model()`)
- `run_config`/`scaling_config` injected (not `maf.ray.train.create_*`)
- Column projection via post-load `.select_columns()` (OSS `load_ray_dataset()` takes no kwargs)

## Test plan

- [x] 42 unit tests in `tests/trainer_task_test.py` — all passing
- [x] Existing `workflow/variables/tests/` — 85 tests still passing (127 total)
- [x] `ruff format --check` + `ruff check` clean

## Related

- Closes #1359
- Stacked on 7a (#1415 merged) + 7b (#1421 merged)


sallycr:
## OSS Team Review — PR 7c: tabular_trainer dispatcher + ModelMetadata fields

Reviewed by: OSS Architect, OSS Dev Advocate, OSS Coder

All **blocking** issues have been addressed in the follow-up commit `46092c45`. Summary below for the record.

---

### BLOCKING (fixed in 46092c45)

**B1 · `initial_weights_path` pointed to a directory, not a file**
`storage_backend.upload(upload_tmp, key)` uploads a *directory* containing `model.pt`. On a subsequent warm-start run, `download()` reconstructs that directory in `_weights_tmp/`, but `initial_weights_path = _weights_tmp` passed the *directory* to `_load_weights_from_path`, which calls `torch.load` on the path — crashing at runtime. Fixed: `initial_weights_path = os.path.join(_weights_dir, "model.pt")`.

**B2 · `ExperimentTrackerConfig(mlflow=...)` silently dropped**
`_train_lightning` checked only `config.experiment_tracker.comet`. When a user configured MLflow, the config was ignored with no error and no log line — a silent no-op. Fixed: added a `NotImplementedError` gate matching the pattern already used for `save_every_n_steps` and `transfer_learning_spec`. New test `test_mlflow_config_raises_not_implemented` pins the gate.

**B3 · `_weights_tmp = tempfile.mkdtemp()` leaked on every warm-start run**
The download temp dir was never cleaned up (unlike the upload path which used `TemporaryDirectory`). Fixed: wrapped in `contextlib.ExitStack` so the dir is automatically removed after `trainer.train()` returns.

**B4 · `ConfigurationError` not re-exported from `__init__.py`**
`train_tabular()` raises `ConfigurationError` but it wasn't importable from `michelangelo.workflow.tasks.tabular_trainer`. Fixed: added to imports and `__all__`.

**B5 · Duplicate `patch(trainer_task.torch)` in `_run_train` test helper**
Two identical context managers. The first was shadowed (unreachable), masking a missing `LightningTrainerParam` patch in the helper. Fixed: removed the duplicate.

---

### SUGGESTIONS (non-blocking, carry-forward)

- **Export `ApplyIncrementalTrainingMetadataFn`** from `__init__.py` — needed by anyone writing a custom hook
- **`storage_backend` annotation** — declared as `StorageBackend` but `None` is runtime-guarded; annotation should be `StorageBackend | None` or the docstring note about `ConfigurationError(None)` removed
- **`train_tabular()` `Raises` section** — `NotImplementedError` gates not listed; add entries for `save_every_n_steps`, `transfer_learning_spec`, and `mlflow` cases
- **`baseline_model_identifier` docstring** — "None for BASELINE-start models" is misleading since those runs *do* set `is_incremental_training=True`; clarify as "None for the first run of a new incremental chain"
- **`import dataclasses` inside function** — move to module level (stdlib, no cost, inconsistent with rest of codebase)
- **`sample_data[:5] if sample_data else []`** — `get_sample_data` always returns a list; simplify to `sample_data[:5]`
- **`initial_model(incremental) + BASELINE` combination** — behavior (chain wins over BASELINE re-declaration) is correct but undocumented and untested; add a test case and docstring note

---

### No Uber internals
All 5 PR files clean — zero references to Jira, internal URLs, employee names, or Uber-specific infra.

python/michelangelo/workflow/tasks/tabular_trainer/task.py
```diff
@@ -0,0 +1,397 @@
+"""Dispatcher and lightning glue for the tabular trainer workflow task."""
```
kenns29, line 1: can this file just be called task.py

kenns29 (approved):

sallycr:
> @kenns29 — good catch on the naming. Checked both repos:

**Internal SDK**: every task module uses `task.py` uniformly (`calibrator/task.py`, `comparator/task.py`, `deploy_model/task.py`, etc.).

**OSS current state**: `pusher/pusher.py` uses a different convention (named after the task). `tabular_trainer/trainer_task.py` is a hybrid that's slightly redundant — "trainer" is already implied by the parent directory.

**OSS Architect view**: The `__init__.py` already hides the internal filename from all public imports — users see `from michelangelo.workflow.tasks.tabular_trainer import train_tabular` regardless. So `task.py` vs `trainer_task.py` is a pure internal-module concern. That said, adopting `task.py` as the OSS standard *does* align with the internal convention and makes the layout more predictable as more task modules land.

**OSS Coder view**: Agree with the direction, but renaming only `tabular_trainer` while leaving `pusher/pusher.py` creates a new inconsistency within OSS. The cleanest outcome is to establish `task.py` as the OSS-wide standard and rename `pusher.py` → `task.py` in the same sweep (or a follow-up). Doing it mid-review in this PR adds mechanical churn without blocking value.

**Recommendation**: rename `trainer_task.py` → `task.py` as part of a follow-up "standardise task filenames" commit that covers both `tabular_trainer` and `pusher`. Happy to do that right after this PR merges — or we can do it now if you feel strongly. Not a blocker for this PR since `__init__.py` fully encapsulates it.

sallycr:
Done — renamed in this PR:  +  committed (3e0cb29). The  public API and all external import paths are unchanged. Issue [#1428](https://github.com/michelangelo-ai/michelangelo/issues/1428) tracks the follow-up rename of `pusher/pusher.py → pusher/task.py` to complete the convention.

PR #1430, opened 2026-07-02
feat(python): migrate tabular_assembler to OSS (buckets A-E)

krishpatel9:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Migrates the `tabular_assembler` workflow task (the Train → Package → Register model-packaging stage) from Uber's internal CanvasFlex SDK to OSS `michelangelo`, across five work buckets (A–E), combined on this one branch/PR for review. It will be split into per-bucket PRs after this review.

- **Bucket A** — extended `DataType` to the full schema enum set; added `data_type_to_torch_dtype`/`torch_dtype_to_data_type`; ported `additional_imports` support; added `additional_import_prefixes`/`sample_data`/`triton_parameters` params to `CustomTritonPackager`.
- **Bucket B** — `TabularAssemblerConfig`/`CustomAssemblerConfig`/`TorchAssemblerConfig` dataclasses. (An earlier revision of this branch added a shared `PackagerBase` ABC; it was removed after review — internal has no equivalent shared base across its own packagers, and both `CustomTritonPackager`/`TorchTritonPackager` remain independent classes.)
- **Bucket C** — the core `task.py` dispatch, the custom (Python-backend) assembler, and `_private/` helpers (schema fuse, data fuse, model_class resolve, file download — the latter now routes through the injected `StorageBackend` rather than a raw fsspec bypass); extended `ModelMetadata` with `schema`/`sample_data`/`transform_spec`/`feature_stats`, propagated onto the packaged model's returned metadata.
- **Bucket D** — the torch (PyTorch/Lightning) assembler and `FusedModel` (model fuser), wired into `task.py`'s dispatch.
- **Bucket E** — the model fuser's `fuse.py` engine and the ported test suite (native-transform-gated tests are present but skipped until a future PR migrates native-transform fusion). Native-transform support is represented only as a documented gap (`_build_tx_hydra_spec` raises `NotImplementedError`) — an earlier revision also added a `TransformSpec` protocol stub with no real caller; it was removed as dead documentation-only code after review.
- Follow-up fixes found during validation and review: (1) the custom packager, and the assembler's own dispatch resolution (`resolve_training_framework`), both accept structurally-conformant model classes via a `runtime_checkable ModelProtocol`, rather than requiring literal inheritance from the OSS `Model` ABC; (2) `CustomAssemblerConfig` gained `include_import_prefixes` so callers can bound the packager's import-dependency walk, which was previously unbounded; (3) both packagers' `model_revision` now defaults to `None` consistently; (4) an unrecognized `training_framework` now raises rather than silently returning an empty package pair.

**Why?**

Part of the broader effort to make uber-one's internal CanvasFlex SDK eventually import this OSS package directly instead of maintaining a parallel internal implementation. `tabular_assembler` is the model-packaging stage of that SDK; this PR ports it with a pluggable-backend design so both internal and external consumers share one implementation.

**How did you test it?**

Unit suite: 144 passed / 2 skipped (native-transform, intentionally deferred) / 0 failed in the `tabular_assembler` package; 376 passed in the broader `model_manager` package. `ruff check` and `ruff format` clean; verified no Uber-specific references leak into the OSS code.

Beyond unit tests, this was validated end-to-end via an internal "import-back" simulation — uber-one importing this OSS package back in place of its current internal implementation and running real local pipelines against it. 2 of 5 pipelines exercising this code went fully green (real model packaging, artifact upload, model registration, and inference); the remaining 3 are blocked only on native-transform fusion, which this PR intentionally defers.

This PR also went through two independent review passes (a 4-agent OSS-quality review and a separate coding-agent review), both of which found and led to fixes for real gaps — see the follow-up fixes listed above.

**Potential risks**

This adds a new task under `workflow/tasks/`; it does not change any existing task's behavior. `CustomTritonPackager`/`TorchTritonPackager` remain independent classes (no shared base); the custom packager's model-class validation was loosened to structural typing, which is additive and covered by the existing packager test suite.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Adds the `tabular_assembler` workflow task to OSS `michelangelo`. No schema or config migration required for existing users.

**Documentation Changes**

All new public classes/functions carry Google-style docstrings (Args/Returns/Raises). No wiki updates in this PR.


sallycr:
While validating this branch end-to-end against a real trained model (`michelangelo-examples`' `california_housing/pytorch_train` pipeline, using `tabular_trainer.train_tabular()`'s real output), I found a real bug in `torch_assembler`/the Triton packaging path here — reporting it, not patching, since this is your branch.

**Scalar (shape-less) `ColumnConfig` columns break packaging.**

Calling `tabular_assembler()` against a real `train_tabular()`-produced `ModelVariable` (all-scalar float features, the common tabular case) raises:

```
ValueError: Shape must be provided for item: ModelSchemaItem(name='MedInc', data_type=<DataType.FLOAT: 20>, shape=[])
```

from `lib/model_manager/_private/schema/triton/validate_schema.py:57-58`:

```python
if not item.shape or len(item.shape) == 0:
    return False, ValueError(f"Shape must be provided for item: {item}")
```

`ColumnConfig`'s own docstring and doctest (`workflow/schema/tabular_trainer.py:77-93`) document `ColumnConfig("torch.float32")` with no `shape` as valid, standard usage for a scalar column, and `get_model_schema()` passes that shape straight through unmodified into the `ModelSchemaItem`s this validator rejects. So any trainer config built the documented way — a plain tabular model with scalar features and no embeddings — can't be packaged by `torch_assembler` today.

Repro: standalone script training a real Lightning checkpoint via `train_tabular(config, train_dv, validation_dv, is_local_run=True)` (config using bare `ColumnConfig("torch.float32")` per-feature, no `shape=`), then calling `tabular_assembler(TabularAssemblerConfig(), raw_model, storage_backend=...)` against it — fails on the first scalar input column.

Seems like it needs either a `shape=[1]` normalization step somewhere in `torch_assembler`/`torch_triton_packager` for scalar columns, or the Triton validator itself accepting an empty shape as "scalar" (shape `[1]`) rather than rejecting it outright. Happy to share the full repro script if useful.

(Separately, unrelated to your code: `train_tabular()` only populates `ModelMetadata`'s serialized `_schema`/`_sample_data` payloads, not the live `schema`/`sample_data` fields `torch_assembler` reads at `torch/assembler.py:146` — callers need to unpickle those themselves before calling in. Worth a docs note on `tabular_assembler`'s expected input, at least, even though it's not this branch's bug to fix.)

python/michelangelo/lib/model_manager/schema/data_type.py
```diff
@@ -37,8 +53,19 @@ class DataType(Enum):
 
     INVALID = 0
     UNKNOWN = 1
+    NUMERIC = 2
+    ARRAY = 3
     BOOLEAN = 4
+    DATE = 5
+    HIVE_STRING = 6
     STRING = 7
+    TIMESTAMP = 8
```
kenns29: these are spark types, we don't want them in OSS repo

python/michelangelo/lib/shared/utils/model_fuser/fused_model.py
```diff
@@ -0,0 +1,98 @@
+"""Fused model composing a native-transform module and a predictor module.
```
kenns29, line 1: The fuser has been moved to uber/ai/michelangelo/shared/utils/model_fuser in internal, fuser, especially FusedModel, should not be under assembler

python/michelangelo/workflow/variables/metadata.py
```diff
@@ -90,6 +108,10 @@ def to_registry_dict(self) -> dict[str, str]:
     deployable: bool = False
     is_incremental_training: bool = False
     baseline_model_identifier: str | None = None
+    schema: ModelSchema | None = None
```
kenns29: these can cause IO error in unflow due to size limit of unix command line arg and cadence, refer to the _schema BytesIO approach in internal

python/michelangelo/lib/model_manager/interface/custom_model.py
```diff
@@ -78,3 +79,59 @@ def predict(
             schema, and each value should be a numpy array containing the
             predictions.
         """
+
+
+@runtime_checkable
+class ModelProtocol(Protocol):
```
kenns29: why do we create another custom model interface?

python/michelangelo/workflow/variables/metadata.py
kenns29: plz address this comment

python/michelangelo/lib/model_manager/_private/utils/module_finder/dependency_files.py
```diff
@@ -82,26 +82,44 @@ def find_dependency_files_internal(
...
+        # dependency walk over one unrelated, unimportable submodule.
+        #
+        # onerror only intercepts Exception subclasses, not SystemExit
+        # (BaseException, not Exception) -- and real-world packages ship
+        # __main__.py submodules with unguarded argparse/sys.exit() calls at
+        # import time (e.g. torch.utils.model_dump.__main__), which
+        # walk_packages() imports while probing whether it is itself a
+        # package. Wrap the walk too, so one such submodule doesn't abort
+        # dependency discovery for the rest of the package.
+        try:
```
kenns29, line 98: corrupted/missed import files can make the model unable to load, did we test this logic?

python/michelangelo/lib/shared/utils/model_fuser/fuse_schema.py
```diff
@@ -0,0 +1,75 @@
+"""Schema fusion for native-transform + predictor models.
```
kenns29, line 1: why is this util in the assembler and not in the fuser?

python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
```diff
@@ -0,0 +1,351 @@
...
+__all__ = ["torch_assembler"]
+
+# Triton backend identifiers accepted by ``TorchAssemblerConfig.backend``.
+# These mirror ``michelangelo.lib.model_manager._private.constants
+# .triton_backend_type.TritonBackendType.{PYTHON,ONNX}`` values, duplicated
+# here as plain strings rather than imported: that module is private to
+# ``model_manager`` and the assembler task layer never crosses that
+# boundary (see ``TorchAssemblerConfig.backend``'s docstring, which already
+# documents the accepted string values).
+_BACKEND_PYTHON = "python"
```
kenns29: why do we create additional constants instead of using the the TritonBackend constant?

python/michelangelo/workflow/tasks/tabular_assembler/_private/data/fuse.py
```diff
@@ -0,0 +1,67 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import numpy as np
+
+
+def fuse_sample_data(
```
kenns29, line 11: why is this util in assembler and not in the fuser?

python/michelangelo/workflow/schema/assembler.py
```diff
@@ -0,0 +1,114 @@
...
+
+Consumers may subclass these to add provider-specific fields (e.g. a custom
+storage toggle) without modifying this module.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+__all__ = [
```
kenns29: you don't need this constant

python/michelangelo/lib/model_manager/schema/data_type.py
```diff
@@ -37,6 +41,7 @@ class DataType(Enum):
 
     INVALID = 0
     UNKNOWN = 1
+    NUMERIC = 2
```
kenns29: Don't add this in OSS, Numeric is simply Double

python/michelangelo/lib/shared/utils/model_fuser/fuse_schema.py
krishpatel9: moved to python/michelangelo/lib/shared/utils/model_fuser/

python/michelangelo/lib/shared/utils/model_fuser/fused_model.py
krishpatel9: moved to python/michelangelo/lib/shared/utils/model_fuser/

python/michelangelo/workflow/tasks/tabular_assembler/_private/data/fuse.py
krishpatel9: now moved to python/michelangelo/lib/shared/utils/model_fuser/

python/michelangelo/lib/model_manager/_private/utils/module_finder/dependency_files.py
krishpatel9: Added unit test, and tested with local pipeline run

python/michelangelo/workflow/variables/metadata.py
```diff
@@ -90,11 +106,45 @@ def to_registry_dict(self) -> dict[str, str]:
     deployable: bool = False
     is_incremental_training: bool = False
     baseline_model_identifier: str | None = None
+    transform_spec: dict[str, Any] | None = None
```
kenns29: these needs to be handled by the BytesIO too

python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
```diff
@@ -0,0 +1,353 @@
...
+from michelangelo.workflow.variables.types import AssembledModel, ModelArtifact
+
+if TYPE_CHECKING:
+    from michelangelo.lib.artifact_manager.storage_backend import StorageBackend
+    from michelangelo.workflow.schema.assembler import TabularAssemblerConfig
+
+__all__ = ["torch_assembler"]
+
+
+def _normalize_scalar_shapes(
```
kenns29: these should be private, check the internal impl, and we should not alter the structure too much

python/michelangelo/lib/shared/utils/model_fuser/fuse.py
```diff
@@ -0,0 +1,1145 @@
+"""Fuse a PyTorch predictor with a preceding native-transform model.
```
kenns29, line 1: this file is very outdated, check the internal implementation, a lot of torchscript and onnx conversion has been decoupled with fuse.py, try pull the latest main and run again

python/michelangelo/workflow/variables/metadata.py
```diff
@@ -92,16 +127,109 @@ def to_registry_dict(self) -> dict[str, str]:
...
+        the only thing carried across a workflow task boundary is the
+        serialised payload, never a live object passed by value.
+
+        Returns:
+            The unpickled ``ModelSchema``, or ``None`` if ``_schema`` is unset.
+        """
+        if self._schema is None:
+            return None
+        self._schema.seek(0)
+        return pickle.loads(self._schema.read())
```
kenns29: instead of duplicating these pickling and unpickling logic, we need to have a _private util for BytesIO serialization, refer to the implementation in uber/ai/michelangelo/sdk/workflow/variables/_private/utils/serialization.py

python/michelangelo/workflow/tasks/tabular_assembler/_private/schema/scalar_shapes.py
```diff
@@ -0,0 +1,94 @@
...
+
+from dataclasses import replace
+from typing import Any
+
+import numpy as np
+
+from michelangelo.lib.model_manager.schema import ModelSchema, ModelSchemaItem
+
+
+def normalize_scalar_shapes(
```
kenns29: Sample data is to be either inferred from the schema or obtain from the previous steps. We always assume the sample data is in the proper shape in the assembler, we validate sample data against the schema in the packager, but we don't want to modify sample data, we should NEVER modify or rename fields in the sample data or schema

kenns29 (approved):

PR #1441, opened 2026-07-06
fix(trainer): default RunConfig storage to backend bucket on multi-node runs

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`train_tabular()` previously defaulted `ray.train.RunConfig.storage_path` to a local tempdir whenever the caller didn't pass `run_config` explicitly — regardless of `is_local_run`. That path is only visible to the process that created it, so on a real multi-node Ray cluster, worker pods can't read/write it, silently breaking Ray Train's distributed checkpointing.

This PR:
- Adds `MinioStorageBackend.to_ray_storage_target(path_prefix="ray_train")`, which builds a `pyarrow.fs.S3FileSystem` from the backend's own connection details and returns a `(storage_path, storage_filesystem)` pair pointing at the same bucket already used for the final model upload.
- Updates `train_tabular()`'s default-`RunConfig` construction: when `run_config` is `None` and `is_local_run=False`, it now uses `storage_backend.to_ray_storage_target()` if the backend implements it (checked via `getattr(..., None)` — `StorageBackend` does not declare this method, so it's opt-in per backend). Otherwise (local runs, or backends without object storage) it falls back to the previous local-tempdir behavior, unchanged.

**Why?**

Found via sandbox testing: training via `train_tabular()` worked locally but the default `RunConfig` couldn't share checkpoint state across nodes on a real cluster. Pointing the default at the same backend already configured for model upload fixes this without requiring every caller to construct their own `RunConfig` by hand.

**How did you test it?**

`pytest michelangelo/lib/artifact_manager/ michelangelo/workflow/tasks/tabular_trainer/ michelangelo/lib/trainer/ -q` — 323 passed (9 new tests: 5 for `to_ray_storage_target()`, 4 for the `train_tabular` default-`RunConfig` branch, covering remote+capable-backend, local-run-always-tempdir, remote+incapable-backend, and explicit-`run_config`-not-overridden). `ruff format --check` / `ruff check` clean.

**Potential risks**

Low. The only behavior change is for callers who (a) don't pass `run_config`, (b) pass `is_local_run=False` (the default), and (c) use a `storage_backend` that implements `to_ray_storage_target()` (currently only `MinioStorageBackend`) — these now get cluster-shared storage instead of a broken local tempdir, which is strictly a fix. All other combinations (local runs, other backends, explicit `run_config`) are unchanged.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

See CHANGELOG.md — Fixed: default `RunConfig` storage no longer breaks on multi-node clusters.

**Documentation Changes**

`train_tabular()`'s `run_config` docstring updated to describe the new default-selection logic. `MinioStorageBackend.to_ray_storage_target()` fully documented (Google-style Args/Returns/Raises).


python/michelangelo/workflow/tasks/tabular_trainer/task.py
```diff
@@ -293,13 +299,27 @@ def _train_lightning(
...
+    # multi-node cluster. Duck-typed since StorageBackend does not declare
+    # to_ray_storage_target(); only object-store backends implement it.
     if run_config is None:
-        _tmp_storage = tempfile.mkdtemp(prefix="michelangelo_train_")
-        run_config = ray.train.RunConfig(
-            checkpoint_config=checkpoint_config,
-            storage_path=_tmp_storage,
-        )
+        to_ray_storage_target = getattr(storage_backend, "to_ray_storage_target", None)
+        if to_ray_storage_target is not None and not is_local_run:
```
kenns29: this type of logic should not be in the task.py level, it should be inside the DatasetVariable, we have close to 10 tasks and each one have multiple interactions with storage IO and having each one to deal with multiple storage backend in the task code is not sustainable

kenns29: Internally, we use fsspec to deal with different type of storage backend and in fact we never worry about storage backend in the application code

sallycr: Good catch, agreed - refactored. Added michelangelo.uniflow.plugins.ray.create_run_config(), a shared helper that resolves storage_path/storage_filesystem from the UF_STORAGE_URL env var (the same convention DatasetVariable/ModelVariable already use for their own storage location), built on the filesystem resolver already in uniflow/plugins/ray/io.py (_fs_path/resolve_fs, already used by RayDatasetIO).

Removed MinioStorageBackend.to_ray_storage_target() and the per-task getattr/duck-typing entirely. train_tabular() now just calls create_run_config() with no storage_backend involvement in RunConfig construction, so this generalizes to future tasks without each one re-deriving storage on its own.

kenns29 (approved):

PR #1438, opened 2026-07-04
ci: notify Slack on scheduled/release workflow failures

sallycr:
## Summary
- Adds a `notify-on-failure` job to `nightly.yml` that posts to Slack via `slackapi/slack-github-action` when any nightly job fails, with a direct link to the failing run.
- **Consolidates the channel-id fix** for the same `notify-on-failure` pattern that was already merged into `main` via #1398 (cve-scan.yml), #1399 (release.yaml), #1400 (cleanup-nightlies.yml), and #1403 (compat-matrix.yml) — those PRs merged while this one was still open, carrying the placeholder `#michelangelo-releases` channel-id.
- Per paulzim's review feedback (same note across all 5 PRs): `channel-id` must be a real Slack channel ID (`Cxxxxxxxxxx`), not a `#channel-name` string, and CI failures should route to `#ma-oss-ci`. All 5 workflows now use the actual channel ID `C0BER0RCQ8P`.

Requires the `SLACK_BOT_TOKEN` repo secret (already configured) with `chat:write` scope, and the bot invited to the `#ma-oss-ci` channel.

Closes #1436.

## Test plan
- [x] `actionlint` passes on all 5 modified workflow files
- [ ] After merge, trigger via `workflow_dispatch` with a deliberately broken step (or wait for a real failure) to confirm the Slack message is posted to `#ma-oss-ci`

Issue #1436, opened 2026-07-03
ci: add failure notifications for scheduled workflows


sallycr:
## Problem

Scheduled workflows (compat matrix, nightly cleanup, CVE scan) have no notification mechanism beyond the default GitHub email. When a weekly workflow fails, the only signal is an email to the repo owner — easy to miss, not actionable for the team.

Currently there are zero notification integrations in `.github/workflows/`.

## Proposed solution

Add a failure notification step to each scheduled workflow that posts to a team Slack channel (e.g. `#michelangelo-releases` or `#ci-alerts`). This gives the team a visible, actionable signal when a scheduled job fails.

```yaml
- name: Notify Slack on failure
  if: failure()
  uses: slackapi/slack-github-action@v1
  with:
    channel-id: '#michelangelo-releases'
    slack-message: |
      ❌ *${{ github.workflow }}* failed
      Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
  env:
    SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
```

## Affected workflows

- `.github/workflows/compat-matrix.yml` (weekly Monday)
- `.github/workflows/cleanup-nightlies.yml` (daily)
- `.github/workflows/cve-scan.yml` (weekly Monday)
- `.github/workflows/nightly.yml` (daily)

## Pre-requisites

- Slack bot token stored as `SLACK_BOT_TOKEN` repo secret
- Target channel decided (`#michelangelo-releases`, `#ci-alerts`, or similar)

## Alternatives considered

- **GitHub Issue on failure** — creates a trackable issue per failure; good for OSS but noisy for internal teams
- **Email only** — current state; insufficient for a team workflow

Tracking: #1340

sallycr:
Work started: #1438 adds Slack notification to `nightly.yml`. The other three target workflows (`cve-scan.yml` in #1398, `cleanup-nightlies.yml` in #1400, `compat-matrix.yml` in #1403) already got the same `notify-on-failure` job added as follow-up commits since those files only exist on those open PR branches. Also added to the tag-triggered `release.yaml` in #1399 since it's the highest-value release-critical workflow to alert on.

All notification jobs post to `#michelangelo-releases` using the `SLACK_BOT_TOKEN` secret (now configured). This issue will close once #1438, #1398, #1400, #1403, and #1399 are all merged.

PR #1444, opened 2026-07-07
fix(ci): CVE Scan fails to resolve trivy-action/trivy binary version

### LINKED ISSUE #1445: ci: CVE Scan fails — trivy-action tag unresolvable and trivy binary release pulled
## Problem

Every run of the `CVE Scan` workflow (`.github/workflows/cve-scan.yml`) fails immediately, before any actual scanning happens. Found while investigating run https://github.com/michelangelo-ai/michelangelo/actions/runs/28831990997.

## Root causes (two stacked issues)

1. **Unresolvable action tag** — `aquasecurity/trivy-action@0.28.0` was never a valid ref; the action only publishes `v`-prefixed tags (`v0.28.0`). This was wrong since the workflow was introduced in #1398, but went unnoticed because the `workflow_run` trigger (fires after Dev Release / UI Release / Nightly Build complete) hadn't actually fired until now.
2. **Pulled upstream release** — after fixing the tag, `trivy-action@v0.32.0`'s default `version: v0.64.1` (the Trivy scanner binary, not the action) points at a Trivy release whose GitHub Release/assets have since been removed upstream (the git tag still exists, but the Release page 404s). This broke the install step immediately after resolving the version tag, with no clear error surfaced in the log.

## Fix

- Add the `v` prefix to the trivy-action ref.
- Explicitly pin `version: v0.72.0` (a currently published Trivy release) instead of relying on the action's stale default.

Verified via `workflow_dispatch`: the scan now actually runs, downloads its vulnerability DB, and scans all 4 images. All 4 `scan (*)` jobs still report `failure`, but that's the *intended* signal — `exit-code: '1'` is configured to fail when CRITICAL/HIGH CVEs are found, and the images do have real findings (a separate, genuine security issue, out of scope here). The `notify-on-failure` job (#1436) correctly fires.

## Fix PR

#1444

sallycr:
Closes #1445

## Summary

Investigated https://github.com/michelangelo-ai/michelangelo/actions/runs/28831990997 (`CVE Scan` failing on every job).

Two separate, stacked upstream breakages, both now fixed:

1. **`aquasecurity/trivy-action@0.28.0` was never a valid ref** — the action only publishes `v`-prefixed tags (`v0.28.0`). This has apparently been wrong since the workflow was first added in #1398, but went unnoticed because the `workflow_run` trigger (fires after Dev Release / UI Release / Nightly Build complete) hadn't actually run until now.
2. **Even after fixing the tag**, `trivy-action@v0.32.0`'s default `version: v0.64.1` (the Trivy *scanner binary*, not the action) points at a Trivy release whose GitHub Release/assets have since been pulled upstream (the git tag still exists, but `gh api repos/aquasecurity/trivy/releases/tags/v0.64.1` 404s). This caused the install script to fail immediately after resolving the tag, with no useful error surfaced. Pinned to `v0.72.0`, which is currently published.

## Verification

Tested via `workflow_dispatch` on the fix branch (3 iterations, since the second failure only appeared after fixing the first):
- Confirmed action now resolves and Trivy actually runs, downloads its vulnerability DB, and scans all 4 images.
- All 4 `scan (*)` jobs still show `failure` — but that's now the *intended* signal: `exit-code: '1'` is configured so the job fails when CRITICAL/HIGH CVEs are found, and the images do have real CRITICAL/HIGH findings (a separate, genuine finding — not a CI infra bug).
- Confirmed `notify-on-failure` (added in #1436 / #1438) correctly fired and would have posted to `#ma-oss-ci`.

## Test plan
- [x] `actionlint` passes on `cve-scan.yml`
- [x] `workflow_dispatch` run confirms Trivy actually executes and scans images
- [x] `notify-on-failure` job fires correctly on scan failure
- [ ] Separate follow-up needed to actually remediate the CRITICAL/HIGH CVEs found in the container images (tracked separately, out of scope for this CI fix)

### DIFF
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -36,6 +36,7 @@ jobs:
           output: trivy-${{ matrix.service }}.sarif
           severity: CRITICAL,HIGH
           exit-code: '1'
+          version: v0.72.0
 
       - name: Upload SARIF to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v3
@@ -52,6 +53,7 @@ jobs:
           format: table
           severity: CRITICAL,HIGH
           exit-code: '0'
+          version: v0.72.0
 
       - name: Upload Trivy SARIF report as artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Run Trivy vulnerability scanner (${{ matrix.service }})
-        uses: aquasecurity/trivy-action@v0.28.0
+        uses: aquasecurity/trivy-action@v0.32.0
         with:
           image-ref: ${{ env.REGISTRY }}/${{ env.ORG }}/${{ matrix.service }}:latest
           format: sarif
@@ -45,7 +45,7 @@ jobs:
           category: trivy-${{ matrix.service }}
 
       - name: Run Trivy table output for summary
-        uses: aquasecurity/trivy-action@v0.28.0
+        uses: aquasecurity/trivy-action@v0.32.0
         if: always()
         with:
           image-ref: ${{ env.REGISTRY }}/${{ env.ORG }}/${{ matrix.service }}:latest
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - name: Run Trivy vulnerability scanner (${{ matrix.service }})
-        uses: aquasecurity/trivy-action@0.28.0
+        uses: aquasecurity/trivy-action@v0.28.0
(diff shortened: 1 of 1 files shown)


paulzim (approved):
Straightforward fix for a broken scan workflow — both changes are forced by upstream breakage. Approved.

**`aquasecurity/trivy-action@0.28.0` → `@v0.32.0` ✓**

`@0.28.0` is unresolvable (the tag no longer exists or was never published without the `v` prefix). `@v0.32.0` uses the correct conventional tag format for this action. Applied consistently to both scan steps (SARIF and table).

**`version: v0.72.0` added to both steps ✓**

The underlying Trivy binary `v0.64.1` was pulled from upstream. Adding an explicit `version:` pin is also better hygiene than relying on whatever binary the action wrapper bundles — it makes the scan reproducible and the binary version auditable. Both steps pinned identically.

**No other changes.** The rest of `cve-scan.yml` is untouched.

Approved.

PR #1455, opened 2026-07-07
Fold successOperations into useStudioMutation

craigmarker:
**Summary**
Move successOperations from MutationActionConfig onto MutationConfig, and run them automatically inside useStudioMutation instead of requiring each caller to wire useSuccessOperations by hand. useStudioMutation runs the mutation's successOperations on success, and also invokes clientOptions.onSuccess if one is supplied.

ConfirmDispatcher was the only caller, and it had to remember to invoke useSuccessOperations itself after every successful mutation. That pattern doesn't scale to other dispatch paths (e.g. a future bare mutation action with no confirm modal) without duplicating the same wiring at each new call site.

**Test Plan**
Updated trigger.ts to include toast successOperations and verified it appeared on successful trigger kill action.

### DIFF
--- a/javascript/packages/core/hooks/__tests__/use-studio-mutation.test.ts
+++ b/javascript/packages/core/hooks/__tests__/use-studio-mutation.test.ts
@@ -1,4 +1,4 @@
-import { renderHook, waitFor } from '@testing-library/react';
+import { renderHook, screen, waitFor } from '@testing-library/react';
 import { vi } from 'vitest';
 
 import { GrpcStatusCode } from '#core/constants/grpc-status-codes';
@@ -10,6 +10,7 @@ import {
   createQueryMockRouter,
   getServiceProviderWrapper,
 } from '#core/test/wrappers/get-service-provider-wrapper';
+import { getSnackbarProviderWrapper } from '#core/test/wrappers/get-snackbar-provider-wrapper';
 import { ApplicationError } from '#core/types/error-types';
 
 import type { ErrorNormalizer } from '#core/types/error-types';
@@ -27,6 +28,7 @@ describe('useStudioMutation', () => {
         getErrorProviderWrapper(),
         getRouterWrapper({ location: '/ma-dev-test' }),
         getServiceProviderWrapper({ request: mockRequest }),
+        getSnackbarProviderWrapper(),
       ])
     );
 
@@ -49,6 +51,7 @@ describe('useStudioMutation', () => {
         getErrorProviderWrapper(),
         getRouterWrapper({ location: '/ma-dev-test' }),
         getServiceProviderWrapper({ request: mockRequest }),
+        getSnackbarProviderWrapper(),
       ])
     );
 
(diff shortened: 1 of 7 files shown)


javascript/packages/core/hooks/__tests__/use-studio-mutation.test.ts
```diff
@@ -60,6 +63,59 @@ describe('useStudioMutation', () => {
...
+        getRouterWrapper(),
+        getServiceProviderWrapper({ request: mockRequest }),
+        getSnackbarProviderWrapper(),
+      ])
+    ).result.current.mutate({ name: 'test-run' });
+
+    expect(await screen.findByText('Pipeline run created')).toBeInTheDocument();
+  });
+
+  test('clientOptions.onSuccess fully replaces default successOperations', async () => {
```
donovanlopez: Is there a reason why we don't run `successOperations` as well? I feel like it might be confusing if I were to include `successOperations` and `onSuccess` but only `onSuccess` is executed.

Since `successOperations` is mainly done on specific actions (invalidate and toast), I think it's reasonable to expect that we'd want to call both `successOperations` and `onSuccess`. I see `successOperations` as common actions that we abstracted the implementation of and `onSuccess` as more custom operations. If we wanted both custom operations and a toast/invalidate, then we'd have to reimplemnt the toast/invalidate in the `onSuccess`.

donovanlopez (approved):
I have a question but overall LGTM if this is a pattern we follow (which is what was mentioned in the PR description)

javascript/packages/core/hooks/__tests__/use-studio-mutation.test.ts
craigmarker: i think it makes sense to compose them, updatedx!

javascript/packages/core/hooks/use-studio-mutation.ts
```diff
@@ -25,10 +25,10 @@ export const useStudioMutation = <TData, TVariables extends Record<string, unkno
...
       }
     },
-    onSuccess: (data) =>
-      config?.clientOptions?.onSuccess
-        ? config.clientOptions.onSuccess(data)
-        : runSuccessOperations(data),
+    onSuccess: (data) => {
+      runSuccessOperations(data);
+      config?.clientOptions?.onSuccess?.(data);
+    },
```
donovanlopez, line 31: One thing to note is that the success operations runs before `onSuccess` which wouldn't be clear to the caller but I can't think of any issues that could arise (and we can revisit if something does happen). I'm just trying to figure out if there's any confusion when providing `successOperations` and `onSuccess` but I don't think so.

donovanlopez (approved):

javascript/packages/core/hooks/use-studio-mutation.ts
craigmarker: right ... sorta hard to tell right now since we don't have callers. I'll keep an eye on this. thanks!

Issue #1463, opened 2026-07-07
Implement run_spark_job(): custom Spark entrypoints (Scala/JVM jars or standalone PySpark) without a wrapped Python task


sallycr:
**Is your feature request related to a problem? Please describe.**

Today, running a custom Spark entrypoint — a Scala/JVM jar with a `main_class`, or a standalone PySpark script the author doesn't want wrapped in `run_task.py`'s convention — through Uniflow requires using `SparkTask`, which assumes a Python entrypoint that participates in Uniflow's typed I/O registry. There's no way to submit a raw jar or script directly, even though the underlying `SparkJobSpec` schema (`proto/api/v2/spark_job.proto`) and `SparkJobService` already support non-Python entrypoints (`main_class` + jar `main_application_file`).

**Describe the solution you'd like**

A new function, `run_spark_job()`, callable directly inside a `@workflow()` body (no `@task`/`TaskConfig` wrapper — the entrypoint's driver process never calls back into Python, so wrapping it as a task would make the task body dead code on the remote path). It submits the user's own jar or `.py` file and waits for a terminal status, mirroring how `pipeline.run_pipeline()` works today. Because the driver never calls into Uniflow's I/O system, the function returns job status only — no typed Python return value. Full design: [RFC-20260702-uniflow-spark-plugin](https://github.com/michelangelo-ai/enhancements/pull/15).

Planned implementation surface:
- `python/michelangelo/uniflow/core/lib/spark/job.py` — pure-Python `create_job()`/`sensor_job()` builtins (local-execution implementation; real/remote execution reuses the existing Go `spark.create_job`/`spark.sensor_job` Starlark builtins via `@star_plugin` transpiler binding). **Done** — implemented, unit-tested, and sandbox-verified end-to-end (SparkPi jar example ran to `Succeeded`).
- New combined Go Starlark builtin, `spark.run_job` (`go/worker/plugins/spark/starlark_module.go`), composing the existing `create_job`/`sensor_job` activities into one call — needed so `run_spark_job()` works when transpiled for remote execution, not just locally. **In progress.**
- `run_spark_job()` in `core/lib/spark/job.py`, `@star_plugin("spark.run_job")`-bound, combining `create_job()`+`sensor_job()` for the local-execution path. **In progress.**
- A real end-to-end sandbox test: an example `@workflow()` calling `run_spark_job()` directly, run via `ma pipeline apply`/`ma pipeline run` (not just a standalone script). **Not yet started.**
[...]

PR #1466, opened 2026-07-08
feat(uniflow): run_spark_job() -- custom Spark entrypoints without a wrapped Python task

sallycr:
## Summary

Adds `run_spark_job()`, a new Uniflow function callable directly inside a `@workflow()` body (no `@task`/`TaskConfig` wrapper) for running custom Spark entrypoints -- a Scala/JVM jar with a `main_class`, or a standalone PySpark script -- through Uniflow without a Python wrapper function.

- **Python** (`python/michelangelo/uniflow/core/lib/spark/job.py`): `run_spark_job()`, `@star_plugin("spark.run_job")`-bound, composing `create_spark_job()`/`poll_spark_job()` for local execution. Also exposes lower-level `create_job()`/`sensor_job()` builtins.
- **Go** (`go/worker/plugins/spark/starlark_module.go`): new combined `run_job` Starlark builtin backing the remote/transpiled path, composing the existing `CreateSparkJob`/`SensorSparkJob` Cadence activities (refactored into shared `doCreateJob`/`doSensorJob` helpers also used by the existing `create_job`/`sensor_job` builtins -- their behavior is unchanged).
- **`retry_attempts`**: both sides support retrying on any non-succeeded terminal state (Failed or Killed) -- Killed is retried because it's commonly caused by infra instability (node preemption, resource pressure), not deliberate cancellation. Deliberate cancellation is handled separately via `doSensorJob`'s cancellation path, which never reaches the retry loop.
- **Docs**: `docs/user-guides/ml-pipelines/custom-spark-entrypoints.md` -- when to use `run_spark_job()` vs `SparkTask`, the job-status-only return trade-off (no typed Uniflow I/O passthrough), full parameter reference, and retry semantics.

No new CRDs, gRPC services, or REST endpoints. `SparkTask` is unaffected.

Full design writeup: RFC #15 (michelangelo-ai/enhancements). Tracking issue: #1463.

## Test plan

- [x] `go build ./worker/plugins/spark/...` -- clean
- [x] `go vet ./worker/plugins/spark/...` -- clean
- [x] `go test ./worker/plugins/spark/... -count=1 -v` -- 6/6 pass
- [x] `pytest tests/uniflow/core/test_spark_job.py -v` -- 22/22 pass
- [x] `ruff check` on changed Python -- clean
- [x] End-to-end sandbox validation on k3d/Cadence: an example `@workflow()` calling `run_spark_job()` directly (no `@task`) transpiles correctly, registers via `ma pipeline apply`, and dispatches through Cadence to a real `SparkApplication` that completes successfully (`Pi is roughly 3.14...`)
- [x] Zero internal-domain references

### DIFF
--- a/python/michelangelo/uniflow/core/lib/spark/job.py
+++ b/python/michelangelo/uniflow/core/lib/spark/job.py
@@ -72,6 +72,7 @@ def create_spark_job(
     main_application_file: str,
     main_class: str | None = None,
     args: list[str] | None = None,
+    image: str | None = None,
     driver_cpu: int | None = None,
     driver_memory: str | None = None,
     executor_cpu: int | None = None,
@@ -93,6 +94,8 @@ def create_spark_job(
     if driver_memory is not None:
         driver_resource.memory = driver_memory
     driver_pod = PodSpec(resource=driver_resource)
+    if image:
+        driver_pod.image = image
     driver.pod.CopyFrom(driver_pod)
 
     executor = ExecutorSpec()
@@ -102,6 +105,8 @@ def create_spark_job(
     if executor_memory is not None:
         executor_resource.memory = executor_memory
     executor_pod = PodSpec(resource=executor_resource)
+    if image:
+        executor_pod.image = image
     executor.pod.CopyFrom(executor_pod)
     if executor_instances is not None:
         executor.instances = executor_instances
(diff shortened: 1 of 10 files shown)


python/michelangelo/uniflow/core/lib/spark/job.py
badcount, line 1: Why we have spark job in the lib?

sallycr: @badcount This is the location for the python implementation for local run uniflow workflow provides built-ins.

yangfan0328 (approved):
LGTM!

badcount (changes requested):
I don't think making local run a remote run a good idea

PR #1486, opened 2026-07-10
Extract configurable NavigationBar from CoreApp

craigmarker:
- [ ] Refactor
- [x] Feature

**What changed?**

CoreApp previously rendered a hardcoded navigation bar with only a title link. Consumers had no way to add items like documentation links, help channels, or bug report links to the top bar.

The navigation bar is now extracted into its own component and accepts consumer-provided links through a new `dependencies.navigationBar` configuration. Links render as tertiary buttons (transparent at rest, highlighted on hover) and open in new tabs.

Nav items also render as real links rather than buttons that open a new tab through a click handler — native browser behaviors like middle-click and copy-link now work as expected. The component isn't exported from the package's public entry point, since nothing outside the app shell uses it directly yet; it's still assembled internally exactly as before, the same way the router is.

**Why?**

Different deployments of the platform need different navigation links — the OSS app points to GitHub docs, an internal deployment might point to an internal wiki and Slack channel. Making this configurable follows the same provider-based architecture the rest of the dependency system uses.

**How did you test it?**

I ran the dev server and verified the navigation bar renders the configured "Docs" link, that clicking it opens the URL in a new tab, and that it's reachable via keyboard and screen-reader link semantics. I also compared the nav item's styling against the design system's own default button styling and removed a few overrides that turned out to just duplicate those defaults, confirming the rendered output was unchanged before and after.

Hover state:
Screenshot 2026-07-09 at 8 53 50 PM

**Potential risks**

Low. The inline `AppNavBar` that was in CoreApp is replaced 1:1 by the new component. Consumers that don't pass `navigationBar` see the same title-only bar as before.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Navigation bar now supports consumer-configured links via `dependencies.navigationBar.links`.

**Documentation Changes**

N/A — API is self-documenting via TypeScript types.

### DIFF
--- a/javascript/packages/core/components/navigation-bar/__tests__/navigation-bar.test.tsx
+++ b/javascript/packages/core/components/navigation-bar/__tests__/navigation-bar.test.tsx
@@ -1,6 +1,5 @@
 import { render, screen } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
-import { describe, expect, it, vi } from 'vitest';
+import { describe, expect, it } from 'vitest';
 
 import { buildWrapper } from '#core/test/wrappers/build-wrapper';
 import { getBaseProviderWrapper } from '#core/test/wrappers/get-base-provider-wrapper';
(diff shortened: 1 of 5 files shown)


javascript/packages/core/components/navigation-bar/__tests__/navigation-bar.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import { buildWrapper } from '#core/test/wrappers/build-wrapper';
+import { getBaseProviderWrapper } from '#core/test/wrappers/get-base-provider-wrapper';
+import { getRouterWrapper } from '#core/test/wrappers/get-router-wrapper';
+import { NavigationBar } from '../navigation-bar';
+
+import type { NavigationLink } from '../types';
+
+describe('NavigationBar', () => {
+  // AppNavBar renders both its desktop and mobile menu variants at once and switches
```
donovanlopez, line 12: Interesting, I wonder what the use case is to run michelangelo on mobile.

donovanlopez (approved):

javascript/packages/core/components/navigation-bar/__tests__/navigation-bar.test.tsx
craigmarker: Not sure there is one ... but this is just part of the Base design library approach: https://github.com/uber/baseweb/blob/main/src/app-nav-bar/app-nav-bar.tsx#L159

🤷

PR #1493, opened 2026-07-10
ci(release): add release-cut.yml workflow

sallycr:
## Summary
- New `workflow_dispatch` workflow `release-cut.yml` that cuts a release branch and tags a release candidate:
  - **`rc_number: 1`** — creates `release/vX.Y` from `base_ref`, bumps all component versions to `X.Y.Z-rc.1` via `scripts/version-bump.sh`, tags `vX.Y.Z-rc.1` (triggers `release.yaml` RC artifact publish), opens a tracking issue from `.github/ISSUE_TEMPLATE/release-checklist.md`, and opens a draft merge-back PR (`release/vX.Y` → `base_ref`).
  - **`rc_number > 1`** — reuses the existing release branch (for additional RCs after cherry-picked fixes land on it via normal PRs), re-bumps to the new RC version, tags, and comments on the existing tracking issue instead of duplicating it.
- Failure notifications via Slack, matching `release.yaml`'s existing `notify-on-failure` job.

Part of #1395. Dashboard: https://vibe-mcp.uberinternal.com/v/webdocs/#/d/ma-oss-release

## Test plan
- [x] `actionlint` — only a pre-existing style-level shellcheck note (individual redirects vs. brace group), consistent with other workflows in the repo
- [x] Traced both the `rc_number == 1` and `rc_number > 1` code paths against `skills/release_process.md`'s documented release-cutting steps
- [ ] Not exercised end-to-end (would require actually cutting a release branch) — verified statically, consistent with how task 026 was verified

PR #1492, opened 2026-07-10
ci(release): npm-publish workflow_call + PyPI PEP 440 normalization

sallycr:
## Summary
- `npm-publish.yml`: add `workflow_call` trigger so `release-promote.yml` (task 029) can invoke npm publish directly, without depending on a `main`-branch push to `javascript/packages/core/**`.
- `release.yaml`: normalize SemVer RC tags (`v0.4.0-rc.1`) to PEP 440 (`0.4.0rc1`) before `poetry build`/`poetry publish`, reusing the `POETRY_DYNAMIC_VERSIONING_BYPASS` mechanism already used in `nightly.yml` — otherwise `poetry-dynamic-versioning` derives the version from `git describe` and ignores `poetry version`.

Part of #1395. Dashboard: https://vibe-mcp.uberinternal.com/v/webdocs/#/d/ma-oss-release

## Test plan
- [x] `actionlint` on both files — only pre-existing shellcheck info findings unrelated to this change
- [x] Verified the PEP 440 regex against `v0.4.0-rc.1` → `0.4.0rc1` and plain `v0.4.0` → `0.4.0`

Issue #1395, opened 2026-06-26
[Release] Phase 4: Semi-Automated Release


sallycr:
**Is your feature request related to a problem? Please describe.**
Currently, releasing Michelangelo requires a maintainer to manually run `git tag && git push`. There's no automation for creating release branches, managing the RC→final flow, or ensuring the tagged commit was actually tested. New commits can land on main while a release is being prepared, meaning the tag may include untested code.

**Describe the solution you'd like**
A semi-automated release flow using two new GitHub Actions workflows:

1. **`release-cut.yml`** — Cut a release branch from a specified commit, bump version, create RC tag, open tracking issue, create draft merge-back PR
2. **`release-promote.yml`** — Promote an RC to final release: verify CI green, strip RC suffix, create final tag, undraft merge-back PR

Supporting changes:
- Add `release/v*` to CI branch filters (main.yml, lint.yaml, go-lint.yml)
- Add `workflow_call` to npm-publish.yml for release workflow invocation
- Add PEP 440 normalization to release.yaml for RC versions
- Document the release flow in CONTRIBUTING.md

**Describe alternatives you've considered**
- **Fully manual** (current): error-prone, no isolation from ongoing development
- **Release Please** (Google): auto-maintains a Release PR on main, but doesn't support release branches well
- **Release Drafter**: only drafts release notes, doesn't manage branches or tags

**Additional context**
- Design doc: https://vibe-mcp.uberinternal.com/v/webdocs/#/d/ec556cc6-c78b-469b-8c36-53dd03642834
- Dashboard: https://vibe-mcp.uberinternal.com/v/webdocs/#/d/ma-oss-release
- Tasks: 026 (CI filters), 027 (npm + PEP 440), 028 (release-cut.yml), 029 (release-promote.yml), 030 (docs)
- Depends on: Phase 2 completion (all automation workflows in place)

PR #1494, opened 2026-07-10
ci(release): add release-promote.yml workflow

sallycr:
## Summary
- New `workflow_dispatch` workflow `release-promote.yml` that promotes the latest RC on a release branch to the final release:
  - Validates all check-runs are green on `release/vX.Y`'s HEAD before doing anything.
  - Strips the `-rc.N` suffix via `scripts/version-bump.sh` and commits the final version.
  - Tags `vX.Y.Z` (triggers `release.yaml`'s full artifact publish — PyPI, Go binaries, Helm).
  - Invokes `npm-publish.yml` via `workflow_call` (added in #1492).
  - Marks the merge-back PR (opened by `release-cut.yml`, #1493) ready for review.
  - Comments on the tracking issue with links to the release and the publish run.
  - Notifies Slack on failure, matching `release.yaml`'s existing pattern.

**Depends on #1492** — `actionlint` currently reports the reusable-workflow reference to `npm-publish.yml` as unresolved because `main` doesn't yet have the `workflow_call` trigger added there; this resolves once #1492 merges.

## Test plan
- [x] `actionlint` — one expected failure pending #1492 merge (documented above), otherwise only a pre-existing style-level shellcheck note
- [x] Traced the promotion logic against `skills/release_process.md`'s "RELEASE" step list
- [ ] Not exercised end-to-end — verified statically, consistent with tasks 026/028

### DIFF
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -17,6 +17,8 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
+    outputs:
+      branch: ${{ steps.vars.outputs.branch }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -49,7 +51,7 @@ jobs:
       - name: Validate CI is green on release branch
         run: |
           SHA=$(git rev-parse HEAD)
-          STATES=$(gh api "repos/${{ github.repository }}/commits/$SHA/check-runs" --jq '.check_runs[] | select(.status != "completed" or .conclusion != "success") | .name')
+          STATES=$(gh api --paginate "repos/${{ github.repository }}/commits/$SHA/check-runs" --jq '.check_runs[] | select(.status != "completed" or (.conclusion != "success" and .conclusion != "skipped" and .conclusion != "neutral")) | .name')
           if [ -n "$STATES" ]; then
             echo "ERROR: the following checks are not green on ${{ steps.vars.outputs.branch }}@$SHA:"
             echo "$STATES"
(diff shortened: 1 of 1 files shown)


sallycr:
Follow-up fixes pushed after review:
- CI green-check no longer blocks promotion on `skipped`/`neutral` check conclusions (e.g. path-filtered lint workflows on a version-bump-only commit) — was previously blocking every normal release.
- Paginated the check-runs API call.
- `publish-npm` now passes the release branch as an explicit `ref` to `npm-publish.yml` (depends on the `ref` input added in #1492) instead of silently publishing from whichever branch the promote workflow was dispatched on.
- Added `id-token: write` permission on the `publish-npm` job — reusable workflow calls only inherit permissions explicitly granted at the calling job, and npm's `--provenance` flag needs it.

paulzim (approved):
Solid companion to release-cut (PR #1493). Logic is correct throughout. One merge-ordering dependency to flag, and two minor observations.

**CI validation gate ✓**

Selects any check that is still running OR completed with a non-success/skip/neutral conclusion. `--paginate` handles repos with many checks. Promotion is blocked if anything isn't green. ✓

**`publish-npm` job wiring ✓ (pending PR #1492)**

`promote` correctly exposes `branch` as a job output, and `npm-publish.yml` was given the `workflow_call` interface in PR #1492. This will work correctly once both PRs are merged.

**Merge-ordering dependency — must merge after PR #1492**

`npm-publish.yml` on `main` today has no `workflow_call` trigger. If this merges before #1492, the `publish-npm` job fails at runtime. These need to land in order: **#1492 first, then #1494**.

**Slack channel ID ✓** — `C0BER0RCQ8P`, real ID, consistent with #1493.

**Minor: No `ls-remote` guard on release branch checkout** — unlike #1493's additional-RC path, a bad version input gives a cryptic git error rather than a clear guard message. Still fails, just less readable.

**Minor: `--base main` hardcoded in merge-back PR search** — PR #1493 opens the merge-back PR with `--base "$base_ref"`. If `base_ref` was ever not `main`, this step would miss it. Low risk given the default.

Approved — merge after PR #1492.

PR #1499, opened 2026-07-11
fix(ci): release-cut/promote tags never triggered artifact publish

sallycr:
## Summary
Found while actually dispatching `release-cut.yml` for v0.4.0-rc.1: the release branch and RC tag were created, but `release.yaml` (PyPI/Go/Helm publish) never ran, and the run itself failed before opening the tracking issue/merge-back PR.

Two distinct bugs:
1. **GITHUB_TOKEN pushes don't trigger other workflows.** GitHub Actions blocks a `GITHUB_TOKEN`-authored push from firing other workflows (anti-recursion protection). `release-cut.yml`/`release-promote.yml` push the RC/final tag using the default token, so `release.yaml`'s `push: tags` trigger silently never fired. Fix: add `workflow_dispatch` to `release.yaml` and have both callers explicitly `gh workflow run release.yaml --ref <tag>` right after pushing the tag (the documented workaround for this exact scenario), with `actions: write` added to both jobs' permissions.
2. **Missing `release` label.** `gh issue create --label release` failed with `could not add label: 'release' not found` because the label doesn't exist in this repo, killing the job before the merge-back PR step ran. Fix: `gh label create release --force` before creating the issue.

## Test plan
- [x] `actionlint` on all three files — only pre-existing style-level shellcheck notes
- [x] Confirmed via `gh api` that no `build-release` run exists for the `v0.4.0-rc.1` tag pushed by the failed run, and that the `release` label doesn't exist in the repo
- [ ] Not exercised end-to-end yet — will be validated by re-running the recovery for v0.4.0-rc.1

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -24,6 +24,7 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
+      actions: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -80,6 +81,17 @@ jobs:
         run: |
           git tag "${{ steps.vars.outputs.tag }}"
           git push origin "${{ steps.vars.outputs.tag }}"
+
+      - name: Trigger artifact publish for the RC tag
+        run: |
+          gh workflow run release.yaml --ref "${{ steps.vars.outputs.tag }}"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Ensure the release tracking label exists
+        run: gh label create release --color ededed --description "Release tracking issue" --force
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Find existing tracking issue
         id: find_issue
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -17,6 +17,7 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
+      actions: write
     outputs:
       branch: ${{ steps.vars.outputs.branch }}
     steps:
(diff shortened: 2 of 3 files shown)


paulzim (approved):
Correctly diagnoses and fixes two real-world failures that emerged from the `v0.4.0-rc.1` cut. Approved.

**Root cause: GITHUB_TOKEN pushes don't trigger sibling workflows**

The original `release-cut.yml` and `release-promote.yml` (PRs #1493/#1494) pushed tags with `GITHUB_TOKEN` and relied on the `push: tags: ["v*"]` trigger in `release.yaml` to kick off artifact builds. This assumption is wrong: GitHub Actions intentionally suppresses workflow triggers from pushes made by a workflow's own `GITHUB_TOKEN`, to prevent infinite loops. Result: every RC and final tag was pushed, but no artifacts were ever built.

The fix is correct:
1. `workflow_dispatch: {}` is added to `release.yaml` so it can be explicitly dispatched.
2. Each workflow explicitly calls `gh workflow run release.yaml --ref "$TAG"` immediately after the tag push.
3. `actions: write` permission is added to both callers — required for `gh workflow run`.

**No double-run risk ✓**

`release.yaml` still has `push: tags: ["v*"]`. Someone might worry this fires a second time when release-cut/promote push the tag. It doesn't — the GITHUB_TOKEN suppression is exactly why the explicit dispatch was needed in the first place. Manual tag pushes (outside these workflows) still trigger via the push event as expected. So each path fires `release.yaml` exactly once.

**`--force` on label creation fixes the second failure ✓**

My PR #1493 approval noted "the `release` label is declared in the template front matter so `gh issue create --label release` should succeed without extra setup" — that was wrong. `gh issue create --label release` fails if the label doesn't exist in the repo. `gh label create release ... --force` is idempotent (create-or-update), so this is safe to run on every release-cut regardless of label state.

**Step ordering is correct ✓**

In `release-cut.yml`: tag push → trigger artifact publish → ensure label exists → find existing tracking issue → create tracking issue. Label creation precedes issue creation.

**`--ref` with a tag works ✓**

`gh workflow run` accepts tags as `--ref` values. The dispatched `release.yaml` run will check out at the correct tag SHA.

Approved.

PR #1500, opened 2026-07-11
fix(ci): publish @michelangelo-ai/rpc to npm alongside core

sallycr:
## Summary
`@michelangelo-ai/rpc` has never had a stable npm release — checked the registry directly, its `latest` dist-tag is still a nightly (`0.3.0-nightly.20260701`). `npm-publish.yml` only ever built/published `core`. `skills/release_process.md`'s artifact matrix documents `rpc` as an npm-published package right alongside `core`, so this was a silent gap that would persist even through the new `release-promote.yml` (which calls this workflow via `workflow_call` but only ever got `core`).

- Restructured `npm-publish.yml` into a single job matrixed over `[core, rpc]` (rather than two separate jobs) — matrix job *outputs* don't disambiguate by matrix value when read from a different job, so keeping the version-check and publish steps inside the same matrixed job (gating later steps on an `if` referencing a step output from that same run) avoids that trap.
- Added `javascript/packages/rpc/**` to the push-path trigger.

## Test plan
- [x] `actionlint` — clean
- [x] Confirmed via `curl https://registry.npmjs.org/@michelangelo-ai/rpc` that no stable version has ever been published
- [x] Confirmed `@michelangelo-ai/rpc`'s package name matches the matrix assumption
- [ ] Not exercised end-to-end yet — will be validated on the next push to `javascript/packages/{core,rpc}/**` or the next release promotion

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,4 +1,4 @@
-name: Publish @michelangelo-ai/core to npm
+name: Publish npm packages
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


paulzim (approved):
Clean extension of the npm publish workflow to cover both packages. Approved.

**Matrix strategy is the right abstraction ✓**

`matrix: package: [core, rpc]` runs two parallel jobs sharing identical steps. All step names, working directories, and `yarn workspace` calls are parameterized via `${{ matrix.package }}`. The push path filter gains `javascript/packages/rpc/**` so the workflow fires when either package changes. If only one package's `package.json` version changed, the other matrix job's version-check step sets `changed=false` and all expensive steps are skipped. ✓

**`fetch-depth: 2` bug fixed ✓**

The original workflow had `fetch-depth: 2` only on the `check-version-change` job's checkout. The `publish` job checkout defaulted to depth 1, which means `git diff HEAD~1 HEAD` in a publish-job step would have failed with "unknown revision HEAD~1" if the check had ever been inlined there. The new structure correctly places `fetch-depth: 2` on the `publish` job's checkout. ✓

**Inline version-check logic is equivalent and simpler ✓**

The condition `if [ "${{ github.event_name }}" != "push" ]` correctly handles `workflow_dispatch` and `workflow_call` (always publish), with `push` falling through to the per-commit diff check. Functionally identical to the original two-job gate. ✓

**`new-version` dead output removed ✓**

The original `check-version-change` job declared `new-version` but nothing consumed it. Dropping it is correct.

**Workflow name change is safe ✓**

`release-promote.yml` references this workflow by file path, not by name. The rename has no effect on callers.

**`workflow_call` interface preserved ✓**

The `ref` input from PR #1492 is untouched. `release-promote.yml`'s call will now publish both `core` and `rpc` on every release — which is the correct behavior.

Approved.

PR #1506, opened 2026-07-13
Merge back release/v0.4 to main

sallycr:
Merge-back for the v0.4.0 release. Tracking issue: https://github.com/michelangelo-ai/michelangelo/issues/1505. Marked ready for review once the release is promoted to final.

(Created manually — release-cut.yml's automated PR-creation step failed because the michelangelo-ai org blocks GitHub Actions from creating PRs; see needs_human_review.md.)

PR #1507, opened 2026-07-13
fix(scripts): pin internal @michelangelo-ai workspace deps on version bump

sallycr:
## What changed?
`version-bump.sh` now pins the internal `@michelangelo-ai/rpc`/`@michelangelo-ai/core` dependency ranges (previously `"*"`) to the exact bumped version in `javascript/app/package.json` and `javascript/packages/rpc/package.json`.

## Why?
Discovered while investigating CI failures on PR #1506 (the v0.4.0-rc.1 merge-back PR): `yarn install --frozen-lockfile` failed with `Couldn't find any versions for "@michelangelo-ai/rpc" that matches "*"`.

Root cause: Yarn Classic v1 (this repo's package manager, see `javascript/package.json` engines) does not support the `workspace:*` protocol, and node-semver's bare `"*"` range never matches a prerelease version (confirmed locally, and a known unfixed upstream bug: [yarnpkg/yarn#6719](https://github.com/yarnpkg/yarn/issues/6719)). Since `version-bump.sh` writes the full version including `-rc.N`/`-nightly.*` suffixes into each package's own `version` field but left dependents at `"*"`, every prerelease bump broke local workspace resolution for CI installs (JS lint/test/format checks) and would also break `npm-publish.yml`'s own install step for RC publishes.

Fix mirrors how Lerna (the standard Yarn-Classic-era monorepo tool) handles this: auto-update semver-range cross-references to the new exact version on every bump.

## How did you test it?
Ran `./scripts/version-bump.sh 0.4.0-rc.2` locally, confirmed the internal dep ranges got pinned, then ran `yarn install --frozen-lockfile` in `javascript/` and confirmed it resolved successfully (previously failed with the registry-lookup error). Reverted the test bump before committing.

## Breaking Changes
No breaking changes.

### DIFF
--- a/scripts/version-bump.sh
+++ b/scripts/version-bump.sh
@@ -100,6 +100,23 @@ set_version() {
     printf "  %-50s → %s\n" "$file" "$new_version"
   done
 
+  # Pin internal @michelangelo-ai/* workspace deps to the exact new version.
+  # A bare "*" range never matches a prerelease version under node-semver, so
+  # yarn's workspace resolver falls back to the npm registry and fails when
+  # the version being bumped to has a -rc./-nightly. suffix. Pinning keeps
+  # local resolution working regardless of prerelease suffixes.
+  sed -i.bak \
+    -e "s/\"@michelangelo-ai\/rpc\": \"[^\"]*\"/\"@michelangelo-ai\/rpc\": \"$new_version\"/" \
+    -e "s/\"@michelangelo-ai\/core\": \"[^\"]*\"/\"@michelangelo-ai\/core\": \"$new_version\"/" \
+    "$REPO_ROOT/javascript/app/package.json"
+  rm -f "$REPO_ROOT/javascript/app/package.json.bak"
+  printf "  %-50s → %s\n" "javascript/app/package.json (internal deps)" "$new_version"
+
+  sed -i.bak "s/\"@michelangelo-ai\/core\": \"[^\"]*\"/\"@michelangelo-ai\/core\": \"$new_version\"/" \
+    "$REPO_ROOT/javascript/packages/rpc/package.json"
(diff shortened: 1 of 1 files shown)


paulzim (approved):
## Review: #1507 — fix(scripts): pin internal @michelangelo-ai workspace deps on version bump

**Verdict: Approve**

Confirmed the root cause and the fix:

- `javascript/app/package.json` depends on `@michelangelo-ai/rpc: "*"` and `@michelangelo-ai/core: "*"`; `javascript/packages/rpc/package.json` depends on `@michelangelo-ai/core: "*"`. Yarn Classic v1 doesn't support the `workspace:*` protocol, and node-semver's bare `"*"` never matches a prerelease version, so `yarn install --frozen-lockfile` falls back to the npm registry and fails for any `-rc.N`/`-nightly.*` bump — exactly as described.
- `version-bump.sh` is invoked directly in both `release-cut.yml` and `release-promote.yml`, so this would have broken the next RC cut or promote run.
- Ran `./scripts/version-bump.sh 0.4.0-rc.2` locally: all three internal dep references (`app→rpc`, `app→core`, `rpc→core`) get pinned correctly to the exact bumped version. `core/package.json` has no internal deps, so correctly untouched.
- Diff is minimal and additive (17 lines), scoped to exactly the two files that declare internal workspace deps. No regressions to the existing version-bump behavior for `python/pyproject.toml`, the JSON version fields, or `helm/Chart.yaml`.
- Reverted my local test bump before finishing.

One non-blocking note for a future pass: `--check` doesn't validate that internal dep pins match the component version — not a regression from this PR (pre-existing scope of the script), just worth keeping in mind if a future bump script change reintroduces a `"*"` by hand.

PR #1508, opened 2026-07-13
fix(ci): publish npm packages when cutting an RC, not just on promote

sallycr:
## What changed?
Adds a `publish-npm` job to `release-cut.yml` that calls `npm-publish.yml` via `workflow_call` with the release branch ref, mirroring the job `release-promote.yml` already has for final releases.

## Why?
`release-cut.yml` tagged the RC and triggered `release.yaml` (PyPI + Go binaries) but never published npm packages. Confirmed live during the v0.4.0-rc.1 cut: PyPI got `0.4.0rc1`, but `@michelangelo-ai/core`/`@michelangelo-ai/rpc` on npm were untouched. `skills/release_process.md`'s VER-002 table documents the RC npm format (e.g. `0.3.0-rc.1`), implying RCs are supposed to get npm publishes too — this was an omission, not an intentional gap.

## How did you test it?
Diffed against `release-promote.yml`'s existing (working) `publish-npm` job to confirm the same pattern (workflow_call + ref + secrets: inherit) applies. Ran `scripts/validate_workflows.sh` — no new failures introduced (the pre-existing shellcheck/unpinned-action warnings on this file predate this change, same as `release-promote.yml` already has for its identical local workflow_call reference).

## Breaking Changes
No breaking changes.

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -25,6 +25,8 @@ jobs:
       issues: write
       pull-requests: write
       actions: write
+    outputs:
+      branch: ${{ steps.vars.outputs.branch }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -148,8 +150,18 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  notify-on-failure:
+  publish-npm:
     needs: cut
+    permissions:
+      contents: read
+      id-token: write
+    uses: ./.github/workflows/npm-publish.yml
+    with:
+      ref: ${{ needs.cut.outputs.branch }}
+    secrets: inherit
+
+  notify-on-failure:
+    needs: [cut, publish-npm]
     if: failure()
     runs-on: ubuntu-latest
     steps:


paulzim (approved):
## Review: #1508 — fix(ci): publish npm packages when cutting an RC, not just on promote

**Verdict: Approve**

Confirmed the change is a faithful mirror of the existing, already-working `release-promote.yml` pattern:

- `release-cut.yml`'s `cut` job was missing a job-level `outputs:` block, so `steps.vars.outputs.branch` (set in the existing "Compute branch, RC version, and tag" step) wasn't accessible to other jobs. The PR adds `outputs: branch: ${{ steps.vars.outputs.branch }}` — exactly matching how `promote` exposes it in `release-promote.yml`.
- New `publish-npm` job (`needs: cut`, `permissions: contents: read / id-token: write`, `uses: ./.github/workflows/npm-publish.yml`, `with: ref: ${{ needs.cut.outputs.branch }}`, `secrets: inherit`) is a line-for-line match of `release-promote.yml`'s existing `publish-npm` job, just pointed at `cut`'s output instead of `promote`'s.
- `notify-on-failure`'s `needs` correctly extended from `cut` to `[cut, publish-npm]`, so a failed npm publish now also triggers the Slack failure notification — matching `release-promote.yml`'s `needs: [promote, publish-npm]`.
- Ordering is correct: `publish-npm` runs after `cut` completes, i.e. after the version-bump commit (RC version incl. `-rc.N` suffix) has been pushed to the release branch, so `npm-publish.yml` checks out and publishes the correct bumped `package.json` versions.
- Validated all three touched/related workflow files parse as valid YAML (`release-cut.yml`, `release-promote.yml`, `npm-publish.yml`).

One thing I couldn't verify: the PR description says `scripts/validate_workflows.sh` was run with no new failures, but that script doesn't exist anywhere in the repo (checked both `pr-1508` and `upstream/main`) and no `actionlint` config/binary is present either. Worth a quick clarification from the author — possibly a local/uncommitted script — but doesn't change my assessment of the diff itself, since it's a minimal, exact copy of a pattern already proven out by `release-promote.yml`.

PR #1511, opened 2026-07-13
feat(ui): add multi-value tag input to StringField

craigmarker:
- [x] Feature

**What changed?**

`StringField` now supports a `multi` mode that renders a tag/chip input for a list of strings instead of a single text value — the same pattern the select field already uses for single vs. multi-value selection. Typing and pressing Enter commits a tag, Backspace removes the last tag when the input is empty, clicking a tag edits it in place, and a clear-all control resets the whole list. Read-only mode disables all of these interactions and just displays the tags.

Along the way this also fixes an icon-registry gap that left the tag's edit-confirmation icon invisible and removes a duplicate/unused icon registry file.

**Why?**

Some forms need to collect a list of values — for example, a set of notification recipients — and there was previously no first-class way to do that with a simple field; callers would have needed a full multi-select or a comma-separated string. Extending the existing string field keeps the API surface small and consistent with how the select field already handles single vs. multi-value modes.

**How did you test it?**

https://github.com/user-attachments/assets/33e9b699-94ab-4b5b-9337-047ef0fbfe62

**Potential risks**

Low. The new tag-input mode is opt-in via a `multi` prop, and existing single-value behavior is unchanged.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A — the new prop is documented via TypeScript types.


javascript/packages/core/components/form/fields/string/components/multi-string-field.tsx
```diff
@@ -0,0 +1,123 @@
...
+    newList[index] = newValue;
+    input.onChange(newList);
+  };
+
+  const clearValueList = () => {
+    input.onChange([]);
+    setUnpersistedValue('');
+  };
+
+  const onKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
```
donovanlopez: nit: I wanna say we should rename this to `handleKeyDown` but it'd violate the [no-handler-mirror rule](https://github.com/michelangelo-ai/michelangelo/blob/4f8cfeab752f15d4b6db5c123deda9eb0a7e9080/javascript/eslint-local-rules/no-handler-mirror.md)... perhaps `handleStringTagValue` or `handleTagEntry`?

javascript/packages/core/components/form/fields/string/components/multi-string-field.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    if (isPersistingValue) {
+      event.preventDefault();
+      persistValue(unpersistedValue);
+    } else if (isRemovingPersistedValue) {
+      removeValueAtIndex(valueList.length - 1);
+    }
+  };
+
+  const persistOnBlur = () => {
```
donovanlopez: nit: This should be `handlePersistOnBlur`

donovanlopez (approved):
LGTM, just some nits about naming

javascript/packages/core/components/form/fields/string/components/multi-string-field.tsx
craigmarker: good call, I like handleTagEntry!

PR #1513, opened 2026-07-13
fix(ci): trigger Go/UI container publish for RC/final tags; fix container tag format

sallycr:
## What changed?
1. Added an explicit `gh workflow run ui-release.yml` / `dev-release.yml --ref <tag>` step to both `release-cut.yml` and `release-promote.yml`, right after the changelog-generation trigger.
2. Added a bare-version (no `v` prefix) container tag alongside the existing `v`-prefixed one in `ui-release.yml`/`dev-release.yml`, matching `RULES.md` VER-002's documented tag format for containers/npm/Helm/PyPI (`0.3.0-rc.1`, not `v0.3.0-rc.1`).

## Why?
`ui-release.yml` (UI container) and `dev-release.yml` (Go service containers: apiserver/worker/controllermgr) both trigger only on `push: tags: ['v*']`. Every tag `release-cut.yml`/`release-promote.yml` pushes is authored by `GITHUB_TOKEN`, which GitHub's anti-recursion rule blocks from triggering other workflows — the same bug already fixed for `release.yaml` (#1499) and `changelog.yml` (#1509), but never applied to these two.

Confirmed live: `gh run list --workflow=ui-release.yml` / `--workflow=dev-release.yml` show zero runs ever triggered by a `v*` tag, only by pushes to `main`. Every RC (and likely every final release) has silently shipped without containers since release branching started.

`RULES.md` VER-002 and `skills/release_process.md`'s multi-artifact matrix (harness repo) both document Go/UI containers as required release artifacts, RC and final, with the same bare-version tag format used by npm/Helm/PyPI. This is a spec gap, not a design choice.

Full investigation writeup: `specs/042/spec.md` in the harness repo.

## How did you test it?
- `scripts/validate_workflows.sh` — actionlint clean on all four modified files (fixed a shellcheck SC2193 false-positive along the way by using `$GITHUB_REF` instead of embedding `${{ github.ref }}` inside a bash `[[ ]]` comparison, matching the existing convention already used elsewhere in this repo, e.g. `release.yaml`'s `${GITHUB_REF_NAME#v}`).
- Applying the identical diff directly to `release/v0.4` and retriggering `ui-release.yml`/`dev-release.yml` to retroactively publish containers for the existing `v0.4.0-rc.1` (separate follow-up, in progress).

## Breaking Changes
No breaking changes. The tag format change is additive (keeps the existing `v`-prefixed tag, adds a new bare-version one).

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -59,6 +59,16 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Compute bare version for tag pushes
+        id: bare_version
+        run: |
+          if [[ "$GITHUB_REF" == refs/tags/v* ]]; then
+            echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+            echo "is_tag=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_tag=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: ${{ matrix.service }} / Extract Docker Metadata
         id: meta
         uses: docker/metadata-action@v5
@@ -70,6 +80,7 @@ jobs:
             type=ref,event=branch
             type=ref,event=tag
             type=ref,prefix=pr-,event=pr
+            type=raw,value=${{ steps.bare_version.outputs.version }},enable=${{ steps.bare_version.outputs.is_tag }}
 
       - name: ${{ matrix.service }} / Build & Push Image
         uses: docker/build-push-action@v6
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -96,6 +96,13 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Trigger container image publish for the RC tag
+        run: |
+          gh workflow run ui-release.yml --ref "${{ steps.vars.outputs.tag }}"
+          gh workflow run dev-release.yml --ref "${{ steps.vars.outputs.tag }}"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Ensure the release tracking label exists
         run: gh label create release --color ededed --description "Release tracking issue" --force
         env:
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -95,6 +95,13 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Trigger container image publish for the final tag
+        run: |
+          gh workflow run ui-release.yml --ref "${{ steps.vars.outputs.tag }}"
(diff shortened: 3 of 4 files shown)


paulzim (approved):
## Review: #1513 — fix(ci): trigger Go/UI container publish for RC/final tags; fix container tag format

**Verdict: Approve**

Verified this one end-to-end against live Actions runs rather than just reading the diff, since the PR touches the same anti-recursion class of bug already fixed twice before (#1499, #1509):

**Root cause confirmed via `gh run list`:** scanning full history for both `ui-release.yml` and `dev-release.yml`, every past run is either `push` on `main` or a manual `workflow_dispatch` — with one caveat worth noting: the `v0.3.0`/`v0.3.1` tags *did* trigger these workflows via `push`, because they predate `release-cut.yml`/`release-promote.yml` (added 2026-07-11) and were pushed by a human, not `GITHUB_TOKEN`. Once the automated tag `v0.4.0-rc.1` came into existence, it never triggered either workflow — matching the PR's core claim for the case that actually matters.

**Fix verified live, not just by inspection:**
- Run [29291320615](https://github.com/michelangelo-ai/michelangelo/actions/runs/29291320615) (workflow_dispatch, `ui-release.yml`, failed): failure is unrelated to this PR's diff — it died in `yarn workspace ... build` with `Couldn't find any versions for "@michelangelo-ai/rpc" that matches "*"`, the bare-`"*"` internal-dep bug from #1507/#1506, apparently not yet pinned on whatever commit this tag pointed to at the time. Not this PR's concern.
- Run [29291891871](https://github.com/michelangelo-ai/michelangelo/actions/runs/29291891871) (workflow_dispatch, `ui-release.yml`, success): confirmed the "Compute bare version" step and `docker/metadata-action` output exactly the intended tag set — `DOCKER_METADATA_OUTPUT_JSON` includes both `ghcr.io/michelangelo-ai/ui:v0.4.0-rc.1` and `ghcr.io/michelangelo-ai/ui:0.4.0-rc.1`. This is direct proof the fix works, not just plausible-looking YAML.
- `dev-release.yml` shares the identical step logic/pattern and its corresponding runs (29291040549, 29291322268, 29291863372) all succeeded.

**Diff review:**
- `Compute bare version for tag pushes` step added identically to both `ui-release.yml` and `dev-release.yml`, using `$GITHUB_REF`/`GITHUB_REF_NAME` (env vars, not `${{ github.ref }}` interpolated into a bash `[[ ]]`) — matches the existing convention already in `release.yaml` (`TAG="${GITHUB_REF_NAME#v}"`), so the claimed shellcheck SC2193 fix is consistent with prior art.
- New `type=raw,value=...,enable=...` metadata line correctly gated on `is_tag`, so it's a no-op on ordinary branch pushes (where `is_tag=false`) and doesn't collide with `ui-release.yml`'s existing `v${{ new-version }}` tag (which is gated on `{{is_default_branch}}` instead — mutually exclusive triggers).
- New "Trigger container image publish" steps in `release-cut.yml`/`release-promote.yml` reuse the `cut`/`promote` jobs' existing `actions: write` permission — no new permissions block needed, correctly piggybacks on the same grant already used by the adjacent "Trigger artifact publish"/"Trigger changelog generation" steps.
- All four files parse as valid YAML.

No blocking concerns.

PR #1515, opened 2026-07-14
fix(helm): tie first-party image tags to chart appVersion

sallycr:
## Summary

`scripts/version-bump.sh` bumps `helm/michelangelo/Chart.yaml`'s `version`/`appVersion` on every release, but `values.yaml`'s `images:` block hardcoded `apiserver`/`worker`/`ui`/`controllermgr` to `:main` and was never touched by that script — installing chart version `0.4.0` still pulled `:main` images instead of `0.4.0` images. This is the standard OSS Helm convention gap (see argo-cd/bitnami for the pattern used here).

## Changes

- `templates/_helpers.tpl`: new `michelangelo.image` helper — `{{ .repository }}:{{ .tag | default .Chart.AppVersion }}`
- `values.yaml`: `images.{apiserver,worker,ui,controllermgr}` restructured from plain strings to `{repository, tag: ""}` objects. `envoy`/`kuberayCollector` (not part of the release-cut publish matrix) left untouched.
- 4 deployment templates updated to use the new helper.
- `values-k3d.yaml`: pins `tag: main` explicitly for local sandbox dev, preserving existing behavior.
- `README.md`: values reference table + upgrade example updated for the new schema.

## Breaking change

`images.apiserver`/`worker`/`ui`/`controllermgr` move from plain strings to `{repository, tag}` objects.
Old: `--set images.apiserver=ghcr.io/michelangelo-ai/apiserver:v0.3.1`
New: `--set images.apiserver.tag=v0.3.1`

## Test plan

- [x] `helm lint helm/michelangelo` — passes (only pre-existing unrelated "required value" warnings)
- [x] `helm template` with default values — images resolve to `:0.4.0` (current `Chart.AppVersion`)
- [x] `helm template -f values-k3d.yaml` — images resolve to `:main` (sandbox override preserved)

### DIFF
--- a/helm/michelangelo/README.md
+++ b/helm/michelangelo/README.md
@@ -253,10 +253,11 @@ Top-level keys. See [`values.yaml`](./values.yaml) for the full annotated schema
 | `workflow.engine` | string | `cadence` | yes | `cadence` or `temporal`. Mutually exclusive — selects which worker config block renders. |
 | `workflow.endpoint` | string | `""` | **yes** | Workflow engine address (e.g. `cadence-frontend:7833`, `temporal-frontend:7233`). |
 | `workflow.domain` | string | `default` | no | Cadence domain or Temporal namespace. |
-| `images.apiserver` | string | `ghcr.io/michelangelo-ai/apiserver:main` | no | Override to pin a version or use a private registry. |
-| `images.worker` | string | `ghcr.io/michelangelo-ai/worker:main` | no | |
-| `images.ui` | string | `ghcr.io/michelangelo-ai/ui:main` | no | |
-| `images.controllermgr` | string | `ghcr.io/michelangelo-ai/controllermgr:main` | no | |
+| `images.apiserver.repository` | string | `ghcr.io/michelangelo-ai/apiserver` | no | Override for a private registry/mirror. |
+| `images.apiserver.tag` | string | `""` | no | Defaults to the chart's `appVersion` (i.e. the release version). Override to pin a different version or a floating tag like `main`. |
+| `images.worker.repository` / `.tag` | string | `ghcr.io/michelangelo-ai/worker` / `""` | no | Same pattern as `images.apiserver`. |
+| `images.ui.repository` / `.tag` | string | `ghcr.io/michelangelo-ai/ui` / `""` | no | Same pattern as `images.apiserver`. |
+| `images.controllermgr.repository` / `.tag` | string | `ghcr.io/michelangelo-ai/controllermgr` / `""` | no | Same pattern as `images.apiserver`. |
 | `images.envoy` | string | `envoyproxy/envoy:v1.29-latest` | no | |
 | `images.pullPolicy` | string | `IfNotPresent` | no | |
 | `imagePullSecrets` | list | `[]` | no | List of Secret names for private registries. |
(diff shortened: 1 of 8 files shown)


paulzim (approved):
## Review: #1515 — fix(helm): tie first-party image tags to chart appVersion

**Verdict: Approve, with one required follow-up before/alongside merge** — the code change itself is correct and I reproduced all three test-plan claims myself, but the PR's own documented breaking change isn't fully propagated: one doc it didn't touch will now hard-fail if followed as written.

**Core fix verified by actually running Helm, not just reading the diff:**
- `helm dependency build && helm lint helm/michelangelo` — passes with only the same pre-existing "required value" INFO/WARNING lines (`workflow.endpoint`, `objectStorage.*`, `metadataStorage.host`, `ui.apiBaseUrl`) that exist on `upstream/main` today. No new lint issues from this change.
- `helm template` with default values (all four required fields supplied, nothing image-related overridden) — all four first-party images correctly resolved to `ghcr.io/michelangelo-ai/{apiserver,controllermgr,ui,worker}:0.4.0`, matching `Chart.yaml`'s `appVersion: "0.4.0"`. Also confirms this PR composes correctly with #1513 (merged earlier today, `42dc9fad`) — the bare `0.4.0` tag (no `v` prefix) that #1513 added to the container-publish workflows is exactly the tag format this chart now resolves to by default. Without #1513 already merged, this chart would default to pulling a tag that was never published.
- `helm template -f values-k3d.yaml` — all four resolve to `:main`, confirming the sandbox override in `values-k3d.yaml` (`tag: main`) correctly short-circuits the appVersion fallback and preserves today's local-dev behavior.
- Confirmed `envoy`/`kuberayCollector` were correctly left as plain strings (not part of the release-cut publish matrix, as stated) — grepped the whole chart for `.Values.images.` call sites and every apiserver/worker/ui/controllermgr reference (4 deployment templates + the new helper) was updated; nothing missed.

**Real, reproducible gap found — `docs/contributing/custom-docker-images.md` was not updated and now documents a syntax that hard-errors:**

That doc (not touched by this PR) has two call sites still using the pre-PR flat-string form:
```
--set images.apiserver=ghcr.io/michelangelo-ai/apiserver:local-dev \
--set images.controllermgr=ghcr.io/michelangelo-ai/controllermgr:local-dev \
--set images.worker=ghcr.io/michelangelo-ai/worker:local-dev
```
I ran this exact `--set` form against the new chart to check severity:
```
coalesce.go:316: warning: cannot overwrite table with non table for michelangelo.images.apiserver (map[repository:... tag:])
Error: template: .../apiserver-deployment.yaml:111:75: executing ... at <.Values.images.apiserver.repository>: can't evaluate field repository in type interface {}
```
This isn't a silent degrade — it's a hard `helm template`/`helm upgrade` failure. `docs/contributing/custom-docker-images.md` documents exactly the local custom-image dev loop (build your own image, point the sandbox at it) that's actively used for this kind of local iteration, so this is a real workflow this PR would break for anyone following the doc verbatim, not a hypothetical.

**Recommendation:** update the two `--set images.*=` blocks in `docs/contributing/custom-docker-images.md` (lines ~96-98 and ~209-211) to the new `--set images.*.tag=` form before merging, or as an immediate same-day follow-up. The chart/template code itself doesn't need any changes.

No other blocking concerns.

PR #1519, opened 2026-07-14
fix(python): explicitly set local_rank in RayTrainReportCallback

### LINKED ISSUE #1518: RayTrainReportCallback references self.local_rank, which is never assigned
## Summary

`python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py`
defines `RayTrainReportCallback.__init__` setting only:

```python
self.world_rank = ray.train.get_context().get_world_rank()
```

But `on_train_epoch_end()` later references `self.local_rank` (line 71), which is
never assigned anywhere in the class or its Ray base class
(`ray.train.lightning.RayTrainReportCallback` — confirmed by reading its source
directly, it does not define `local_rank` either). The subclass
`RayTrainReportPerNodeCallback` also references `self.local_rank` twice more
(lines 151, 166), and its docstring explicitly describes per-node ("local rank 0")
reporting, so `local_rank` is clearly an intended concept here — it's just never
actually captured as an instance attribute.

## Impact

Any training run using `RayTrainReportCallback` (e.g. via
`tabular_trainer.train_tabular()`) fails at the end of the first training epoch:

```
File ".../michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py", line 71, in on_train_epoch_end
    if self.local_rank == 0:
AttributeError: 'RayTrainReportCallback' object has no attribute 'local_rank'
```

Reproduced with `ray==2.56.0`, `pytorch_lightning==2.6.5`, `torch==2.13.0` (versions
satisfying michelangelo's own declared `>=2.55.1,<3.0.0` / `>=2.2.0,<3.0.0` pins for
`ray`/`pytorch_lightning`), running `michelangelo-examples`'s
`california_housing`/`pytorch_lightning_train` pipeline through a Michelangelo
sandbox (Ray-dispatched `train` task calling `train_tabular()`).

## Suggested fix

Add the missing assignment to `RayTrainReportCallback.__init__`, alongside the
existing `world_rank` line:

```python
self.world_rank = ray.train.get_context().get_world_rank()
self.local_rank = ray.train.get_context().get_local_rank()
```

This single fix in the base class resolves all three call sites, since
`RayTrainReportPerNodeCallback` inherits from `RayTrainReportCallback` and calls
`super().__init__(...)`.

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`RayTrainReportCallback.__init__` now explicitly assigns `self.local_rank = ray.train.get_context().get_local_rank()`, mirroring the existing explicit `self.world_rank` assignment. `on_train_epoch_end()` and the `RayTrainReportPerNodeCallback` subclass read `self.local_rank`, but it was never set directly in this class — it only existed because `super().__init__()` (the upstream `ray.train.lightning.RayTrainReportCallback.__init__`) happens to set it as a side effect.

**Why?**

Fixes #1518, which reports that `self.local_rank` is "never assigned" and would crash `on_train_epoch_end()` with an `AttributeError`.

Investigation: I confirmed via direct instantiation (mocking `ray.train.get_context()`/`ray.get_runtime_context()`) that in real production usage `self.local_rank` **is** actually populated today, because it's set inside the upstream Ray base class's `__init__`, which our subclass calls via `super().__init__()`. So this isn't a live crash in normal operation. However:

- It's fragile and non-obvious — the attribute's existence depends on an undocumented side effect of the upstream base class rather than being owned by this class, and would silently break if Ray ever changes that internal behavior.
- It already had to be worked around in this repo's own test suite (`tests/test_callbacks.py`), which constructs callback instances via `object.__new__(...)` to avoid needing a live Ray Train session, and therefore has to set `local_rank` by hand — a footgun for future tests/usages that don't know to do this.
- I checked the internal Uber SDK equivalent (`RayTrainReportCallback`/`RayTrainReportPerNodeCallback` in the internal monorepo) and it has the exact same pattern (only `world_rank` set explicitly, `local_rank` relies on the same upstream side effect) — so this isn't a regression introduced during the OSS port, it's a pre-existing implicit dependency in both codebases. Making it explicit here removes the fragility for the OSS side regardless.

**How did you test it?**

- Added `test_init_sets_local_rank_from_context` verifying `local_rank` is captured from `ray.train.get_context().get_local_rank()`.
- Added `test_on_train_epoch_end_does_not_raise_attributeerror` for both `RayTrainReportCallback` and `RayTrainReportPerNodeCallback`, constructing the callback through its real `__init__` (with only the upstream base `__init__` stubbed out, since it requires a live Ray Train session) and running `on_train_epoch_end()` end-to-end to confirm no `AttributeError`.
- Ran the full `tests/test_callbacks.py` suite: 24 passed.
- `ruff format --check` and `ruff check` clean on both changed Python files.

**Potential risks**

None — this only adds an explicit assignment of an attribute that was already implicitly available; no behavior change.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

CHANGELOG.md updated under Unreleased > Fixed.

**Documentation Changes**

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,11 @@ All notable changes to this project will be documented in this file.
   what `LightningTrainerParam.initial_weights_path` always expected. No
   storage backend is involved, and a missing file now raises
   `ConfigurationError` eagerly instead of failing deep inside Ray Train.
+- `RayTrainReportCallback.__init__` now explicitly sets `self.local_rank`
+  alongside `self.world_rank` instead of relying on the implicit assignment
+  in the upstream Ray base class `__init__`, guarding `on_train_epoch_end()`
+  and `RayTrainReportPerNodeCallback` against an `AttributeError` if that
+  upstream behavior ever changes.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,10 +230,15 @@ All notable changes to this project will be documented in this file.
   storage backend is involved, and a missing file now raises
   `ConfigurationError` eagerly instead of failing deep inside Ray Train.
 - `RayTrainReportCallback.__init__` now explicitly sets `self.local_rank`
-  alongside `self.world_rank` instead of relying on the implicit assignment
-  in the upstream Ray base class `__init__`, guarding `on_train_epoch_end()`
-  and `RayTrainReportPerNodeCallback` against an `AttributeError` if that
-  upstream behavior ever changes.
+  alongside `self.world_rank` instead of relying on an implicit side-effect
+  assignment in the upstream Ray base class `__init__`. On `ray>=2.56` that
+  upstream side-effect is gone, so `on_train_epoch_end()` (and every
+  `RayTrainReportPerNodeCallback` checkpoint path) raised
+  `AttributeError: 'RayTrainReportCallback' object has no attribute 'local_rank'`
+  during real training. The `ray` dependency is declared without an upper
+  bound, so a fresh install resolves to the latest Ray and hits this live;
+  reading `local_rank` from the public `ray.train.get_context()` API fixes it
+  independent of the Ray version.
 
 ### Removed
 
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/callbacks.py
@@ -30,6 +30,7 @@ class RayTrainReportCallback(ray.train.lightning.RayTrainReportCallback):
     def __init__(self, training_observer: TrainingObserver | None = None) -> None:
         super().__init__()
         self.world_rank = ray.train.get_context().get_world_rank()
+        self.local_rank = ray.train.get_context().get_local_rank()
         self._training_observer = training_observer
 
     def on_train_epoch_end(self, trainer, pl_module) -> None:
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/tests/test_callbacks.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/tests/test_callbacks.py
@@ -103,7 +103,7 @@ def patched_io():
 
 
 class TestRayTrainReportCallbackInit:
-    """``RayTrainReportCallback.__init__`` captures the world rank."""
+    """``RayTrainReportCallback.__init__`` captures the world and local rank."""
 
     def test_init_sets_world_rank_from_context(self):
         """``__init__`` records the world rank from the Ray Train context."""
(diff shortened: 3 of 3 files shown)


sallycr:
## Severity correction: this is now a confirmed live production bug on `ray>=2.56`

This PR was initially framed as *defensive robustness* (the crash only reproduced under test scaffolding that bypasses `__init__` via `object.__new__`, since at `ray==2.51.2` the upstream `RayTrainReportCallback.__init__` still sets `self.local_rank` as a side effect). That framing is now outdated.

**Live reproduction:** the `AttributeError: 'RayTrainReportCallback' object has no attribute 'local_rank'` was reproduced in *real Ray-dispatched training* (not a test) in the `michelangelo-examples` repo's `california_housing/pytorch_lightning_train` pipeline via `train_tabular()`, with:
- `ray==2.56.0`
- `pytorch_lightning==2.6.5`
- `torch==2.13.0`

Upstream Ray removed the implicit `self.local_rank` assignment in `RayTrainReportCallback.__init__` somewhere between `2.51.2` and `2.56.0`. Once that side-effect is gone, `on_train_epoch_end()` (line 72, `if self.local_rank == 0`) and every `RayTrainReportPerNodeCallback` checkpoint path (`_create_and_report_checkpoint`) raise `AttributeError` during actual training.

**Blast radius:** the `ray` dependency in `python/pyproject.toml` is declared **without an upper bound** (bare `ray` in the `trainer`/`plugin`/`example` extras; only a transitive `>=2.7`, lockfile currently at `2.51.2`). A fresh install therefore resolves to the latest Ray on PyPI (currently `>=2.56`) and hits this on the first real training run. This is not gated by an intentional pin.

**Why internal Uber SDK never saw it:** internal pins `ray==2.32.0`, an old version where the implicit side-effect assignment still exists — which is why it "works internally" and only surfaced for OSS users on modern Ray.

**Why the fix holds regardless of Ray version:** the fix reads `self.local_rank = ray.train.get_context().get_local_rank()` — a public Ray Train API that reads the ambient train-session context. It does **not** depend on `super().__init__()` having run or on any upstream-callback-internal state, so it is correct whether or not a given Ray version performs the implicit assignment. It mirrors the existing `self.world_rank` line directly above it.

**Verification just completed:**
- Full module test suite green: `139 passed` (`michelangelo/lib/trainer/torch/pytorch_lightning/tests/`), including the new `test_init_sets_local_rank_from_context` and the two `test_on_train_epoch_end_does_not_raise_attributeerror` regression tests covering both `RayTrainReportCallback` and `RayTrainReportPerNodeCallback`.
- `ruff format --check` and `ruff check` both clean on the touched files.
- CHANGELOG entry reworded from defensive-guard framing to confirmed-live-bug framing (pushed as `0a24b398`).

**Recommendation:** given this is live breakage for any OSS user on modern Ray (which is the default resolution), I'd recommend **expedited human review and merge**. The change is a one-line fix plus regression tests, low risk, and fixes a hard crash on the primary training path.

dkurra (approved):

PR #1528, opened 2026-07-15
Wire revisioned_in and content_index annotations

donovanlopez:
- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Makes `revisioned_in` actually generate code. Three commits:

1. **Refactor** — extract `buildIndexedField` from `ParseIndexedFields` (behavior-preserving) so the per-(key,path) resolution can be reused.
2. **Wiring** —
   - `util.ParseContentIndexedFields` parses `revisioned_in[].content_index[]` into per-wrapper `IndexedField` sets, resolved against the base message.
   - `protoc-gen-sql` emits a `<base>_<wrapper>_unmarshaled` sidecar table per wrapper kind (one column per content-index field, uid FK + indexes).
   - `protoc-gen-kubeproto` generates `GetContentIndexedKeyValuePairs()` and `ContentIndexFieldSpecs()` on revisioned base types.
   - `storage` gains the `ContentIndexable` / `ContentIndexDescribable` interfaces and `ContentIndexField` / `ContentIndexFieldSpec` types the generated code returns.
   - Also fixes a pre-existing bug where multiple `has_resource` messages in one proto file overwrote each other's generated SQL (`buf = append`, not `buf =`); updates `object_expected_output.sql` accordingly.
3. **Templates** — the `mysql_unmarshaled_table.tmpl` header and the`CRDMySQLUnmarshaledTable` / content-index field-spec templates.

Wiring up the annotations should result in creating tables like this (using `Pipeline` as an example):

```
CREATE TABLE `pipeline_revision_unmarshaled`
(
    `revision_uid` VARCHAR(255) NOT NULL,
    `name`    VARCHAR(255),
    `pipeline_type`    VARCHAR(255),
    `branch`    VARCHAR(255),
    `state`    VARCHAR(255),
}

CREATE TABLE `pipeline_draft_unmarshaled`
(
    `draft_uid` VARCHAR(255) NOT NULL,
    `name`    VARCHAR(255),
    `pipeline_type`    VARCHAR(255),
    `branch`    VARCHAR(255),
}
```

**Why?**
This is the mechanism that turns the annotation into filterable sidecar tables and the runtime extractor methods, so a wrapper CRD's List can filter on fields inside its opaque content blob. Grouped together because none of the pieces is meaningful alone — the annotation, the generator, the templates, and the storage interfaces only make sense as a unit.

**How did you test it?**
- `bazel test //go/kubeproto/util/... //go/cmd/kubeproto/protoc-gen-sql/... //go/cmd/kubeproto/protoc-gen kubeproto/...` — all pass.
- The multi-CRD `buf`-fix is covered by the updated `object_expected_output.sql` golden.
- End-to-end generation of an actual sidecar table + extractor method (against real fixtures) is exercised by the follow-up fixtures PR; the full stack was verified locally together before splitting.

**Potential risks**
- Codegen behavior change: the multi-CRD `buf` fix changes generated SQL for any proto file that declares more than one `has_resource` message (previously only the last message's schema was emitted). Files with one CRD each (the common case, incl. go-code) are unaffected.
- Content-index generation only triggers for types with `revisioned_in`; since no proto sets that yet, existing generated output is otherwise unchanged.
- New storage interfaces are additive (implemented by downstream consumers).

**Breaking Changes**

- [x] No breaking changes

**Release notes**

`protoc-gen-sql` / `protoc-gen-kubeproto` now generate `*_unmarshaled` content sidecar tables and `GetContentIndexedKeyValuePairs` / `ContentIndexFieldSpecs` for CRDs annotated with `revisioned_in`. Also fixes SQL generation for protos with multiple resources in one file.

**Documentation Changes**
Recommend a wiki section documenting `revisioned_in` usage and the generated sidecar table naming/shape (not included in this PR).


go/cmd/kubeproto/protoc-gen-sql/main.go
```diff
@@ -109,7 +176,7 @@ func generateSQL(reqData []byte) *pluginpb.CodeGeneratorResponse {
 			}
 
 			if options.Bool("has_resource") {
-				buf = generateSQLSchema(msg, options)
+				buf = append(buf, generateSQLSchema(msg, options)...)
```
craigmarker: i'm pretty sure the convention is one CRD/protobuf, what am I missing?

go/cmd/kubeproto/protoc-gen-kubeproto/main.go
```diff
@@ -434,27 +466,116 @@ func genCRDIndexedFields(crdName string, crdRootMsg *protogen.Message, crdBuf *b
...
+			baseCrdName, wrapperKind, err)
+	}
+	if wrapperOptions == nil || !wrapperOptions.Bool("has_resource") {
+		logger.Panicf("Invalid revisioned_in annotation on %v: kind %q resolves to message %v, which is not "+
+			"a CRD resource (missing michelangelo.api.resource)", baseCrdName, kind, wrapperKind)
+	}
+	return wrapperKind
+}
+
+func snakeToPascalCase(s string) string {
```
craigmarker: consider relocating alongside toSnakeCase in go/api/utils/utils.go

go/kubeproto/util/index.go
```diff
@@ -90,78 +90,152 @@ func ParseIndexedFields(crdRootMsg *protogen.Message, crdOptions *pboptions.Opti
...
+func SidecarTable(baseTableName, kind string) string {
+	return baseTableName + "_" + kind + "_unmarshaled"
+}
+
+// SidecarUIDColumn returns the primary-key column of a wrapper kind's sidecar
+// table (e.g. "revision" -> "revision_uid"). Shared by both generators for the
+// same reason as SidecarTable.
+func SidecarUIDColumn(kind string) string {
+	return kind + "_uid"
+}
```
craigmarker, line 178: thinking out loud: what do you think about consolidating these into a struct to clarify their relationship on one another.

craigmarker (approved):

craigmarker (approved):

PR #1543, opened 2026-07-17
feat(native_transform): add pyarrow conversion helpers (PR A)

### LINKED ISSUE #1537: native_transform PR A: pyarrow conversion helpers in _internal/numpy_utils.py
Part of the `native_transform` migration (see [migration plan](https://webdocs.uberinternal.com/#/d/ba39d9f9-d6ee-4cb5-a6cc-ede1b45dd725)).

**Scope:** Add `pyarrow_to_numpy` and `assemble_output_table` helpers to `python/michelangelo/lib/_internal/numpy_utils.py`. These are hard prerequisites for PR C (the Ray compute lib) — ported from the internal SDK's `shared/utils/numpy_utils/pyarrow_conversion.py`.

**Est. size:** ~230 src LOC, ~150 test LOC
**Priority:** P0
**Depends on:** none

**Out of scope:** Everything else in `native_transform` — this PR is intentionally the smallest, dependency-free slice.

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Ports the two pyarrow-conversion helpers listed as **Hard Prerequisites (Missing in OSS)** item #1 in the `native_transform` migration plan (PR A), into `michelangelo/lib/_internal/numpy_utils.py`:

- `pyarrow_to_numpy(data)` — converts a PyArrow `ChunkedArray`/`Array` to numpy. For uniformly-shaped nested list columns (`list<T>`, `large_list<T>`, `fixed_size_list<T, N>`, any depth) it extracts the flat values buffer and reshapes directly to an N-D typed array, avoiding object-dtype arrays that trigger expensive recursive padding. Falls back to PyArrow's `to_numpy()` for ragged/scalar/unreshapeable inputs.
- `assemble_output_table(input_table, predictions, ...)` — builds a model-output table from zero-copy passthrough input columns plus numpy prediction arrays encoded to native Arrow. Predictions overwrite same-named inputs by default (in-place transform semantics); `raise_on_collision=True` rejects overlaps. Supports `columns_to_keep` and pre-built `extra_columns`.

Supporting dependencies ported alongside: `_try_extract_uniform_nested` (private recursive shape/buffer extractor) and `numpy_to_pyarrow` (numpy→Arrow encoder; renamed from `numpy_to_pyarrow_array` during review for naming symmetry with `pyarrow_to_numpy` — no external callers).

**Review follow-up (pushed after initial open):**
- Fixed a `fixed_size_list` offset bug in `pyarrow_to_numpy` — it read `data.values` without accounting for the parent array's offset, crashing on a sliced bare `Array`. Now slices by `data.offset * size`, mirroring the variable-length-list branch.
- Hardened `numpy_to_pyarrow`'s `tolist()` fallback: log level `debug` → `warning`, and the original exception is now chained via `raise ... from e` if the fallback itself fails.
- Documented (and pinned with a test) a known limitation: null rows in `fixed_size_list` columns currently become sentinel/NaN rather than preserved nulls — a full fix needs a masked-array redesign that would conflict with this PR's behavior-preservation goal, so it's tracked as a documented precondition instead.

This unblocks the Ray compute library port (PR C) in the same migration.

**Why?**

`native_transform` migration PR C (Ray compute lib) depends on these two helpers, which don't yet exist in OSS. Porting them standalone first (small, dependency-free) keeps PR C's diff focused on the Ray-specific logic.

**How did you test it?**

- `pytest michelangelo/lib/_internal/tests/test_numpy_utils.py` → 52 passed (was 46 pre-review-fixes)
- Full `_internal` suite: 54/54 green
- Module coverage: 90.72% (up from 88.46%), above the 85% bar
- `ruff format` + `ruff check` clean on both touched files (also enforced by pre-commit)
- Independently reviewed by two harness agents (OSS Coder for correctness/parity against the internal source, OSS Dev Advocate for API ergonomics) — see PR comments for full findings; all actionable items addressed in the review-follow-up commit above

**Potential risks**

Low. Both functions live in `_internal` (not public API) and have no existing callers yet in this repo — nothing downstream can regress. The one known behavior gap (null handling in `fixed_size_list`, see above) is pre-existing in the internal source this was ported from, not a regression, and is now explicitly documented rather than silently present.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A — internal (`_internal` package) helper additions with no existing callers; not user-facing.

**Documentation Changes**

Google-style docstrings (Args/Returns/Raises) added on both new public functions and the renamed `numpy_to_pyarrow`. No wiki-facing docs needed (internal helper, not public API).

Closes #1537


sallycr:
## DX Review (OSS Dev Advocate)

Reviewed for API ergonomics, docstring clarity, naming consistency, type safety, and test-as-documentation. Overall this is high-quality work — clean signatures, complete type hints, no ambiguous `Any`, runnable doctest examples, and tests that read as living documentation. A few actionable items below, none blocking. **Not approving/merging** — flagging for the author.

### 1. Naming asymmetry between the inverse pair (medium)
`pyarrow_to_numpy` and `numpy_to_pyarrow_array` are inverse operations but named asymmetrically (one has the `_array` suffix, the other doesn't). The next engineer wiring up the Ray compute lib (PR C) will reach for the symmetric name and guess wrong. Since these are `_internal`, renaming is free now before PR C imports them. Suggest one of:
- `pyarrow_to_numpy` / `numpy_to_pyarrow`  (preferred — shortest, symmetric)
- `pyarrow_array_to_numpy` / `numpy_to_pyarrow_array`

### 2. `raise_on_collision=False` default — the less-safe direction (medium, defensible)
The overwrite-on-name-collision default is well-documented and matches the torch path's in-place semantics, so it's a reasonable choice. But the *safe* behavior (fail loud) requires the opt-in, which is backwards from safe-by-default. Rather than flip it (would break the torch path), suggest both consumers **pass the flag explicitly at their call sites** — `raise_on_collision=True` in the inference path, `raise_on_collision=False` in the torch path — so intent is visible where it matters and nobody inherits the unsafe default by accident.

### 3. `numpy_to_pyarrow_array` swallows all conversion errors (medium)
The broad `except Exception` logs at `debug` and silently drops to the slow `tolist()` path. Two DX concerns:
- If the fallback *also* fails, the original error is lost (only in debug logs). Chain it: `raise ... from e`, or re-raise after the fallback fails.
- Silently degrading to the slow path can mask correctness/perf regressions in the fast path. Consider `warning` over `debug` for the fallback log.
- Add a `Raises:` section (currently absent) noting the fallback can still raise, and an `Examples:` block to match `pyarrow_to_numpy`.

### 4. Test gaps for documented-but-untested behavior (low)
Tests are excellent overall. Three documented behaviors lack a locking test:
- `columns_to_keep=[]` (empty collection) keeps all columns — only the `None` case is tested.
- `columns_to_keep` is a **filter, not a reorder** — `["score","id"]` still yields assembled order `["id","score"]`. This is subtle and a consumer may expect it to control order; worth a test to lock the contract.
- The `except`/`tolist()` fallback branch in `numpy_to_pyarrow_array` is never exercised.

### 5. `_try_extract_uniform_nested` offset contract (low)
Docstring correctly notes the `offset=0` contract holds after `combine_chunks()`. Note it only holds for the `ChunkedArray` entry path — a directly-passed *sliced* `pa.Array` skips `combine_chunks`. Fine for the intended callers; just flagging so PR C doesn't pass a pre-sliced array in.

### Positives worth keeping
- Signatures are minimal and intuitive; full type annotations; no ambiguous `Any`.
- Docstrings explain *when* raises fire and *why* defaults are chosen — notably more thorough than the existing terse-prose siblings, which raises the module's bar.
- 23 tests organized per-function with descriptive names; `test_prediction_overwrites_input_by_default` and `test_roundtrip_nd_shape` are great living-doc examples.

sallycr:
## Review feedback addressed

Pushed `3e09fabf` addressing all findings from the OSS Coder and OSS Dev Advocate reviews above:

1. **Naming symmetry (DX #1):** `numpy_to_pyarrow_array` → `numpy_to_pyarrow`. Grepped the whole repo — no external callers, so this is a clean rename (no deprecated alias needed).
2. **Exception handling (DX #3):** the `tolist()` fallback in `numpy_to_pyarrow` now logs at `warning` (was `debug`) and chains the original exception via `raise ... from e` if the fallback itself fails. Added a `Raises:` docstring section.
3. **Fixed-size-list offset bug (Coder #1) — fixed:** `pyarrow_to_numpy` previously read `data.values` without accounting for the parent array's offset, crashing on a sliced bare `Array` (e.g. `fsl_array.slice(1)`). Now slices by `data.offset * size`, mirroring how the variable-length-list branch already handled this. Verified against sliced, length-bounded, and nested/3D sliced fixed-size-list arrays.
4. **Null-row handling (Coder #2) — documented + pinned, not deep-fixed:** proper null preservation would need a masked-array/sentinel redesign across arbitrary nesting depth, which conflicts with this PR's behavior-preservation goal (matches the fallback route both reviews offered). Documented the limitation on `pyarrow_to_numpy`'s docstring and added a regression test pinning current behavior (null FSL row → NaN, int leaf promoted to float) so it can't silently change later.
5. **Coverage gaps (bot + DX #4):** added tests for the offset-fix regression, the fallback warning/chaining path (covers previously-uncovered lines), the ragged-padding fallback, and `columns_to_keep=[]` semantics (confirmed: empty/None keeps all columns, filter-not-reorder).

**Results:** 52 tests pass (was 44) · full `_internal` suite 54/54 green · module coverage 90.72% (was 88.46%) · ruff clean.

Not approved/merged by me — leaving for human review per this team's standing rule.

python/michelangelo/lib/_internal/utils/numpy_utils/pyarrow_conversion.py
kechen2013, line 1: Wondering if we would also need numpy_batch_to_pyarrow_table from the source file uber/ai/michelangelo/shared/utils/numpy_utils/pyarrow_conversion.py (maybe in future PR)?

kechen2013 (approved):

sallycr:
@kechen2013 good question — I traced every caller of `numpy_batch_to_pyarrow_table` in the internal SDK to check.

**Not needed for this PR or any planned native_transform PR (A–F).** Within `native_transform`'s scope (`torch/`, `ray/`, `workflow/tasks/tabular_native_transform/`), the only production file that imports from `pyarrow_conversion.py` is `native_transform/ray/torch_inference.py`, and it only calls `pyarrow_to_numpy` and `assemble_output_table` — both already ported here in PR A. `numpy_batch_to_pyarrow_table` shows up in that module's test file, but only in two tests that re-test the utility function in isolation (`test_numpy_batch_to_pyarrow_table_mixed_shapes`, `test_numpy_batch_to_pyarrow_table_accepts_python_lists`) — they don't exercise `torch_inference.py`'s own logic, so they're not something PR C needs to port over.

Its only *real* production caller internally is `sdk/inference/ray/inference.py`'s `_numpy_dict_to_pyarrow_table` wrapper — but `inference/` is a separate, not-yet-scoped module in the migration plan (maps to OSS `canvas/`, listed "Not started"), unrelated to `native_transform`. There's also one downstream product consumer (`earner_foundation`) outside the SDK entirely.

So: correctly out of scope for PR A, and I don't see it becoming in-scope for B/C/E/F either based on current native_transform code. If/when the `inference/`→`canvas/` migration is scoped, it'd land there instead, not here.

python/michelangelo/lib/_internal/utils/numpy_utils/pyarrow_conversion.py
```diff
@@ -2,9 +2,16 @@
 
 from __future__ import annotations
```
kenns29, line 3: can we moe this whole util to _internal/utils/numpy_utils/...
Internal we have much more numpy utils live under uber/ai/michelangelo/shared/utils/numpy_utils

python/michelangelo/lib/_internal/utils/numpy_utils/pyarrow_conversion.py
sallycr: Confirmed — only torch_inference.py (PR C scope) uses these helpers, and only pyarrow_to_numpy/assemble_output_table (both already ported in PR A). numpy_batch_to_pyarrow_table doesn't appear in any native_transform or tabular_native_transform production source.

We can introduce that function when we use.

sallycr:
@kenns29 good call — done in b406131e. Split `_internal/numpy_utils.py` into a package mirroring the internal layout (`shared/utils/numpy_utils/{sentinel,type,pad,pyarrow_conversion}.py`):

- `_internal/numpy_utils/sentinel.py` — sentinel constants + `sentinel_for_numpy_dtype`
- `_internal/numpy_utils/type.py` — `infer_dtype`
- `_internal/numpy_utils/pad.py` — `pad_ragged_tensor`
- `_internal/numpy_utils/pyarrow_conversion.py` — `pyarrow_to_numpy`, `numpy_to_pyarrow`, `assemble_output_table` (this PR's additions)
- `_internal/numpy_utils/__init__.py` re-exports the public names, so `from michelangelo.lib._internal.numpy_utils import ...` still works unchanged — no caller updates needed.

Tests split the same way under `numpy_utils/tests/`. All 87 tests in `_internal/` + `trainer/torch` (the one existing consumer) pass, ruff clean. This gives room to grow the package as more numpy utils get ported later, matching how much more lives under the internal path you mentioned.

python/michelangelo/lib/_internal/utils/numpy_utils/pyarrow_conversion.py
sallycr: Addressed feedback: split _internal/numpy_utils.py into a utils/numpy_utils/ package (sentinel.py, type.py, pad.py, pyarrow_conversion.py), mirroring the internal SDK's shared/utils/numpy_utils/ layout. The __init__.py re-exports the public API so the existing import path (michelangelo.lib._internal.numpy_utils) and the one existing caller (trainer/torch/data_collate_functions.py) needed zero changes.

python/michelangelo/lib/constants/sentinel.py
```diff
@@ -0,0 +1,38 @@
+"""Type-native sentinel values used to fill padded/missing numpy positions."""
+
+from __future__ import annotations
+
+import numpy as np
+
+INT32_SENTINEL = -(2**31)  # -2147483648, np.iinfo(np.int32).min
```
kenns29, line 5: These sentinel needs to be in a shared constants like internally we have uber/ai/michelangelo/shared/constants/sentinel.py, and over the long term, they shouldn't be private and should be in the public interface

sallycr:
@kenns29 Following up on the numpy_utils location — I moved it one level deeper to match your literal suggestion and the internal layout exactly:

`michelangelo.lib._internal.numpy_utils` → `michelangelo.lib._internal.utils.numpy_utils`

**Why the extra `utils/` segment (not flat under `_internal/`):** the codebase already has an established `<private-bucket>/utils/<name>_utils/` convention in two places:
- top-level `michelangelo/_internal/utils/{file_utils, reflection_utils}`
- `michelangelo/lib/model_manager/_private/utils/{data_utils, module_utils, pickle_utils, ...}`

So `_internal/` is *not* itself the utils bucket — it nests a `utils/` layer, exactly like internal's `shared/utils/`. In fact `file_utils` and `reflection_utils` already live under OSS `_internal/utils/` and are the direct ports of internal's `shared/utils/{file_utils,reflection_utils}`, so putting `numpy_utils` alongside them mirrors `shared/utils/numpy_utils/` 1:1 and leaves room to grow the other numpy utils you mentioned. `errors.py` stays at `_internal/errors.py` since it's a shared exception, not a util.

Public import path is unchanged in spirit — callers now use `from michelangelo.lib._internal.utils.numpy_utils import ...`; the package `__init__` still re-exports the full API. Updated the two callers in `trainer/torch`, ran the suite (87 passed), and ruff format+check are clean. Pushed as a3531c7f. Thanks for the nudge!

python/michelangelo/lib/constants/sentinel.py
sallycr: Good call — done in this PR. I promoted all five sentinels **and** their dispatch helper to a new public top-level constants package:

`michelangelo.lib._internal.utils.numpy_utils.sentinel` → **`michelangelo.lib.constants.sentinel`**

- Moved (not duplicated): `INT32_SENTINEL`, `FLOAT_SENTINEL`, `STRING_SENTINEL`, `BYTES_SENTINEL`, `BOOL_SENTINEL`, and `sentinel_for_numpy_dtype`.
- `lib/constants/__init__.py` re-exports them, mirroring the existing `model_manager/constants/` public-constants pattern. It's a lib-wide (cross-module) package rather than living under any one feature, matching internal's `shared/constants/sentinel.py` being consumed by many unrelated modules.
- `numpy_utils.pad` and the trainer/torch collate helpers + tests now import from the public location; `numpy_utils` no longer re-exports the sentinels. Test moved to `lib/constants/tests/test_sentinel.py`.

Now genuinely public and no longer private. Pushed as 5050a05d.

sallycr:
@kenns29 Both of your review points are addressed now — summary of the two follow-up commits:

**1. numpy_utils path (commit a3531c7f)** — moved to `michelangelo.lib._internal.utils.numpy_utils` (added the `utils/` segment). This matches the repo's established `<private-bucket>/utils/<name>_utils/` convention (top-level `_internal/utils/{file_utils,reflection_utils}` and `model_manager/_private/utils/*`) and mirrors internal's `shared/utils/numpy_utils/` 1:1. `errors.py` stays at `_internal/errors.py` since it's a shared exception, not a util.

**2. Sentinel constants made public (commit 5050a05d)** — promoted all five sentinels (`INT32/FLOAT/STRING/BYTES/BOOL_SENTINEL`) plus `sentinel_for_numpy_dtype` out of the private numpy_utils package into a new public top-level **`michelangelo.lib.constants.sentinel`**, re-exported via `lib/constants/__init__.py` (same pattern as `model_manager/constants/`). It's a lib-wide package since these aren't owned by any single module — matching internal's `shared/constants/sentinel.py`. Callers (`numpy_utils.pad`, trainer/torch collate + tests) now import from the public location; test moved to `lib/constants/tests/`.

Full suite passes (87 tests) and ruff format+check are clean on both. Thanks for the review!

sallycr:
@kenns29 one correction to the earlier sentinel promotion (5050a05d) — checked the internal source more closely: internally, only the raw sentinel *values* live in the public `shared/constants/sentinel.py`. The dispatch function `sentinel_for_numpy_dtype` actually lives separately in `shared/utils/numpy_utils/sentinel.py`, importing those public constants — it's never itself promoted to the public constants module internally.

I'd over-broadened the promotion by moving the function alongside the values. Fixed in e7d6164b:
- `lib/constants/sentinel.py` now holds only the 5 raw values (public).
- `sentinel_for_numpy_dtype` moved back to `_internal/utils/numpy_utils/sentinel.py`, importing the public constants — matching internal's actual public/private split. It takes on a numpy dependency and dtype-branching logic that doesn't belong in a generic, dependency-light constants module.

Tests/ruff still green (87 passed, clean). Thanks for prompting the closer look.

sallycr:
Independent OSS Coder + Dev Advocate review of the module restructure (commits b406131e → 6d01b4f3):

**OSS Coder:** clean and ready — import graph acyclic, no stale references to any of the intermediate paths, test coverage identical pre/post split (87 tests, zero dropped/duplicated), `__all__` matches actual exports on both `lib/constants/` and `_internal/utils/numpy_utils/`, sanitization clean, ruff clean.

**OSS Dev Advocate: APPROVED.** Naming reads naturally against existing conventions (`_internal/utils/<name>_utils/`, `lib/constants/` vs `model_manager/constants/`), docstrings on the public pyarrow helpers are exemplary (Google-style, runnable examples, the null-handling limitation called out explicitly), and the three rounds of restructuring landed at a proportionate end state — not over-engineered for what PR A needed. One nit: `lib/constants/__init__.py`'s docstring didn't point to `sentinel_for_numpy_dtype` for dtype-based selection, which could leave a reader wondering how to pick the right sentinel. Fixed in 6d01b4f3.

No other changes needed. This should be ready for a final look.

PR #1548, opened 2026-07-18
feat(native_transform): add IDHashTokenizer torch layer (PR A2)

### LINKED ISSUE #1538: native_transform PR A2: port IDHashTokenizer to lib/native_transform/torch/
Part of the `native_transform` migration (see [migration plan](https://webdocs.uberinternal.com/#/d/ba39d9f9-d6ee-4cb5-a6cc-ede1b45dd725)).

**Scope:** Port `IDHashTokenizer` (pure torch, 251 LOC) from `sdk/models/layers/pretrained_embedding` to `python/michelangelo/lib/native_transform/torch/`. This is a hard prerequisite for PR B (torch core), which references it. Alternatively, descope and stub with `NotImplementedError` if not needed at launch — flagged as an open question in the migration plan.

**Est. size:** ~251 src LOC, ~100 test LOC
**Priority:** P0
**Depends on:** none

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds a new `michelangelo/lib/native_transform/torch/` package containing `IDHashTokenizer`, a small, pure-`torch` `nn.Module`:

- `IDHashTokenizer(nn.Module)` in `id_hash_tokenizer.py` — maps arbitrary integer IDs to contiguous, zero-based vocabulary indices, mapping any out-of-vocabulary value to a dedicated unknown index (`len(vocabulary)`). This is the standard "tokenize integer IDs into an embedding-table row index" building block used when embedding categorical/id features in a model.
  - Accepts an **unsorted** vocabulary and preserves the provided-list position as the output index (`vocabulary[i]` maps to `i`); internally it sorts for an efficient `torch.bucketize` lookup and remaps back, so caller ordering is honored.
  - Deduplicates the vocabulary, keeping the first occurrence (and warns via `warnings.warn` when duplicates are dropped).
  - Is both **TorchScript-** and **ONNX-exportable** via buffer-backed lookup tables, so the exact same transform runs at train and serve time.
- Seeds the new `lib/native_transform/` and `lib/native_transform/torch/` packages (with `__init__.py`s), and re-exports `IDHashTokenizer` at the package level so consumers can `from michelangelo.lib.native_transform.torch import IDHashTokenizer`.

Code-quality notes:
- `forward` builds the unknown-index constant with `input_ids_long.new_full((), self.unk_index)` so the constant follows the input's device instead of baking a literal device into a TorchScript trace.
- Google-style docstrings (Args/Returns/Raises) and full type hints on the public API.

**Why?**

`IDHashTokenizer` is a self-contained, dependency-free (`torch` / `torch.nn` only) layer that a larger follow-up PR will build on to assemble native feature-transform graphs. Landing it standalone keeps that follow-up's diff focused on the layer-registry logic, and makes this primitive independently reviewable and reusable on its own.

**How did you test it?**

- `pytest michelangelo/lib/native_transform/torch/tests/test_id_hash_tokenizer.py` → 9 passed, 1 skipped (the ONNX round-trip test skips when `onnx`/`onnxruntime` are absent)
- Tests cover 2D and 1D lookups, all-out-of-vocabulary input, an unsorted vocabulary, a **TorchScript `torch.jit.script` + save/load round-trip** (guards the serving-path contract), a duplicate-vocabulary warning, empty/non-integer vocabulary rejection, non-integer input rejection, and an **ONNX export + onnxruntime** round-trip
- Module coverage: 100% on `id_hash_tokenizer.py`, above the 85% bar
- `ruff format --check` + `ruff check` clean on all touched files (also enforced by pre-commit)

**Potential risks**

Low. `IDHashTokenizer` is a new public class with no existing callers in this repo yet — nothing downstream can regress. Pure-`torch`, no external infra coupling. TorchScript and ONNX export are both covered by tests to protect the serving path.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A — new layer in the `native_transform` library; no existing callers, not yet wired into a user-facing entry point.

**Documentation Changes**

Google-style docstrings (Args/Returns/Raises) added on the new public class, its `__init__`, and `forward`, plus module- and package-level docstrings.

Closes #1538

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
+++ b/python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
@@ -1,0 +1,187 @@
+"""ID-hash tokenizer layer for native feature transforms.
+
+Maps arbitrary integer input values to contiguous, zero-based indices based on a
+provided vocabulary, mapping out-of-vocabulary values to a dedicated unknown
+index. The layer is TorchScript- and ONNX-exportable so it can be embedded in a
+model graph and run identically at train and serve time.
+"""
+
+from __future__ import annotations
+
+import warnings
+
+import torch
+from torch import nn
+
+__all__ = ["IDHashTokenizer"]
+
+
+class IDHashTokenizer(nn.Module):
+    """Map integer IDs to contiguous vocabulary indices.
+
+    Maps arbitrary input integer values to new, contiguous integer indices based
+    on a provided vocabulary. Values not found in the vocabulary are mapped to an
+    unknown index, which is set to the size of the (deduplicated) vocabulary.
+
+    The input ``vocabulary`` may be unsorted. The mapping from an original
+    vocabulary value to its new index is based on its position in the *provided*
+    vocabulary list (i.e. ``vocabulary[i]`` maps to ``i``). Internally the values
+    are sorted for an efficient :func:`torch.bucketize` lookup, then remapped back
+    to their original positions, so ordering of the provided list is preserved in
+    the output indices.
+
+    The layer is compatible with both TorchScript and ONNX export.
+
+    Args:
+        vocabulary: List of integer values to map to contiguous indices. Duplicate
+            values are removed, preserving the index of their first occurrence.
+
+    Raises:
+        TypeError: If ``vocabulary`` is not a list of integers.
+
+    Example:
+        >>> tokenizer = IDHashTokenizer(vocabulary=[-10, -3, 0, 2, 4, 6])
+        >>> tokenizer(torch.tensor([-10, 0, 5], dtype=torch.long))
+        tensor([0, 2, 6])
+    """
+
+    def __init__(self, vocabulary: list[int]) -> None:
+        """Initialize the tokenizer from a vocabulary of integer values.
+
+        Args:
+            vocabulary: List of integer values to map to contiguous indices.
+                Duplicate values are removed, preserving the index of their first
+                occurrence.
+
+        Raises:
+            TypeError: If ``vocabulary`` is not a list of integers.
+        """
+        super().__init__()
+        # Validate input vocabulary type.
+        if not isinstance(vocabulary, list) or not all(
+            isinstance(v, int) for v in vocabulary
+        ):
+            raise TypeError("Vocabulary must be a list of integers.")
+
+        # Deduplicate while preserving original order. If duplicates exist (e.g.
+        # [10, 20, 10]), only the first occurrence is kept and its original index
+        # is used for the mapping.
+        unique_vocab_values: list[int] = []
+        seen_values: set[int] = set()
+        for v in vocabulary:
+            if v not in seen_values:
+                unique_vocab_values.append(v)
+                seen_values.add(v)
+
+        if len(unique_vocab_values) != len(vocabulary):
+            warnings.warn(
+                "Duplicate values found in vocabulary. Only unique values will be "
+                "used, preserving the index of their first occurrence. Original "
+                f"size: {len(vocabulary)}, effective unique size: "
+                f"{len(unique_vocab_values)}.",
+                stacklevel=2,
+            )
+
+        # Store the effective unique vocabulary (preserving original order).
+        self.vocabulary = unique_vocab_values
+
+        # The unknown index is one past the last valid mapped index.
+        self.unk_index = len(self.vocabulary)
+        self.output_vocab_size = len(self.vocabulary)
+
+        # Map each unique vocabulary value to its original desired output index
+        # (its position in ``self.vocabulary``).
+        value_to_original_idx_map = {
+            val: idx for idx, val in enumerate(self.vocabulary)
+        }
+
+        # Sorted unique values used by torch.bucketize for efficient lookup.
+        sorted_unique_values = sorted(self.vocabulary)
+        self.register_buffer(
+            "_sorted_unique_values_tensor",
+            torch.tensor(sorted_unique_values, dtype=torch.long),
+            persistent=True,  # Saved with the model and moved by ``.to(device)``.
+        )
+
+        # Map from an index in ``_sorted_unique_values_tensor`` back to the value's
+        # *original* index in ``self.vocabulary``, preserving the caller's desired
+        # mapping order. Example: vocabulary=[0, -10] -> sorted=[-10, 0]; -10
+        # (sorted index 0) was at original index 1 and 0 (sorted index 1) was at
+        # original index 0, so this mapping is [1, 0].
+        sorted_to_original_idx_mapping = [
+            value_to_original_idx_map[val] for val in sorted_unique_values
+        ]
+        self.register_buffer(
+            "_sorted_to_original_idx_mapping_tensor",
+            torch.tensor(sorted_to_original_idx_mapping, dtype=torch.long),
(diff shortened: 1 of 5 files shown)


sallycr:
Addressed DX review findings in 78e4e5df:

**Blocking fix — empty vocabulary now rejected early.** `IDHashTokenizer(vocabulary=[])` previously constructed fine (the `all(...)` type check is vacuously true on an empty list) and only failed on the first `forward()` call, where `output_vocab_size - 1 == -1` collapsed indices to `-1` and indexing the empty lookup buffer raised a bare `IndexError` — worst case surfacing post-export at serve time. `__init__` now raises `ValueError("vocabulary must be non-empty.")` up front, alongside the existing `TypeError` validation. Added `test_rejects_empty_vocabulary` asserting this.

**Minor (applied):**
- Class docstring now notes that despite the name "Hash", this is an exact vocabulary lookup via `torch.bucketize`, not a hash (name kept for parity with the internal SDK layer).
- Package docstring (`native_transform/__init__.py`) expanded to note the full transform DAG (TransformSpec, TorchTransformModule, etc.) lands in a follow-up PR. Left `torch/__init__.py` untouched as suggested.

Not acted on (per review, non-blocking): `bool`/int validation quirk, dual-location `vocabulary` doc, and the optional `Attributes:` section.

Tests: `native_transform/` now 9 passed, 1 skipped (ONNX). `ruff format --check` + `ruff check` clean.

sallycr:
### Second review round applied (from the blind-review pass)

This branch went through a review pass where reviewers evaluated it purely as a standalone external contribution, with no project-specific context. That surfaced a few things worth tightening, now addressed in `7fa824c1`:

1. **PR description rewritten** to be self-contained — it now describes what `IDHashTokenizer` does, why it's useful, and its dependency-free / no-existing-callers status without relying on any project-internal shorthand. Also corrected a stale count (the suite is **9 passed, 1 skipped**, not 8, after the empty-vocabulary fix landed).
2. **ONNX test skip-gating fixed** — `torch.onnx.export(...)` needs the `onnx` package, but the test only did `pytest.importorskip("onnxruntime")`. If `onnxruntime` were installed without `onnx`, the test would ERROR instead of cleanly SKIP. It now `importorskip`s both.
3. **Docstring precision fix** — the `torch.bucketize` comment said a result equal to `len(sorted)` means the input is `>=` the last sorted element. That's imprecise: a value *equal to* the last element lands at index `size - 1`; only values *strictly greater* land at `size`. Reworded to `>`.
4. **Convenience re-export added** — `from michelangelo.lib.native_transform.torch import IDHashTokenizer` now works (with `__all__`), matching the re-export convention used by sibling packages, so users don't need the deep module path.

Non-blocking style suggestions from the pass (the `bool`-passes-`int` check, collapsing `unk_index`/`output_vocab_size`, dropping the unused `vocabulary` attribute, adding `onnx`/`onnxruntime` as hard test deps) were intentionally left as-is to keep this PR focused.

Verification: `pytest michelangelo/lib/native_transform/` → **9 passed, 1 skipped**; `ruff format --check` + `ruff check` clean.

python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
```diff
@@ -0,0 +1,199 @@
...
+        Raises:
+            TypeError: If ``input_ids`` is not of integer type (``torch.int32`` or
+                ``torch.long``).
+        """
+        # Ensure input tensor is of integer type.
+        if input_ids.dtype != torch.int32 and input_ids.dtype != torch.long:
+            raise TypeError(
+                "Input tensor must be of integer type (torch.int32 or "
+                f"torch.long), but got {input_ids.dtype}."
+            )
```
kechen2013, line 155: Wondering if it's intended to raise TypeError and allow only integer inputs?  I think this is cleaner with strong type requirement, however not sure if it would break for the case where user implicitly set float as input.
The internal base_layers.IDHashTokenizer silently casts float inputs to int32 before tokenizing (to handle native transform pipelines that emit float tensors).

python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
```diff
@@ -0,0 +1,199 @@
...
+
+import warnings
+
+import torch
+from torch import nn
+
+__all__ = ["IDHashTokenizer"]
+
+
+class IDHashTokenizer(nn.Module):
```
kechen2013, line 19: Seems the open-source IDHashTokenizer does not have support for sequence feature (format_inputs) and ragged tensor (replace sentinel value) as in internal version?  https://sg.uberinternal.com/r/code.uber.internal/uber-code/uber-one/-/blob/uber/ai/michelangelo/sdk/native_transform/torch/base_layers.py?L855-871

python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
sallycr: Good catch, and I traced this through to double-check.

This class is the *bare* tokenizer core — a 1:1 port of the standalone `IDHashTokenizer` (integer-vocab -> contiguous index, unknown -> unk_index). Float-to-int normalization is a caller concern, not something this core should do implicitly: silently truncating/casting a float tensor hides a real bug if a caller passes floats by mistake (e.g. an upstream dtype error), and `TypeError` surfaces that immediately instead of producing quietly-wrong indices.

The pipeline layer that wraps this core with dict I/O (a `forward(dict[str, Tensor]) -> dict[str, Tensor])` transform layer) is a separate, larger piece of work — that's where input normalization for float-emitting upstream layers belongs, since it already owns dtype handling for the whole transform DAG. This PR intentionally does not include that wrapper; it's scoped to only the core class so the wrapper's PR isn't blocked on it. I'll make sure the wrapper's design carries this normalization forward when it lands.

python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py
sallycr: Confirmed by reading the internal wrapper directly — this is intentional scoping, not a gap.

`format_inputs`/`format_outputs` (dict-of-tensors <-> tensor) and the ragged-fill/sentinel preservation (`pad_mask`, NaN-vs-`INT32_SENTINEL` handling) both live on the *wrapper* class that subclasses the shared pipeline base and does `forward(dict[str, Tensor]) -> dict[str, Tensor]`, not on the bare tokenizer core this PR ports. The wrapper composes the core (constructs it internally, delegates the actual index lookup to it) and layers the dict I/O + sentinel plumbing on top.

That wrapper belongs to the ~20-layer transform-DAG PR (all layers built on the same base class, TransformSpec, schema derivation) — a much larger, separate piece of work that depends on this PR landing first (this PR seeds the package). Baking sequence/ragged handling into the core class now would duplicate logic the wrapper needs to own anyway, and would tie this small, self-contained PR to unrelated design decisions in the DAG layer. I've noted this in our migration plan so it isn't lost: when that wrapper PR lands, it needs to reproduce the sentinel-capture-before-cast ordering (capturing the pad mask before the float->int cast, since NaN is lost after casting) — worth flagging it there too so it isn't missed during that implementation.

kechen2013 (approved):

Issue #1555, opened 2026-07-19
Incorrect history server image leads to ImagePullBackOffError


Capiru:
**Describe the bug**
When deploying a local k3d cluster using `ma sandbox create`, the final deployment is in a Failed State. The non existing `ghcr.io/michelangelo-ai/kuberay-historyserver:main` image leads to `ImagePullBackOffError`. The root cause is that `scripts/kuberay/build-kuberay-images.sh` builds the image in the local k3d cluster under the name `kuberay-historyserver:v0.1.0` instead.

**To Reproduce**
1. Create kuberay images using `bash scripts/kuberay/build-kuberay-images.sh`
2. Deploy local sandbox cluster with `ma sandbox create`
3. Verify that the pods from the `history-server` deployment have the failed events `ImagePullBackOffError`

**Expected behavior**
`ma sandbox create` should scaffold the local cluster using the expected image generated via `scripts/kuberay/build-kuberay-images.sh`.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: MacOS

**Smartphone (please complete the following information):**
N/A

**Additional context**
The sandbox tutorial at https://michelangelo-ai.org/docs/getting-started/sandbox-setup#history-server-stuck-in-imagepullbackoff did not solve the issue described above.

The fix in the meantime is simple: Users/developers can use `kubectl set image deployment/history-server history-server=kuberay-historyserver:v0.1.0` to use the expected image.

And a PR only needs to change this line here: https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/cli/sandbox/resources/history-server.yaml#L48

Capiru:
I tested locally, and my proposed fix seems to work. Opened up PR #1556 .

Capiru:
```
And a PR only needs to change this line here: https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/cli/sandbox/resources/history-server.yaml#L48
```
After re-thinking a little bit about this proposed implementation, the CI would likely fail in any integration or e2e test in which it deploys the sandbox as part of it. A better proposal could be using the local image name/tag as a fallback in case a user cannot pull from the ghcr registry. 


From the function docstring, I saw the below section:

```
 ... since the collector sidecar and history server are optional for basic sandbox usage
```
[...]

PR #1556, opened 2026-07-19
Fix: Update history server image name and version to v0.1.0

Capiru:
- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fix the incorrect `history-server` image reference as referenced by https://github.com/michelangelo-ai/michelangelo/issues/1555 .

**Why?**
This should make it easier and smoother to deploy a `sandbox` cluster locally.

**How did you test it?**
I fixed the reference locally, and deleted and re-created the cluster. It then successfully was able to start the pod as expected.
image

**Potential risks**
This implementation might fail the e2e test if it pulls from `ghcr.io/michelangelo-ai/kuberay-historyserver:main` and uses `ma sandbox create` as part of it.

**Breaking Changes**

- [X] No breaking changes

**Migration guide**

**Release notes**
Fixed a reference mismatch on sandbox clusters where history-server was pointing to an incorrect reference.

**Documentation Changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -239,8 +239,8 @@ All notable changes to this project will be documented in this file.
   bound, so a fresh install resolves to the latest Ray and hits this live;
   reading `local_rank` from the public `ray.train.get_context()` API fixes it
   independent of the Ray version.
-- Fixed a reference mismatch on sandbox clusters where `history-server` was
-  pointing to an incorrect reference.
+- The sandbox History Server now falls back to its locally built image when
+  GHCR cannot be reached.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -239,6 +239,8 @@ All notable changes to this project will be documented in this file.
   bound, so a fresh install resolves to the latest Ray and hits this live;
   reading `local_rank` from the public `ray.train.get_context()` API fixes it
   independent of the Ray version.
+- Fixed a reference mismatch on sandbox clusters where `history-server` was
+  pointing to an incorrect reference.
 
 ### Removed
 
--- a/python/michelangelo/cli/sandbox/resources/history-server.yaml
+++ b/python/michelangelo/cli/sandbox/resources/history-server.yaml
@@ -45,7 +45,7 @@ spec:
       serviceAccountName: historyserver
       containers:
       - name: history-server
-        image: kuberay-historyserver:v0.1.0
+        image: ghcr.io/michelangelo-ai/kuberay-historyserver:main
         imagePullPolicy: IfNotPresent
         command:
         - historyserver
--- a/python/michelangelo/cli/sandbox/resources/history-server.yaml
+++ b/python/michelangelo/cli/sandbox/resources/history-server.yaml
@@ -45,7 +45,7 @@ spec:
       serviceAccountName: historyserver
       containers:
       - name: history-server
-        image: ghcr.io/michelangelo-ai/kuberay-historyserver:main
+        image: kuberay-historyserver:v0.1.0
         imagePullPolicy: IfNotPresent
         command:
         - historyserver
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -28,6 +28,12 @@ _michelangelo_sandbox_kube_cluster_name = "michelangelo-sandbox"
 _cadence_domain = "default"
 _default_compute_kube_cluster_name = "michelangelo-compute-0"
 
+# Keep this in sync with IMAGE_TAG in scripts/kuberay/build-kuberay-images.sh.
+_kuberay_local_image_tag = "v0.1.0"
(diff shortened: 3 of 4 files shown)


paulzim (changes requested):
## Review: #1556 — Fix: Update history server image name and version to v0.1.0

**Base:** `michelangelo-ai:main` · **Head:** `Capiru:fix/history-server-image-tag` · fixes #1555 · 3 commits

### Summary
Adds a fallback in `_import_kuberay_images()`: if the GHCR-hosted `kuberay-historyserver` image can't be pulled/imported, switch the already-applied `history-server` Deployment to the image built locally by `scripts/kuberay/build-kuberay-images.sh` (`kuberay-historyserver:v0.1.0`) via `kubectl set image`. Good direction, includes tests and a CHANGELOG entry, and the final commit correctly reverted an earlier, unrelated resources/manifest edit.

### Requesting changes

**1. Fallback fires on the wrong image (functional bug)**

`_import_kuberay_images()` loops over `_KUBERAY_IMAGES = [kuberay-collector:main, kuberay-historyserver:main]` and calls `_use_local_historyserver_image()` whenever **any** image in that loop fails to pull or import — not just when `kuberay-historyserver` itself fails. `kuberay-collector` has no local-build fallback and is documented (`values-k3d.yaml`) as "pulled from GHCR automatically, no local build needed" — it's an unrelated sidecar image injected into RayCluster pods, not the History Server.

Repro (collector pull fails, historyserver pull+import both succeed):
```
Importing ghcr.io/michelangelo-ai/kuberay-collector:main into k3d...
Warning: could not pull ... Using local History Server image.
Importing ghcr.io/michelangelo-ai/kuberay-historyserver:main into k3d...
Successfully imported ghcr.io/michelangelo-ai/kuberay-historyserver:main into k3d.
fallback call count: 1
```
So a transient failure pulling the *collector* image needlessly downgrades the History Server Deployment to the local image via `kubectl set image` — even though the GHCR historyserver image just imported fine, and even though most sandboxes won't have pre-built the local image (per the existing local-build prerequisite), turning a working deployment into a broken one.

Suggested fix: scope the fallback to the historyserver image specifically, e.g.
```python
if pull.returncode != 0:
    print(f"Warning: could not pull {image}.")
    if image == _KUBERAY_HISTORYSERVER_IMAGE:
        _use_local_historyserver_image()
    continue
```
(and same for the k3d-import failure branch). The new tests don't catch this because both cover "every image in the loop fails," which happens to also produce the intended call count — a mixed pass/fail case across the two images would have caught it.

**2. CI-blocking lint/format failures**

- `poetry run ruff check .` → 3× `D102` (missing docstring) on the new `test_uses_local_historyserver_when_ghcr_pull_fails`, `test_uses_local_historyserver_when_k3d_import_fails`, `test_imports_and_sets_local_historyserver_image` methods. This repo enforces `D` rules on new code (`pyproject.toml`), so this fails the PR checklist's `ruff check` gate as-is.
- `poetry run ruff format --check .` → `sandbox.py`'s new `_kuberay_local_historyserver_image` assignment gets reformatted to a single line; currently fails the formatter check.

Both are one-line fixes (add docstrings, run `ruff format .`), just flagging since they'll fail CI in their current form.

### What looks right
- `_kuberay_local_historyserver_image` tag is kept in sync with `IMAGE_TAG` in `build-kuberay-images.sh` via a comment — reasonable given there's no shared source of truth.
- `history-server.yaml`'s image reference is untouched (still `ghcr.io/michelangelo-ai/kuberay-historyserver:main`, already fully-qualified) — confirms the earlier "revert resources" commit correctly walked back an unrelated manifest edit.
- CHANGELOG entry is appropriately scoped.

### Verdict
Request changes — the collector/historyserver conflation is a real regression risk (can break a working sandbox on an unrelated image's transient failure), and the lint/format gates need to pass before merge. The overall approach (local-image fallback via `kubectl set image`) is sound once scoped correctly.

sallycr (approved):
Thanks for improving the sandbox reliability.

sallycr (changes requested):
Can you address the feedback that was provided by @paulzim? Thanks again for the improvement.

dragod812:
Hi @Capiru , Thanks for looking into this, but i ran the build kuberay images pipeline on main branch, and now we have the main tagged history server images -> https://github.com/michelangelo-ai/michelangelo/actions/runs/29815300314
You can try running sandbox locally once the main images are available.

dragod812 (changes requested):
This is not needed, we have the main tagged image now -> https://github.com/michelangelo-ai/michelangelo/pkgs/container/kuberay-historyserver

Thanks for pointing out that it was missing before.

Capiru:
Hey @dragod812 , thanks for giving some more context and deploying those images. Now that the images are public `ma sandbox create` is self contained (i.e. no need for `bash scripts/kuberay/build-kuberay-images.sh`). This PR is no longer needed.

I would suggest also as a chore to update the docs regarding the sandbox as described above, as it should simplify the UX.

PR #1561, opened 2026-07-21
fix(release): exclude RC/nightly tags from changelog diff boundary, commit CHANGELOG.md via PR

sallycr:
Found while checking v0.5.0's GitHub Release notes after promoting from rc.2.

**Bug 1:** `cliff.toml`'s `tag_pattern` matches every `v*` tag including RCs; `skip_tags` only excluded `.dev` strings, not the actual `-rc.N`/`-nightly.<date>` git tag formats. So `changelog.yml`'s `--latest` mode anchored `v0.5.0`'s release notes on `v0.5.0-rc.2` instead of `v0.4.0`, producing notes covering only the last RC's fixes instead of the whole minor release. Confirmed live: https://github.com/michelangelo-ai/michelangelo/releases/tag/v0.5.0 currently shows only 2 entries. Fixed by adding `-rc.*`/`-nightly.*` to `skip_tags`.

**Bug 2:** `changelog.yml` regenerates `CHANGELOG.md` via git-cliff but only ever uploads it as a workflow artifact, never commits it to the repo. Confirmed: `CHANGELOG.md` has been stuck at the `0.4.0` section since it was hand-updated during v0.4.0 prep. Fixed by opening a PR with the updated file via `peter-evans/create-pull-request` (both `main` and release branches require PRs via an org ruleset, so direct push isn't possible), auto-detecting the correct base branch from which branch contains the tagged commit.

Verified: `actionlint` and YAML syntax check both pass clean on the modified workflow.


paulzim (changes requested):
## Review: #1561 — fix(release): exclude RC/nightly tags from changelog diff boundary, commit CHANGELOG.md via PR

**Verdict: Request changes**

This PR bundles two independent fixes. One is sound. The other — the one in the title, and the one addressing the actual reported symptom — doesn't work, and I can reproduce that byte-for-byte.

### Bug 2 (CHANGELOG.md committed via PR): looks correct, keep it

Verified independently: the new `Determine base branch for tag` step's logic —

```bash
git branch -r --contains "${{ steps.tag.outputs.tag }}" | sed 's#^[* ]*origin/##' | grep -E '^release/v[0-9]' | sort -V | tail -1
```

— correctly resolves `v0.5.0` to `release/v0.5` when run against the real repo history. The fallback to `main` when no `release/v*` branch contains the tag is reasonable. `pull-requests: write` is the correct added permission for `peter-evans/create-pull-request@v6`. I have no objections to this half landing.

### Bug 1 (RC/nightly changelog boundary): symptom is real, fix does nothing

**The symptom is real.** I pulled the actual CI log for the `v0.5.0` tag push (run `29788544720`, job `88505178340`) and confirmed the `--latest --strip header` step really did generate only a 2-entry release body at 2026-07-20T23:51:50Z — matching the PR description's complaint exactly.

**The fix does not address it.** I tested this empirically against a fresh clone of the repo at the `v0.5.0` tag, using git-cliff 2.13.1 (matching the version `orhun/git-cliff-action@v4` resolves to in CI) — not just read the diff:

```
$ git-cliff --config cliff.toml            --latest --strip header   > out_old.md   # main's current cliff.toml
$ git-cliff --config <PR's cliff.toml>     --latest --strip header   > out_new.md   # this PR's skip_tags change
$ diff out_old.md out_new.md && md5sum out_old.md out_new.md
IDENTICAL - confirmed byte-for-byte no difference
dad89bec5c6001f50fe82491c622459c  out_old.md
dad89bec5c6001f50fe82491c622459c  out_new.md
```

Adding `.*-rc\..*|.*-nightly\..*` to `skip_tags` has **zero effect** on the exact command the workflow runs. The release notes for `v0.5.0` would still ship as the same 2-entry list this PR claims to fix. Two independent reasons why:

**1. Wrong git-cliff option.** git-cliff's own docs draw a hard line between `skip_tags` and `ignore_tags`:
- `skip_tags` **drops** the matched tag's commits from processing entirely — they disappear from output, full stop.
- `ignore_tags` **folds** those commits forward into the next non-ignored tag's section.

The PR's stated intent ("RC/nightly tags shouldn't reset the diff boundary, their commits should count toward the next real release") is exactly what `ignore_tags` does and `skip_tags` does not. I confirmed this concretely: switching `0.5.0-rc.2`'s commits to `ignore_tags` instead of `skip_tags` makes PR #1553's commit (`Allow x-user-email header in envoy CORS preflight`) reappear in the full changelog output; under this PR's actual `skip_tags`-based diff, it's gone — 0 matches, confirmed via `grep -c`.

**2. Even the correct option wouldn't cleanly fix this repo.** `v0.4.0` is not an ancestor of `v0.5.0` in this repo's commit graph (`git merge-base --is-ancestor v0.4.0 v0.5.0` → false) — the release branches diverge (`release/v0.4` vs `release/v0.5`), so no amount of `cliff.toml` tag-regex tuning can make `--latest` walk cleanly from "the last real release" to `v0.5.0`. I confirmed this too: even with `ignore_tags` set correctly, the full changelog's `[0.5.0]` section balloons all the way back past `v0.4.0` to `v0.3.1` (over 100 commits), because git-cliff's forward-fold walks the raw commit graph, not release-branch identity. So the fix that *would* work for `skip_tags` vs `ignore_tags` in isolation still produces a wrong/oversized boundary here — the root problem is structural, not a regex tuning problem.

### New regression introduced by this PR

Before this PR, RC-only commits get their own visible section in `CHANGELOG.md` (e.g. `## [0.5.0-rc.2]` at line 22 of the current full changelog, with PR #1553's commit at line 28 under it). After this PR's `skip_tags` change, that section — and #1553's commit — **vanishes from `CHANGELOG.md` entirely**, confirmed via `grep -c "1553"` → `0` on the full changelog generated with this PR's config. That's a silent loss of changelog content that exists today, introduced by the fix meant to make the changelog more complete.

### Suggested path forward

- Split this into two PRs — Bug 2's mechanism is good and shouldn't wait on Bug 1.
- For Bug 1, `skip_tags` → `ignore_tags` is a necessary change but not sufficient given the `release/v0.4` vs `release/v0.5` branch divergence. Given that, tuning `cliff.toml` regexes alone can't reliably solve the boundary problem in this repo — worth considering computing the previous "real" tag explicitly in the workflow (e.g. via a version-sorted `git tag -l 'v*'` filtered to exclude `-rc`/`-nightly`/`.dev`, then passing `--tag`/an explicit range to git-cliff) rather than relying on git-cliff's own tag-adjacency walk.

Happy to help verify a revised version against the same clean-clone repro I used here.

sallycr:
Thanks for the thorough repro — confirmed both findings independently (skip_tags vs ignore_tags byte-for-byte, and the v0.4.0/v0.5.0 non-ancestor issue).

Pushed a revised fix (`52914f8b`) that replaces the cliff.toml-only approach with an explicit range computed in the workflow, per your suggestion:

- `cliff.toml`: switched `skip_tags` → `ignore_tags` for rc/nightly (verified locally: PR #1553/#1559's commits now correctly reappear in the full changelog, folded into their release's section, instead of vanishing).
- `changelog.yml`: new "Determine changelog range" step. Instead of relying on git-cliff's own `--latest` tag-adjacency walk, it explicitly resolves the immediately-preceding **final** tag and checks whether it's a real ancestor of the tag being released:
  - If yes (the common case, and correctly handles patches too — e.g. `v0.5.0` is `v0.5.1`'s immediately-preceding tag *and* a real ancestor, so a patch's notes are scoped to `v0.5.0..v0.5.1`, not re-walked back further), use that tag as the range start.
  - If no (the `v0.4.0`/`v0.5.0` case you found), fall back to the *previous release branch's fork point off `main`* — always a real ancestor since `main` only grows forward — rather than searching further back for any older ancestor tag (which would silently re-include an entire prior cycle's content, e.g. landing on `v0.3.1` and re-listing all of `v0.4.0`'s changes in `v0.5.0`'s notes).

Verified locally against this repo's real history with git-cliff 2.13.1: `v0.5.0` now resolves to `release/v0.4`'s fork point (28 commits — matches the full real changelog I hand-curated for that release), and a simulated `v0.5.1` tag on `release/v0.5` correctly resolves to `v0.5.0..v0.5.1` instead of ballooning back to `v0.4.0`.

Kept both fixes in one PR since Bug 1 no longer blocks — happy to split if you'd still prefer that. Let me know if you want to re-run your repro against `52914f8b`.

sallycr:
## Simulated verification, all 3 boundary scenarios (per maintainer request)

Ran the workflow's exact range-selection logic against this repo's real history (and one isolated-worktree simulation for the patch case), using `git-cliff` 2.13.1 directly.

### 1. `v0.5.0` final (cut from `release/v0.4`'s branch point, through v0.5.0 changes + cherry-picks)
```
candidate=v0.4.0 -> NOT a real ancestor -> fallback
range = c4e8936..v0.5.0   (release/v0.4's fork point off main)
```
28 commits — matches the hand-curated real release notes for this tag.

### 2. `v0.5.0-rc.2` (same branch point, through the cherry-picks that existed at rc.2 time)
```
candidate=v0.4.0 -> NOT a real ancestor -> fallback
range = c4e8936..v0.5.0-rc.2
```
Same list as #1 minus `#1559` (not yet merged at rc.2 time), ending in the rc.2 version-bump commit — correctly reflects what existed at that point in time.

### 3. `v0.5.1` patch (simulated: one new fix commit on top of `release/v0.5`'s real tip)
```
candidate=v0.5.0 -> IS a real ancestor (same branch) -> used directly
range = v0.5.0..v0.5.1
```
```
### Bug Fixes
- **api:** Correct pagination off-by-one in ListPipelineRun
```
Correctly scoped to just the patch's own commit, not re-including anything since `v0.4.0`.

### Extra hardening found while re-testing scenario 2 retroactively (`6dd38938`)

With both `v0.5.0-rc.2` and `v0.5.0` now existing as real tags, I noticed the original "immediately-preceding tag" lookup (via inserting the current tag into a `sort -V`'d list) picked `v0.5.0` as `v0.5.0-rc.2`'s candidate — backwards, since GNU `sort -V` doesn't understand semver pre-release ordering (`v0.5.0` sorts lexically before `v0.5.0-rc.2`). It didn't produce a wrong final range here (the wrong candidate still correctly failed the ancestor check for an unrelated reason, so it fell through to the same correct fallback), but it's a latent bug for any future `workflow_dispatch` re-run against an old RC tag after its final has shipped. Replaced the lookup with one based on commit timestamps (`git log --format=%at`) instead of tag-string sorting — re-verified all three scenarios above still resolve identically, and `v0.5.0-rc.2`'s candidate now correctly resolves to `v0.4.0` rather than `v0.5.0`.

paulzim (approved):
## Review: #1561 (updated) — fix(release): exclude RC/nightly tags from changelog diff boundary, commit CHANGELOG.md via PR

**Verdict: Approve**

This supersedes my earlier "Request changes" review. That review found two concrete problems with the original version: (1) the fix used the wrong git-cliff option (`skip_tags` instead of `ignore_tags`), verified to have zero effect on the actual `--latest` output, and (2) even the right option wouldn't have worked in this repo because `v0.4.0` isn't an ancestor of `v0.5.0` (`release/v0.4` and `release/v0.5` diverged from `main` at different points, and the v0.4 merge-back never landed before v0.5 was cut). The updated version fixes both, and I re-verified everything empirically against a fresh clone with git-cliff 2.13.1 (matching CI's pin) rather than re-reading the diff.

### Fix 1: `cliff.toml` now uses `ignore_tags`, correctly

```toml
skip_tags = ".*\\.dev.*"
ignore_tags = ".*-rc\\..*|.*-nightly\\..*"
```

This is the semantically correct option — `ignore_tags` folds an ignored tag's commits forward into the next real release's section instead of dropping them. Confirmed both effects:
- RC-only hotfixes (#1553, #1559) still appear in the full changelog — now folded into the `[0.5.0]` section instead of living under their own `[0.5.0-rc.2]`/`[0.5.0-rc.1]` headers. No content loss (this fixes the regression I flagged in the original review, which was specific to `skip_tags` dropping them entirely).
- `.dev` tags remain on `skip_tags`, unchanged — correct, those really are noise (nightly dev builds), not folded content.

### Fix 2: new range-computation step correctly handles the non-ancestor topology

The new `Determine changelog range` step doesn't rely on git-cliff's own tag-adjacency walk at all — it computes an explicit range and passes it positionally instead of `--latest`. I traced the logic against the real `v0.5.0` tag line-by-line in a fresh clone:

1. Finds the most recent preceding **final** tag by commit timestamp (not `sort -V`, which the PR's own comment correctly notes would rank `v0.5.0` before `v0.5.0-rc.2` since it doesn't understand semver pre-release ordering) → correctly picks `v0.4.0`.
2. Checks `git merge-base --is-ancestor v0.4.0 v0.5.0` → false, as expected.
3. Falls back to the fork point of the *previous release branch* off `main`: `git merge-base origin/main origin/release/v0.4` → `c4e89368...`. I confirmed this commit really is a common ancestor of both `v0.4.0` and `v0.5.0`, unlike `v0.4.0` itself.
4. Final range: `c4e89368...^..v0.5.0`. Running `git-cliff` with this range and the new `cliff.toml` produces a clean 28-commit release-notes body — a large improvement over the original's byte-identical 2-entry output — and, critically, it **includes** the RC hotfixes (#1553, #1559) that would otherwise have been silently absent from `v0.5.0`'s release notes.

The patch-release case (e.g. `v0.5.0 → v0.5.1`) isn't affected by the fallback at all, since `v0.5.0` actually is an ancestor of `v0.5.1` — the first branch (step 2 above) handles it directly.

I also checked the guard for the "no earlier release branch" and "tag lives directly on `main`" cases (both fall back to empty range → `--latest`, via the `&&`/`||` expression in the `Generate release notes` step's `args:`) — both are sound.

### One thing I chased down and ruled out

I initially suspected the unmodified "Generate full changelog" step (still just `--output CHANGELOG.md`, no range restriction — this PR doesn't touch it) might have a related bug: running it against a plain `main` checkout produced a `## [Unreleased]` top section with no `[0.5.0]`/`[0.4.0]` headers at all. That turned out to be a testing artifact on my end — `main` has commits past the `v0.5.0` tag, so "Unreleased" is the *correct* label there. Re-running against the actual `v0.5.0` tag (matching what CI's checkout does for a tag-push trigger) gives the right header, `## [0.5.0]`. `v0.4.0` still doesn't get its own section in the full changelog either way (folded into `[0.5.0]`'s content) — confirmed this is pre-existing behavior, identical with and without this PR's changes, since it's the same non-ancestor topology issue and not something Bug 1's fix was scoped to address for the *full* changelog (only for the range-restricted release-notes step). Not a regression, not something this PR needs to fix.

### Bug 2 (commit CHANGELOG.md via PR) — unchanged from before, still sound

No changes here since my last review; base-branch detection still correctly resolves `v0.5.0 → release/v0.5`, `pull-requests: write` permission is correctly added.

### Minor note

`actionlint` isn't available in my environment so I couldn't independently verify the PR body's claim that it passes clean — I did confirm the workflow YAML parses correctly (`yaml.safe_load`). Not blocking.

Nice fix — this is a real, verified improvement over the current behavior, not just a plausible-looking one.

PR #1562, opened 2026-07-21
fix(rayjob): right-size submitter pod instead of cloning the head

dragod812:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`mapRay` no longer clones the RayCluster head pod template into the KubeRay `RayJob.Spec.SubmitterPodTemplate`. Instead it builds a minimal, right-sized submitter template (`buildSubmitterPodTemplate`): a single `ray-job-submitter` container using the head image with modest resources (500m/200Mi requests, 1/1Gi limits) and `RestartPolicy: Never`. It reuses only the pod-level settings the submitter needs to start in the same environment as its target cluster — `imagePullPolicy`, `imagePullSecrets`, and `serviceAccountName` — and deliberately omits the head's GPU request, `nodeSelector`, `tolerations`, and `affinity`. The stale comment about upgrading to kuberay v1.2.2 (already pinned in `go.mod`) is removed, and the mapper unit test is updated accordingly.

**Why?**

The submitter pod only runs `ray job submit` against an already-running cluster (we always select an existing cluster via `ClusterSelector`). Cloning the head spec made this trivial pod reserve head-sized CPU/memory — and because the head spec can carry GPU requests and GPU node scheduling, it could needlessly reserve scarce GPU compute as well.

KubeRay can default the submitter itself when the template is nil, and its default sizing matches what we set here (500m/200Mi requests, 1/1Gi limits, head image, no GPU — identical on [1.2.2](https://github.com/ray-project/kuberay/blob/v1.2.2/ray-operator/controllers/ray/common/job.go#L121-L145) and 1.6.1). But that nil-default drops the container `imagePullPolicy` (and pull secrets / service account): an image tagged `:latest` then falls back to `Always` and cannot use a preloaded or local-only image, which breaks environments that don't push the task image to a registry. Building the template ourselves keeps the right-sizing while preserving the pull/auth settings the submitter needs.

**How did you test it?**

- `bazel test //go/components/jobs/client/k8sengine:go_default_test` — updated to assert the submitter inherits the head image + pull/auth but is right-sized (500m/200Mi, no GPU).
- End-to-end on the OSS sandbox: ran a remote Uniflow pipeline and confirmed the submitter pod is `500m/200Mi` with no GPU, pulls the local-only image via `IfNotPresent` (no `ImagePullBackOff`), and the RayJob reaches `SUCCEEDED`.

**Potential risks**

Low. The change only affects the submitter pod created for new Ray jobs; existing clusters and running jobs are unaffected. The submitter still derives its image and pull/auth from the head, so it starts in the same environment as the cluster it targets.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Ray job submitter pods are now right-sized (modest CPU/memory, no GPU) instead of inheriting the head pod's resources, while still reusing the head image and image pull/service-account settings so they start correctly.

**Documentation Changes**

None.

### DIFF
--- a/go/components/jobs/client/k8sengine/ray.go
+++ b/go/components/jobs/client/k8sengine/ray.go
@@ -41,11 +41,7 @@ func (m Mapper) mapRay(rayJob *v2pb.RayJob, jobClusterObject runtime.Object, clu
 	if !ok {
 		return nil, fmt.Errorf("expected *v2pb.RayCluster, got %T", jobClusterObject)
 	}
-	pod := rayCluster.GetSpec().Head.GetPod()
-	submitterPod := k8sptr.Deref(pod, corev1.PodTemplateSpec{})
-	// Kubernetes Jobs require restartPolicy to be either "OnFailure" or "Never"
-	submitterPod.Spec.RestartPolicy = corev1.RestartPolicyNever
-
+	head := k8sptr.Deref(rayCluster.GetSpec().Head.GetPod(), corev1.PodTemplateSpec{})
 	kubeRayJob := &rayv1.RayJob{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       RayJobKind,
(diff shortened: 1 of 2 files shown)


badcount (approved):

go/components/jobs/client/k8sengine/ray.go
```diff
@@ -63,16 +59,74 @@ func (m Mapper) mapRay(rayJob *v2pb.RayJob, jobClusterObject runtime.Object, clu
...
+// drops pod-level settings the submitter needs to run in the same environment as the
+// cluster it targets — most importantly the container image pull policy (KubeRay's
+// default leaves it unset, so an image tagged :latest falls back to Always and cannot
+// use a preloaded/local-only image), plus image pull secrets and the service account.
+//
+// So instead of cloning the head pod (which would reserve head-sized, possibly GPU,
+// compute) or leaving the template nil, we construct a minimal submitter that reuses
+// the head image and its pull/auth settings while keeping modest resources and none of
+// the head's GPU or scheduling constraints.
+func buildSubmitterPodTemplate(head corev1.PodTemplateSpec) *corev1.PodTemplateSpec {
```
dragod812, line 96: fyi @badcount i will have to construct a default submitter pod template with specific image pull policy and secrets. It can't be same as the default kuberay one.

PR #1570, opened 2026-07-22
feat(native_transform): add foundation transform layers (PR B1)

sallycr:
## Summary

Establishes the `native_transform` PyTorch transform-layer package: a small set of shared leaf modules plus the first tier of stateless, elementwise transform layers. These layers form the base pattern (dict-of-tensors I/O, device-safe constants, TorchScript export) that the remaining transform layers build on.

~500 source LOC. Follow-up work adds structural, fitted-statistics, tokenizer-wrapper, spec, and DAG-engine layers on top of this foundation.

## What's included

- **`torch/constants.py`** — shared dtype-mapping dicts and default values (a dependency-free leaf module).
- **`torch/utils.py`** — dtype resolution (`resolve_torch_dtype`, `initialize_dtype`), layer-name generation (`generate_layer_name`, `to_snake_case`, `id_generator`), the type-native sentinel lookup, and the `format_inputs` / `format_outputs` pair that defines the package's `dict[str, torch.Tensor]` in/out contract.
- **`torch/base_layers.py`** — the abstract `TorchTransformBaseLayer` plus 10 concrete layers: `Concatenate`, `Stack`, `Cast`, `Constant`, `Divide`, `LogTransform`, `Subtract`, `Floor`, `Ceil`, `IdentityTransform`.
- Package exports updated in `torch/__init__.py`.

## Design notes

- **TorchScript/serve parity:** every layer is `torch.jit.script`-able so the exact transform runs at both train and serve time.
- **Device safety:** `Constant` uses `new_full` (not `torch.tensor(...).to(device)`) so a traced graph does not bake in a literal device.
- **Numerical safety:** `Divide` / `Subtract` upcast to float64; `Divide` substitutes an epsilon for zero denominators and forces `0/0 -> 0`.

## Testing

`python -m pytest michelangelo/lib/native_transform/ -v` → **112 passed, 1 skipped** (the skip is the pre-existing ONNX-export test, which needs `onnx`/`onnxruntime`).

New tests cover each layer's forward semantics and validation errors, all utility functions, and — for every layer — a `torch.jit.script` save/load round-trip asserting scripted output matches eager execution.

**Coverage** (`--cov=michelangelo.lib.native_transform --cov-report=term-missing`): **99.04% total** — `constants.py`, `utils.py`, `id_hash_tokenizer.py`, and `base_layers.py` all at **100%**. The only uncovered lines are the re-export statements in `torch/__init__.py`, which execute at import time before the coverage tracer starts measuring.

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/base_layers.py
+++ b/python/michelangelo/lib/native_transform/torch/base_layers.py
@@ -21,6 +21,7 @@ from michelangelo.lib.native_transform.torch.constants import DEFAULT_EPSILON
 from michelangelo.lib.native_transform.torch.utils import (
     format_inputs,
     format_outputs,
+    generate_layer_name,
     initialize_dtype,
 )
 
@@ -50,7 +51,9 @@ class TorchTransformBaseLayer(torch.nn.Module, abc.ABC):
         input_cols: Column names of the input tensors.
         output_cols: Column names of the output tensors.
         **kwargs: Additional options. ``name`` (str) sets the layer name, which
-            must be unique within a model; it defaults to ``"transform_layer"``.
+            must be unique within a model. When omitted, a unique name is
+            generated automatically from the layer's class name (e.g.
+            ``"stack_A1B2C3D4E5"``).
     """
 
     def __init__(self, input_cols: list[str], output_cols: list[str], **kwargs) -> None:
@@ -59,12 +62,16 @@ class TorchTransformBaseLayer(torch.nn.Module, abc.ABC):
         Args:
             input_cols: Column names of the input tensors.
             output_cols: Column names of the output tensors.
-            **kwargs: Additional options; ``name`` (str) sets the layer name.
+            **kwargs: Additional options; ``name`` (str) sets the layer name. When
+                omitted, a unique name is generated from the class name.
         """
         super().__init__()
         self.input_cols = input_cols
         self.output_cols = output_cols
-        self.name = kwargs.get("name", "transform_layer")
+        name = kwargs.get("name")
+        self.name = (
+            name if name is not None else generate_layer_name(self.__class__.__name__)
+        )
 
     @abc.abstractmethod
     def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
@@ -146,21 +153,27 @@ class Stack(TorchTransformBaseLayer):
     Args:
         input_cols: Column names of the input tensors.
         output_cols: Single-element list naming the stacked output column.
-        **kwargs: Additional options. ``dim`` (int) is the dimension along which
-            to stack (default ``-1``); ``name`` sets the layer name.
+        dim: The dimension along which to stack (default ``-1``).
+        **kwargs: Additional base-layer options (e.g. ``name``).
     """
 
-    def __init__(self, input_cols: list[str], output_cols: list[str], **kwargs) -> None:
+    def __init__(
+        self,
+        input_cols: list[str],
+        output_cols: list[str],
+        dim: int = -1,
+        **kwargs,
+    ) -> None:
         """Initialize the Stack layer.
 
         Args:
             input_cols: Column names of the input tensors.
             output_cols: Single-element list naming the stacked output column.
-            **kwargs: Additional options; ``dim`` (int, default ``-1``) selects
-                the new dimension, and ``name`` sets the layer name.
+            dim: The new dimension along which to stack (default ``-1``).
+            **kwargs: Additional base-layer options (e.g. ``name``).
         """
         super().__init__(input_cols, output_cols, **kwargs)
-        self.dim: int = kwargs.get("dim", -1)
+        self.dim: int = dim
 
     def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         """Stack the input columns along ``dim``.
@@ -185,12 +198,15 @@ class Cast(TorchTransformBaseLayer):
         input_cols: Column names of the input tensors.
         output_cols: Column names of the output tensors; must match the length
             of ``input_cols``.
-        dtype: Target dtype to cast to. Defaults to ``torch.int64`` when not a
-            recognized dtype or string alias.
+        dtype: Target dtype to cast to. May be a ``torch.dtype`` or a string
+            alias (e.g. ``"float32"`` or ``"torch.float32"``). Defaults to
+            ``torch.int64`` when ``None``. An unrecognized string alias raises
+            ``ValueError``.
         **kwargs: Additional base-layer options (e.g. ``name``).
 
     Raises:
-        ValueError: If ``input_cols`` and ``output_cols`` differ in length.
+        ValueError: If ``input_cols`` and ``output_cols`` differ in length, or if
+            ``dtype`` is a string that names no recognized dtype.
     """
 
     def __init__(
@@ -206,11 +222,13 @@ class Cast(TorchTransformBaseLayer):
             input_cols: Column names of the input tensors.
             output_cols: Column names of the output tensors; must match the
                 length of ``input_cols``.
-            dtype: Target dtype; defaults to ``torch.int64``.
+            dtype: Target dtype (``torch.dtype`` or string alias); defaults to
+                ``torch.int64`` when ``None``.
             **kwargs: Additional base-layer options (e.g. ``name``).
 
         Raises:
-            ValueError: If ``input_cols`` and ``output_cols`` differ in length.
+            ValueError: If ``input_cols`` and ``output_cols`` differ in length,
+                or if ``dtype`` is a string that names no recognized dtype.
         """
         if len(input_cols) != len(output_cols):
             raise ValueError(
@@ -325,10 +343,10 @@ class Divide(TorchTransformBaseLayer):
         output_cols: Column names of the quotient outputs.
         add_constant_to_divisor: Constant added to every denominator before
             division.
-        **kwargs: Additional options. ``eps`` (float) is the small value
-            substituted for a zero denominator (default
-            :data:`~michelangelo.lib.native_transform.torch.constants.DEFAULT_EPSILON`);
-            ``name`` sets the layer name.
+        eps: Small value substituted for a zero denominator to avoid division by
+            zero (default
+            :data:`~michelangelo.lib.native_transform.torch.constants.DEFAULT_EPSILON`).
+        **kwargs: Additional base-layer options (e.g. ``name``).
 
     Raises:
         ValueError: If ``input_cols`` is not even, or ``output_cols`` is not half
@@ -340,6 +358,7 @@ class Divide(TorchTransformBaseLayer):
         input_cols: list[str],
         output_cols: list[str],
         add_constant_to_divisor: float = 0.0,
+        eps: float = DEFAULT_EPSILON,
         **kwargs,
     ) -> None:
         """Initialize the Divide layer.
@@ -348,8 +367,9 @@ class Divide(TorchTransformBaseLayer):
             input_cols: Column names as ``(numerator, denominator)`` pairs.
             output_cols: Column names of the quotient outputs.
             add_constant_to_divisor: Constant added to every denominator.
-            **kwargs: Additional options; ``eps`` (float) overrides the zero-safe
-                epsilon and ``name`` sets the layer name.
+            eps: Small value substituted for a zero denominator (default
+                :data:`~michelangelo.lib.native_transform.torch.constants.DEFAULT_EPSILON`).
+            **kwargs: Additional base-layer options (e.g. ``name``).
 
         Raises:
             ValueError: If ``input_cols`` is not even, or ``output_cols`` is not
@@ -362,7 +382,7 @@ class Divide(TorchTransformBaseLayer):
                 "input columns."
             )
         self.add_constant_to_divisor = add_constant_to_divisor
-        self.eps = kwargs.get("eps", DEFAULT_EPSILON)
+        self.eps = eps
 
     def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         """Divide numerators by denominators, pairwise and zero-safe.
@@ -405,34 +425,39 @@ class LogTransform(TorchTransformBaseLayer):
         input_cols: Column names of the input tensors.
         output_cols: Column names of the output tensors; must match the length of
             ``input_cols``.
-        **kwargs: Additional options. ``add_constant`` (float) is added before
-            the logarithm to avoid ``log(0)`` (default ``1.0``); ``name`` sets
-            the layer name.
+        add_constant: Value added before the logarithm to avoid ``log(0)``
+            (default ``1.0``).
+        **kwargs: Additional base-layer options (e.g. ``name``).
 
     Raises:
         ValueError: If ``input_cols`` and ``output_cols`` differ in length.
     """
 
-    def __init__(self, input_cols: list[str], output_cols: list[str], **kwargs) -> None:
+    def __init__(
+        self,
+        input_cols: list[str],
+        output_cols: list[str],
+        add_constant: float = 1.0,
+        **kwargs,
+    ) -> None:
         """Initialize the LogTransform layer.
 
         Args:
             input_cols: Column names of the input tensors.
             output_cols: Column names of the output tensors; must match the
                 length of ``input_cols``.
-            **kwargs: Additional options; ``add_constant`` (float, default
-                ``1.0``) is added before the logarithm and ``name`` sets the
-                layer name.
+            add_constant: Value added before the logarithm (default ``1.0``).
+            **kwargs: Additional base-layer options (e.g. ``name``).
 
         Raises:
             ValueError: If ``input_cols`` and ``output_cols`` differ in length.
         """
-        super().__init__(input_cols, output_cols)
+        super().__init__(input_cols, output_cols, **kwargs)
         if len(input_cols) != len(output_cols):
             raise ValueError(
                 "Input columns and output columns must have the same length."
             )
-        self.add_constant = kwargs.get("add_constant", 1.0)
+        self.add_constant = add_constant
 
     def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         """Apply the log transform to each input column.
--- a/python/michelangelo/lib/native_transform/torch/base_layers.py
+++ b/python/michelangelo/lib/native_transform/torch/base_layers.py
@@ -1,0 +1,643 @@
+"""PyTorch native transform layers.
+
+TorchScript- and ONNX-exportable ``nn.Module`` transform layers that operate on a
+``dict[str, torch.Tensor]`` in/out contract so the exact same transform runs at
+train and serve time. Every layer subclasses :class:`TorchTransformBaseLayer` and
+uses :func:`~michelangelo.lib.native_transform.torch.utils.format_inputs` /
+:func:`~michelangelo.lib.native_transform.torch.utils.format_outputs` to map its
+declared input/output columns to and from a single stacked tensor.
+
+This module provides the foundation (stateless, elementwise) layers. Structural,
+fitted-statistics, and tokenizer layers are added in follow-up modules.
+"""
+
+from __future__ import annotations
+
+import abc
+
+import torch
+
+from michelangelo.lib.native_transform.torch.constants import DEFAULT_EPSILON
+from michelangelo.lib.native_transform.torch.utils import (
+    format_inputs,
+    format_outputs,
+    initialize_dtype,
+)
+
+__all__ = [
+    "Cast",
+    "Ceil",
+    "Concatenate",
+    "Constant",
+    "Divide",
+    "Floor",
+    "IdentityTransform",
+    "LogTransform",
+    "Stack",
+    "Subtract",
+    "TorchTransformBaseLayer",
+]
+
+
+class TorchTransformBaseLayer(torch.nn.Module, abc.ABC):
+    """Abstract base for native PyTorch transform layers.
+
+    All layers consume and produce ``dict[str, torch.Tensor]`` so they compose
+    into a single TorchScript-exportable graph. Subclasses select their inputs by
+    ``input_cols`` and write their results under ``output_cols``.
+
+    Args:
+        input_cols: Column names of the input tensors.
+        output_cols: Column names of the output tensors.
+        **kwargs: Additional options. ``name`` (str) sets the layer name, which
+            must be unique within a model; it defaults to ``"transform_layer"``.
+    """
+
+    def __init__(self, input_cols: list[str], output_cols: list[str], **kwargs) -> None:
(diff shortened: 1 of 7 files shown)


python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+
+        Args:
+            input_cols: Column names of the input tensors.
+            output_cols: Column names of the output tensors.
+            **kwargs: Additional options; ``name`` (str) sets the layer name.
+        """
+        super().__init__()
+        self.input_cols = input_cols
+        self.output_cols = output_cols
+        self.name = kwargs.get("name", "transform_layer")
```
sallycr: Two issues on this line:

1. `name` is a meaningful, documented option but only reachable via `**kwargs`, so it's invisible to autocomplete/type-checkers and typos are silently dropped. Consider an explicit `name: str | None = None` parameter. Same applies to `dim` (Stack), `eps` (Divide), `add_constant` (LogTransform).
2. The class docstring says `name` "must be unique within a model," but this default is a shared constant, so every layer built with defaults collides. `generate_layer_name()` already exists and is tested but isn't used here — either default to it (e.g. `generate_layer_name(type(self).__name__)`) or relax the docstring.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+        """Initialize the Stack layer.
+
+        Args:
+            input_cols: Column names of the input tensors.
+            output_cols: Single-element list naming the stacked output column.
+            **kwargs: Additional options; ``dim`` (int, default ``-1``) selects
+                the new dimension, and ``name`` sets the layer name.
+        """
+        super().__init__(input_cols, output_cols, **kwargs)
+        self.dim: int = kwargs.get("dim", -1)
```
sallycr: `dim` via `kwargs.get` means a misspelled kwarg is silently ignored: `Stack(..., dimension=1).dim` is `-1`, not `1` (verified). Promoting `dim: int = -1` to an explicit keyword arg makes it discoverable and lets a typo surface as a `TypeError`.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+
+class Cast(TorchTransformBaseLayer):
+    """Cast input tensors to a target dtype.
+
+    Args:
+        input_cols: Column names of the input tensors.
+        output_cols: Column names of the output tensors; must match the length
+            of ``input_cols``.
+        dtype: Target dtype to cast to. Defaults to ``torch.int64`` when not a
+            recognized dtype or string alias.
```
sallycr: This docstring says dtype "Defaults to `torch.int64` when not a recognized dtype or string alias," but that's not the behavior. An unrecognized string flows through `initialize_dtype` -> `None`, and `tensor.to(None)` is a no-op, so `Cast(..., dtype='flaot32')` leaves the tensor's dtype unchanged with no error (verified: int32 stays int32). A silently-skipped cast is a dangerous failure mode — recommend raising on an unresolvable dtype rather than falling through to `None`.

python/michelangelo/lib/native_transform/torch/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Returns:
+        The resolved ``torch.dtype``. When ``raw_dtype`` is an unrecognized
+        string alias, ``None`` is returned (the caller decides how to handle an
+        unresolved dtype).
+    """
+    if isinstance(raw_dtype, torch.dtype):
+        return raw_dtype
+    elif isinstance(raw_dtype, str):
+        return STRING_DATA_TYPE_TO_TORCH_TYPE_MAP.get(raw_dtype, None)
```
sallycr: `initialize_dtype` silently returns `None` for an unrecognized string, while the sibling `resolve_torch_dtype` (also public) *raises* and additionally accepts the `torch.`-prefixed forms this one rejects. So `Cast(dtype='torch.float32')` silently no-ops but `Cast(dtype='float32')` works — an inconsistency a user can't discover without reading source. Consider a single resolver with one documented set of accepted strings and one (raising) failure mode.

python/michelangelo/lib/native_transform/torch/constants.py
```diff
@@ -0,0 +1,82 @@
...
+TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP: dict[torch.dtype | str, str] = {
+    torch.int32: "torch.int32",
+    torch.int64: "torch.int64",
+    torch.float32: "torch.float32",
+    torch.float64: "torch.float64",
+    torch.bool: "torch.bool",
+    "string": "string",
+}
+
+NUMERICAL_TRANSFORM = "numerical_transform"
```
sallycr, line 47: This block (and several defaults below: `DEFAULT_VECTOR_TYPE_DICT`, the gibberish-score constants, the `*_TRANSFORM` name strings, `TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP`, `DEFAULT_NUM_PARTITIONS`, `DEFAULT_PARQUET_BLOCK_SIZE`, etc.) has zero references anywhere in the tree. For a PR scoped as "foundation layers," shipping unused constants makes it hard to tell what's actually load-bearing. Suggest trimming to what these layers use and introducing the rest alongside the layers that consume them.

sallycr (commented):
## DX / API-ergonomics review (external contributor perspective)

I looked at this cold, as someone who might adopt `native_transform` from PyPI. Overall the craft is high: docstrings are thorough Google-style (Args/Returns/Raises everywhere), the dict-in/dict-out contract is coherent, and the TorchScript save-load round-trip tests are exactly what I'd want to see for a serve-parity claim. The PR description is also genuinely self-contained. Below are the things that would trip me up when actually writing code against these classes.

### 1. Behavior-affecting parameters are hidden in `**kwargs`
The interesting knobs are not in the signatures: `name` (base), `dim` (Stack), `eps` (Divide), `add_constant` (LogTransform). Because they go through `kwargs.get(...)`, they're invisible to autocomplete/type-checkers, and — worse — a typo is silently swallowed. I confirmed `Stack(input_cols=['a','b'], output_cols=['out'], dimension=1).dim` is `-1` (my `dimension=` typo was dropped, no error). For a first-time user this is a debugging trap. These are all fixed, well-understood options; making them explicit keyword args with defaults would fix discoverability, type safety, and typo-detection in one move. It would also make the surface consistent — `dtype` already *is* an explicit arg on Concatenate/Cast/Constant, so it's odd that `dim`/`eps`/`add_constant` are not.

### 2. `Cast` silently no-ops on an unrecognized/typo'd dtype (and the docstring is wrong here)
The `Cast` docstring says dtype "Defaults to `torch.int64` when not a recognized dtype or string alias." That isn't what happens. `initialize_dtype` returns `None` for an unrecognized *string*, and `tensor.to(None)` is a no-op, so `Cast(..., dtype='flaot32')` leaves an int32 tensor as int32 with no error and no int64 fallback (verified). A silent no-op cast is about the worst failure mode for a data-transform layer. I'd raise on an unresolvable dtype instead of returning `None`.

### 3. Two dtype resolvers with different accepted inputs and different failure modes
`resolve_torch_dtype` accepts the `torch.`-prefixed names (`'torch.float32'`) and *raises* on bad input; `initialize_dtype` (the one actually wired into the layers) accepts only the bare aliases (`'float32'`) and *silently returns None*. Net effect: `Cast(dtype='torch.float32')` silently does nothing, while `Cast(dtype='float32')` works — confirmed. As an external user I can't tell which string forms are legal without reading source, and the public `resolve_torch_dtype` accepts a form the layers reject. Consider having the layers use one resolver with one documented set of accepted strings and one (raising) failure mode.

### 4. Default layer name contradicts the documented uniqueness rule
The base docstring says `name` "must be unique within a model," but the default is the constant `'transform_layer'`, so every layer built with defaults collides. Meanwhile `generate_layer_name()` (which exists, is public, and is tested) would produce unique names but is never wired in. Either default to `generate_layer_name(type(self).__name__)`, or soften the docstring — right now the code and the contract disagree.

### 5. Public API shipped without consumers
`resolve_torch_dtype`, `sentinel_for_torch_dtype`, and `generate_layer_name` are all in `__all__` and tested, but nothing in this PR calls them. Similarly `torch/constants.py` ships a large block of constants with zero references anywhere in the tree (e.g. `SHAPE`, `DEFAULT_NUM_PARTITIONS`, `DEFAULT_PARQUET_BLOCK_SIZE`, `DEFAULT_NUM_SAMPLES_TO_EXPLAIN`, the `*_TRANSFORM` name strings, `DEFAULT_VECTOR_TYPE_DICT`, the gibberish-score defaults, `TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP`). For a PR scoped as "foundation layers," this speculative surface is confusing — a reader can't tell what's load-bearing now vs. staged for later work. I'd trim to what these layers use and add the rest alongside the layers that need them.

### 6. No end-to-end usage example
The PR says these layers "form the base pattern" the rest build on, but there's no runnable example of that pattern anywhere — not in the module docstrings, not a README, not a doctest. One tiny snippet (construct a layer, call it with a `{col: tensor}` dict, show the returned dict, maybe a `torch.jit.script` line) would get a new user from zero to working far faster than reading ten class docstrings.

### 7. The interleaved even/odd pairing for Divide/Subtract is easy to get wrong
Reading `Divide(input_cols=['a','b','c','d'], output_cols=['x','y'])`, it's not obvious that `a/b` and `c/d` are the pairs. It's documented, but getting the order wrong yields silently-wrong numbers rather than an error. A two-line example in each docstring would de-risk this a lot.

### Minor
- `format_inputs` uses `torch.stack`, which requires all selected columns to share a shape. So multi-column `Cast`/`Floor`/`Ceil`/`LogTransform` require equal-shaped inputs, whereas `Concatenate` (using `torch.cat`) tolerates differing last dims. That difference in constraint between sibling layers isn't documented and would surprise me.
- Naming the subpackage `torch` (`native_transform.torch`) shadows the library name conceptually; absolute imports keep it working, but `from ...native_transform.torch import Cast` reads oddly. Fine if a non-torch backend is really coming, but worth a one-line note on the intent.

None of these block the concept — the layer set and test rigor are solid. Items 1-4 are the ones I'd want addressed before I'd trust this in my own pipeline, since they're the silent-wrong-behavior cases.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+            output_cols: Column names of the output tensors; must match the
+                length of ``input_cols``.
+            **kwargs: Additional options; ``add_constant`` (float, default
+                ``1.0``) is added before the logarithm and ``name`` sets the
+                layer name.
+
+        Raises:
+            ValueError: If ``input_cols`` and ``output_cols`` differ in length.
+        """
+        super().__init__(input_cols, output_cols)
```
sallycr: **Bug:** this is the only layer that doesn't forward `**kwargs` to the base class. As written, `LogTransform(input_cols=['a'], output_cols=['b'], name='my_log').name` returns `'transform_layer'`, not `'my_log'` — the `name` kwarg (and any future base kwarg) is silently dropped. Should be `super().__init__(input_cols, output_cols, **kwargs)` like every sibling layer. Worth a regression test asserting `name` is honored on `LogTransform`.

sallycr (commented):
## Review — foundation transform layers (PR B1)

Thanks for this. The package is well-structured, the docstrings are thorough, and I verified the test claims locally: `pytest michelangelo/lib/native_transform/` gives **112 passed, 1 skipped** and coverage is **99.04%** with `base_layers.py`, `constants.py`, and `utils.py` all at 100%. The TorchScript round-trip parametrization across every layer is a nice touch and is exactly the kind of test this code needs.

That said, I found one real bug and several correctness/API concerns I'd want addressed before merge. Details are in inline comments; summarizing here:

**Bug**
- `LogTransform.__init__` calls `super().__init__(input_cols, output_cols)` without `**kwargs`, so it silently drops the `name` (and any other base) kwarg — unlike every other layer. `LogTransform(..., name="x").name` returns `"transform_layer"`. Confirmed by running it. No test covers `name=` on `LogTransform`.

**Correctness / API consistency**
- Two overlapping dtype resolvers with *different* accepted inputs: `resolve_torch_dtype` accepts both `"torch.float32"` and `"float32"`, but the layers use `initialize_dtype`, which only accepts the bare aliases. `Cast(dtype="torch.float32")` therefore resolves to `None` and silently does **not** cast (`.to(None)` is a no-op). Confirmed by running it.
- `Cast`'s docstring says the dtype "Defaults to `torch.int64` when not a recognized dtype or string alias." That's inaccurate: an unrecognized *string* resolves to `None` (no cast), not `int64` — the `int64` default only applies to non-string inputs.
- `resolve_torch_dtype` / `sentinel_for_torch_dtype` are exported and tested but unused anywhere in the package. Likewise ~35 of the ~40 names in `constants.py` are unused by anything in this PR. The 100% coverage on `constants.py` is just module-level assignment execution and doesn't indicate any of it is exercised.

**Minor**
- `to_snake_case("")` raises `IndexError` (unguarded `insecure[0]`).
- `LogTransform` clamps output to `[1.0, 1e20]` — the `min=1.0` floor is unusual and unexplained (see inline).
- Inconsistent naming: `LogTransform`/`IdentityTransform` carry a `Transform` suffix while `Cast`/`Divide`/`Subtract`/`Floor`/`Ceil`/`Concatenate`/`Stack`/`Constant` don't.
- Layers accept `**kwargs` and silently ignore unknown keys (`epsilon=`, `nam=` typos pass unnoticed).

The scope (a foundation layer package that stands on its own) is coherent and the follow-up plan is clearly stated. Not approving pending the `LogTransform` kwargs bug and the dtype-resolution inconsistency.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+
+        Raises:
+            ValueError: If ``input_cols`` and ``output_cols`` differ in length.
+        """
+        if len(input_cols) != len(output_cols):
+            raise ValueError(
+                "Input columns and output columns must have the same length."
+            )
+        super().__init__(input_cols, output_cols, **kwargs)
+        self.dtype = initialize_dtype(dtype, torch.int64)
```
sallycr, line 238: `initialize_dtype` only recognizes the bare aliases in `STRING_DATA_TYPE_TO_TORCH_TYPE_MAP` (`"float32"`, `"int64"`, ...). So `Cast(dtype="torch.float32")` resolves to `None` and the forward pass calls `.to(None)`, which is a silent no-op — no cast happens. Meanwhile `resolve_torch_dtype` (exported from the same utils module) *does* accept `"torch.float32"`. Two dtype resolvers accepting different string sets is a footgun. Also, the class docstring claims the dtype "Defaults to `torch.int64` when not a recognized dtype or string alias" — that's only true for non-string inputs; an unrecognized string yields `None`, not `int64`. Consider routing all layers through a single resolver and raising on unrecognized strings rather than silently no-op'ing.

python/michelangelo/lib/native_transform/torch/utils.py
```diff
@@ -0,0 +1,188 @@
...
+        ValueError: If ``dtype_spec`` cannot be resolved.
+    """
+    if isinstance(dtype_spec, torch.dtype):
+        return dtype_spec
+    if isinstance(dtype_spec, str):
+        if dtype_spec in TORCH_DTYPE_CLASS_NAME_TO_TORCH_TYPE_MAP:
+            return TORCH_DTYPE_CLASS_NAME_TO_TORCH_TYPE_MAP[dtype_spec]
+        if dtype_spec in STRING_DATA_TYPE_TO_TORCH_TYPE_MAP:
+            return STRING_DATA_TYPE_TO_TORCH_TYPE_MAP[dtype_spec]
+    raise ValueError(f"Unsupported dtype specification: {dtype_spec}")
```
sallycr, line 130: `resolve_torch_dtype` and `initialize_dtype` overlap heavily but disagree: `resolve_torch_dtype` accepts `"torch.float32"` *and* `"float32"` and raises on the unknown; `initialize_dtype` accepts only `"float32"` and returns `None` on the unknown. The concrete layers all use `initialize_dtype`, so the more capable `resolve_torch_dtype` is currently unused (see its definition below) and tested in isolation. Recommend consolidating to one function so callers get consistent behavior.

python/michelangelo/lib/native_transform/torch/utils.py
```diff
@@ -0,0 +1,188 @@
...
+        name: The name to convert (e.g. a class name in ``CamelCase``).
+
+    Returns:
+        The snake_case form of ``name``.
+    """
+    intermediate = re.sub("(.)([A-Z][a-z0-9]+)", r"\1_\2", name)
+    insecure = re.sub("([a-z])([A-Z])", r"\1_\2", intermediate).lower()
+    # A leading underscore (from a private class name) is not a valid scope
+    # name, so prefix it with "private".
+    if insecure[0] != "_":
```
sallycr: `insecure[0]` raises `IndexError` for an empty string, so `to_snake_case("")` (and `generate_layer_name("")`) crash. Layer names are class-derived so this is unlikely in practice, but a guard (`if insecure and insecure[0] != "_"`) or an explicit precondition would be safer for a public helper.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -0,0 +1,643 @@
...
+
+        Args:
+            inputs: Mapping from column name to tensor.
+
+        Returns:
+            A mapping from each output column to its transformed, clamped tensor.
+        """
+        stacked_inputs = format_inputs(self.input_cols, inputs)
+        stacked_outputs = torch.log(stacked_inputs + self.add_constant)
+        stacked_outputs = torch.clamp(stacked_outputs, min=1.0, max=1e20)
```
sallycr, line 473: The `min=1.0` clamp is worth a comment explaining the rationale. With the default `add_constant=1.0`, any input where `x + add_constant <= e` (~`x <= 1.72`) maps to exactly `1.0`, so the transform is non-monotonic / loses resolution across the low end of the range. If that's intentional (e.g. matching a specific serving contract) a one-line note would save the next reader a double-take; if `0.0` was meant, this is a subtle bug.

python/michelangelo/lib/native_transform/torch/constants.py
sallycr, line 47: Most of this file is unused by anything in the PR. I grepped the package: only `DEFAULT_EPSILON`, `STRING_DATA_TYPE_TO_TORCH_TYPE_MAP`, and `TORCH_DTYPE_CLASS_NAME_TO_TORCH_TYPE_MAP` are referenced anywhere; the ~30 transform-name strings, `DEFAULT_VECTOR_TYPE_DICT`, the gibberish-score constants (`DEFAULT_NAME_GIBBERISH_SCORE = 0.113792509289` is a particularly opaque magic number), `DEFAULT_PARQUET_BLOCK_SIZE`, `DEFAULT_NUM_PARTITIONS`, etc. have no consumer in this foundation layer. For a package that's meant to be the clean base others build on, I'd prefer these land alongside the code that uses them rather than up front — as-is they're dead weight, and the 100% coverage number on this file only reflects import-time assignment, not use.

sallycr:
## Review summary

Two independent reviews have gone through this PR (posted as separate review threads above) and converged on the same core issue: several silent-failure/dtype-handling gaps in `base_layers.py`/`utils.py`. All findings below have been reproduced by running the code directly (not just read), so these are confirmed bugs, not stylistic preferences:

1. **Hidden params in `**kwargs`, no validation** — `Stack`'s `dim`, `Divide`'s `eps`, `LogTransform`'s `add_constant` aren't explicit constructor params, so a typo (e.g. `dimension=1` instead of `dim=1`) is silently swallowed with no error.
2. **`Cast` silently no-ops on an unrecognized dtype string**, and its docstring's claimed `torch.int64` fallback doesn't apply to string inputs.
3. **`resolve_torch_dtype` and `initialize_dtype` disagree on identical input** — `Cast(dtype="torch.float32")` silently no-ops while `Cast(dtype="float32")` works, despite referring to the same dtype.
4. **Default layer name collides** — every layer defaults to the literal `"transform_layer"`, contradicting the base class's own docstring that names "must be unique in the model." `generate_layer_name()` exists and is tested but isn't wired in as the default.
5. **`LogTransform.__init__` drops `**kwargs`** — doesn't forward to `super().__init__()`, so `name=` is silently ignored for this one layer only.
6. **`to_snake_case("")` raises `IndexError`** — unguarded index access on empty input.

Fixes for all 6 are in progress. Tagging for review once pushed.

/cc @dkurra — flagging for your review given your familiarity with the trainer/lightning trainer snapshot work; this PR seeds the sibling `native_transform` package with a similar pattern (pluggable transform layers over a shared base class).

sallycr:
## Addressed review feedback (commit d1b29a52)

Fixed 6 verified bugs surfaced by review, each with a regression test.

1. **Hidden params in `**kwargs`** — `Stack.dim`, `Divide.eps`, and `LogTransform.add_constant` are now explicit, type-hinted, documented constructor parameters instead of `kwargs.get(...)` reads, so a typo like `Stack(dimension=1)` can no longer silently leave the default in place. (`name` still flows through `**kwargs` via the base class, as before.)
2. **`Cast` silent no-op on bad dtype string + wrong docstring** — `initialize_dtype` now raises `ValueError` on an unrecognized string dtype instead of returning `None`; the `Cast` docstring now accurately describes the `torch.int64`-on-`None` default and the raise-on-bad-string behavior.
3. **`resolve_torch_dtype` / `initialize_dtype` disagreement** — `initialize_dtype` now resolves string inputs through `resolve_torch_dtype`, so `"float32"` and `"torch.float32"` behave identically and consistently with `resolve_torch_dtype`.
4. **Colliding default layer name** — the base class default `name` is now `generate_layer_name(self.__class__.__name__)`, so default-constructed layers get unique names, matching the base docstring's uniqueness contract.
5. **`LogTransform` dropped `**kwargs`** — `LogTransform.__init__` now forwards `**kwargs` to `super().__init__()`, so `name=` is honored (regression test added).
6. **`to_snake_case("")` `IndexError`** — guarded against empty input; returns `""` unchanged.

### Verification
- `ruff format` + `ruff check`: clean on all touched files.
- `pytest michelangelo/lib/native_transform/ --cov`: **122 passed, 1 skipped**; **100%** coverage on `base_layers.py` and `utils.py` (99.04% total).

Scope limited to these 6 fixes + regression tests; no unrelated refactors.

python/michelangelo/lib/native_transform/torch/constants.py
sallycr: it will be used by upcoming PRs.

sallycr: it will be used by upcoming PRs.

python/michelangelo/lib/native_transform/torch/utils.py
sallycr: not a concern for initialize.

kechen2013 (approved):

PR #1571, opened 2026-07-22
feat(trainer): add pluggable ExperimentStore for auto-resume (PR 4)

sallycr:
## Summary

Adds an **opt-in auto-resume** capability to `LightningTrainer` through a pluggable `ExperimentStore` seam, plus a filesystem-backed default. A re-run with the same `RunConfig(name=..., storage_path=...)` can pick up where a prior run left off, instead of starting from scratch.

## What's new

- **`ExperimentStore`** — a `@runtime_checkable` `Protocol` (sibling to the existing `TrainingObserver`) with two keyword-only methods:
  - `track(*, storage_path, run_name, experiment_path) -> None` — records where this run's Ray Train experiment directory lives (runs on worker rank 0).
  - `locate_resumable(*, storage_path, run_name) -> str | None` — returns a *candidate* directory to resume from (runs on the driver).
  - The stable cross-run identity is `(storage_path, run_name)`, taken from the driver's `RunConfig` and passed verbatim to both sides, so worker and driver always agree on the key regardless of how Ray names the on-disk experiment directory.

- **`FsspecExperimentStore`** — the default implementation. It persists a small JSON marker at `{storage_path}/.michelangelo_resume/{run_name}.json` using whatever fsspec scheme `storage_path` selects (local, `s3://`, `gs://`, ...). The marker path is derived purely from the stable identity, so it does not depend on Ray's (possibly timestamp-suffixed) experiment directory name.
  - **Never raises:** `track` logs-and-swallows on a write failure (a failed marker write must not fail an otherwise-successful run); `locate_resumable` returns `None` on a missing/corrupt/unreadable marker.
  - Holds only a plain `storage_options` dict, so it is picklable for transport to Ray workers.

## Behavior

- **Opt-in, default `None`.** No tracking and no resume happen unless you set `experiment_store` *and* your `RunConfig` carries both a `name` and a `storage_path`. This mirrors the `training_observer` no-op-by-default precedent and avoids surprising an existing user with implicit filesystem writes or a silent "resume instead of start fresh" behavior change.
- **The store never validates checkpoints.** `train()` takes the store's candidate path and gates it through `ray.train.torch.TorchTrainer.can_restore()` before enabling restoration. If the candidate has no restorable checkpoint (e.g. a first run that died before its first checkpoint), the run simply starts fresh. Staleness is therefore never an error — resume is an explicit opt-in.

## Usage

```python
from michelangelo.lib.trainer.torch.pytorch_lightning import (
    LightningTrainer,
    LightningTrainerParam,
    FsspecExperimentStore,
)

trainer = LightningTrainer(
    trainer_param=LightningTrainerParam(
        create_model_fn=my_model_factory,
        create_model_fn_kwargs={},
        train_data=train_ds,
        val_data=val_ds,
        experiment_store=FsspecExperimentStore(),
    ),
    run_config=ray.train.RunConfig(name="my_run", storage_path="s3://bucket/runs"),
)
result = trainer.train()  # a later re-run with the same name/storage_path resumes
```

## Implementation notes

- `LightningTrainer.__init__` pops `experiment_store` before `asdict()` (a Protocol instance can't survive dataclass recursion) and re-injects the original object plus the `(storage_path, run_name)` identity into the worker config — mirroring the existing `training_observer` handling.
- `train()` sets Ray's internal `BaseTrainer._restore_path` / `_restore_storage_filesystem` to enable an in-place resumable run. This is a deliberate, documented private-Ray coupling; the public alternative (`TorchTrainer.restore`) rebuilds a fresh trainer, which is awkward from inside an instance's `train()`.
- The worker-side `track` call derives the experiment path from `ray.train.get_context().get_storage()` and is guarded to rank 0.

## New public exports

`ExperimentStore`, `FsspecExperimentStore`.

## Tests & coverage

Full trainer suite: **222 passed**. New code (`experiment_store.py`, `schema.py`, `lightning_trainer.py`) at **100%** line coverage; **94%** overall for `michelangelo.lib.trainer`. `ruff format` + `ruff check` clean on all touched files.

Covered:
- `ExperimentStore` protocol conformance (accept/reject).
- `FsspecExperimentStore`: marker-path derivation, track→locate round-trip on `tmp_path`, run-name isolation, overwrite, missing/corrupt/`experiment_path`-less markers → `None`, best-effort swallow on filesystem errors in both methods, `storage_options` forwarding, picklability.
- `LightningTrainer.__init__`: pop/re-inject + identity injection (with and without a `RunConfig`).
- Auto-resume gating: resumes when `can_restore` is `True`, no resume when `False`/no candidate/store raises, and skips cleanly when the store, `run_config`, `storage_path`, or `name` is missing; `train()` invokes resume before `fit()`.
- Worker `track`: rank-0 guard, derived experiment path, identity/store skips, and swallowed `track` failure.

### DIFF
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
@@ -1,0 +1,123 @@
+"""Default :class:`~schema.ExperimentStore` implementation for auto-resume.
+
+Public module: users import and construct :class:`FsspecExperimentStore` and
+pass it as ``LightningTrainerParam.experiment_store`` to opt into auto-resume.
+See :class:`michelangelo.lib.trainer.torch.pytorch_lightning.schema.ExperimentStore`
+for the seam it implements.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from datetime import datetime, timezone
+
+from fsspec.core import url_to_fs
+
+_logger = logging.getLogger(__name__)
+
+
+class FsspecExperimentStore:
+    """Filesystem-backed :class:`~schema.ExperimentStore` using an fsspec marker file.
+
+    Persists a small JSON marker at a deterministic path so a re-run with the
+    same ``RunConfig(storage_path=..., name=...)`` can locate and resume the
+    previous run's experiment directory. Works with any fsspec scheme the
+    ``storage_path`` selects (local, ``s3://``, ``gs://``, ...).
+
+    The marker lives at ``{storage_path}/.michelangelo_resume/{run_name}.json``
+    rather than inside the Ray experiment directory, because that path is
+    derivable purely from the stable ``(storage_path, run_name)`` identity and
+    does not depend on Ray's (possibly timestamp-suffixed) experiment directory
+    name. It inherits the same permissions and lifecycle as the run data it sits
+    beside.
+
+    Neither method raises: :meth:`track` logs and swallows any write failure,
+    and :meth:`locate_resumable` returns ``None`` on a missing, corrupt, or
+    unreadable marker. Staleness is not treated as an error — the trainer runs
+    the recorded path through ``TorchTrainer.can_restore()`` and starts fresh if
+    it does not point at a restorable checkpoint.
+
+    Attributes:
+        _MARKER_DIR: Subdirectory (under ``storage_path``) holding marker files.
+        _SCHEMA_VERSION: Marker payload schema version, for forward evolution.
+    """
+
+    _MARKER_DIR = ".michelangelo_resume"
+    _SCHEMA_VERSION = 1
+
+    def __init__(self, storage_options: dict | None = None) -> None:
+        """Initialize the store.
+
+        Args:
+            storage_options: Optional keyword arguments forwarded to
+                ``fsspec.core.url_to_fs`` (e.g. credentials or endpoint
+                overrides for the backing filesystem). Kept as a plain dict so
+                the store remains picklable for transport to Ray workers.
+        """
+        self._storage_options = storage_options or {}
+
+    def _marker_path(self, storage_path: str, run_name: str) -> str:
+        """Return the deterministic marker path for ``(storage_path, run_name)``."""
+        return f"{storage_path.rstrip('/')}/{self._MARKER_DIR}/{run_name}.json"
+
+    def track(self, *, storage_path: str, run_name: str, experiment_path: str) -> None:
+        """Write the marker recording this run's experiment directory.
+
+        Best-effort: any failure is logged and swallowed so a failed marker
+        write can never fail an otherwise-successful training run.
+
+        Args:
+            storage_path: The driver's ``RunConfig.storage_path``.
+            run_name: The driver's ``RunConfig.name``.
+            experiment_path: Absolute path of this run's Ray Train experiment
+                directory.
+
+        Returns:
+            ``None``.
+        """
+        try:
+            fs, path = url_to_fs(
+                self._marker_path(storage_path, run_name), **self._storage_options
+            )
+            payload = json.dumps(
+                {
+                    "schema_version": self._SCHEMA_VERSION,
+                    "run_name": run_name,
+                    "experiment_path": experiment_path,
+                    "written_at": datetime.now(timezone.utc).isoformat(),
+                }
+            )
+            fs.makedirs(path.rsplit("/", 1)[0], exist_ok=True)
+            with fs.open(path, "w") as f:
+                f.write(payload)
+        except Exception:  # best-effort: never fail training
+            _logger.warning("FsspecExperimentStore.track failed", exc_info=True)
+
+    def locate_resumable(self, *, storage_path: str, run_name: str) -> str | None:
+        """Read the marker and return the recorded experiment path, or ``None``.
+
+        Returns ``None`` (never raises) when the marker is missing, corrupt, or
+        unreadable, or when it carries no ``experiment_path``.
+
+        Args:
+            storage_path: The driver's ``RunConfig.storage_path``.
+            run_name: The driver's ``RunConfig.name``.
+
+        Returns:
+            The recorded candidate experiment directory, or ``None``.
+        """
+        try:
+            fs, path = url_to_fs(
+                self._marker_path(storage_path, run_name), **self._storage_options
+            )
+            if not fs.exists(path):
+                return None
+            with fs.open(path, "r") as f:
+                data = json.loads(f.read())
+            return data.get("experiment_path") or None
+        except Exception:  # missing / corrupt / unreadable -> nothing to resume
+            _logger.warning(
+                "FsspecExperimentStore.locate_resumable failed", exc_info=True
+            )
+            return None
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
@@ -10,11 +10,16 @@ from __future__ import annotations
 
 import json
 import logging
+import uuid
 from datetime import datetime, timezone
 
 from fsspec.core import url_to_fs
 
 _logger = logging.getLogger(__name__)
+
+
+class _InvalidRunNameError(ValueError):
+    """Raised internally when a ``run_name`` is unsafe for use in a marker path."""
 
 
 class FsspecExperimentStore:
@@ -34,9 +39,9 @@ class FsspecExperimentStore:
 
     Neither method raises: :meth:`track` logs and swallows any write failure,
     and :meth:`locate_resumable` returns ``None`` on a missing, corrupt, or
-    unreadable marker. Staleness is not treated as an error — the trainer runs
-    the recorded path through ``TorchTrainer.can_restore()`` and starts fresh if
-    it does not point at a restorable checkpoint.
+    unreadable marker. Staleness is not treated as an error — the trainer only
+    seeds a resume when the recorded directory still holds a restorable Ray
+    Train checkpoint, and otherwise starts fresh.
 
     Attributes:
         _MARKER_DIR: Subdirectory (under ``storage_path``) holding marker files.
@@ -58,28 +63,54 @@ class FsspecExperimentStore:
         self._storage_options = storage_options or {}
 
     def _marker_path(self, storage_path: str, run_name: str) -> str:
-        """Return the deterministic marker path for ``(storage_path, run_name)``."""
+        """Return the deterministic marker path for ``(storage_path, run_name)``.
+
+        Args:
+            storage_path: The storage root under which markers are kept.
+            run_name: The stable run identity. Used verbatim as the marker
+                filename stem, so it must not contain path separators.
+
+        Returns:
+            The marker path ``{storage_path}/.michelangelo_resume/{run_name}.json``.
+
+        Raises:
+            _InvalidRunNameError: If ``run_name`` is empty or contains a path
+                separator or ``..`` component that could escape the marker
+                directory.
+        """
+        if (
+            not run_name
+            or "/" in run_name
+            or "\\" in run_name
+            or run_name in (".", "..")
+        ):
+            raise _InvalidRunNameError(
+                f"run_name must be a single path-safe component, got {run_name!r}"
+            )
         return f"{storage_path.rstrip('/')}/{self._MARKER_DIR}/{run_name}.json"
 
     def track(self, *, storage_path: str, run_name: str, experiment_path: str) -> None:
         """Write the marker recording this run's experiment directory.
 
         Best-effort: any failure is logged and swallowed so a failed marker
-        write can never fail an otherwise-successful training run.
+        write can never fail an otherwise-successful training run. The payload
+        is written to a temporary sibling file and atomically renamed into
+        place, so a crash mid-write can never leave a partial/corrupt marker
+        that a later :meth:`locate_resumable` would read.
 
         Args:
             storage_path: The driver's ``RunConfig.storage_path``.
             run_name: The driver's ``RunConfig.name``.
-            experiment_path: Absolute path of this run's Ray Train experiment
-                directory.
+            experiment_path: Absolute path (scheme-qualified for remote
+                filesystems, e.g. ``s3://bucket/runs/my_run``) of this run's Ray
+                Train experiment directory.
 
         Returns:
             ``None``.
         """
         try:
-            fs, path = url_to_fs(
-                self._marker_path(storage_path, run_name), **self._storage_options
-            )
+            marker = self._marker_path(storage_path, run_name)
+            fs, path = url_to_fs(marker, **self._storage_options)
             payload = json.dumps(
                 {
                     "schema_version": self._SCHEMA_VERSION,
@@ -88,9 +119,14 @@ class FsspecExperimentStore:
                     "written_at": datetime.now(timezone.utc).isoformat(),
                 }
             )
-            fs.makedirs(path.rsplit("/", 1)[0], exist_ok=True)
-            with fs.open(path, "w") as f:
+            parent = path.rsplit("/", 1)[0]
+            fs.makedirs(parent, exist_ok=True)
+            # Write to a unique temp sibling then atomically rename into place so
+            # a crash mid-write cannot leave a partial marker behind.
+            tmp_path = f"{path}.{uuid.uuid4().hex}.tmp"
+            with fs.open(tmp_path, "w") as f:
                 f.write(payload)
+            fs.mv(tmp_path, path)
         except Exception:  # best-effort: never fail training
             _logger.warning("FsspecExperimentStore.track failed", exc_info=True)
 
@@ -98,7 +134,10 @@ class FsspecExperimentStore:
         """Read the marker and return the recorded experiment path, or ``None``.
 
         Returns ``None`` (never raises) when the marker is missing, corrupt, or
-        unreadable, or when it carries no ``experiment_path``.
+        unreadable, or when it carries no ``experiment_path``. A missing or
+        corrupt marker is the normal "nothing to resume yet" case and is logged
+        at ``DEBUG``; only genuinely unexpected filesystem errors are logged at
+        ``WARNING``.
 
         Args:
             storage_path: The driver's ``RunConfig.storage_path``.
@@ -116,7 +155,12 @@ class FsspecExperimentStore:
             with fs.open(path, "r") as f:
                 data = json.loads(f.read())
             return data.get("experiment_path") or None
-        except Exception:  # missing / corrupt / unreadable -> nothing to resume
+        except (json.JSONDecodeError, _InvalidRunNameError) as exc:
+            # A corrupt marker or an unusable run_name means "nothing to
+            # resume" — an expected, benign outcome, not an error.
+            _logger.debug("No resumable marker for run_name=%r: %s", run_name, exc)
+            return None
+        except Exception:  # unexpected filesystem error -> nothing to resume
             _logger.warning(
                 "FsspecExperimentStore.locate_resumable failed", exc_info=True
             )
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
@@ -28,6 +28,7 @@ Typical use::
 
 from __future__ import annotations
 
+import json
 import logging
 import os
 import uuid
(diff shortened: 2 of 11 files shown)


sallycr (commented):
**DX / API-ergonomics review** (evaluating this as an external adopter, docs + public surface only — not a code-correctness pass).

Overall this is a pleasant seam to adopt. The `ExperimentStore` Protocol mirrors the existing `TrainingObserver` precedent cleanly, the opt-in default is the right call, the docstrings on both `ExperimentStore` and `FsspecExperimentStore` are genuinely good (they explain *where* each method runs, the picklability constraint, the never-raise contract, and why the marker path is derived from identity rather than Ray's dir name), and the PR description stands on its own — I could merge this without asking "what does it do" or "how do I turn it on." The `can_restore()` gating so staleness is never an error is a nice touch. Credit where due.

A few DX rough edges an outside adopter would hit:

**1. Silent no-op when you opt in but under-configure (biggest one).** If I set `experiment_store=FsspecExperimentStore()` but forget `RunConfig.name` or `.storage_path` (or pass no `RunConfig` at all), *nothing happens and nothing tells me*. Both `_maybe_enable_auto_resume` and the worker `_maybe_track_experiment` just `return`. From my seat, I turned the feature on, my re-run started from scratch, and I have no signal why. A one-time `logger.info`/`warning` on the "store is set but identity is incomplete, auto-resume disabled" path would turn a frustrating silent debugging session into a one-line answer. The never-raise contract is right; total silence on a *misconfiguration* (as opposed to normal "nothing to resume") is the part that bites. Line comment below.

**2. The `train(run_config=...)` override silently defeats tracking.** `train()` accepts a `run_config` override and `_maybe_enable_auto_resume()` reads it live via `self.run_config`, but the `(storage_path, run_name)` identity handed to the worker's `track()` is frozen into `train_loop_config` at `__init__` time. So the natural pattern `LightningTrainer(param).train(run_config=rc)` resumes on `rc`'s identity but *never writes a marker* (no identity was injected at init) — auto-resume becomes a permanent silent no-op. Even `__init__(run_config=a).train(run_config=b)` makes track and resume key off different identities. All the docs/examples pass `run_config` to the constructor (which works), but the API openly invites the override path. Worth either re-deriving identity in `train()` or documenting that `run_config` must be set at construction for auto-resume. Line comment below.

**3. No "bring your own store" example.** The CHANGELOG advertises "bring your own `ExperimentStore` for other backends" and the whole point of the Protocol is custom implementations, but every example (PR body, both docstrings) only shows `FsspecExperimentStore()`. A ~10-line example implementing `track`/`locate_resumable` against a custom backend would make the seam's value obvious without reading source — especially since the methods are keyword-only and have a non-obvious "return a *candidate*, don't validate" contract.

**4. Minor: `track` as a method name.** `ExperimentStore.track()` reads like MLflow/W&B-style experiment *tracking* (metrics/params), when it actually just records a resume pointer. `record`/`locate_resumable` or `save`/`locate` would signal the narrower intent. Not blocking, but a new reader may over-assume what the store does.

None of these block merge; #1 and #2 are the ones I'd want addressed since both produce a silent "why didn't my run resume?" with no breadcrumb.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -213,6 +238,48 @@ def train(
...
+        :meth:`fit` restores instead of starting fresh. A store that raises, a
+        missing candidate, or a candidate with no restorable checkpoint all fall
+        through to a fresh run.
+        """
+        store = self._experiment_store
+        if store is None or self.run_config is None:
+            return
+        storage_path = self.run_config.storage_path
+        run_name = self.run_config.name
+        if not storage_path or not run_name:
```
sallycr, line 297: DX: this is the silent-misconfiguration path. If a user opts in (`experiment_store=...`) but `run_config` lacks `name`/`storage_path`, we `return` with no signal — their re-run starts fresh and there's no breadcrumb explaining why. Consider a one-time `_logger.info("experiment_store set but RunConfig lacks name/storage_path; auto-resume disabled")` here (and symmetrically in the worker's `_maybe_track_experiment`). Keeps the never-raise contract while making a misconfiguration debuggable instead of invisible.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -191,6 +214,8 @@ def train(
         if run_config is not None:
             self.run_config = run_config
```
sallycr, line 247: DX trap: this override updates `self.run_config`, which `_maybe_enable_auto_resume()` (line 217) reads live for *resume*. But the `(storage_path, run_name)` identity passed to the worker's `track()` was frozen into `train_loop_config` back in `__init__` (lines 182-184, guarded by `run_config is not None` *at construction*). So `LightningTrainer(param).train(run_config=rc)` resumes on `rc`'s identity but never writes a marker — auto-resume becomes a permanent silent no-op. And `__init__(run_config=a).train(run_config=b)` makes track and resume key off different identities. Either re-inject the identity from `self.run_config` here before `_maybe_enable_auto_resume()`, or document that `run_config` must be supplied at construction for auto-resume to work.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -213,6 +238,48 @@ def train(
...
+
+        if not candidate:
+            return
+
+        fs = self.run_config.storage_filesystem
+        # ``_restore_path`` / ``_restore_storage_filesystem`` are Ray-internal
+        # BaseTrainer attributes; setting them is how a resumable run is enabled
+        # in-place. The public alternative (``TorchTrainer.restore``) rebuilds a
+        # fresh trainer, which is awkward from inside an instance's ``train()``.
+        if TorchTrainer.can_restore(candidate, fs):
```
sallycr: **Blocking.** With the installed Ray (`2.51.2`, Train V2), `TorchTrainer.can_restore` is `@Deprecated` and unconditionally `raise DeprecationWarning(...)` — it never inspects `candidate`. This call is not wrapped in try/except, so the first re-run that finds a marker raises straight out of `_maybe_enable_auto_resume` -> `train()`, crashing the job. That directly contradicts the PR's "staleness is never an error / never fail training" claim. Reproduced end-to-end against real Ray. The tests only pass because they mock this method.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -213,6 +238,48 @@ def train(
...
+            self._restore_path = candidate
+            self._restore_storage_filesystem = fs
```
sallycr: `_restore_path` / `_restore_storage_filesystem` are Ray Train **V1** `BaseTrainer` internals. This class's MRO here is `ray.train.v2...TorchTrainer -> DataParallelTrainer -> object` (no `BaseTrainer`), and neither attribute is referenced anywhere in the V2 trainer — so setting them is a silent no-op and `fit()` will never restore. The V2 resume path is the constructor arg `resume_from_checkpoint: ray.train.Checkpoint`, a different (checkpoint-level) model. This seam likely needs to be redesigned around that.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -160,6 +171,18 @@ def __init__(
...
+        # re-inject it plus the (storage_path, run_name) identity taken from the
+        # driver's RunConfig, so the worker can record this run's experiment
+        # directory keyed off byte-identical strings the driver later resumes on.
+        self._experiment_store = trainer_param.experiment_store
+        train_loop_config.pop("experiment_store", None)
+        if self._experiment_store is not None:
+            train_loop_config["experiment_store"] = self._experiment_store
+            if run_config is not None:
+                train_loop_config["storage_path"] = run_config.storage_path
+                train_loop_config["run_name"] = run_config.name
```
sallycr, line 198: The worker-side `(storage_path, run_name)` identity is injected only here, from the **constructor's** `run_config`. But `train(run_config=...)` is a documented override that updates `self.run_config` (used by the driver's `locate_resumable`) without refreshing this worker identity. Result: if a user passes `run_config` to `train()` instead of `__init__`, `track()` writes no marker and resume silently never works. The two sides should agree on a single authoritative `run_config`.

python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
```diff
@@ -0,0 +1,123 @@
...
+            payload = json.dumps(
+                {
+                    "schema_version": self._SCHEMA_VERSION,
+                    "run_name": run_name,
+                    "experiment_path": experiment_path,
+                    "written_at": datetime.now(timezone.utc).isoformat(),
+                }
+            )
+            fs.makedirs(path.rsplit("/", 1)[0], exist_ok=True)
+            with fs.open(path, "w") as f:
```
sallycr: This write isn't atomic (`open("w")` + `write`, no temp-file + rename). A concurrent reader — or two runs sharing the same `(storage_path, run_name)` — can observe a truncated marker. It degrades safely to `None` (start fresh) so severity is low, but writing to a temp path and renaming into place would make markers robust on NFS/object stores and matches the "never surprise the user" spirit of the rest of the design.

python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
```diff
@@ -0,0 +1,123 @@
...
+            storage_options: Optional keyword arguments forwarded to
+                ``fsspec.core.url_to_fs`` (e.g. credentials or endpoint
+                overrides for the backing filesystem). Kept as a plain dict so
+                the store remains picklable for transport to Ray workers.
+        """
+        self._storage_options = storage_options or {}
+
+    def _marker_path(self, storage_path: str, run_name: str) -> str:
+        """Return the deterministic marker path for ``(storage_path, run_name)``."""
+        return f"{storage_path.rstrip('/')}/{self._MARKER_DIR}/{run_name}.json"
```
sallycr, line 90: `run_name` is interpolated into the marker path with no sanitization. A `run_name` of `../evil` escapes `.michelangelo_resume/` (verified: the marker landed one directory up), and any `/` in a name silently redirects the write and behaves differently across fsspec backends. `run_name` comes from user config so the security risk is low, but consider validating (reject `/`, `..`) or at least documenting the constraint.

python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
```diff
@@ -0,0 +1,123 @@
...
+            fs, path = url_to_fs(
+                self._marker_path(storage_path, run_name), **self._storage_options
+            )
+            if not fs.exists(path):
+                return None
+            with fs.open(path, "r") as f:
+                data = json.loads(f.read())
+            return data.get("experiment_path") or None
+        except Exception:  # missing / corrupt / unreadable -> nothing to resume
+            _logger.warning(
```
sallycr, line 164: A missing/corrupt/unreadable marker is an expected, benign "nothing to resume" condition, but this logs at `warning` with `exc_info=True`, emitting a full stack trace on every resume attempt where no marker exists. That's noisy for the common first-run case. Consider a message-only `debug`/`info` log for the benign paths and reserve the traceback for genuinely unexpected errors.

sallycr (commented):
(Posting as COMMENT — the API blocks REQUEST_CHANGES from this account. Treat the following as a request-changes-level review; I am not approving.)

Thanks for the thorough write-up and docstrings — the intent is clear and the `ExperimentStore` seam is a clean design on paper. However, I ran the code against the Ray version this repo actually depends on and the auto-resume feature does not work; worse, its resume path raises and crashes `train()`, which is the opposite of the PR's stated "staleness is never an error / never fail training" guarantee. The green test suite masks this because every auto-resume test mocks `can_restore`.

## Blocking: built on Ray Train **V1** APIs that are removed/hard-deprecated in the installed Ray Train **V2** (`ray==2.51.2`)

`LightningTrainer`'s actual MRO here is `ray.train.v2.torch.torch_trainer.TorchTrainer -> ray.train.v2.api.data_parallel_trainer.DataParallelTrainer -> object` (no `BaseTrainer`). In that V2 trainer:

1. `TorchTrainer.can_restore(...)` is `@Deprecated` and its body is literally `raise DeprecationWarning(...)` — it **unconditionally raises before looking at the path**, on every backend (local included). `_maybe_enable_auto_resume` calls it **unguarded** (`lightning_trainer.py:278`), so the *first re-run where a marker exists* raises straight out of `train()` and fails the job. I reproduced this end-to-end:

```
$ python -c "...build trainer, store returns a candidate, call _maybe_enable_auto_resume()..."
train() would CRASH on resume path: DeprecationWarning - The `restore` and `can_restore` APIs are deprecated ...
```

2. Even if that call were guarded, `_restore_path` / `_restore_storage_filesystem` are **not referenced anywhere** in the V2 trainer (I grepped `DataParallelTrainer`), so setting them (`lightning_trainer.py:279-280`) is a silent no-op — resume would never actually happen. The V2 resume mechanism is the constructor arg `resume_from_checkpoint: ray.train.Checkpoint` (checkpoint-level, not experiment-directory-level) — a materially different design that this PR would need to be rebuilt around.

3. Every `_maybe_enable_auto_resume` test mocks `TorchTrainer.can_restore` and asserts the private `_restore_*` attributes get set — i.e. the tests validate a Ray **V1** contract the installed runtime does not honor. So the "100% coverage / 222 passed" numbers (which I did reproduce) don't tell you the feature works; they tell you the mock works. I'd want at least one test that exercises the real Ray `can_restore`/resume path (even if xfail-marked) so this can't regress silently.

## Secondary issues (independent of the V2 problem)

- **Silent no-op when `run_config` is passed to `train()` instead of `__init__`.** The worker-side `(storage_path, run_name)` identity is injected only in `__init__` from the constructor's `run_config` (`lightning_trainer.py:182-184`). `train(run_config=...)` — a documented override — updates `self.run_config` (used by the driver's `locate_resumable`) but never refreshes the worker identity, so `track()` writes nothing and resume never works. The two call sites disagree on which `run_config` is authoritative.
- **Remote-scheme mismatch in the PR's own Usage example.** `track` records `os.path.join(storage_context.storage_fs_path, experiment_dir_name)`, and `storage_fs_path` is **scheme-stripped** (`s3://bucket/runs` -> `bucket/runs`). `locate_resumable` returns that scheme-less path, and `can_restore(scheme_less_path, None)` -> `pyarrow.lib.ArrowInvalid: URI has empty scheme` whenever `storage_filesystem` isn't explicitly set — which is exactly the documented `RunConfig(storage_path="s3://bucket/runs")` example. (Moot given the V1/V2 issue, but it shows the remote path was never actually run.)
- **Non-atomic marker write** (`experiment_store.py:92`): `fs.open(path,"w")` + `write` isn't atomic. A reader mid-write, or two concurrent same-identity runs, can observe a truncated marker. It degrades safely to `None` (start fresh), so low severity, but a temp-file + rename would make it robust on NFS/object stores.
- **`run_name` interpolated into a path unsanitized** (`experiment_store.py:62`): a `run_name` of `../evil` escapes `.michelangelo_resume/` (verified: marker written one dir up). `run_name` is user config so this is low security risk, but a name with `/` silently writes somewhere unexpected and behaves differently per fsspec backend — worth validating or documenting.
- **Noisy logging**: a missing/corrupt marker is an expected, benign condition, but `locate_resumable` logs `warning(..., exc_info=True)` (`experiment_store.py:120`), emitting a full stack trace on every resume attempt with no marker. Consider a message-only debug/info log for the benign paths.

Happy to be corrected if I'm misreading the Ray version pinning — but as it stands against `ray==2.51.2` this feature is non-functional and the tests don't catch it. Not approving.

dkurra (approved):

python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
dkurra: agree with this review

python/michelangelo/lib/trainer/torch/pytorch_lightning/experiment_store.py
dkurra: nit pick! not a real issue

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
dkurra: agreeed!

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
dkurra: agreed!

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
dkurra: agreed!

sallycr:
## Review fixes pushed (commit 5c08b969)

### The blocking bug — reproduced, then fixed

**Before:** confirmed against the installed **Ray 2.51.2**, where `ray.train.torch.TorchTrainer` is the **V2** API (`ray.train.v2.torch.torch_trainer.TorchTrainer`). `TorchTrainer.can_restore()` is `@Deprecated` and unconditionally `raise DeprecationWarning(...)`; the old `_maybe_enable_auto_resume` called it **unguarded**, so the first re-run that found a resume marker crashed `train()`. The V1-only `_restore_path` / `_restore_storage_filesystem` attributes don't exist on the V2 trainer, so even without the crash **resume never happened**. The old tests passed only because they mocked `can_restore` and asserted the V1 attributes.

**Important:** the originally-suggested fix — the `resume_from_checkpoint` constructor arg — **is also deprecated in V2** and raises `DeprecationWarning` unconditionally (`DataParallelTrainer.__init__`: `if resume_from_checkpoint is not None: raise DeprecationWarning(...)`). So it could not be used either.

**The redesign (this is a design change, not just a bug fix):** Ray Train V2 resumes **natively** from a reused run directory (`storage_path/name`) via its on-disk `checkpoint_manager_snapshot.json` — no `can_restore`, no `resume_from_checkpoint`. I verified this end-to-end (run 1 → `checkpoint.path=None`; run 2 same identity → resumes from the run-1 checkpoint and continues).

Because `FsspecExperimentStore` is keyed by `(storage_path, run_name)`, it can only ever locate the same directory V2 already auto-resumes from — **so the store is non-load-bearing for the common same-identity case; native V2 resume covers it.** To keep the feature meaningful (and per maintainer guidance not to remove it here), the store now has a real role for the case native resume does *not* cover: resume resolution runs at `__init__` (V2 freezes the run context at construction), and a store-located checkpoint is threaded to the worker as `resume_checkpoint_path`, consumed as `ray.train.get_checkpoint() or <seed>` — **native restoration always takes priority**. The latest checkpoint inside a located dir is resolved from Ray's `checkpoint_manager_snapshot.json` (`latest_checkpoint_result.checkpoint_dir_name`), since V2's public `Result.from_path` is unimplemented. Verified end-to-end: a run with a fresh dir + a custom store pointing at another run's dir seeds from that checkpoint and continues.

### Reproduction confirmation
- **Before my fix:** a re-run with a marker present crashes in `train()` at the unguarded `can_restore` (DeprecationWarning).
- **After my fix:** the same two-run scenario does not crash and resumes (epoch continues past the first run) — asserted in a real, non-mocked integration test.

### Secondary issues
1. **Silent no-op on incomplete identity** — now logs (driver `_resolve_resume_checkpoint`, worker `_maybe_track_experiment`) when `storage_path`/`name` is missing; never-raise contract preserved.
2. **`train(run_config=...)` override** — V2 freezes `run_config` into `TrainRunContext` at construction, so an override can't re-trigger resume. `train()` now warns clearly instead of silently disagreeing.
3. **`run_name` path traversal** — `_marker_path` rejects empty / `/` / `\` / `..` / `.` run names (`_InvalidRunNameError`); `track`/`locate` swallow it to `None`.
4. **Non-atomic marker write** — `track()` writes to a unique temp sibling then `fs.mv()` into place.
5. **Noisy logging** — benign missing/corrupt marker is now `DEBUG`; only unexpected filesystem errors are `WARNING`.
6. **Remote-scheme mismatch** — `track()` records a **scheme-qualified** experiment path (from the driver's `storage_path` + `experiment_dir_name`) instead of the scheme-stripped `storage_fs_path`, so remote seeding addresses the right filesystem.

DX notes: added a "bring your own backend" custom `ExperimentStore` example to the protocol docstring. I did **not** rename `track` → `record` — it ripples through the protocol, default impl, worker, and tests for a cosmetic gain; left as-is.

### Verification
- Full trainer suite: **195 passed** (incl. 2 real Ray integration tests).
- Coverage (pure-Python tracer to avoid a torch/coverage C-tracer conflict): **100%** on `lightning_trainer.py`, `experiment_store.py`, `schema.py`. `_private/util.py` is omitted from the report by project config (worker-internal) but is exercised by the integration tests.

PR #1588, opened 2026-07-23
feat(python): additional_columns + filter_field_map hooks on CRD

yaoyueyue:
## What type of PR is this?

- [x] Feature

## What changed?

Adds three opt-in extension registries on `michelangelo.cli.mactl.crd.CRD` so plugins can extend `<crd> get` without monkey-patching the framework:

| attr | purpose |
|---|---|
| `additional_columns: list[dict]` | Extra table columns. Table-only; `-o yaml`/`-o json` unchanged. |
| `additional_get_args: list[dict]` | Extra argparse flags on `get`, same shape as `func_signature["get"]["args"]`. |
| `filter_field_map: dict[str, str]` | argparse-dest → server-field-name. Non-empty values become `Criterion(op=EQUAL, match_value=Any(StringValue))` on `list_options_ext.operation`. |

`<crd> get -n <ns>` with no positional name **and** a filter dest set now falls through to `list` with that filter applied (previously the filter kwarg was silently dropped). `list_func_signature` is extended with each `filter_field_map` dest so the bound-method call passes `bind_signature` cleanly.

Defaults are empty → CRDs that don't opt in see zero behavior change.

`configure_parser("get", …)` calls `_validated_additional_get_args()` which raises `ValueError` if any extra dest collides with a built-in `get` dest, or if `filter_field_map` references an undeclared dest.

## Why?

Plugins that want a `STATE` column or a `--pipeline-name` filter on their `get` command currently monkey-patch `prepare_column_info` / `_list_func_impl` — several near-identical copies drift silently across CRDs. This lifts the pattern into the framework, mirroring k8s `AdditionalPrintColumns` / `--field-selector`.

## How did you test it?

- **Unit tests:** 52/52 pass in `crd_test.py` (15 new — 4 column hook, 3 filter hook, 4 validation guards, 2 signature folding, 2 end-to-end wiring).
  ```
  PYTHONPATH=. python3 -m pytest michelangelo/cli/mactl/crd_test.py
  ```
- **No-regression E2E:** dev-installed this change into an existing PEX cache and ran `<crd> get` against a live cluster in both table and yaml modes — output byte-identical to pre-change baseline (no CRD opts in yet, so the hook is inert end-to-end).

## Potential risks

None for existing CRDs — all three registries default to empty and the render/list paths short-circuit on empty. Table rendering now `str()`-coerces `retrieve_func` return values, which is strictly more forgiving than pre-change.

## Breaking Changes

- [x] No breaking changes

## Release notes

Framework: `CRD` now exposes `additional_columns`, `additional_get_args`, and `filter_field_map` registries so plugins can extend `<crd> get` output and filters without monkey-patching.

## Documentation Changes

None — self-documenting via the empty-default `__init__` block.

### DIFF
--- a/python/michelangelo/cli/mactl/crd.py
+++ b/python/michelangelo/cli/mactl/crd.py
@@ -314,8 +314,9 @@ def prepare_column_info(extra: Sequence[dict] = ()) -> list[dict]:
 
     ``extra`` is a per-CRD ``additional_columns`` list; each entry
     ``{"column_name": str, "retrieve_func": Callable[[Message], str]}`` is
-    appended after the built-in columns with ``max_length`` seeded from the
-    header length.
+    appended after the built-in columns. Each column's ``max_length`` is
+    seeded from ``len(column_name) + 1`` (the header string plus 1 for
+    trailing space) and grown per-row inside ``print_list_formatted``.
     """
     res = [
         {
--- a/python/michelangelo/cli/mactl/crd.py
+++ b/python/michelangelo/cli/mactl/crd.py
@@ -380,8 +380,7 @@ def _render_list_items(
 ) -> None:
     """Render list of CRD items in the requested output format.
 
-    Matches Go mactl `-o {table|yaml|json}` behavior. ``extra_columns`` is
-    table-only; yaml/json emit the raw proto fields.
+    ``extra_columns`` is table-only; yaml/json emit the raw proto fields.
     """
     if output_format == "yaml":
         docs = [MessageToDict(m, preserving_proto_field_name=True) for m in items]
--- a/python/michelangelo/cli/mactl/crd.py
+++ b/python/michelangelo/cli/mactl/crd.py
@@ -290,8 +290,8 @@ def get_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Mess
     _LOG.debug("No name argument passed. List CRD in the namespace.")
     _self.generate_list(crd_method_info.channel)
     # Forward any filter dest args declared in _self.filter_field_map so
-    # `<crd> get --filter foo=bar` (no name) falls through to list with the
-    # filter applied, not silently dropped.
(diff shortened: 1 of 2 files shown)


python/michelangelo/cli/mactl/crd.py
```diff
@@ -296,8 +297,14 @@ def get_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Mess
...
 
-def prepare_column_info() -> list[dict]:
-    """Prepare column info for formatted printing of CRD items."""
+def prepare_column_info(extra: Sequence[dict] = ()) -> list[dict]:
+    """Prepare column info for formatted printing of CRD items.
+
+    ``extra`` is a per-CRD ``additional_columns`` list; each entry
+    ``{"column_name": str, "retrieve_func": Callable[[Message], str]}`` is
+    appended after the built-in columns with ``max_length`` seeded from the
+    header length.
```
sishilong: what is header length in this context

python/michelangelo/cli/mactl/crd.py
sishilong, line 303: from reading the diff, is ur goal adding this new mechanism
 (filter_field_map -> build a filter -> APIserver) 

do u also need to change here?

python/michelangelo/cli/mactl/crd.py
yaoyueyue: It's a UI design for table when user run ma <crd> get -n<ns> whose default output is a table,  the initial length of each colum in the table is the width of the header string plus 1 char of trailing space, and it grows to max(col["max_length"], len(retrieve_func(item)) + 1). so the table stays condensed and clean.

python/michelangelo/cli/mactl/crd.py
yaoyueyue: good catch, fixed by forwarding filter dest kwargs from get_func_impl to _self.list(...), and extend list_func_signature to accept those per-CRD dests dynamically from filter_field_map

sishilong (approved):

PR #1590, opened 2026-07-23
feat(python): framework -r/--root and -R/--recursive on apply+create

yaoyueyue:
- [x] Feature

**What changed?**
Adds `-r/--root` and `-R/--recursive` to `<crd> apply` at the framework level.

- `-r/--root <path>` — prepend to yaml path before file read; forwarded to per-CRD Construct hook via new `external_root` kwarg on `generate_create`.
- `-R/--recursive` — when `-f` is a directory, walk matching yamls sorted, filter by `kind:`, apply each with continue-on-error aggregate. Per-file server response output is suppressed via `redirect_stdout` in recursive mode; on failure the captured context is dumped alongside `Error: ...`.

Helpers `resolve_yaml_path` + `walk_crd_yamls` + `apply_recursive(directory, kind, per_file_fn)` live inline in `crd.py`. Framework `apply_func_impl` is the reference consumer.

**Why?**
Users applying CRD YAMLs from outside the repo root or batching a directory of yamls need both flags; standard shape for both use cases.

**How did you test it?**

**Unit** — 16 new tests in `crd_test.py`; 56/56 pass. Ruff clean. Coverage 71.2% on `crd.py` (>66%).

**Integration (staging)** — via `<crd> apply` on a CRD served by the framework consumer:

| command | result | verified by |
|---|---|---|
| `ma <crd> apply -r <root> -f <rel>` from foreign CWD | server returned spec | get-before → get-after: `resourceVersion` changed, `uid` unchanged |
| `ma <crd> apply -f <rel>` from foreign CWD (no `-r`) | `FileNotFoundError` at CWD-relative path | proves `-r` was load-bearing (negative control) |
| `ma <crd> apply -R -f <3-file-dir>` | 3× `Applying...`, per-file protos suppressed, `Successfully applied all 3 files ...` | per-file `UpdateTimestamp` bumped on all 3 |
| `ma <crd> apply -R -f /nonexistent` | `FileNotFoundError: directory not found: /nonexistent`, non-zero exit | traceback origin in `walk_crd_yamls` proves no gRPC fired |
| `ma <crd> apply -R -f <3-valid+1-broken>` | 4× `Applying...`, error dumped on broken only, `apply failed on 1 of 4 files: <path>`, non-zero exit | server's own rejection string proves round-trip; continue-on-error confirmed |

**Potential risks**
None.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
- `ma <crd> apply -r <path>` — prepends `<path>` to `-f`; lets apply run outside the repo.
- `ma <crd> apply -R -f <dir>` — apply every YAML in a directory (sorted, kind-filtered, continue-on-error).

**Documentation Changes**
None.

### DIFF
--- a/python/michelangelo/cli/mactl/crd.py
+++ b/python/michelangelo/cli/mactl/crd.py
@@ -2,12 +2,14 @@
 
 import json
 from argparse import ArgumentParser
-from collections.abc import MutableMapping, Sequence
+from collections.abc import Iterator, MutableMapping, Sequence
+from contextlib import redirect_stdout
 from copy import deepcopy
 from dataclasses import dataclass
 from datetime import datetime
 from functools import partial
 from inspect import Parameter, Signature
+from io import StringIO
 from logging import getLogger
 from pathlib import Path
 from types import MethodType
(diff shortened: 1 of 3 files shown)


python/michelangelo/cli/mactl/crd.py
```diff
@@ -495,7 +559,7 @@ def apply_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Me
         # Create new CRD
         _LOG.info("Create a new CRD")
         _self.generate_create(crd_method_info.channel)
-        return _self.create(_file)
+        return _self.create(_file, external_root=external_root)
```
sishilong: will it change the current behavior of create path? create a new CRD
will it cause double prepends the path

python/michelangelo/cli/mactl/crd.py
```diff
@@ -507,13 +571,52 @@ def apply_func_impl(crd_method_info: CrdMethodInfo, bound_args: Signature) -> Me
...
+            # debug context. On success it stays suppressed to keep recursive
+            # output readable.
+            print(buf.getvalue(), end="")
+            print(f"Error: {e}")
+    if failed:
+        raise RuntimeError(
+            f"apply failed on {len(failed)} of {total} files: "
+            + ", ".join(str(p) for p in failed)
+        )
+    if total:
```
sishilong: lets also print if total = 0, but no error case.

yaoyueyue: updated

python/michelangelo/cli/mactl/crd.py
yaoyueyue: Good catch, fixed by making resolve_yaml_path idempotent (early-return when file is already under root). We keep both resolve at apply and create impl because they're independent public entry points in case downstream caller call it without pre-resolved path.

sishilong (approved):

PR #1603, opened 2026-07-25
feat(native_transform): add structural/shape layers (PR B2)

sallycr:
## Summary

Second of an 8-part split (B1–B8) of a larger `native_transform` change, each part adding one family of layers. Builds on **PR B1** (foundation layers), merged as #1570. Adds the **structural / shape-manipulating** torch transform layers to `python/michelangelo/lib/native_transform/torch/base_layers.py`.

### Scope — 7 classes
- `TensorColFillNone` — fill missing (NaN / int-min sentinel) positions with a default
- `CaseWhen` — SQL-like conditional selection over (condition, value) pairs
- `Compare` — pairwise element-wise comparison operators
- `Tile` — repeat along an axis by explicit count or inferred from a target tensor
- `PadOrCrop1D` — pad/crop to fixed length with sentinel handling and left/right align **(riskiest — see below)**
- `Scale` — scalar multiplication
- `Clip` — clamp to a range with optional value exemption

All extend `TorchTransformBaseLayer` from B1 (same `dict[str, torch.Tensor]` forward contract, same `format_inputs`/`format_outputs` pattern). No B3–B8 classes touched.

## Implementation notes
- `INT32_SENTINEL` is imported from the public `michelangelo.lib.constants.sentinel`.
- **Docstrings** are Google style (Args / Returns / Raises) and **full type hints** are present on all public APIs.
- **`CaseWhen`** builds a non-scalar default with `torch.as_tensor` rather than `torch.tensor`, avoiding a copy-construct `UserWarning`; still TorchScript-scriptable for scalar/tensor/list defaults.
- Three genuinely-unreachable defensive `else` branches (the validated `compare_op` in `Compare`; the count/target guard and the matching `multiples` guard in `Tile`) are kept for TorchScript control-flow totality and marked `# pragma: no cover`.
- All seven layers validate their configuration eagerly in `__init__`, matching the B1 layers.

### ⚠️ Reviewer attention: `PadOrCrop1D` sentinel handling
This is the most involved layer. It builds a boolean **sentinel mask** over positions holding upstream ragged padding rather than real data — `NaN` for float dtypes, `INT32_SENTINEL` for int dtypes, auto-detected from the input's dtype — then rewrites those positions to `pad_value`.

The mask is **retained**, not just applied: once a sentinel becomes `pad_value` it is indistinguishable from padding, so `align="right"` uses the mask to find where real content ends and gathers the last `max_length` elements of the *content* (per batch row), left-padding when the content is shorter. `align="left"` pads on the right and keeps the first `max_length` elements, which are always real content.

Tested edge cases: over-padding, zero-length/empty last-dim input (batch dim and dtype preserved), left/right align pad+crop, NaN vs `INT32_SENTINEL` collation sentinels, a non-standard numeric value correctly left as real data, per-row content lengths in a 2D batch, all-sentinel input, and sentinel replacement before cropping.

## Review fixes

`b8afd6eb` fixes three correctness bugs found in review of the initial commit — [full write-up in the commit comment](https://github.com/michelangelo-ai/michelangelo/pull/1603#issuecomment-5098555880):

1. **`PadOrCrop1D` with `align="right"` dropped real data and kept padding.** Sentinel conversion ran before the crop, so trailing `NaN`/`INT32_SENTINEL` became indistinguishable from `pad_value` and the "keep the last `max_length` elements" crop retained the padding. `[1,2,3,nan,nan]` with `max_length=3, pad_value=-1` returned `[3,-1,-1]`; now returns `[1,2,3]`. Fixed by retaining the sentinel mask (described above). `align="left"` was unaffected.
2. **`TensorColFillNone` int64 sentinel detection had a ~±2.7e11 false-positive radius.** `x / 2 == int64_min / 2` promotes int64 to float32, whose 24-bit mantissa cannot resolve values near `int64_min`, so e.g. `int64_min + 1_000_000_000` was wrongly filled. The comment justifying the halved comparison as a TorchScript constant-overflow workaround was incorrect — the direct comparison scripts cleanly (verified with `torch.jit.script`). Replaced with the exact comparison.
3. **`Tile` and `Clip` skipped eager `__init__` validation.** `Tile` without `count` or `target_tensor_provided`, and `Clip` with neither bound, both constructed *and scripted* fine and only failed inside `forward()`. Both now raise `ValueError` at construction, with `Raises:` docstring sections added.

Regression tests were added for all three, covering `align="right"` with float NaN / int sentinel / content shorter than `max_length` / all-sentinel input / per-row content lengths, near-`int64_min` real values, and the two new construction-time errors.

## Tests
Appended to `test_base_layers.py` (the B1 file, not replaced): per-layer coverage of validation errors, dtypes, broadcasting, multi-column inputs, and edge cases, plus a **parametrized TorchScript `script` + save/load round-trip for each of the 7 layers**, following the pattern established in B1.

## Verification
- `pytest michelangelo/lib/native_transform/` → **212 passed, 1 skipped**
- Coverage: **`base_layers.py` 100% statement and branch**
- `ruff format --check` and `ruff check` → **clean** on all touched files
- Full B1 suite (`test_base_layers.py`, `test_utils.py`, `test_id_hash_tokenizer.py`) re-run → **green**, no regressions

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/base_layers.py
+++ b/python/michelangelo/lib/native_transform/torch/base_layers.py
@@ -14,7 +14,6 @@ fitted-statistics, and tokenizer layers are added in follow-up modules.
 from __future__ import annotations
 
 import abc
-import math
 
 import torch
 
@@ -1048,15 +1047,9 @@ class PadOrCrop1D(TorchTransformBaseLayer):
     dimension: shorter sequences are padded with ``pad_value`` and longer ones
     are cropped. ``align`` controls which end is kept and padded.
 
-    Before padding, two kinds of sentinel are converted to ``pad_value`` so that
-    padding introduced upstream (ragged-array collation) is not mistaken for real
-    data:
-
-    - When ``ragged_fill_value`` is set, positions equal to it (or ``NaN`` when
-      it is ``float('nan')``) are replaced with ``pad_value``.
-    - Regardless of ``ragged_fill_value``, remaining ``NaN`` positions (float
-      dtypes) and ``INT32_SENTINEL`` positions (integer dtypes) are replaced with
-      ``pad_value``.
+    Sentinel positions from upstream ragged-batch collation (``NaN`` for float
+    dtypes, ``INT32_SENTINEL`` for integer dtypes) are automatically replaced
+    with ``pad_value`` before the pad/crop logic runs.
 
     Args:
         input_cols: Column names of the input tensors.
@@ -1066,9 +1059,6 @@ class PadOrCrop1D(TorchTransformBaseLayer):
         dtype: Optional output dtype. When ``None``, the input dtype is
             preserved.
         pad_value: The value used for padding (default ``0``).
-        ragged_fill_value: Optional upstream padding sentinel to convert to
-            ``pad_value`` before pad/crop. Supports ``float('nan')`` (detected
-            via ``isnan``) and numeric values (detected via ``==``).
         align: ``"left"`` (default) pads on the right and keeps the first
             ``max_length`` elements; ``"right"`` pads on the left and keeps the
             last ``max_length`` elements of the real content, i.e. sentinel
@@ -1089,7 +1079,6 @@ class PadOrCrop1D(TorchTransformBaseLayer):
         max_length: int,
         dtype: torch.dtype | str | None = None,
         pad_value: int | float = 0,
-        ragged_fill_value: int | float | None = None,
         align: str = "left",
         **kwargs,
     ) -> None:
@@ -1102,8 +1091,6 @@ class PadOrCrop1D(TorchTransformBaseLayer):
             max_length: The fixed target length; must be positive.
             dtype: Optional output dtype; preserves input dtype when ``None``.
             pad_value: The value used for padding (default ``0``).
-            ragged_fill_value: Optional upstream padding sentinel to convert to
-                ``pad_value`` before pad/crop.
             align: ``"left"`` pads/crops on the right; ``"right"`` pads/crops on
                 the left.
             **kwargs: Additional base-layer options (e.g. ``name``).
@@ -1125,12 +1112,6 @@ class PadOrCrop1D(TorchTransformBaseLayer):
         self.dtype = initialize_dtype(dtype, None)
         # Store as float for TorchScript compatibility.
         self.pad_value = float(pad_value)
-        self.ragged_fill_value = (
-            float(ragged_fill_value) if ragged_fill_value is not None else None
-        )
-        self._ragged_sentinel_is_nan = ragged_fill_value is not None and math.isnan(
-            float(ragged_fill_value)
-        )
         self._int_sentinel = INT32_SENTINEL
         self.align = align
 
@@ -1162,23 +1143,11 @@ class PadOrCrop1D(TorchTransformBaseLayer):
         # where the real content ends: once a sentinel is rewritten to
         # ``pad_value`` it is indistinguishable from padding, and cropping to the
         # "last max_length elements" would keep padding over real data.
-        sentinel_mask = torch.zeros_like(stacked_input, dtype=torch.bool)
-        if self.ragged_fill_value is not None:
-            if self._ragged_sentinel_is_nan:
-                sentinel_mask = torch.isnan(stacked_input)
-            elif stacked_input.is_floating_point():
-                incoming = stacked_input.new_full((), self.ragged_fill_value)
-                sentinel_mask = (stacked_input == incoming) | torch.isnan(stacked_input)
-            else:
-                incoming = stacked_input.new_full((), self.ragged_fill_value)
-                sentinel_mask = stacked_input == incoming
-
-        # Remaining collation sentinels: NaN for floats, INT32_SENTINEL for ints.
         if stacked_input.is_floating_point():
-            sentinel_mask = sentinel_mask | torch.isnan(stacked_input)
+            sentinel_mask = torch.isnan(stacked_input)
         else:
             sentinel = stacked_input.new_full((), self._int_sentinel)
-            sentinel_mask = sentinel_mask | (stacked_input == sentinel)
+            sentinel_mask = stacked_input == sentinel
 
         replacement = stacked_input.new_full((), self.pad_value)
         stacked_input = torch.where(sentinel_mask, replacement, stacked_input)
--- a/python/michelangelo/lib/native_transform/torch/base_layers.py
+++ b/python/michelangelo/lib/native_transform/torch/base_layers.py
@@ -740,10 +740,8 @@ class TensorColFillNone(TorchTransformBaseLayer):
             # The int32 minimum encodes a missing value.
             condition = stacked_input == self.int32_min
         elif stacked_input.dtype == torch.int64:
-            # Comparing ``x / 2 == int64_min / 2`` avoids embedding the literal
-            # ``int64`` minimum, which overflows TorchScript's signed-64-bit
-            # constant range.
-            condition = stacked_input / 2 == self.int64_min / 2
+            # The int64 minimum encodes a missing value.
+            condition = stacked_input == self.int64_min
         else:
             # NaN encodes a missing value in floating-point tensors.
             condition = torch.isnan(stacked_input)
@@ -948,6 +946,10 @@ class Tile(TorchTransformBaseLayer):
         target_tensor_provided: When ``True`` and ``count`` is ``None``, infer
             the count from the last input column's size along ``axis``.
         **kwargs: Additional base-layer options (e.g. ``name``).
+
+    Raises:
+        ValueError: If neither ``count`` is set nor ``target_tensor_provided``
+            is ``True``.
     """
 
     def __init__(
@@ -971,8 +973,16 @@ class Tile(TorchTransformBaseLayer):
             target_tensor_provided: When ``True`` and ``count`` is ``None``,
                 infer the count from the last input column's size along ``axis``.
             **kwargs: Additional base-layer options (e.g. ``name``).
+
+        Raises:
+            ValueError: If neither ``count`` is set nor
+                ``target_tensor_provided`` is ``True``.
         """
         super().__init__(input_cols, output_cols, **kwargs)
+        if count is None and not target_tensor_provided:
+            raise ValueError(
+                "Either count must be specified or target_tensor_provided must be True."
+            )
         self.axis = axis
         self.count = count
         self.target_tensor_provided = target_tensor_provided
@@ -996,7 +1006,9 @@ class Tile(TorchTransformBaseLayer):
         elif self.target_tensor_provided:
             source_cols = self.input_cols[:-1]
             target_tensor = inputs[self.input_cols[-1]]
-        else:
+        else:  # pragma: no cover
+            # Unreachable: validated in ``__init__``. Kept so TorchScript sees
+            # every branch bind ``source_cols`` and ``target_tensor``.
             raise ValueError(
                 "Either count must be specified or target_tensor_provided must be True."
             )
@@ -1059,8 +1071,9 @@ class PadOrCrop1D(TorchTransformBaseLayer):
             via ``isnan``) and numeric values (detected via ``==``).
         align: ``"left"`` (default) pads on the right and keeps the first
             ``max_length`` elements; ``"right"`` pads on the left and keeps the
-            last ``max_length`` elements. Retained elements keep their original
-            order.
+            last ``max_length`` elements of the real content, i.e. sentinel
+            positions trailing the data are stripped before the crop rather
+            than being cropped to. Retained elements keep their original order.
         **kwargs: Additional base-layer options (e.g. ``name``).
 
     Raises:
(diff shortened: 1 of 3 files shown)


sallycr:
## Code-correctness review (independent, blind)

Ran the suite locally on `native-transform-pr-b2-structural-layers`: **154 passed**, `base_layers.py` at **100.00% (287 stmts, 84 branches, 0 missed)**, `ruff format --check` and `ruff check` both clean on all 3 touched files. TorchScript round-trip (`TestB2TorchScriptRoundTrip`, 16 cases) does cover all 7 new layers, and the device discipline is respected throughout — every constant is materialized via `ones_like` / `new_full` / `new_empty().copy_()`, no `torch.tensor(...).to(device)` anywhere.

That said, 100% line coverage here coexists with three genuine behavioral bugs, all in untested *combinations* of tested options.

### Must fix

**1. `PadOrCrop1D` with `align="right"` discards real data and keeps padding.**

The sentinel pass rewrites NaN/`INT32_SENTINEL` to `pad_value` *before* the pad+crop, so converted positions become indistinguishable from content. Ragged collation padding is always trailing, so right-alignment then keeps exactly the wrong end:

```python
layer = PadOrCrop1D(input_cols=["f"], output_cols=["o"], max_length=3,
                    pad_value=-1.0, align="right")
layer({"f": torch.tensor([1., 2., 3., nan, nan])})["o"]
# actual:   [3.0, -1.0, -1.0]   <- real values 1,2 dropped, pad retained
# expected: [1.0,  2.0,  3.0]
```

`align="left"` is correct (`[1.0, 2.0, 3.0]`). Reproduces with and without `ragged_fill_value` — the blanket NaN pass alone is enough. For right-alignment the sentinel positions need to be stripped (or the crop needs to be computed against the true length) rather than folded into `pad_value` first.

No test combines `align="right"` with sentinel input; the two dimensions are only ever tested separately.

**2. `TensorColFillNone` int64 path silently destroys genuine values near the minimum.**

`stacked_input / 2 == self.int64_min / 2` runs through float32, whose ~7 significant digits give a false-positive radius of roughly ±5e11 around `int64_min`:

```python
TensorColFillNone(input_cols=["f"], output_cols=["o"], default_value=0)
# input  [int64_min+1, int64_min+1e12, -9.0e18, 5]
# output [0,           -9223371036854775808, -9000000000000000000, 5]
#         ^ int64_min+1 was treated as missing
```

The stated justification doesn't hold: `self.int64_min` already stores the literal as an int64 attribute, and the direct comparison scripts fine — I verified `torch.jit.script` on a module doing `x == self.int64_min` where `self.int64_min = torch.iinfo(torch.int64).min`, which returns exact `[True, False, False]`. Recommend dropping the division and comparing directly (`stacked_input == self.int64_min`), matching the int32 branch.

**3. `Tile` defers config validation to `forward`; `Clip` doesn't validate at all.**

- `Tile(input_cols=["s"], output_cols=["o"])` (no `count`, no `target_tensor_provided`) constructs fine and even `torch.jit.script`s fine — it only raises at call time. Every sibling layer rejects bad config in `__init__`; this one should too.
- `Clip(input_cols=["a"], output_cols=["b"])` with both bounds `None` constructs fine and then dies in `forward` with a raw torch error: `RuntimeError: torch.clamp: At least one of 'min' or 'max' must not be None`. Untested. Needs an `__init__` check.

On the specific question of the two `# pragma: no cover` branches: both **are** genuinely unreachable, not silently wrong. `Compare`'s else is guarded by the identical 6-tuple check in `__init__`; `Tile`'s else is guarded by the `raise` earlier in the same `forward`. Keeping them for TorchScript control-flow totality is correct. But `Tile`'s guard belongs in `__init__` (see above), and the pragma should stay either way.

### Nice to have

- **`CaseWhen` doesn't validate `len(output_cols) == 1`.** Two output cols silently drops the second (`['o1']` returned); zero output cols raises `IndexError` from `self.output_cols[0]` in `forward`. Cheap `__init__` check.
- **`PadOrCrop1D` accepts `dtype="string"`.** `initialize_dtype` legitimately returns the string `"string"`, which then surfaces as a baffling `RuntimeError: Expected one of cpu, cuda, ... at start of device string: string` from `.to()`. Reject non-`torch.dtype` results in `__init__`.
- **`Clip.ignore_value` uses `torch.isclose`** (default rtol 1e-5), so `-1.000005` is preserved as if it were `ignore_value=-1.0`. A sentinel exemption should be exact `==`.
- **`TensorColFillNone` silently promotes dtype.** A float `default_value` on an int32 input yields float32 output (`ones_like(int32) * 0.5` promotes). Similarly `PadOrCrop1D` stores `pad_value` as `float`, so `pad_value=2.7` on an int column pads with `2`. Both are probably fine in practice but worth an explicit cast or a validation.
- **`CaseWhen`'s non-scalar `default_value` isn't a registered buffer.** `state_dict()` is empty, so the constant isn't module state and `.to(device)` doesn't move it — the `new_empty().copy_()` pattern keeps it *correct* at runtime, so this isn't a device-baking bug, but it re-allocates and host→device copies on every forward. `register_buffer` would be both cheaper at serve time and more consistent with the discipline for constants.
- **TorchScript round-trip gaps (no latent failures — I checked).** Not scripted: `Tile(target_tensor_provided=True)` (both axis=0 and axis=1), `Clip` with a single bound, `CaseWhen` with a list default, `PadOrCrop1D` on empty input. I scripted all of these manually and they all pass, so these are coverage gaps rather than bugs.
- `_b2_layer_cases()` is invoked twice in the `parametrize` decorator (once for values, once for `ids`), building and discarding a second set of layer instances. Harmless, but a module-level list would be tidier.

### Verified clean

- `dict[str, Tensor] -> dict[str, Tensor]` contract holds for all 7 layers; extra input keys ignored, missing keys `KeyError` — consistent with the pre-existing layers. `CaseWhen` returning a dict literal instead of `format_outputs` is fine since it's single-output.
- No device baking, no untyped Python objects crossing into scripted forwards (`isinstance` on `self.default_value`, the generator in `Tile`'s `all(...)`, and the dict comprehension all script correctly under type refinement — I confirmed each).
- `PadOrCrop1D` negative pad amounts are correctly floored by `max(..., 0)`; the empty-last-dim early return preserves batch dims and dtype as tested.

sallycr:
## DX review (independent pass — API ergonomics, naming, docs)

Reviewed the diff and the resulting `base_layers.py` at head (`bc721fa`) against the B1 layers already merged in #1570. Docstring coverage is genuinely good: all 7 new classes carry class-level Args/Raises plus per-method Args/Returns, matching B1's class-docstring + `__init__`-docstring convention exactly, and type hints are modern throughout (`int | None`, `torch.dtype | str | None` — no `Optional[X]` anywhere). Code is clean of internal references. Findings below are mostly about naming and where errors surface.

### Must fix

- **PR description leaks internal paths/names.** The body spells out the internal SDK's fully-qualified module path for the shared sentinel constants (Porting notes, bullet 1), plus several other references to the internal codebase ("the internal CanvasFlex SDK", "Uber-internal", "deviations from internal"). The code is clean — I grepped the full diff for `uber|jira|uberinternal|@uber.com` and found nothing — but the PR body is permanent public repo metadata and is held to the same bar. Suggest rewording to something like "`INT32_SENTINEL` now imports from the public `michelangelo.lib.constants.sentinel`" and dropping the upstream-codebase references.

- **Module docstring is now stale.** `base_layers.py:10-11` still reads *"This module provides the foundation (stateless, elementwise) layers. Structural, fitted-statistics, and tokenizer layers are added in follow-up modules."* This PR adds the structural layers to *this* module, so the sentence now actively misdirects a reader looking for `PadOrCrop1D`/`Tile`. Please update it (and consider whether `torch/__init__.py:1-5` needs a matching touch-up).

- **`TensorColFillNone` name.** This is the one class name that reads as a direct carry-over rather than an OSS-designed name, and it breaks from B1's clean vocabulary (`Concatenate`, `Cast`, `Constant`, `IdentityTransform`, …). Two concrete problems: (a) `TensorCol` is redundant — every layer in this file operates on tensor columns; (b) `None` is inaccurate — the layer never sees a Python `None`, it detects `NaN` and the int-dtype minimum (`base_layers.py:739-749`), which the docstring itself has to explain away. `FillMissing` (or `FillNull`) would describe the actual behavior and match the sibling naming. Worth doing now, before the symbol is exported and externally imported — it's already in `__all__` and `torch/__init__.py`.

- **`Tile` validates configuration too late.** The `count is None and not target_tensor_provided` combination is a pure construction-time config error, but it is only raised inside `forward()` (`base_layers.py:999-1002`). Every other layer in the file — including the other six added here — validates in `__init__`. The practical cost is that a misconfigured `Tile` constructs fine and fails at first inference (or at `torch.jit.script` time), instead of failing where the user typed the mistake. Suggest adding the check to `__init__` and keeping the `forward` branch as-is for TorchScript totality; the class docstring would then also gain the `Raises:` section that all its siblings have and it currently lacks.

### Nice to have

- **`CaseWhen` bare `list` annotation.** `default_value: int | float | bool | list | torch.Tensor` (`base_layers.py:785`, repeated at 794) is the only unparameterized generic in the file — everything else is fully parameterized. `list[int] | list[float]` would be consistent and gives users real editor completion.

- **`Compare` error message doesn't enumerate the valid options.** `raise ValueError(f"compare_op: {compare_op} is not supported")` (`base_layers.py:893`) leaves the user to go read the source for the six accepted strings. B1's messages are equally terse so this is consistent rather than a regression — but this is a closed six-value set that's cheap to list, and a typo like `"greater_than"` is the likeliest way to hit it.

- **`PadOrCrop1D` says 1D but isn't.** The implementation operates on the last dimension of an arbitrary-rank stacked tensor (`base_layers.py:1139-1190`), so the `1D` suffix slightly undersells it. Not worth churn on its own given it mirrors a known upstream name; just noting it in case the rename above is being considered anyway.

### Checked and fine

- **The two `# pragma: no cover` comments are adequately explained.** `Compare` (`base_layers.py:920-923`) is exemplary — it names *why* it's unreachable (validated in `__init__`) and *why* it must stay (TorchScript needs `stacked_output` defined on every path). A maintainer with zero context could act on that. `Tile`'s (`base_layers.py:1023-1026`) is thinner — "keeps TorchScript's control flow total" — but "unreachable given the guard above" points at a guard 20 lines up that's easy to find. It would read marginally better if it named the assignment target the way `Compare`'s does. Not a blocker.

- **No internal references in the code.** Verified by grep over the full diff, not just spot-reading.

- **PR description structure** is otherwise strong — the scope table, the explicit deviations list, and the `PadOrCrop1D` reviewer-attention callout are exactly what a reviewer needs. The `B1`–`B8` shorthand is unexplained for a drive-by contributor, but the link to #1570 mostly covers it.

### Out of scope for this PR (flagging for the B-series)

`native_transform` has no page under `docs/` and no entry in `python/examples/`, so these layers are currently discoverable only by reading source. B1 shipped none either, so this is a module-level gap rather than something B2 introduced — but it's worth a tracked follow-up before the B-series wraps, since a user-facing transform library with no prose docs is a real adoption barrier.

sallycr:
Fixed all three issues from the [code-correctness review](https://github.com/michelangelo-ai/michelangelo/pull/1603#issuecomment-5098396255) in b8afd6eb.

**1. `PadOrCrop1D` with `align="right"` dropped real data and kept padding.** Sentinel conversion ran before pad/crop, so a trailing `NaN`/`INT32_SENTINEL` became indistinguishable from `pad_value` and the "keep the last `max_length` elements" crop retained the padding. `[1,2,3,nan,nan]` with `max_length=3, pad_value=-1` returned `[3,-1,-1]`; now returns `[1,2,3]`. The sentinel mask is now retained rather than only applied, and `align="right"` gathers the last `max_length` elements of the *real content*, per batch row, left-padding when content is shorter. `align="left"` is unchanged — its crop always keeps real content.

**2. `TensorColFillNone` int64 sentinel detection had a ~±2.7e11 false-positive radius.** `x / 2 == int64_min / 2` promotes int64 to float32, whose 24-bit mantissa cannot resolve values near `int64_min`; `int64_min + 1_000_000_000` was wrongly filled. The comment justifying this as a TorchScript constant-overflow workaround was incorrect — `int64_min` is a plain Python int attribute and `stacked_input == self.int64_min` scripts cleanly (verified with `torch.jit.script`). Replaced with the exact comparison.

**3. `Tile` and `Clip` skipped eager `__init__` validation.** `Tile` without `count` or `target_tensor_provided`, and `Clip` with neither bound, both constructed *and scripted* fine and only failed inside `forward()` — `Clip` with a raw `torch.clamp` `RuntimeError`. Both now raise `ValueError` at construction, matching every other layer in this file, with `Raises:` docstring sections added. The forward-level `Tile` guard is kept for TorchScript control-flow totality and marked `# pragma: no cover`.

**Tests:** added `align="right"` + sentinel regression coverage (float NaN, int sentinel, explicit `ragged_fill_value`, content shorter than `max_length`, all-sentinel input, and per-row content lengths in a 2D batch), near-`int64_min` real values, and the two new `__init__` errors. The existing `Tile` forward-raise test was moved to construction time.

**Results:** 219 passed, 1 skipped across `native_transform`; `base_layers.py` at 100% statement and branch coverage. `ruff format --check` and `ruff check` clean.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -666,3 +675,678 @@ def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+        Returns:
+            A mapping from each output column to its filled tensor.
+        """
+        stacked_input = format_inputs(self.input_cols, inputs)
+        if stacked_input.dtype == torch.int32:
+            # The int32 minimum encodes a missing value.
+            condition = stacked_input == self.int32_min
+        elif stacked_input.dtype == torch.int64:
+            # The int64 minimum encodes a missing value.
+            condition = stacked_input == self.int64_min
```
kechen2013, line 743: Comparing ``x / 2 == int64_min / 2` was trying to detect the int64 sentinel without writing -9223372036854775808 into the scripted code, so it divided both sides by 2 first -- but that concern only applies to a literal typed in the source, and here the value comes from the self.int64_min attribute, which TorchScript accepts fine. This was from Canvas v1 NTX, and removing it probably is fine.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -666,3 +675,678 @@ def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+            raise ValueError(f"max_length must be a positive integer, got {max_length}")
+        self.max_length = max_length
+        self.dtype = initialize_dtype(dtype, None)
+        # Store as float for TorchScript compatibility.
+        self.pad_value = float(pad_value)
+        self.ragged_fill_value = (
+            float(ragged_fill_value) if ragged_fill_value is not None else None
+        )
+        self._ragged_sentinel_is_nan = ragged_fill_value is not None and math.isnan(
+            float(ragged_fill_value)
```
kechen2013: Note we just did cleanup to remove customized ragged_fill_value in https://github.com/uber-code/uber-one/pull/34869.  Wondering if the PR could be in sync with latest code in main branch?

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -666,3 +675,678 @@ def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+                ``target_tensor_provided`` when set.
+            target_tensor_provided: When ``True`` and ``count`` is ``None``,
+                infer the count from the last input column's size along ``axis``.
+            **kwargs: Additional base-layer options (e.g. ``name``).
+
+        Raises:
+            ValueError: If neither ``count`` is set nor
+                ``target_tensor_provided`` is ``True``.
+        """
+        super().__init__(input_cols, output_cols, **kwargs)
```
kechen2013, line 980: Just a note this fixes something vs internal: internal Tile calls super().__init__(input_cols, output_cols) without **kwargs, so name= is dropped and every Tile ends up named "transform_layer".

python/michelangelo/lib/native_transform/torch/base_layers.py
sallycr: pushed 375f212b removing `ragged_fill_value` from `PadOrCrop1D`, replaced by the same automatic sentinel detection (NaN for floats, INT32_SENTINEL for ints). All 157 tests in `test_base_layers.py` pass, ruff clean. `Bucketization`/`IDHashTokenizer` aren't ported here yet, so their equivalent cleanup will land with those layers in a follow-up PR.

python/michelangelo/lib/native_transform/torch/base_layers.py
```diff
@@ -666,3 +674,648 @@ def forward(self, inputs: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
...
+        Returns:
+            A mapping from each output column to its filled tensor.
+        """
+        stacked_input = format_inputs(self.input_cols, inputs)
+        if stacked_input.dtype == torch.int32:
+            # The int32 minimum encodes a missing value.
+            condition = stacked_input == self.int32_min
+        elif stacked_input.dtype == torch.int64:
+            # The int64 minimum encodes a missing value.
+            condition = stacked_input == self.int64_min
```
kechen2013, line 743: Just a note, not blocking: the system only ever emits `INT32_SENTINEL` for integers, so this `int64_min` check never matches and an int64 column carrying the real sentinel comes back unfilled -- `[5, INT32_SENTINEL]` gives `[5, -2147483648]`, while `PadOrCrop1D` in this same file uses `INT32_SENTINEL` for every int dtype and fills it. Internal main has the same `int64_min` check though, so probably better as a follow-up fixing both sides rather than diverging here.

kechen2013 (approved):

PR #1610, opened 2026-07-27
feat(revision): type Manager interface to *v2.Revision

craigmarker:
## Summary

The Revision Manager previously accepted `client.Object` — callers had to read the implementation to know which fields mattered, and nothing prevented passing an arbitrary object. This PR pins the interface to `*v2.Revision`, so the type system enforces that callers construct the right thing.

Version conversion infrastructure already exists in the repository (`go/kubeproto/conversion/`, `go/api/webhook/`) to handle multi-version translation transparently, so controllers can work with v2 types directly.

## Test plan

I deployed the controllermgr with these changes to the local sandbox and verified the pipeline controller creates revisions end-to-end.

**Unit tests** — all passing:
```
=== RUN   TestBuildRevision          --- PASS
=== RUN   TestBuildRevision_NoGitInfo --- PASS
=== RUN   TestUpsertRevision_Create  --- PASS
=== RUN   TestUpsertRevision_CreateImmutable --- PASS
=== RUN   TestUpsertRevision_DedupImmutable  --- PASS
=== RUN   TestUpsertRevision_RejectImmutableToMutable --- PASS
=== RUN   TestUpsertRevision_UpdateMutable   --- PASS
=== RUN   TestUpsertRevision_MutableThenImmutable --- PASS
ok  	.../go/base/revision	1.127s

=== RUN   TestReconcile_RevisioningDisabled  --- PASS
=== RUN   TestReconcile_RevisioningEnabled   --- PASS
=== RUN   TestReconcile_RevisioningEnabled_NoCommit --- PASS
=== RUN   TestFormatRevisionName     --- PASS
ok  	.../go/components/pipeline	1.154s
```

**Sandbox integration** — deployed controllermgr with `revisioningEnabled: true`, created a test pipeline with commit info, and verified the revision was created:
```
$ kubectl apply -f - <<EOF
kind: Pipeline
metadata:
  name: test-upsertparams-rename
  namespace: default
spec:
  owner: { name: craig.marker }
  commit: { gitRef: "abc123def456", branch: "main" }
EOF

$ kubectl get pipeline test-upsertparams-rename -o jsonpath='{.status.state}'
PIPELINE_STATE_READY

$ kubectl get revisions
NAME                                             AGE
pipeline-test-upsertparams-rename-abc123def456   12s

$ kubectl get revision pipeline-test-upsertparams-rename-abc123def456 -o json | jq '.spec'
{
  "baseType": { "kind": "Pipeline" },
  "baseResource": { "name": "test-upsertparams-rename" },
  "revisionId": "abc123def456",
  "source": "Git",
  "gitCommit": { "branch": "main", "gitRef": "abc123def456" },
  "owner": { "name": "craig.marker" }
}
```

### DIFF
--- a/go/base/revision/input.go
+++ b/go/base/revision/input.go
@@ -1,0 +1,48 @@
+package revision
+
+import (
+	apipb "github.com/michelangelo-ai/michelangelo/proto-go/api"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// RevisionInput carries the caller-facing fields needed to create or update
+// a Revision. The Manager builds the full v2 Revision proto internally —
+// callers never import or construct a *v2pb.Revision directly.
+//
+// The caller owns Content preparation: BaseCR is marshaled as-is, so strip
+// any unwanted metadata (e.g. ManagedFields) before passing it in.
+type RevisionInput struct {
+	// Name is the Revision object's metadata name. The caller controls the
+	// naming scheme (e.g. "pipeline-foo-abc123", "my-model-42").
+	Name string
+
+	// BaseCR is the resource being revisioned. The manager derives Namespace,
+	// BaseType (Kind/APIVersion), BaseResource (namespace/name), and Content
+	// (proto-marshaled) from this object.
+	BaseCR client.Object
+
+	// Owner is the username recorded as the revision owner.
+	Owner string
+
+	// RevisionID is a caller-chosen identifier for this revision point
+	// (git SHA, generation number, semver string, etc.).
+	RevisionID string
+
+	// Labels applied to the Revision's ObjectMeta.
+	Labels map[string]string
+
+	// Annotations applied to the Revision's ObjectMeta.
+	Annotations map[string]string
+
+	// Source indicates how the revision was produced (e.g. SourceGit).
+	Source string
+
+	// GitRef populates CommitInfo.GitRef (e.g. a commit SHA).
+	GitRef string
+
+	// GitBranch populates CommitInfo.Branch.
+	GitBranch string
+
+	// Parent points to the previous revision, if any.
+	Parent *apipb.ResourceIdentifier
+}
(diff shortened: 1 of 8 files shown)


go/base/revision/input.go
```diff
@@ -0,0 +1,48 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UpsertParams carries the caller-facing fields needed to create or update
+// a Revision. The Manager builds the full v2 Revision proto internally —
+// callers never import or construct a *v2pb.Revision directly.
+//
+// The caller owns Content preparation: BaseCR is marshaled as-is, so strip
+// any unwanted metadata (e.g. ManagedFields) before passing it in.
+type UpsertParams struct {
```
uzbq, line 14: if we decide to move forward with scheme-driven interface, how can this param can be helpful? 

When users use this interface, they still need to deal with the general client.Object.

go/base/revision/interface.go
```diff
@@ -2,30 +2,22 @@ package revision
...
-	// complete Revision object; the Manager orchestrates the state machine
-	// (get existing, create if absent, check immutability, update if mutable).
+	// UpsertRevision creates or updates a Revision built from the given input.
+	// The Manager marshals the BaseCR, derives BaseType and BaseResource, and
+	// orchestrates the state machine (get existing, create if absent, check
+	// immutability, update if mutable).
 	// Returns (true, nil) on create or update, (false, nil) on dedup (an
 	// existing immutable Revision with the same name already exists).
-	UpsertRevision(ctx context.Context, rev client.Object, opts UpsertOpts, options ...yarpc.CallOption) (bool, error)
+	UpsertRevision(ctx context.Context, input UpsertParams, opts UpsertOpts) (bool, error)
```
badcount, line 17: Let's use Revision instead of a param which is pretty much same as Revision proto

craigmarker: agree. one benefit of the existing interface is that the manager owns marshaling the base resource (eg pipeline) into proto any and base resource. we have two options:

1. Add interface to manager that takes crd-like (pipeline, model, etc) and marshals into proto any
2. Require callers to marshal into proto any, which pushes more error handling to clients

What do you think @uzbq @badcount ?

PR #1614, opened 2026-07-28
feat(trainer): re-introduce profiler subsystem (PR 5)

sallycr:
## Scope

Adds opt-in PyTorch Lightning profiling to `LightningTrainer`, configured through `lightning_trainer_kwargs["profiler"]`. This is the last unimplemented piece of the trainer module migration.

### Profiler construction (`_private/util.py`)

- `_build_profiler` accepts `None`, a shorthand string (`"simple"` / `"advanced"` / `"pytorch"` / `"xla"`), or a dict setting **exactly one** of `simple` / `advanced` / `pytorch` / `xla` / `custom`. Zero or multiple selections raise `UserInputError`. `custom` resolves a dotted class path and rejects anything that is not a `pl.profilers.Profiler` subclass.
- All profilers write to a fixed `profiler_logs/` directory under the worker cwd, with a per-world-rank trace filename (Lightning appends the local rank), so results are locatable after `fit()` returns.
- Lightning profiler classes are imported lazily inside the functions, matching the existing style for optional/heavy deps in this module.

### Schedule derivation

- `_compute_steps_per_epoch` estimates `ceil(rows / world_size / batch_size)`, then applies `limit_train_batches` (float scales, int caps) and `max_steps` (scaled by `accumulate_grad_batches`) in Lightning's own order.
- `_compute_default_schedule` picks a sane `wait`/`warmup`/`active` window for the epoch length; `_validate_profiler_schedule` bounds an explicit schedule (`active` <= 10, `repeat` <= 3) and rejects one that overruns the epoch with a message spelling out the arithmetic so users can fix their config.
- `LightningTrainer.__init__` counts training rows **only** when a profiler is configured, since `count()` is cheap on Parquet-metadata-backed datasets but forces execution otherwise. A failed count degrades to `None` (unbounded default schedule) rather than failing the run.

### Result export: `profiler_sink`

New optional `LightningTrainerParam.profiler_sink`, a `Callable[[Profiler, str, Logger], None]` invoked on each node-local rank 0 after `fit()`:

- The generic profiler builders stay tracker-agnostic; the sink is the only tracker-aware piece, consistent with the pluggable-backend approach used in earlier PRs in this migration.
- Gated on `profiler is not None`, the config's `upload_to_comet` flag, and `local_rank == 0` (one call per node, since `profiler_logs/` is shared by all workers on a node).
- Exceptions from the sink are logged and swallowed — profiling is observability and must not fail a completed run.
- `comet_profiler_sink` ships the standard Comet export (`log_tensorflow_folder` for `PyTorchProfiler`, per-file `log_asset` for `SimpleProfiler`/`AdvancedProfiler`, skip for XLA/custom) and is re-exported from the package so users can reference it without importing from `_private`.

The `TrainingObserver` Protocol is deliberately **not** extended — it is `@runtime_checkable`, so adding a required method would break existing implementations that currently satisfy it.

## Deviations from the source snapshot

1. **Sink receives the logger as a call argument** rather than being a factory closing over it. The logger is built inside the worker by `_resolve_logger`, while `profiler_sink` is set on the driver, so a driver-side closure cannot capture it. Passing it at call time also keeps the sink a plain module-level function, which pickles cleanly to workers.
2. **Row count is read off `trainer_param.train_data`**, not the value popped from the loop config — `dataclasses.asdict()` has already deep-copied the latter.
3. Validation failures raise the repo's `UserInputError` rather than the source's bespoke exception type; `print` diagnostics became `_logger` calls.
4. Environment-specific logger subclasses and pipeline-annotation hooks from the source were dropped; the generic `profiler_sink` seam replaces them.

### Known limitation, documented in `comet_profiler_sink`

Lightning decorates `CometLogger.experiment` with `@rank_zero_experiment`, so on any worker whose **global** rank is not 0 the property yields a no-op dummy. In a multi-node run only the node holding global rank 0 actually exports its profile; other node leaders' uploads are silently dropped. This is called out in the docstring, with a pointer to writing a custom sink that resolves its own experiment if per-node profiles are needed.

## Verification

- Full trainer suite: **353 passed** (up from 293), 0 failures.
- Coverage on the new profiler code in `_private/util.py`: **100%** of statements and branches outside `_train_loop_per_worker`. Package-level coverage for `lightning_trainer.py`, `schema.py`, `experiment_store.py`, and `__init__.py` remains **100.00%**. (`_private/` is excluded from the project coverage gate by `pyproject.toml`, as its worker-loop body needs a live Ray runtime; the two new call sites inside `_train_loop_per_worker` share that exclusion.)

### DIFF
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -536,6 +536,16 @@ def _compute_steps_per_epoch(
        accumulate_grad_batches`` (``max_steps`` counts optimizer steps, the
        estimate counts mini-batches).
 
+    The division by ``world_size`` assumes each worker iterates roughly
+    ``1 / world_size`` of the rows. That holds for every configuration this
+    trainer supports: the split is performed by Ray Train's default
+    ``DataConfig`` (``datasets_to_split="all"``), which shards the dataset
+    evenly across workers before the Lightning strategy sees it, so it is
+    unaffected by the choice of DDP / FSDP / DeepSpeed. A caller supplying a
+    custom ``Strategy`` under which several workers form one model replica and
+    must therefore see identical batches would break the assumption, and the
+    estimate would then be low by the replication factor.
+
     Args:
         trainer_kwargs: Keyword arguments destined for ``pl.Trainer``. Read
             only; ``limit_train_batches``, ``max_steps``, and
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -64,7 +64,14 @@ _MAX_PROFILER_ACTIVE_STEPS = 10
 _MAX_PROFILER_REPEAT = 3
 
 # Shorthand strings accepted in place of a full profiler config dict.
-_PROFILER_SHORTHANDS = ("simple", "advanced", "pytorch", "xla")
+#
+# Only ``pytorch`` is supported today — Michelangelo OSS is PyTorch-heavy, and
+# the other Lightning profilers (simple / advanced / xla) and a custom-class
+# escape hatch add config surface without a concrete user yet. The dict shape
+# (``{"pytorch": {...}, "upload_to_comet": ...}``) is kept as-is so adding a
+# second backend later, as we expand to more platforms, is additive rather
+# than a breaking config change.
+_PROFILER_SHORTHANDS = ("pytorch",)
 
 # Directory (under the worker's cwd) that every profiler writes its output to.
 # Fixed rather than temporary so a profiler sink can find the results after
@@ -765,6 +772,9 @@ def _build_profiler(
 ) -> tuple[pl.profilers.Profiler | None, str]:
     """Build a Lightning profiler from a user-supplied profiler config.
 
+    Only the ``pytorch`` profiler is currently supported (Michelangelo OSS is
+    PyTorch-heavy); more backends can be added as the platform expands.
+
     Every profiler writes into a fixed ``profiler_logs`` directory under the
     worker's current working directory, so a profiler sink can locate the
     results after training. Traces are named per world rank; Lightning's
@@ -774,21 +784,16 @@ def _build_profiler(
         profiler_config: One of
 
             * ``None`` — no profiler.
-            * A shorthand string: ``"simple"``, ``"advanced"``, ``"pytorch"``,
-              or ``"xla"``, equivalent to that key mapped to ``{}``.
-            * A dict setting exactly one of ``simple``, ``advanced``,
-              ``pytorch``, ``xla``, or ``custom`` to a kwargs dict. The
-              ``simple`` / ``advanced`` / ``xla`` kwargs are forwarded verbatim
-              to the corresponding Lightning profiler; ``pytorch`` is handled by
-              :func:`_build_pytorch_profiler`; ``custom`` takes
-              ``profiler_class`` (a dotted path to a ``pl.profilers.Profiler``
-              subclass) and optional ``profiler_kwargs``.
+            * The shorthand string ``"pytorch"``, equivalent to
+              ``{"pytorch": {}}``.
+            * A dict setting ``pytorch`` to a kwargs dict, handled by
+              :func:`_build_pytorch_profiler`.
 
             ``upload_to_comet`` is reserved metadata read by
             :func:`_resolve_profiler` and is never forwarded to a profiler
             constructor.
-        steps_per_epoch: Estimated steps per epoch, used only by the
-            ``pytorch`` branch to derive or validate a schedule.
+        steps_per_epoch: Estimated steps per epoch, used to derive or
+            validate the profiler's schedule.
 
     Returns:
         A ``(profiler, profiler_logs_path)`` tuple. Both are ``None`` / ``""``
@@ -797,9 +802,7 @@ def _build_profiler(
     Raises:
         TypeError: If ``profiler_config`` is neither a str, dict, nor ``None``.
         ValueError: If a shorthand string is not a recognized profiler name.
-        UserInputError: If the dict sets zero or more than one profiler, or if
-            ``custom`` resolves to something that is not a Lightning
-            ``Profiler`` subclass.
+        UserInputError: If the dict sets zero or more than one profiler.
     """
     if profiler_config is None:
         return None, ""
@@ -817,19 +820,17 @@ def _build_profiler(
 
     # Exactly one profiler flavor must be selected.
     selected = [
-        name
-        for name in ("simple", "advanced", "pytorch", "xla", "custom")
-        if profiler_config.get(name) is not None
+        name for name in _PROFILER_SHORTHANDS if profiler_config.get(name) is not None
     ]
     if not selected:
         raise UserInputError(
-            "Profiler config must set exactly one of: simple, advanced, "
-            "pytorch, xla, custom. None were set."
+            "Profiler config must set exactly one of: "
+            f"{', '.join(_PROFILER_SHORTHANDS)}. None were set."
         )
     if len(selected) > 1:
         raise UserInputError(
-            "Profiler config must set exactly one of: simple, advanced, "
-            f"pytorch, xla, custom. Multiple were set: {selected}."
+            "Profiler config must set exactly one of: "
+            f"{', '.join(_PROFILER_SHORTHANDS)}. Multiple were set: {selected}."
         )
 
     profiler_type = selected[0]
@@ -846,39 +847,8 @@ def _build_profiler(
     # Lightning's Profiler base class appends the local rank to this filename.
     filename = f"profile-world-rank-{world_rank}-local-rank"
 
-    from pytorch_lightning.profilers import (
-        AdvancedProfiler,
-        SimpleProfiler,
-        XLAProfiler,
-    )
-
-    if profiler_type == "simple":
-        return SimpleProfiler(
-            dirpath=profiler_logs_path, filename=filename, **profiler_kwargs
-        ), profiler_logs_path
-    if profiler_type == "advanced":
-        return AdvancedProfiler(
-            dirpath=profiler_logs_path, filename=filename, **profiler_kwargs
-        ), profiler_logs_path
-    if profiler_type == "pytorch":
-        return _build_pytorch_profiler(
-            profiler_logs_path, filename, profiler_kwargs, steps_per_epoch
-        ), profiler_logs_path
-    if profiler_type == "xla":
-        return XLAProfiler(**profiler_kwargs), profiler_logs_path
-
-    profiler_class = profiler_kwargs.get("profiler_class")
-    profiler_cls = get_module_attr(profiler_class)
-    if not (
-        isinstance(profiler_cls, type)
-        and issubclass(profiler_cls, pl.profilers.Profiler)
-    ):
-        raise UserInputError(
-            f"Custom profiler class {profiler_class!r} must be a subclass of "
-            f"pytorch_lightning.profilers.Profiler, got {profiler_cls!r}."
-        )
-    return profiler_cls(
-        **(profiler_kwargs.get("profiler_kwargs") or {})
+    return _build_pytorch_profiler(
+        profiler_logs_path, filename, profiler_kwargs, steps_per_epoch
     ), profiler_logs_path
 
 
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -7,9 +7,11 @@ logger / callback resolution helpers. Public APIs live in
 
 from __future__ import annotations
 
+import glob
 import hashlib
 import inspect
 import logging
+import math
 import os
 import re
 from tempfile import TemporaryDirectory
@@ -54,6 +56,20 @@ _PLUGIN_INPUT = Union[Precision, ClusterEnvironment, CheckpointIO, LayerSync]
 
 CALLBACK_REPORT_PER_NODE = "callback_report_per_node"
 CHECKPOINT_FILENAME = "checkpoint.ckpt"
+
+# Upper bounds on the ``torch.profiler`` schedule. Both windows multiply the
+# volume of trace data written per worker, and traces are large; these caps
+# keep a mis-typed config from filling the run's storage.
+_MAX_PROFILER_ACTIVE_STEPS = 10
+_MAX_PROFILER_REPEAT = 3
+
+# Shorthand strings accepted in place of a full profiler config dict.
+_PROFILER_SHORTHANDS = ("simple", "advanced", "pytorch", "xla")
+
+# Directory (under the worker's cwd) that every profiler writes its output to.
+# Fixed rather than temporary so a profiler sink can find the results after
+# ``trainer.fit()`` returns.
+_PROFILER_LOGS_DIRNAME = "profiler_logs"
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


sallycr:
**Developer-experience / documentation review** (independent pass; correctness is being reviewed separately).

The implementation reads well and the new helpers are unusually well documented — every new function has Args/Returns and a Raises section where it can raise, the `_MAX_PROFILER_*` constants carry a real "why", and the table in `_compute_default_schedule` is a genuinely nice touch. My findings are almost entirely about **public discoverability and the public type surface**, not the internals.

I also grepped the diff and the PR body for internal identifiers (ticket IDs, employee names, internal hostnames/URLs, internal service names): **clean**, nothing to redact.

---

### Must-fix

- **`lightning_trainer_kwargs` docstring is now inaccurate.** `LightningTrainerParam.lightning_trainer_kwargs` (`lightning_trainer.py`) still says *"Extra keyword arguments forwarded verbatim to `pytorch_lightning.Trainer(...)`"*. After this PR that is false for the `profiler` key: `_train_loop_per_worker` pops it and reshapes it into a config that `pl.Trainer` would reject outright. A user following the docstring literally will write `profiler=PyTorchProfiler(...)` (the Lightning-native form) and get a confusing failure. The attribute doc needs to call out that `profiler` is intercepted.

- **The profiler config schema has no public documentation.** The entire accepted shape — shorthand strings, the exactly-one-of dict, nested `schedule`, the dotted-path `on_trace_ready`, `custom` + `profiler_class` + `profiler_kwargs`, and `upload_to_comet` — lives only in the docstring of `_build_profiler`, inside `_private/util.py`. That is the one place a user of the package will never look. This is the single biggest DX gap in the PR: the feature is effectively undiscoverable from the public surface. Please mirror the schema (or at least a worked example plus a pointer) into the `lightning_trainer_kwargs` attribute docs or the package docstring, alongside the existing `comet_profiler_sink` bullet in `__init__.py`.

- **`upload_to_comet` is a tracker-specific name on a deliberately tracker-agnostic seam.** The PR description is explicit that the sink is the only tracker-aware piece and that the builders stay generic — but the flag that gates *any* sink is named after one vendor. A user who writes an S3 or MLflow sink has to set `upload_to_comet: False` to turn their own export off, which reads as a bug. `_resolve_profiler` already calls the internal variable `upload_results`; consider naming the config key to match (`export_results` / `upload_results`). Since this is a new key, now is the only cheap moment to rename it.

- **Public type hints fall back to `Callable` / `Any` where a real type exists.**
  - `LightningTrainerParam.profiler_sink: Callable | None` — the PR body states the contract precisely (`Callable[[Profiler, str, Logger], None]`), but the annotation carries none of it, so neither an IDE nor a type checker will help a user get the three-arg signature right.
  - `comet_profiler_sink(..., logger: Any)` — this is a **re-exported public function**, and `Logger` is already imported at module scope in `_private/util.py`; `Any` is not needed here.
  - `_maybe_export_profiler_results(..., profiler_sink: Any)` — same, and its docstring already spells out the callable type in prose.

  Package precedent for a named seam type is right next door: `TrainingObserver` and `ExperimentStore` are `@runtime_checkable` Protocols in `schema.py`, and the two comparable `LightningTrainerParam` fields are annotated with them. A `ProfilerSink` Protocol (or at minimum a `ProfilerSink` type alias) in `schema.py` would match that convention, be self-documenting, and give the sink the same discoverability the other two seams have. I'd note the PR body's stated reason for not extending `TrainingObserver` — that it is `@runtime_checkable` and adding a required method breaks existing implementers — is correct and I agree with the separate-param decision; this is about *typing* the new param, not merging it into the observer.

- **Cross-reference points into a private module.** The `profiler_sink` docstring uses ``:func:`~_private.util.comet_profiler_sink` ``. The function is re-exported at the package root by this same PR, so the reference should use the public path — otherwise rendered docs link users at `_private`.

### Nice-to-have

- **Three exception types for one config field.** `profiler` config mistakes surface as `ValueError` (unrecognized shorthand; bad `limit_train_batches`; bad `accumulate_grad_batches`), `TypeError` (config is not str/dict/None), or `UserInputError` (zero/multiple profilers selected; out-of-range schedule). These are all the same category of mistake — a human typed the wrong thing in a config dict — and a caller who wants to catch them has to catch three types. `UserInputError` looks like the right single answer.

- **`world_sz` vs `world_size`.** `_resolve_profiler` takes `world_sz` while the `_compute_steps_per_epoch` it immediately calls takes `world_size`. `world_sz` mirrors the pre-existing local in the worker loop so it is defensible, but two adjacent new functions disagreeing on the same concept is avoidable.

- **`_resolve_profiler` has no `Raises:` section** despite being the entry point that propagates every exception the profiler subsystem can produce. One line pointing at `_build_profiler` / `_compute_steps_per_epoch` would complete it.

- **The known-limitation note is clearly written but stops just short of actionable.** The description and the `comet_profiler_sink` docstring both explain the rank-zero-gated `experiment` property well — I understood the failure mode on one read, and the "silently dropped" framing is exactly right since that is the part users would otherwise waste hours on. The remedy, "pass a custom sink that resolves its own experiment," is the correct advice but gives no starting point. Three lines sketching what such a sink does would turn a warning into a fix.

- **Title framing.** "re-introduce ... (PR 5)" is migration-sequence language; from this repository's history nothing was removed, and `(PR 5)` will not mean anything to an external reader or in the generated changelog. Something like `feat(trainer): add opt-in PyTorch Lightning profiler support` describes the same change in terms a reader of this repo can act on. Relatedly, the "Deviations from the source snapshot" section is unreviewable for anyone without the source; the package docstring already frames the trainer as a one-time snapshot, so consider presenting those four points as design rationale rather than as deltas from something invisible.

### Reads well — no change needed

- `_maybe_export_profiler_results` is a clear name and matches the existing `_maybe_track_experiment` convention in the same module.
- `comet_profiler_sink` sits fine next to `build_comet_logger`; the differing prefix correctly signals "this *is* the callable" vs "this *builds* one."
- The test files are well organized and the docstrings-as-specs style makes the intended behavior legible without reading assertions.

sallycr:
## Code-correctness review (independent, blind)

Reviewed the diff and the checked-out branch at `a03288b6`, traced the schedule arithmetic by hand and with a throwaway script, and ran the suite locally. **The core logic is correct** — the gating in `_maybe_export_profiler_results`, the mutual-exclusivity check in `_build_profiler`, the `steps_per_epoch` threading, the `asdict()` row-counting fix, and the `rank_zero_experiment` limitation all check out exactly as described. Findings below are edge cases and one coverage-reporting caveat; nothing blocking.

### Verified claims

- **Sink gating** (`_private/util.py:1258-1262`) is exactly `profiler is not None and upload_profiler_results and profiler_sink is not None`, then `get_local_rank() == 0`. No inverted condition. Exceptions from the sink are caught by a bare `except Exception` and logged (`util.py:1263-1268`) — a raising sink cannot fail the run.
- **Row-counting fix** (`lightning_trainer.py:190-211`) genuinely counts off `trainer_param.train_data`, i.e. the real object, not the `asdict()`-deep-copied value popped into `train_loop_config`. The `try/except Exception` falls back to `None` without escaping `__init__`. The count is correctly skipped entirely when no profiler is configured.
- **`upload_to_comet` is never forwarded** to a profiler constructor: `profiler_kwargs = profiler_config[profiler_type]` only reads the selected flavor's sub-dict, so sibling metadata keys can't leak. Confirmed by `test_upload_to_comet_is_not_a_profiler_selection`.
- **`custom` branch handles a resolved non-class gracefully**: the `isinstance(profiler_cls, type) and issubclass(...)` guard at `util.py:872-878` short-circuits before `issubclass` would raise `TypeError` on a function/instance. Verified with a resolved lambda.
- **`_resolve_profiler` call-site values are right** (`util.py:1177-1184`): `batch_size` is the per-worker `train_loop_config["batch_size"]` fed to `iter_torch_batches`, and `world_sz` is `get_world_size()` — so `ceil(rows / world_size / batch_size)` matches the actual per-worker shard. No off-by-one, and `trainer_kwargs` is passed after `profiler` is popped but while `limit_train_batches` / `max_steps` / `accumulate_grad_batches` are still present, which is what the estimator reads.
- **Default schedule always fits its epoch.** I enumerated n=1..20 independently: every default satisfies `(wait+warmup+active)*repeat + skip_first <= n`. `warmup = min(2, remaining-1)` guarantees `active >= 1` for all n>=3, so the schedule can never degenerate to `active=0`.
- **`rank_zero_experiment` limitation is real and unaddressed.** `pytorch_lightning/loggers/comet.py:260-262` stacks `@property` / `@rank_zero_experiment` on `experiment`, and the decorator (in `lightning_fabric/loggers/logger.py`) returns `_DummyExperiment()` whenever `rank_zero_only.rank > 0`. Worth noting the second-order effect: because the dummy is a *non-None* object with a catch-all `__getattr__`, the `experiment is None` guard at `util.py:918` never fires for a `CometLogger` on a non-zero rank — the upload calls succeed as no-ops. The docstring note is accurate; the code is not silently broken in some other way.
- **Picklability holds.** `profiler_sink` is left in `train_loop_config` (functions are atomic for `deepcopy`, so `asdict()` preserves identity — asserted by `test_profiler_sink_reaches_the_worker_config`). I additionally round-tripped both a module-level and a locally-defined sink through `ray.cloudpickle` successfully, so the claim survives past the identity assertion the test makes.

### Nice-to-have (no must-fix items found)

- **`{"custom": {}}` raises a raw `AttributeError`, not `UserInputError`.** `profiler_kwargs.get("profiler_class")` returns `None`, which reaches `get_module_attr(None)` and dies on `None.rpartition`. Every other malformed-config path in `_build_profiler` produces a typed error with a helpful message; a missing `profiler_class` should too. Currently untested.
- **`steps_per_epoch == 0` + an explicit schedule produces an unsatisfiable error.** A zero-row dataset, or a small one with `limit_train_batches=0.001` (`int(10 * 0.001) == 0`), yields `steps_per_epoch=0`. `_validate_profiler_schedule` then always raises, telling the user to make the total `<= 0` — impossible, since `active >= 1` is separately enforced. The default-schedule path dodges this only because it skips validation. Consider treating `steps_per_epoch <= 0` the same as `None` (skip the bounds check, warn), or emitting a dedicated "no training steps" message.
- **`_validate_profiler_schedule` doesn't type-check its fields.** `{"active": "3"}` raises a bare `TypeError` from the comparison rather than `UserInputError`, and floats slip through entirely: `active=2.5` and `repeat=1.5` both pass validation and are then silently accepted by `torch.profiler.schedule` with surprising behavior (I confirmed `active=2.5` records 3 steps and `repeat=1.5` starts a second cycle). An `isinstance(v, int) and not isinstance(v, bool)` check per field would close both.
- **`comet_profiler_sink` doesn't handle a list of loggers.** `_resolve_logger` explicitly supports `list[Logger]` (`util.py:399-404`), and a Comet-plus-something config is a normal setup. In that case `getattr(logger, "experiment", None)` on the list returns `None`, so the sink logs "has no Comet experiment" and skips. Scanning a list for the first logger exposing `experiment` would make the shipped sink work for the multi-logger case.
- **Coverage caveat, not a code bug.** `pyproject.toml` omits `*/lib/trainer/**/_private/*`, so essentially all of this PR's new code is excluded from the default coverage report — which reads a clean `100.00%` while measuring none of it. Re-running with that omit lifted, `_private/util.py` is **76.49%**, but every uncovered line is either a pre-existing helper or the worker loop (`1064-1234`); the profiler block (`520-1001`, `1243+`) has **no missed lines**. Separately: `comet_profiler_sink` is now a public re-export from `__init__.py`, but it lives in a module the repo's own config describes as worker-internal and not-for-direct-import. Might be worth relocating it, or narrowing the omit pattern, so the public surface isn't inside the excluded path.
- Minor doc nit: the `_compute_default_schedule` table jumps from 5 to "N >= 7"; n=6 (`wait=2, warmup=2, active=2`) falls in the gap.

### Test / lint results observed

- Full suite: `353 passed` in 65.0s — **zero failures**, including both `test_auto_resume_integration.py` tests, so the known-flaky pair passed cleanly on this branch.
- New tests alone (`test_profiler.py` + `TestProfilerRowCounting`): `158 passed` in 3.7s.
- Coverage (parallel mode, `coverage run` + `combine` + `report`): default config `100.00%` over 260 stmts; with the `_private` omit lifted, `85.53%` over 741 stmts, `_private/util.py` at 76.49% with the profiler code fully covered.
- `ruff format --check`: 5 files already formatted. `ruff check`: all checks passed.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
```diff
@@ -501,6 +517,488 @@ def _resolve_callbacks(
...
+    if dataset_num_rows is None:
+        return None
+    if batch_size <= 0:
+        raise ValueError(f"batch_size must be a positive integer, got {batch_size}.")
+    if world_size <= 0:
+        raise ValueError(f"world_size must be a positive integer, got {world_size}.")
+    if dataset_num_rows == 0:
+        _logger.warning("dataset_num_rows is 0; steps_per_epoch will be 0.")
+
+    steps_per_epoch = math.ceil(dataset_num_rows / world_size / batch_size)
```
dkurra, line 583: this is only true for distributed data parallel? other training strategies would fail this assumption

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
```diff
@@ -501,6 +517,488 @@ def _resolve_callbacks(
...
+
+    limit_train_batches = trainer_kwargs.get("limit_train_batches")
```
dkurra, line 585: why do we need to limit_train_batches?

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
```diff
@@ -501,6 +517,488 @@ def _resolve_callbacks(
...
+        if not (
+            isinstance(accumulate_grad_batches, int) and accumulate_grad_batches >= 1
+        ):
+            raise ValueError(
+                f"Invalid accumulate_grad_batches: {accumulate_grad_batches!r}. "
+                "Must be an int >= 1."
+            )
+        return min(max_steps * accumulate_grad_batches, steps_per_epoch)
+
+    return steps_per_epoch
```
dkurra, line 609: can we have max_steps configured instead of `steps_per_epoch`

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
```diff
@@ -501,6 +517,488 @@ def _resolve_callbacks(
...
+        profiler_config = {profiler_config: {}}
+    elif not isinstance(profiler_config, dict):
+        raise TypeError(
+            f"profiler_config must be a str, dict, or None, got {type(profiler_config)!r}"
+        )
+
+    # Exactly one profiler flavor must be selected.
+    selected = [
+        name
+        for name in ("simple", "advanced", "pytorch", "xla", "custom")
```
dkurra: cam we just support pytorch profiler for now? as MA OSS is pytorch heavy we can bring me more profilers as we expand to more platforms?

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
```diff
@@ -175,6 +187,29 @@ def __init__(
...
+        # lineage forces execution of the dataset's lazy transformations.
+        # Counted off trainer_param rather than the popped train_data, which
+        # asdict() has already deep-copied.
+        profiler_config = (train_loop_config.get("lightning_trainer_kwargs") or {}).get(
+            "profiler"
+        )
+        if profiler_config is not None:
+            try:
+                train_loop_config["train_dataset_num_rows"] = (
+                    trainer_param.train_data.count()
```
dkurra, line 204: trainer_param.train_data.count() is this pyspark or raydata construct?

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
sallycr: Good question — I traced where the split actually happens, and it isn't the Lightning strategy that guarantees it, it's Ray Train's data layer.

- The worker loop gets its data from `ray.train.get_dataset_shard("train")` (`_private/util.py:1092`), and this trainer never passes a custom `dataset_config` to `TorchTrainer`, so Ray Train's default `DataConfig(datasets_to_split="all")` applies (Ray 2.51.2). Every dataset is split evenly across workers before Lightning is involved.
- `_resolve_strategy` (`_private/util.py:288-315`) only ever constructs `RayDDPStrategy`, `RayDeepSpeedStrategy`, or `RayFSDPStrategy`. DeepSpeed ZeRO and FSDP shard the *model and optimizer state*, not the input stream — each worker still consumes its own `1/world_size` shard. So `dataset_num_rows / world_size` is correct for all three.

The one case where your concern would bite: `_resolve_strategy` also accepts a caller-supplied `Strategy` instance verbatim (`_private/util.py:302-303`). Under a strategy where several workers form one model replica and must see identical batches (tensor/pipeline parallel), the assumption breaks and this estimate would be low by the replication factor. That is out of reach today — Ray Train would hand those workers disjoint shards, which would be a training-correctness problem long before it was a profiler problem — so I don't think it warrants code, but it was undocumented. Pushed `ad181a57` recording the assumption and that caveat in the docstring.

Worth noting the blast radius either way: `steps_per_epoch` only shapes the profiler's sampling schedule. A bad estimate means a poorly placed trace window, never wrong training.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
sallycr: Because the profiler schedule is derived from `steps_per_epoch`, and `limit_train_batches` changes how many steps actually run. Ignore it and the derived window can land past the end of the epoch, producing an empty trace.

Concrete, using the real output of `_compute_default_schedule`: 100,000 rows, `world_size=4`, `batch_size=32` → 782 steps/epoch → default schedule `wait=260, warmup=2, active=3`. Now add `limit_train_batches=0.1` for a debug run: only 78 steps actually run. If we ignored the limit we would still emit `wait=260`, the profiler would never reach its `active` window, and you'd get nothing. Accounting for it yields 78 → `wait=26, warmup=2, active=3`, which fits. Same for the int form: `limit_train_batches=5` → 5 steps → `wait=1, warmup=2, active=2`.

It matters for user-supplied schedules too. `_validate_profiler_schedule` (`_private/util.py:660-706`) checks `(wait + warmup + active) * repeat + skip_first <= steps_per_epoch`. If `steps_per_epoch` ignored the limit, a schedule needing 200 steps would pass validation against 782 and then silently profile nothing out of the 78 steps that actually run. With the limit applied the user gets an upfront `UserInputError` naming exactly what to reduce.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
sallycr: Reading this as "why compute the row-count estimate at all when `max_steps` already bounds the run?" — I looked at whether the `.count()` could be skipped, and the answer is a qualified no, but you did surface a real gap next to it.

Why `max_steps` alone isn't a substitute: `max_steps` caps optimizer steps for the **entire run**, not per epoch, so it is only an upper bound on the first epoch's length. Dropping the `min()` and using `max_steps * accumulate_grad_batches` directly would overestimate whenever the epoch is shorter than `max_steps`. With `max_steps=1000` over a 50-step epoch you'd derive `wait=333` and capture nothing; today the `min()` gives 50 → `wait=16`. So the estimate has to stay the tighter of the two, which requires knowing the epoch length.

The `.count()` itself is already gated fairly tightly: it only runs when a profiler is configured (`lightning_trainer.py:197-212`), it's metadata-only for `read_parquet`-backed datasets, and it's wrapped in try/except so a failure degrades to the unbounded default rather than failing the run.

The genuine gap your comment points at is the other direction. When the row count is unavailable, `_compute_steps_per_epoch` returns `None` at `_private/util.py:558-559` — before the `max_steps` block — so a perfectly good `max_steps` bound gets discarded. The two consumers want different things there:

- **Validation** would be strictly better off with a `max_steps`-derived ceiling: it would catch a schedule that cannot fit in the whole run, where today it skips the check entirely.
- **The default schedule** is better off with `None`, which yields the conservative `wait=0, active=1` that always fires, rather than an overestimate that might not.

So the fix is to thread a `max_steps` ceiling into `_validate_profiler_schedule` only, leaving default-schedule behavior alone. It's a small change but it touches the `_resolve_profiler` → `_build_profiler` → `_build_pytorch_profiler` signatures, so I'd rather not fold it into this PR silently. Happy to add it here if you'd like, otherwise I'll file it as a follow-up.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
sallycr: Fair question, and it's a scope call rather than a defect, so I don't want to make it unilaterally — flagging the tradeoff and deferring to you.

For keeping all five: `simple`, `advanced`, `pytorch`, and `xla` are exactly Lightning's own built-in profilers, so the current surface is "whatever Lightning offers" rather than an invented set — which is easy to document and holds no surprises for someone arriving from Lightning. `SimpleProfiler` in particular is the cheap first thing to reach for (per-hook wall-clock, no trace files) and is often the right answer when someone just wants to know which stage is slow. The code is written and covered by tests in `test_profiler.py`, so the cost of keeping it is maintenance, not new work.

For narrowing to `pytorch`: it's the only flavor with real machinery behind it here — schedule derivation, schedule validation, trace handler, sink integration. The other four are thin passthroughs to a Lightning constructor. `xla` in particular is untestable in this repo's CI without TPU hardware, so its coverage is construction-level only, and `custom` (a dotted path to an arbitrary `Profiler` subclass) is the widest part of the API surface for the least demonstrated demand. Narrowing would cut the documented surface roughly in half.

My recommendation, for what it's worth: keep `simple` and `pytorch`, drop `xla` and `custom`. That keeps the cheap-first-look profiler and the deep one, removes the branch we can't exercise in CI and the open-ended extension point, and leaves room to re-add either behind a concrete request. But `custom` is also the escape hatch that makes narrowing safe — with it, a user who needs `xla` can still get there — so "keep everything" is defensible too.

I've left all five in place in this PR. Tell me which way you want it and I'll do it as a follow-up commit here.

python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
sallycr: Ray Data, no PySpark anywhere in this path. `LightningTrainerParam.train_data` is annotated `ray.data.Dataset` at `lightning_trainer.py:135`, and it's handed straight to `TorchTrainer(datasets={"train": ..., "val": ...})` in the same `__init__`, so `.count()` is `ray.data.Dataset.count()`.

The reason for the comment above the call is that `count()`'s cost is lineage-dependent. Ray's own docs for the method: for datasets that only read Parquet it reads file metadata and avoids reading the data, but "if this dataset consists of more than a read, or if the row count can't be determined from the metadata provided by the datasource, then this operation will trigger execution of the lazy transformations performed on this dataset." So a `read_parquet(...)` dataset is cheap, while `read_parquet(...).map_batches(...)` pays for the full transformation. That's why the call is gated behind "a profiler was actually requested" and wrapped in try/except.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
dkurra: ack!

dkurra (approved):

PR #1616, opened 2026-07-28
fix(storage/mysql): validate OrderBy.Field identifier before SQL interpolation

sallycr:
## Summary
- `buildOrderBySQL` interpolated a client-supplied `OrderBy.Field` directly into a backtick-quoted SQL identifier with no escaping of backticks in the value itself, allowing a crafted field value to close the identifier early and inject an arbitrary SQL expression.
- This is shared, single-code-path logic (`mysqlMetadataStorage.List`) used by every `List*` RPC that accepts `list_options_ext.order_by` (`ListProject`, `ListPipeline`, `ListModel`, `ListPipelineRun`, etc.) — the fix closes it for all of them, not just one endpoint.
- `indexPathToKeyMap` (an existing per-CRD allowlist mechanism) is not reliably populated at the call sites that reach this path, so the fix instead validates the resolved column name against a strict identifier-syntax allowlist (`^[A-Za-z0-9_]+$`) before it is ever placed inside backticks — the correct general approach for values used as identifiers rather than literals, since identifier positions cannot be `?`-parameterized.
- Confirmed against the real schema (`helm/michelangelo/files/schema/mysql-init-schema.sql`) that no legitimately-sortable column name uses any character outside that set, so this introduces no functional regression for existing sort behavior.
- `indexPathToKeyMap` is still consulted as defense-in-depth when non-nil (mirroring the existing enforcement pattern in `buildFieldCriterionSQL`), so a future caller that does populate real per-CRD maps gets the stricter allowlist automatically.
- `buildOrderBySQL`'s signature changes from `string` to `(string, error)`; its one call site now propagates the error as `codes.InvalidArgument`, consistent with how sibling functions already report unsupported/invalid fields.

## Test plan
- [x] `go test ./go/storage/mysql/...` — all existing cases pass unchanged; added cases cover the literal reported injection-payload shape, general SQL metacharacters, and both branches of `indexPathToKeyMap` enforcement (rejects unmapped fields when non-nil, allows mapped ones).
- [x] `bazel test //go/storage/mysql:go_default_test` passes.
- [x] `go build ./...` and `go vet ./storage/mysql/...` clean.
- [x] `gofmt -l` clean on changed files; `bazel run //:gazelle` produces no diff for this package (no new imports).
- [x] Real-MySQL integration smoke test (`TestIntegration_List_OrderBy_InjectionPayloadRejected`/`ValidFieldStillWorks`) against live, pre-existing sandbox data: malicious payload rejected with zero DB side effect, legitimate ordering unaffected.
- [x] Full local k3d sandbox validation: built the `apiserver` binary/image from this branch, deployed it in place of the running release image, confirmed clean boot and correct end-to-end behavior via a real client over the real wire, then restored the sandbox to its original state.

### DIFF
--- a/go/storage/mysql/mysql.go
+++ b/go/storage/mysql/mysql.go
@@ -290,7 +290,11 @@ func (m *mysqlMetadataStorage) List(ctx context.Context, typeMeta *metav1.TypeMe
 	// that case applies before we start writing the query.
 	orderBySQL := ""
 	if listOptionsExt != nil && len(listOptionsExt.OrderBy) > 0 {
-		orderBySQL = buildOrderBySQL(listOptionsExt.OrderBy)
+		var err error
+		orderBySQL, err = buildOrderBySQL(listOptionsExt.OrderBy, indexPathToKeyMap)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Selector-string path: when the structured proto path isn't being used,
@@ -985,6 +989,17 @@ func buildFieldSelectorSQL(fieldSelectorStr string, indexPathToKeyMap map[string
 	return queryStr.String(), params, nil
 }
 
+// validOrderByColumnName matches MySQL unquoted-identifier-safe characters
+// only (letters, digits, underscore). Applied to the CRD-prefix-stripped
+// remainder of a caller-supplied OrderBy.Field before it is interpolated as
+// a raw backtick-quoted identifier, since identifier positions can't be
+// parameterized like `?`-bound literal values can. This is the load-bearing
+// injection control — indexPathToKeyMap is always nil in both real call
+// sites (apiserver, controllermgr) as shipped, so it cannot be relied on as
+// the primary defense; see indexPathToKeyMap (when non-nil) for how it's
+// layered in as opportunistic defense-in-depth.
+var validOrderByColumnName = regexp.MustCompile(`^[A-Za-z0-9_]+$`)
+
 // buildOrderBySQL builds the ORDER BY clause from a list of OrderBy specs.
 //
 // Each OrderBy.Field is resolved as follows (matches internal buildOrderByQuery):
@@ -994,10 +1009,21 @@ func buildFieldSelectorSQL(fieldSelectorStr string, indexPathToKeyMap map[string
 //  3. Else if the (CRD-stripped) remainder equals orderByLabelField → emit
 //     orderByLabelColumn (which is a CTE column reference; the caller is
 //     responsible for prepending the matching WITH clause).
-//  4. Otherwise the remainder is used as the bare column name.
-func buildOrderBySQL(orderBy []*apipb.OrderBy) string {
+//  4. Else if indexPathToKeyMap is non-nil: use the mapped column if the
+//     remainder is a key in it, otherwise reject (codes.InvalidArgument) —
+//     mirrors buildFieldCriterionSQL's map-enforcement pattern. The map is
+//     always nil in production today (see validOrderByColumnName), so this
+//     branch is opportunistic defense-in-depth, not the primary control.
+//  5. Otherwise (indexPathToKeyMap nil) the remainder is used as the bare
+//     column name, gated by validOrderByColumnName before interpolation.
+//
+// Returns an error (codes.InvalidArgument) if a resolved, non-label column
+// name fails the identifier-syntax check — this is the fix for a SQL
+// injection where an unvalidated OrderBy.Field was interpolated directly
+// into a backtick-quoted identifier with no escaping.
+func buildOrderBySQL(orderBy []*apipb.OrderBy, indexPathToKeyMap map[string]string) (string, error) {
 	if len(orderBy) == 0 {
-		return ""
+		return "", nil
 	}
 	var clauses []string
 	for _, order := range orderBy {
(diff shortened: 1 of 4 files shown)


sallycr:
## Status update

All CI checks are now green:
- Go Test & Coverage (Bazel): **pass** (13m0s) — includes new `TestBuildOrderBySQL` cases covering the reported injection payload, a `DROP TABLE` payload, a blind-injection (`updatexml`) payload, and both branches of the `indexPathToKeyMap` enforcement.
- bazel-build, dirty-check, filter, python-build, Go Lint with TODO Tracking, license/cla, update-release-draft: **pass**
- Proto Breaking Changes / Helm Values Key Diff (advisory): skipped as expected — no proto/Helm changes in this PR

## Self-review notes

Traced `indexPathToKeyMap` back to its two real callers (`cmd/apiserver/providers.go`, `cmd/controllermgr/ingester_providers.go`) to confirm both pass `nil` — the new `validOrderByColumnName` allowlist is the actual load-bearing control in production today, not just defense-in-depth alongside the map. Also verified the allowlist (`^[A-Za-z0-9_]+$`) doesn't reject any real column name in `helm/michelangelo/files/schema/mysql-init-schema.sql`, so no functional regression for existing `ORDER BY` usage.

Not yet done: live sandbox validation via a real apiserver `List` RPC call with a malicious `order_by` field (tracked in the tracking issue/spec). Ready for maintainer review in the meantime.

sallycr:
## Sandbox validation: PASS

Ran a full local sandbox validation (k3d cluster with a real Michelangelo stack + real pre-existing pipeline_run data), in two parts:

**1. Real-MySQL integration smoke test** (new commit 2a472380a, added to this PR): `TestIntegration_List_OrderBy_InjectionPayloadRejected` and `TestIntegration_List_OrderBy_ValidFieldStillWorks` call the actual `List()` RPC path (not the sqlmock unit test) against real, pre-existing `pipeline_run` rows over a live MySQL connection:
- Malicious `order_by` payload → rejected with `InvalidArgument`, confirmed zero DB side effect (row count unchanged).
- Legitimate `order_by=create_time` (ASC/DESC) → real rows returned, correctly sorted, no regression.
- All 4 pre-existing integration tests in the same file still pass (6/6 total).

**2. Full sandbox deploy**: built the `apiserver` binary + Docker image directly from this branch, deployed it into the local k3d cluster in place of the running release image, confirmed clean boot (CRD sync, YARPC dispatcher, `[Fx] RUNNING`, no crash), and exercised it with a real client over the real wire (`ma pipeline_run get`) — correct real data returned end-to-end. Sandbox restored to its original image afterward.

One limitation, documented rather than glossed over: this service runs on YARPC over gRPC transport, so `grpcurl` reflection doesn't work against it and the `ma` CLI has no `--order-by` flag to pass an arbitrary payload through the wire protocol directly. The wire-level malicious-payload proof is therefore via #1 (real MySQL, real `List()` function, in-process) rather than a full YARPC-envelope injection — flagged in `specs/018/sandbox.md` as a follow-up if a stricter proof is ever required for disclosure purposes.

Full writeup: specs/018-fix-orderby-field-sql-injection/sandbox.md in the harness repo.

go/storage/mysql/mysql.go
```diff
@@ -994,10 +1009,21 @@ func buildFieldSelectorSQL(fieldSelectorStr string, indexPathToKeyMap map[string
...
+//     always nil in production today (see validOrderByColumnName), so this
+//     branch is opportunistic defense-in-depth, not the primary control.
+//  5. Otherwise (indexPathToKeyMap nil) the remainder is used as the bare
+//     column name, gated by validOrderByColumnName before interpolation.
+//
+// Returns an error (codes.InvalidArgument) if a resolved, non-label column
+// name fails the identifier-syntax check — this is the fix for a SQL
+// injection where an unvalidated OrderBy.Field was interpolated directly
+// into a backtick-quoted identifier with no escaping.
+func buildOrderBySQL(orderBy []*apipb.OrderBy, indexPathToKeyMap map[string]string) (string, error) {
```
donovanlopez, line 1016: nit: can we remove this comment? it seems too specific for this PR rather than being a general comment for this function.

donovanlopez (approved):
LGTM, just had a small nit about a comment.

go/storage/mysql/mysql.go
sallycr: Agreed, and took it a step further: had our OSS DX reviewer look at both this comment and the neighboring one on `validOrderByColumnName`, and beyond just being too PR-specific, they were also narrating the vulnerability and prior implementation flaw (naming it as a SQL injection fix, and calling out which check is load-bearing vs decorative in production) — detail that belongs in the PR description, not a permanent source comment. Trimmed both to general, timeless documentation of the resolution behavior in 21c3e2e0f.

sallycr:
## Final status: merged ✅

Merged as `58aae830c` (2026-07-28), approved by @donovanlopez.

Recap of the full path to merge:
- All required CI checks green (bazel-build, Go Test & Coverage, Go Lint with TODO Tracking, python-build, dirty-check, filter, license/cla); advisory checks (Proto Breaking Changes, Helm Values Key Diff) skipped as expected.
- Sandbox validation PASS: real-MySQL integration smoke test against live pre-existing data (malicious `order_by` rejected with zero DB side effect, legitimate ordering unaffected), plus a full local k3d deploy of the built `apiserver` binary confirming clean boot and correct end-to-end behavior over the real RPC wire.
- Doc-comment and PR-body disclosure cleanup per review feedback: trimmed vulnerability-narrative and internal call-site detail from permanent source comments and one PR body bullet, keeping only general/timeless documentation and appropriate public-disclosure content.
- `bazel build`/`bazel test` verified at final tip, plus a real gazelle-flagged missing `BUILD.bazel` dep caught and fixed.

Thanks for the review!

PR #1617, opened 2026-07-28
fix(ci): nightly.yml npm/helm jobs no longer hardcode stale base version 0.3.0

sallycr:
## Summary

`npm-nightly` and `helm-nightly` jobs in `nightly.yml` both hardcoded `NIGHTLY_VERSION="0.3.0-nightly.${DATE_TAG}"` instead of reading the current base version — a leftover from when this workflow was authored back when the project's version was `0.3.0` (task 007, PR #1379). It's never been updated since, so nightly npm packages and the nightly Helm chart have been reporting a stale `0.3.0-nightly.*` version through every subsequent minor release, even now at v0.6.0.

- `npm-nightly`: reads `BASE_VERSION` from `packages/core/package.json`'s own `version` field, matching `python-nightly`'s existing dynamic-read pattern (`pyproject.toml`). `packages/rpc/package.json` is versioned off the same `$NIGHTLY_VERSION` variable in this job, so it inherits the fix automatically — confirmed both packages are always kept in sync by `scripts/version-bump.sh` on every real release.
- `helm-nightly`: reads `BASE_VERSION` from `helm/michelangelo/Chart.yaml`'s own `version:` field before overwriting both `version:`/`appVersion:` with the nightly-suffixed value.
- `container-nightly`/`ui-nightly` were confirmed clean — they use bare `nightly-${date_tag}` container tags with no embedded base version, so this bug class doesn't apply to them.

Full investigation and root cause writeup: `specs/059/spec.md` in the harness repo.

## Test plan
- [x] `actionlint` clean on the modified `nightly.yml`
- [ ] `workflow_dispatch` test run (or log review) confirming `@michelangelo-ai/core`/`@michelangelo-ai/rpc` publish with the current base version (e.g. `0.6.x-nightly.YYYYMMDD`, not `0.3.0-nightly.*`)
- [ ] Confirm the pushed nightly Helm chart's `version`/`appVersion` reflect the current base version

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -237,8 +237,10 @@ jobs:
       - name: Set nightly chart version
         run: |
           DATE_TAG="${{ needs.check-changes.outputs.date_tag }}"
-          sed -i "s/^version:.*/version: 0.3.0-nightly.${DATE_TAG}/" helm/michelangelo/Chart.yaml
-          sed -i "s/^appVersion:.*/appVersion: \"0.3.0-nightly.${DATE_TAG}\"/" helm/michelangelo/Chart.yaml
+          BASE_VERSION=$(grep '^version:' helm/michelangelo/Chart.yaml | head -1 | sed 's/^version: *//')
+          NIGHTLY_VERSION="${BASE_VERSION}-nightly.${DATE_TAG}"
+          sed -i "s/^version:.*/version: ${NIGHTLY_VERSION}/" helm/michelangelo/Chart.yaml
+          sed -i "s/^appVersion:.*/appVersion: \"${NIGHTLY_VERSION}\"/" helm/michelangelo/Chart.yaml
 
       - name: Add Helm repos
         run: |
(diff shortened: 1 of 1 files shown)


paulzim (approved):
## Review: #1617 — fix(ci): nightly.yml npm/helm jobs no longer hardcode stale base version 0.3.0

**Verdict: Approve**

Straightforward, correct fix. Every claim in the description verified directly.

### The bug is real

Both `helm-nightly` and `npm-nightly` hardcoded `0.3.0-nightly.${DATE_TAG}` — confirmed by the diff. `python-nightly` (the adjacent job) already reads `BASE_VERSION` dynamically from `pyproject.toml`. The helm and npm jobs were simply never updated to match when the version advanced past 0.3.0.

### The fixes are correct

**`helm-nightly`:**
```bash
BASE_VERSION=$(grep '^version:' helm/michelangelo/Chart.yaml | head -1 | sed 's/^version: *//')
```
`Chart.yaml` has exactly one `^version:` line (line 10: `version: 0.6.0`) — confirmed. The `head -1` guard is safe. Pattern directly mirrors `python-nightly`'s `grep '^version' pyproject.toml | head -1 | sed '...'`.

**`npm-nightly`:**
```bash
BASE_VERSION=$(jq -r '.version' packages/core/package.json)
```
Cleaner than the grep+sed approach (JSON field, `jq` is the right tool). `packages/rpc/package.json` inherits the fix automatically because it's assigned the same `$NIGHTLY_VERSION` variable — and `scripts/version-bump.sh` explicitly keeps both packages versioned together on every real release (lines 8–9 and 97 of the script iterate over both files as a unit).

**`container-nightly` / `ui-nightly`:** confirmed clean — no `version`/`VERSION`/`0.3` references in either job's steps. They use bare `nightly-${date_tag}` container tags with no embedded base version, so the bug class doesn't apply.

### CI

`bazel-build`, `python-build`, `dirty-check`, `filter` all pass. `license/cla` pending (standard lag for this author on release PRs — has cleared on recent PRs). The two unchecked test-plan items (live `workflow_dispatch` run confirming the published version) are post-merge validation, not pre-merge blockers — the code path is simple enough that a `workflow_dispatch` dry-run is more a confirmation than a risk gate.

Safe to merge.

PR #1633, opened 2026-07-30
fix(pusher): use the shared api.v2 name generator for auto-generated model names

### LINKED ISSUE #1632: Auto-generated model names have weak collision resistance and no time ordering
## Summary

When a caller of `ModelPusherPlugin` does not set `config.model_name`, the plugin auto-generates
a name with `_generate_name()` in
`python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py`:

```python
def _generate_name() -> str:
    """Generate a unique model name with an 8-character hex UUID suffix."""
    return f"model-{uuid.uuid4().hex[:8]}"
```

This yields names of the form `model-a1b2c3d4` — 8 hex characters, i.e. a 2^32 name space with
no time component.

## Why this matters

**Collision resistance is weaker than it looks.** By the birthday bound, the probability of at
least one collision crosses 50% at roughly 65,000 auto-named models in a single registry. That is
not an unreasonable number for a registry accumulating models from automated training pipelines
over time, and unlike a random UUID collision it is not a "never happens in practice" number.

**Collisions are currently silent.** Re-registering under an existing model name is an upsert
today, so a collision does not error — it quietly replaces an unrelated model's registration.
That behavior is being addressed separately; until then, and even after, a weak name generator
makes the failure mode both more likely and harder to notice after the fact.

**Names are not time-ordered.** A purely random name carries no information about when the model
was pushed, so a list of auto-named models has no natural chronological ordering. Sorting a
registry listing, or eyeballing which of two auto-named models is newer, requires fetching and
comparing metadata.

## Proposed fix

Prefix the random suffix with a second-granularity UTC timestamp, similar to how many systems
generate sortable unique IDs:

```python
def _generate_name() -> str:
    ts = datetime.now(timezone.utc).strftime("%Y%m%d-%H%M%S")
    return f"model-{ts}-{uuid.uuid4().hex[:8]}"
```

Producing names like `model-20260721-114130-2d9c959d`. A collision now requires two calls landing
in the same second *and* drawing the same 8 hex characters, and the names sort chronologically as
plain strings.

This is a format change, not a semantic one:

- `config.model_name` remains the caller's escape hatch for a stable identity; this code path only
  fires when the caller passes `None` (and a warning is already logged when it does).
- No back-compat concern — the registry treats names as opaque strings and never parses them, so
  previously registered auto-generated names keep working unchanged.

Alongside the code change, two pieces of prose need updating: the `config.model_name` docstring
example (`'model-a1b2c3d4'`), and the `execute()` warning text, which says "auto-generating a
UUID-based name" and would no longer be accurate.

## Priority

P3 on its own — this is a quality-of-life and defense-in-depth improvement, not a correctness bug
in isolation. It is best landed alongside (or shortly after) the fix to the silent
overwrite-on-collision behavior, since the two compound.

sallycr:
## Summary

Closes #1632.

`_generate_name()` in the model pusher plugin produced names of the form `model-a1b2c3d4` — 8 hex
characters, a 2^32 space with no time component. By the birthday bound, collision probability
crosses 50% at roughly 65,000 auto-named models in a single registry, which is reachable for a
registry accumulating models from automated pipelines. Because re-registration under an existing
name is currently a silent upsert, such a collision would quietly replace an unrelated model.

A first pass at this fixed the format inline in the plugin. Further design review caught that
`michelangelo/api/v2/util.py` **already has** a `generate_random_name(prefix)` implementing
exactly the target `{prefix}-{YYYYMMDD}-{HHMMSS}-{8 hex}` format — but it was neither exported
from `michelangelo.api.v2` nor called anywhere in the tree. Fixing the plugin in isolation would
have left two copies of the same logic free to drift.

So instead: delete the private helper, call the shared one, and promote the shared one to public
API so the benefit lands for every `michelangelo.api.v2` consumer rather than just this plugin.
Names now look like `model-20260721-114130-2d9c959d`.

## Changes

- **`workflow/tasks/pusher/plugins/model_plugin.py`** — removed the private `_generate_name()`
  entirely; `execute()` now calls `generate_random_name("model")` directly at its single call
  site. Updated the `config.model_name` validation-error example to the new format, and reworded
  the warning from "auto-generating a UUID-based name" to "auto-generating a
  timestamp+UUID-based name", which is now accurate.
- **`api/v2/__init__.py`** — export `generate_random_name`, so
  `from michelangelo.api.v2 import generate_random_name` works. Previously only `APIClient` was
  exported, leaving the utility unreachable from the package's public surface.
- **`api/v2/util.py`** — added the missing `prefix: str -> str` type hints and documented the
  time-sortability property and an example name. No behavior change.
- **`api/v2/tests/test_util.py`** (new) — the utility had no tests. Covers format, uniqueness
  across successive calls, chronological string sort, prefix lowercasing, underscore-to-hyphen
  normalization, and the empty / over-128-character `RuntimeError` paths, including the
  128-character inclusive boundary.
- **`workflow/tasks/pusher/tests/plugins/test_model_plugin.py`** — the auto-naming test now
  asserts the full format via regex, plus a new test that the plugin delegates to the shared
  utility with the `"model"` prefix. The pre-existing assertion was pinned to the old
  8-character length and had to change either way.

## Implementation note

`model_plugin.py` imports from the `michelangelo.api.v2.util` submodule rather than the
`michelangelo.api.v2` package, because the package `__init__` imports `client`, which imports
`grpc`. Going through the submodule keeps grpc off the pusher plugin's import path. The package
export exists for external consumers, who are already paying that cost.

## Compatibility

Format change only, no semantics change:

- `config.model_name` remains the escape hatch for a stable identity; this path only fires when a
  caller passes `None`, and a warning is already logged when it does.
- The registry treats names as opaque strings and never parses them, so previously registered
  auto-generated names keep working unchanged.
- `generate_random_name` itself is unchanged behaviorally — it moves from unreachable to public.

## Test plan

```
cd python
.venv/bin/python -m pytest michelangelo/api/v2/tests/ michelangelo/workflow/tasks/pusher/tests/ -q
# 150 passed
.venv/bin/python -m pytest michelangelo/api/ michelangelo/workflow/ -q
# 565 passed  (wider sweep, to confirm the new __init__ export introduces no import cycle)
.venv/bin/python -m ruff format --check <touched files>   # clean
.venv/bin/python -m ruff check <touched files>            # clean
```

## Notes for reviewers

This is P3 on its own — a quality-of-life and defense-in-depth improvement rather than a
correctness bug in isolation. It pairs with the separate fix to the silent overwrite-on-collision
behavior in the registry client; the two compound, and landing them together gives the most
benefit.

Worth a look: `generate_random_name` raises `RuntimeError` (not `ValueError`) on a bad prefix.
That is pre-existing behavior now becoming public API, so if it should be `ValueError`, this is
the cheapest moment to change it.

### DIFF
--- a/python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py
+++ b/python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py
@@ -93,6 +93,7 @@ import uuid
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypedDict
 
+from michelangelo.api.v2.util import generate_random_name
 from michelangelo.workflow.schema.exceptions import ConfigurationError
 from michelangelo.workflow.tasks.pusher.plugins.base import PusherPluginBase
 
@@ -346,7 +347,8 @@ class ModelPusherPlugin(PusherPluginBase):
         if config.model_name == "":
             raise ConfigurationError(
                 "ModelPusherPlugin: model_name must be a non-empty string or None "
-                "(None auto-generates a unique name, e.g. 'model-a1b2c3d4')."
+                "(None auto-generates a unique name, e.g. "
+                "'model-20260721-114130-2d9c959d')."
             )
         # Resolve the effective registry list at init time.
         has_injected = registry_client is not None
@@ -405,10 +407,11 @@ class ModelPusherPlugin(PusherPluginBase):
         """
         if self._config.model_name is None:
             _logger.warning(
-                "No model_name set in config — auto-generating a UUID-based name. "
+                "No model_name set in config — auto-generating a "
+                "timestamp+UUID-based name. "
                 "Set config.model_name explicitly for reproducible model identities."
             )
-            model_name = _generate_name()
+            model_name = generate_random_name("model")
         else:
             model_name = self._config.model_name
         push_id = uuid.uuid4().hex[:16]
@@ -515,8 +518,3 @@ class ModelPusherPlugin(PusherPluginBase):
         if self._config.run_id is not None:
             result["run_id"] = self._config.run_id  # run_id wins on collision
         return result
-
-
-def _generate_name() -> str:
-    """Generate a unique model name with an 8-character hex UUID suffix."""
-    return f"model-{uuid.uuid4().hex[:8]}"
--- a/python/michelangelo/api/v2/__init__.py
+++ b/python/michelangelo/api/v2/__init__.py
@@ -7,6 +7,14 @@ Example:
     >>> from michelangelo.api.v2 import APIClient
     >>> APIClient.set_caller('my-application')
     >>> model = APIClient.ModelService.get_model(namespace='default', name='my-model')
+
+The package also exposes ``generate_random_name`` for callers that need a unique,
+time-sortable object name when they have no stable identity of their own:
+
+    >>> from michelangelo.api.v2 import generate_random_name
+    >>> generate_random_name('model')  # doctest: +SKIP
+    'model-20260721-114130-2d9c959d'
 """
 
 from .client import APIClient as APIClient
+from .util import generate_random_name as generate_random_name
--- a/python/michelangelo/api/v2/tests/test_util.py
+++ b/python/michelangelo/api/v2/tests/test_util.py
@@ -1,0 +1,66 @@
+"""Tests for michelangelo.api.v2.util — generated object names."""
+
+from __future__ import annotations
+
+import re
+from unittest import TestCase
+
+from michelangelo.api.v2.util import generate_random_name
+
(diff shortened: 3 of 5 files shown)


python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py
```diff
@@ -518,5 +521,15 @@ def _build_metadata(self) -> dict[str, Any]:
 
 
 def _generate_name() -> str:
```
kenns29: can we have the exact format in internal (refer to `uber/ai/michelangelo/sdk/api_client/v2beta1/utils/crd.py`) to make import back easier. Also, internally we have a tool in APIClient to generate the randome

kenns29 (approved):

PR #1634, opened 2026-07-30
fix(registry): create a revision instead of overwriting on re-registration

### LINKED ISSUE #1631: Model registration silently overwrites prior artifact URIs instead of creating a revision
## Summary

`APIRegistryClient.register_model()` silently overwrites a model's previous artifact URIs when
called again under an existing name, instead of recording a new revision. Prior pushes become
unrecoverable — there is no way to retrieve or audit what was registered under that name before.

## Current behavior

`python/michelangelo/lib/model_manager/registry/api_client.py`

`register_model()` first attempts `create_model`. On `ALREADY_EXISTS` it fetches the existing
model, copies over `metadata.resourceVersion`, and calls `update_model` — but the proto it sends
is the *freshly built* one from `_build_model_proto()`, which always starts
`spec.model_artifact_uri` and `spec.deployable_artifact_uri` empty and appends only the current
call's URI. Two consequences:

1. **The previous artifact URIs are dropped.** After a second registration under the same name,
   the registry holds only the second push's URI. The first artifact is still in object storage
   but is no longer discoverable through the registry.
2. **`spec.revision_id` is never set by the client.** The outgoing proto leaves it at its
   zero-value default, so the client never asks for the next revision — whatever version comes
   back is whatever the server happens to echo.

There is also a latent read bug in `_to_registered_model()`: it extracts
`model.spec.model_artifact_uri[0]` (and the same for `deployable_artifact_uri`). While the list
is always single-element this is harmless, but it means "first URI ever registered", not "the
artifact for this registration" — which becomes wrong as soon as the list can hold history.

## Why it matters

This is silent data loss, not a crash. A caller re-pushing to an existing model name gets a
successful `RegisteredModel` back and no indication that the prior registration's artifact
pointer was discarded.

It also puts the two shipped `ModelRegistryClient` implementations in disagreement about the
ABC's contract: `InMemoryRegistryClient` — the reference implementation its own docstring invites
new client authors to copy — appends a new `RegisteredModel` per call and keeps the full list,
while `APIRegistryClient` replaces in place. The ABC docstring for `register_model()` is silent
on the already-exists case, so neither behavior is currently "wrong" per the written contract.

Version history on re-registration is the norm for model registries generally (MLflow, for
instance, creates a new model version per registration rather than mutating the prior one), so
the append behavior is also the least-surprising choice for anyone coming from another registry.

The auto-generated model-name path makes this sharper: when a caller omits an explicit model
name, a name is generated for them, so an unlucky collision between two unrelated models would
today destroy one of them silently rather than producing a second revision.

## Proposed fix

In `register_model()`'s `ALREADY_EXISTS` branch:

- Set `model.spec.revision_id = existing.spec.revision_id + 1` on the outgoing proto.
- Thread the fetched `existing` model into `_build_model_proto()` so that
  `spec.model_artifact_uri` / `spec.deployable_artifact_uri` are built as *existing entries plus
  the new URI appended*, rather than replaced. The create path is unchanged.

In `_to_registered_model()`:

- Read `[-1]` instead of `[0]` for both artifact URIs, so `RegisteredModel.artifact_uri` means
  "the artifact for this registration" rather than "the oldest one under this name".

In docstrings:

- Document the already-exists contract on the `ModelRegistryClient.register_model()` ABC: a
  collision creates a new revision and prior artifact URIs remain associated with the model.
- Correct `APIRegistryClient`'s "Create vs. update" class-docstring note, which currently
  describes the collision path as an in-place update.

## Explicitly out of scope

Per-revision retrieval. `get_model(version=...)` will still return only the latest revision (it
logs a warning today and ignores `version`) — supporting real per-revision lookup needs the
underlying service surface to expose it, which is a larger change. The ABC docstring wording
above should state this limitation rather than imply it is fixed. Worth a separate issue.

## Priority

P1 — silent data loss on an already-public, already-migrated code path.

sallycr:
Closes #1631.

## Problem

`APIRegistryClient.register_model()` silently discarded a model's previous artifact URIs when
called again under an existing name. On `ALREADY_EXISTS` it rebuilt the outgoing proto from
scratch via `_build_model_proto()` — which always starts `spec.model_artifact_uri` /
`spec.deployable_artifact_uri` empty — and sent that as the update. It also never set
`spec.revision_id`, so the client never asked for the next revision.

Net effect: a re-push under an existing name returned a successful `RegisteredModel` while the
prior registration's artifact pointer became unrecoverable through the registry.

## Changes

- `register_model()`: on `ALREADY_EXISTS`, rebuild the proto with the fetched `existing` model
  threaded in, set `spec.revision_id = existing.spec.revision_id + 1`, and write that back. The
  create path is untouched. The `FAILED_PRECONDITION` retry logic is unchanged.
- `_build_model_proto()`: takes an optional `existing` argument. When given, both artifact URI
  lists start from the existing model's entries and the new URIs are appended. Existing
  deployable URIs are preserved even when this push doesn't supply one, so a raw-only re-push
  doesn't wipe the serving-bundle history.
- `_to_registered_model()`: reads `[-1]` rather than `[0]` from both URI lists. Now that the
  lists accumulate history, index 0 is the oldest entry, not this registration's.
- Docstrings: the already-exists contract is stated on the `ModelRegistryClient.register_model()`
  ABC, and `APIRegistryClient`'s class-docstring note no longer describes the collision path as
  an in-place update.

## Not in scope

Per-revision retrieval. `get_model(version=...)` still returns only the latest revision — that
needs the underlying service surface to expose revision lookup. Both docstrings state this
limitation explicitly rather than implying it is fixed.

## Tests

Four cases added to `api_client_test.py`, backed by a small stateful `ModelService` double that
implements real create/get/update semantics (it echoes the client's proto on update, so
assertions on the returned `RegisteredModel` also assert what the client sent):

- re-registering the same name yields `version="2"` on the second call
- the second call's `artifact_uri` / `deployable_artifact_uri` are the second push's, not the
  first's — regression test for the `[0]` → `[-1]` fix
- both URI lists on the stored model retain the first push's entries after the second
- `_build_model_proto()` with an `existing` model extends both lists rather than replacing them

`python -m pytest michelangelo/lib/model_manager/registry/tests/` — 50 passed.
`michelangelo/workflow/tasks/pusher/tests/` — 119 passed.
`ruff format --check` and `ruff check` clean on all touched files.

### DIFF
--- a/python/michelangelo/lib/model_manager/registry/api_client.py
+++ b/python/michelangelo/lib/model_manager/registry/api_client.py
@@ -94,11 +94,17 @@ class APIRegistryClient(ModelRegistryClient):
     ``rpc-encoding``) on every call via ``DefaultHeaderProvider``, eliminating
     the need for a manual interceptor.
 
-    **Create vs. update (with retry):** :meth:`register_model` attempts
-    ``create_model`` first. On ``ALREADY_EXISTS`` it fetches ``resourceVersion``
-    via ``get_model`` and retries as ``update_model``. On ``FAILED_PRECONDITION``
-    (concurrent write) the whole sequence retries up to
-    :data:`_MAX_REGISTER_RETRIES` times.
+    **Create vs. new revision (with retry):** :meth:`register_model` attempts
+    ``create_model`` first. On ``ALREADY_EXISTS`` it fetches the current model
+    via ``get_model`` and writes back a new revision — ``spec.revision_id`` is
+    incremented and the new artifact URIs are *appended* to the existing ones,
+    so prior pushes stay associated with the model rather than being
+    overwritten. On ``FAILED_PRECONDITION`` (concurrent write) the whole
+    sequence retries up to :data:`_MAX_REGISTER_RETRIES` times.
+
+    Note that :meth:`get_model` still only resolves the latest revision;
+    earlier artifact URIs are retained on the model but are not individually
+    addressable through this client.
 
     **registry_uri format:** ``models:/{namespace}/{name}/{version}`` —
     Michelangelo's three-segment format, not the two-segment MLflow format.
@@ -170,7 +176,12 @@ class APIRegistryClient(ModelRegistryClient):
         labels: Mapping[str, str] | None = None,
         metadata: Mapping[str, Any] | None = None,
     ) -> RegisteredModel:
-        """Register a model via ``create_model``, falling back to ``update_model``.
+        """Register a model via ``create_model``, falling back to a new revision.
+
+        If a model with this name already exists, a new revision is written:
+        ``spec.revision_id`` is incremented and ``artifact_uri`` /
+        ``deployable_artifact_uri`` are appended to the URIs already recorded
+        on the model, so prior pushes remain associated with it.
 
         Args:
             name: Model name. Used as ``model.metadata.name``.
@@ -184,6 +195,7 @@ class APIRegistryClient(ModelRegistryClient):
 
         Returns:
             :class:`~michelangelo.lib.model_manager.registry.client.RegisteredModel`
+            for the newly created revision.
 
         Raises:
             grpc.RpcError: On gRPC errors other than ``ALREADY_EXISTS`` /
@@ -208,14 +220,24 @@ class APIRegistryClient(ModelRegistryClient):
                     raise
 
             _logger.warning(
-                "Model '%s' already exists — fetching resourceVersion and "
-                "updating (attempt %d/%d).",
+                "Model '%s' already exists — fetching it and writing a new "
+                "revision (attempt %d/%d).",
                 name,
                 attempt,
                 _MAX_REGISTER_RETRIES,
             )
             existing = self._svc.get_model(self._namespace, name)
+            model = self._build_model_proto(
+                name=name,
+                artifact_uri=artifact_uri,
+                deployable_artifact_uri=deployable_artifact_uri,
+                description=description,
+                labels=labels,
+                metadata=metadata,
+                existing=existing,
+            )
             model.metadata.resourceVersion = existing.metadata.resourceVersion
+            model.spec.revision_id = existing.spec.revision_id + 1
             try:
                 updated = self._svc.update_model(model)
                 return self._to_registered_model(updated)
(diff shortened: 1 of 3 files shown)


python/michelangelo/lib/model_manager/registry/api_client.py
```diff
@@ -318,11 +346,12 @@ def _to_registered_model(self, model: model_pb2.Model) -> RegisteredModel:
         name = model.metadata.name
         namespace = model.metadata.namespace or self._namespace
         version = str(model.spec.revision_id)
+        # The URI lists accumulate one entry per revision; the last is this one's.
         artifact_uri = (
-            model.spec.model_artifact_uri[0] if model.spec.model_artifact_uri else None
+            model.spec.model_artifact_uri[-1] if model.spec.model_artifact_uri else None
```
kenns29, line 351: the model artifact URI and deployable artifact URI are not for version controlls. Model version control for incremental training are done with the revision_id field in model CRD, and each revision_id corresponds to  a Revision CRD and each revision has its own set of `model_artifact_uri` and `deployable_artifact_uri`. The decision to make both of these fields a list is not due to needs of version control but leave door open to allow multiple artifacts for the same model, e.g. same model but one converted to onnx one converted to tensorRT, etc

kenns29 (changes requested):

sallycr:
Closing as not needed after review feedback.

`model_artifact_uri`/`deployable_artifact_uri` are not a version-history mechanism — per the proto's own documentation and confirmed in review, they hold multiple simultaneous artifact variants of the *same* model version (e.g. an ONNX build alongside a TensorRT build), not a log of URIs across separate pushes. Real version history is a `Revision` CRD/service concept that this client doesn't (and, as scoped, shouldn't) touch.

This PR's approach — appending each re-registration's URI onto the existing list and incrementing `revision_id` — conflates those two things. Worse, it actively breaks the legitimate multi-format case: two separate `register_model()` calls adding an ONNX build and then a TensorRT build to the same version would become indistinguishable from two unrelated pushes accumulating history.

The original premise (re-registration silently loses reachable history) doesn't hold given what these fields are actually for. Real revision support would need `Revision` CRD/service integration in OSS, which is a separate, larger, currently-unscoped feature — not a fix to `register_model()`'s existing upsert behavior. Thanks for the catch.

PR #1628, opened 2026-07-30
docs: add MLOps glossary page

zhoward-1:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**
Adds a new `docs/getting-started/glossary.md` — an alphabetical MLOps glossary (A–W) with a concept mapping table (MLflow/Kubeflow/Ray/Airflow → Michelangelo AI equivalents). Also fixes pre-existing `sidebar_position` collisions in the getting-started section (roadmap.md and sandbox-setup.md were both at position 3).

Final sidebar order:
1. Overview
2. Core Concepts & Key Terms
3. Sandbox Setup
4. Glossary (new)
5. Python IDE Setup
6. FAQ
7. Roadmap
8. Sandbox Ports & Endpoints
9. Support

**Why?**
Users migrating from other ML platforms need a quick reference for Michelangelo AI terminology. Resolves sidebar collision flagged in review.

**How did you test it?**
Verified sidebar positions are unique across all getting-started pages.

**Potential risks**
None.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
New page: `docs/getting-started/glossary.md`. Updated `docs/getting-started/index.md` with glossary link. Sidebar positions renumbered for getting-started section.

PR #1641, opened 2026-07-31
chore: add ma-sandbox-test-plan Claude Code skill

craigmarker:
## Summary

Adds a Claude Code skill for building, testing, and verifying a sandbox change across Go, JS, and Python — producing evidence that a feature actually works, not just that it compiles.

Evidence guidance (`references/evidence-quality.md`) prioritizes positive evidence — showing the feature working through phase transitions, UI screenshots, and kubectl state — over negative evidence like "no errors in the logs," which can't distinguish a working feature from a silent no-op. Build/lint/typecheck results are treated as gates, not evidence, since CI already proves them.

## Test plan

**Full run of skill suite**

https://github.com/user-attachments/assets/a9c6f712-c76e-47bd-bcbb-f8b07910dbd8

### DIFF
--- a/.claude/skills/ma-design-interview/SKILL.md
+++ b/.claude/skills/ma-design-interview/SKILL.md
@@ -159,3 +159,11 @@ The chosen alternative becomes the input to Phase 3.
 
 Generate the highest-quality implementation possible. The goal is a working prototype on a
 branch that either ships as-is (self-service) or gets refined by a frontend engineer.
+
+---
+
+## Phase 4: Verification
+
+Once generation is done, run `/ma-sandbox-test-plan` against the change to get real evidence
+it works — build/lint/test gates plus live integration scenarios in the sandbox — rather than
+declaring the interview complete on the strength of the generated code alone.
--- a/.claude/skills/ma-sandbox-references/evidence-quality.md
+++ b/.claude/skills/ma-sandbox-references/evidence-quality.md
@@ -1,0 +1,57 @@
+# Test Evidence
+
+A test plan's value is evidence that CI can't provide: the feature working in a running environment. A reviewer should be able to reconstruct what you actually observed — not just trust your conclusion.
+
+## Example
+
+```markdown
+## Test plan
+
+### Integration on k3d sandbox
+Image: `michelangelo-controllermgr:local` (branch `craig/revision-manager`, sha `d1046e10`)
+Log level: `debug`
+
+#### 1. PipelineRun reconciles through full lifecycle
+Applied `test-revision-happy` → watched phase transitions:
+
+​```
+{"level":"info","msg":"transitioned phase","name":"test-revision-happy","from":"Pending","to":"Running"}
+{"level":"info","msg":"transitioned phase","name":"test-revision-happy","from":"Running","to":"Succeeded"}
+​```
+
+Verified terminal state:
+​```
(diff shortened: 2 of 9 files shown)


paulzim (approved):
## Review: PR #1641 — chore: add ma-sandbox-test-plan Claude Code skill

**Verdict: Approve (needs rebase + one minor fix)**

---

### Stale merge base

The branch was cut before 8 commits landed on main (including #1297, #1628, #1646, #1647, #1660, #1664, #1665). The apparent large deletions in the diff against main are entirely stale-merge-base artifacts — the net change from the branch point is 8 new `.claude/` files, 615 insertions, 0 deletions. A rebase should be conflict-free since none of those commits touch `.claude/`.

### What's being added

Eight files, all additions:

| File | Purpose |
|------|---------|
| `.claude/skills/ma-sandbox-test-plan/SKILL.md` | Main skill: 5-stage pipeline (detect stack → build/lint → unit test → integration setup → report) |
| `.claude/skills/ma-sandbox-references/evidence-quality.md` | "Positive evidence beats negative" guidance with a concrete worked example |
| `.claude/skills/ma-sandbox-references/routes.md` | Phase/entity keyword tables and URL resolution algorithm for sandbox UI |
| `.claude/skills/ma-sandbox-references/sandbox-data.md` | How to inspect, seed, and create sandbox state; port table |
| `.claude/skills/ma-sandbox-scripts/capture_service_logs.sh` | Filters kubectl logs to signal lines; strips YARPC tracing noise by default |
| `.claude/skills/ma-sandbox-scripts/check-backend.sh` | Detects whether local Go changes exist and if the local image is running |
| `.claude/skills/ma-sandbox-scripts/find-vite-port.sh` | Finds an existing Vite server for the current branch worktree, or starts one |
| `.claude/skills/ma-sandbox-scripts/set_log_level.sh` | Patches the service ConfigMap and restarts the deployment to apply a new zap log level |

### Things that work well

- The multi-stack gating is correct — a Go build failure stops Go integration but lets JS and Python pipelines run independently.
- `find-vite-port.sh` is careful to verify the found server is actually serving the right worktree and branch before trusting it. The fallback-and-detect logic handles the case where Yarn forks deeply and the listening PID isn't a direct child.
- `set_log_level.sh` forces `encoding: json` alongside the level change, which is a necessary precondition for `capture_service_logs.sh` — good that this is done automatically rather than leaving it to the caller.
- The YARPC noise exclusion in `capture_service_logs.sh` (`Handled inbound request.` / `Made outbound call.`) is a real QoL improvement for debug-level log capture.
- `evidence-quality.md`'s "positive evidence beats negative" framing is the right principle: "no errors in logs" is indistinguishable from a silent no-op.

### Minor fix needed: `check-backend.sh` doesn't cover worker

The script only checks `go/cmd/apiserver/` and `go/cmd/controllermgr/`. If there are changes under `go/cmd/worker/`, the script outputs `clean` and the skill skips Go integration entirely for those changes — which is wrong. Worker is one of the three main services and should get the same treatment:

```bash
WORKER_CHANGED=$(echo "$CHANGED_FILES" | grep -E '^go/cmd/worker/' | head -1 || true)
```

The service priority order (which one to report when multiple have changes) should probably be: controllermgr → worker → apiserver, or just report all changed services instead of the first one.

### CI

All checks pass: bazel-build, dirty-check, python-build, filter, update-release-draft. CLA signed.

---

Great addition — the skill fills a real gap between "CI passed" and "I actually verified the feature works in the sandbox." After the rebase and the worker fix, this is ready.

Issue #1173, opened 2026-05-10
mysql: merge column-side res_version / update_time onto runtime objects in List


badcount:
Here’s a cleaner and more structured version:

---

### Issue Summary

`go/storage/mysql/mysql.go` → `executeListQueryAndProcessResult`

The current implementation SELECTs all 8 columns:

* `uid`
* `group_ver`
* `namespace`
* `name`
* `res_version`
* `create_time`
* `update_time`
* `proto`

However, only the `proto` blob is actually used to hydrate the returned `runtime.Object`.
All other columns are scanned but effectively discarded.

---

### Why This Is a Problem

The internal implementation (`storage/pkg/mysql/mysql.go` → `executeListQueryAndProcessResult`) routes through `objManager.ToRuntimeObject`, which merges column-level metadata back into the runtime object.

This behavior is important because several metadata fields may diverge from the serialized proto blob over time.

#### 1. `res_version` Drift

`res_version` and `update_time` can be updated independently from the serialized proto.

Example:

* `directUpdate` paths may increment `res_version`
* but skip full proto re-serialization

Without merging the column-side values:

* callers receive stale `resourceVersion`
* optimistic concurrency checks can silently fail

---

#### 2. Labels / Annotations Drift

The stored proto blob is typically written only during insert/full-update operations.

Subsequent metadata-only updates:

* labels
* annotations

may update dedicated DB columns without refreshing the serialized proto.

As a result:

* `metadata.labels`
* `metadata.annotations`

inside the hydrated object can become stale.

---

### Proposed Fix

After:

```go
proto.Unmarshal(protoBytes, protoMsg)
```

perform a metadata reconciliation step using a `metav1.Object` accessor.

The reconciliation should:

1. Overwrite `ResourceVersion`

   * using the DB column value

2. Set/update the `update_time` annotation

3. Re-hydrate:

   * labels
   * annotations

from the `_labels` / `_annotations` tables (or equivalent metadata sources).

---

### Internal vs OSS Gap

Internally, this reconciliation is handled through:

```go
ObjectManager.ToRuntimeObject(...)
```

OSS currently does not yet have an equivalent abstraction.

This work likely needs one of:

1. An OSS `ObjectManager` equivalent
2. A smaller direct metadata merge layer inside:

   * `executeListQueryAndProcessResult`

---

### Affected File

```text
go/storage/mysql/mysql.go
```

Function:

```go
executeListQueryAndProcessResult
```

PR #1645, opened 2026-07-31
fix(storage/mysql): implement directUpdate for the generic Update RPC

### LINKED ISSUE #1622: Generic Update RPC always fails on MySQL metadata storage backend (direct update not yet implemented)
## Summary

Any resource's generic `Update` RPC (not specific to models — this is the shared metadata-handler path used by all resource types) always fails with `UNIMPLEMENTED` when the apiserver is configured with the MySQL metadata storage backend.

## Where

- `go/api/handler/metadata_handler.go` — `MetadataHandlerImpl.Update()` unconditionally calls `handleUpdate(ctx, obj, m.storage, true, nil, m.blobStorage)`, i.e. always requests a "direct" update.
- `go/storage/mysql/mysql.go` — `directUpdate()` is a stub:
  ```go
  func (m *mysqlMetadataStorage) directUpdate(ctx context.Context, tx *sql.Tx, tableName string, metaObj metav1.Object, object runtime.Object) error {
      return status.Errorf(codes.Unimplemented, "direct update not yet implemented")
  }
  ```
  It returns `Unimplemented` for every call, regardless of input.

By contrast, the controller/ingester reconciliation loop (`go/components/ingester/controller.go`) calls `Upsert(ctx, object, false, indexedFields)` — the "full upsert" path — which works correctly. Only the client-facing generic Update RPC path (`direct=true`) is dead.

## Repro

Run any example pipeline twice against the same model registry name (e.g. the `california_housing` pytorch example, which registers a model under the fixed name `california-housing-lightning`). The second run's push/registration step fails:

```
push_step: register_model('california-housing-lightning'): model already exists — fetching resourceVersion and updating (attempt 1/3)
ERROR: Plugin 'model_plugin' failed for artifact 'model': <_InactiveRpcError ...
  status = StatusCode.UNIMPLEMENTED
  details = "failed to update API object. namespace: <ns>, name: california-housing-lightning: rpc error: code = Unimplemented desc = direct update not yet implemented"
>
```

This causes the whole pipeline run to fail (Spark driver exits 1), even though `preprocess` and `train` completed successfully — so a run's `/steps` page shows a genuine failure, not a display bug.

## Impact

Any client (Python SDK, UI, CLI) calling update on an existing resource via the standard apiserver API — not just models — hits this when the MySQL storage backend is in use. In practice this means re-registering/updating any already-existing resource by name is broken end to end.

## Suggested fix

Implement `directUpdate()` with real optimistic-concurrency semantics (check `resourceVersion`/`res_version` against the stored row, update labels/annotations/resource version, return a conflict error on mismatch), following the same pattern as `fullUpsert`'s indexed-field handling, rather than leaving it as an unconditional stub.

codope:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Implements `mysqlMetadataStorage.directUpdate`, which was an unconditional `codes.Unimplemented` stub, so the generic `Update` RPC failed for every resource kind whenever the apiserver runs with the MySQL metadata storage backend.

It now follows the `MetadataStorage.Upsert` contract: labels, annotations and the resource version are updated, guarded by an optimistic-concurrency check on `resourceVersion`. A mismatch returns `FailedPrecondition`, matching the `StatusReasonConflict` mapping in `K8sStatusReasonToGrpcError`, which is the code `api_client.py` already catches to drive its retry loop. An empty incoming `resourceVersion` is treated as "no precondition", per Kubernetes update semantics.

Two things beyond the stub itself turned out to be required:

1. **`Upsert` never committed on the direct branch.** It did `return m.directUpdate(...)` under a `defer tx.Rollback()`, so even a correct implementation would have been rolled back.

2. **The `proto`/`json` blobs are rewritten, not just the `res_version` column.** `GetByName` and `List` reconstruct objects from the proto blob alone (the existing TODO #1173 ), so a column-only bump leaves the next read reporting a stale version, and since the client re-reads before each retry, its *next* update conflicts forever. Without this the bug returns on the third registration as an exhausted-retry error instead of `UNIMPLEMENTED`.

Two smaller decisions:

- The row is located by `namespace`+`name` rather than `uid`, and child rows are keyed by the `uid` read back from that row. Callers of the generic Update RPC aren't required to echo a UID (`register_model` doesn't), so `metadataStoragePrimaryKey` is usually empty here and would orphan every label/annotation under `obj_uid = ''`.
- Server-managed labels/annotations absent from the incoming object are preserved. Callers build a fresh object rather than doing read-modify-write, and `setUpdateTimestamp` only sets `SpecUpdateTimestampLabel` when the spec changed, so a strict replace would delete it, along with the immutability marker and the migration primary key.

The second commit copies the assigned resource version back onto the caller's object in `apiHandler.Update`. The object is deep-copied before being handed to the metadata handler, so without this the new version lands on a discarded copy and the client sees the version it sent. It's separated out because it's inert without the first commit.

**Why?**

Fixes #1622. Re-registering any already-existing resource by name was broken, which failed the whole pipeline run whenever a `register_model` step ran a second time even though `preprocess` and `train` had succeeded.

**How did you test it?**

Verified before/after against a local MySQL with `scripts/ingester/ingester_schema.sql` loaded: on `main` the second registration returns `Unimplemented` and `res_version` stays at 1; with the fix it succeeds.

- 8 unit tests (no database) covering the resourceVersion precondition -- empty, match, mismatch, non-numeric, negative, overflow, zero-padded -- and the metadata merge, including managed-key preservation and that the spec is untouched.
- 12 `TestIntegration_DirectUpdate_*` tests against live MySQL: happy path, the read-modify-write cycle repeated three times (the #1622 repro), conflict, invalid and empty resource versions, missing and soft-deleted rows, child rows keyed by the stored uid with no orphans under `obj_uid = ''`, managed-annotation preservation, and a corrupt (NULL) proto blob being rejected rather than overwritten.
- `bazel test //go/... //proto/... --build_tests_only`: all tests pass.

Note the integration tests skip when MySQL is unreachable, matching the 6 existing tests in that file, so they do not currently run in CI. I've asked about that in [this comment](https://github.com/michelangelo-ai/michelangelo/issues/1622#issuecomment-5129683888). Happy to drop them from this PR if you'd rather they land alongside a CI job. One useful data point since: the Bazel test sandbox *can* reach a MySQL on the host, so a GitHub Actions service container would work with `bazel test`, needing only `--test_env` for the connection variables.

**Potential risks**

The changed path returned an error unconditionally before this PR, so no working behaviour can regress through it. The three ingester callers use `direct=false` and are untouched; `metadata_handler.go` is the only `direct=true` caller. No dependency, proto, or exported-signature changes.

Three things to consider during review:

- **Spec changes are silently ignored on this path.** `register_model` sends a full `Model` including its spec, and the RPC will now return success without persisting it. I believe that's correct, this path is only reached once the object has been evicted from etcd as immutable, and there's a test covering the behaviour, but it is a quieter failure mode than `UNIMPLEMENTED`. This is the open question from the issue comment linked above; happy to change direction.
- `(namespace, name)` is a non-unique `KEY`, so `GetByName`'s unordered `LIMIT 1` can in principle resolve to a different row than a later update. Pre-existing; I matched `GetByName` exactly rather than invent a different tiebreak. Can file separately.
- `fullUpsert` writes the resourceVersion *string* into a `BIGINT UNSIGNED` column ([mysql.go:1162](https://github.com/michelangelo-ai/michelangelo/blob/277bb0397471ff162255572874df6523429b6c4c/go/storage/mysql/mysql.go#L1162)); an empty or non-numeric value coerces to 0, or errors under `STRICT_TRANS_TABLES`. Also pre-existing, but it would surface looking like a `directUpdate` bug. Can file separately.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Fixes the generic `Update` RPC failing with `UNIMPLEMENTED` on the MySQL metadata storage backend. Updating or re-registering an existing resource by name now works, including repeated `register_model` calls in pipeline runs.

**Documentation Changes**

None. The behaviour now matches the contract already documented on `MetadataStorage.Upsert`.

### DIFF
--- a/go/storage/mysql/mysql.go
+++ b/go/storage/mysql/mysql.go
@@ -145,7 +145,19 @@ func (m *mysqlMetadataStorage) Upsert(ctx context.Context, object runtime.Object
 	if direct {
 		// Direct update: only update labels, annotations, and resource version
 		// Check resource version for optimistic concurrency control
-		return m.directUpdate(ctx, tx, tableName, metaObj, object)
+		newResVersion, directErr := m.directUpdate(ctx, tx, tableName, metaObj, object)
+		if directErr != nil {
+			return directErr
+		}
+		if commitErr := tx.Commit(); commitErr != nil {
+			return status.Errorf(codes.Internal, "failed to commit direct update: %v", commitErr)
+		}
+		// Hand the new resource version back through the object the caller passed in.
+		// Only after a successful commit, so a failed commit never yields a version the
+		// database does not have. Per the MetadataStorage contract, this version must not
+		// be used for listing or watching.
+		metaObj.SetResourceVersion(strconv.FormatUint(newResVersion, 10))
+		return nil
 	}
 
 	// Full upsert: update all fields. Compute the stable primary key once so the main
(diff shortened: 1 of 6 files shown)


go/storage/mysql/mysql.go
```diff
@@ -145,7 +145,19 @@ func (m *mysqlMetadataStorage) Upsert(ctx context.Context, object runtime.Object
...
+			return directErr
+		}
+		if commitErr := tx.Commit(); commitErr != nil {
+			return status.Errorf(codes.Internal, "failed to commit direct update: %v", commitErr)
+		}
+		// Hand the new resource version back through the object the caller passed in.
+		// Only after a successful commit, so a failed commit never yields a version the
+		// database does not have. Per the MetadataStorage contract, this version must not
+		// be used for listing or watching.
+		metaObj.SetResourceVersion(strconv.FormatUint(newResVersion, 10))
```
badcount, line 159: I agree that keeping the stored spec unchanged is consistent with the immutable-resource contract. One concern though: the motivating register_model flow sends a newly uploaded artifact URI in the incoming Model spec when Create falls back to Update. Since this path intentionally ignores spec changes, wouldn't a second registration succeed while MySQL still references the previous artifact? Can we verify the intended behavior here, ideally with an end-to-end register_model → GetModel test using two different artifact URIs?

badcount (approved):

go/storage/mysql/mysql.go
codope: Good point. However, direct update is metadata-only by interface definition, so spec staleness is the immutable-resource contract working as designed, and the real fix belongs in `register_model` (perhaps it should fail loudly, not silently re-register a different artifact). I will verify and add a test to pin the current behavior, and file follow-up issue for the client-side gap.

codope: turns out there is a tracking issue #1631 for this. The fix mentioned in that issue is necessary but not sufficient. I've posted a comment there and i'll follow-up.

PR #1650, opened 2026-07-31
feat(js): config-driven form engine skeleton

craigmarker:
## Summary

The action system proved that declarative config objects that resolve to rendered UI are the right abstraction for this codebase. This PR introduces the same pattern for forms — config authors describe fields and layout, the engine handles rendering and extensibility.

- **FormConfig** separates field definitions (entities keyed by proto path) from structural arrangement (layout tree)
- **FormProvider** registers custom field renderers, validators, and layout renderers — mirrors CellProvider
- **Module augmentation** interfaces (`FieldConfigExtensions`, `LayoutConfigExtensions`) for type-safe consumer extensibility
- **ConfigDrivenForm** walks the layout tree, resolves renderers, composes the form
- Only string fields and group/row/grid layouts wired — enough to prove the resolution chain

## Test plan

- I verified the sandbox renders a config-driven form with two groups, string fields, and a multi-string tags field
- Typecheck and lint pass clean


javascript/packages/core/components/form/fields/string/schema-string-field.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <StringField
+        name={name}
+        label={c.label}
+        required={c.required}
+        disabled={c.disabled}
+        readOnly={c.readOnly}
+        placeholder={c.placeholder}
+        description={c.description}
+        caption={c.caption}
+        multi
```
donovanlopez: nit: is there a reason why we don't do `multi={c.multi}`?

donovanlopez (approved):

PR #1652, opened 2026-07-31
feat(js): declarative validation for form config

craigmarker:
## Summary
Field configs can now specify validation as a declarative object instead of importing and composing validator functions. The resolution layer maps keys (min, max, minLength, maxLength, regex, url) to built-in validators and composes them automatically. A function escape hatch handles anything the declarative keys can't express.

FieldValidationExtensions follows the same module-augmentation pattern as fields and layouts — consumers register domain-specific validators without touching core.

## Test plan

- I added validation (maxLength on title, min/max on replicas) to the sandbox config and verified typecheck passes
- Existing validator tests unaffected

### DIFF
--- a/javascript/packages/core/components/form/validation/resolve-validation.ts
+++ b/javascript/packages/core/components/form/validation/resolve-validation.ts
@@ -1,0 +1,41 @@
+import { combineValidators } from '#core/components/form/validation/combine-validators';
+import {
+  max as maxValidator,
+  maxLength as maxLengthValidator,
+  min as minValidator,
+  minLength as minLengthValidator,
+  regex as regexValidator,
+  url as urlValidator,
+} from '#core/components/form/validation/validators';
+
+import type { FieldValidation } from '#core/components/form/types/config-types';
+import type { FieldValidator } from '#core/components/form/validation/types';
+
+export function resolveValidation(
+  validation: FieldValidation | undefined
+): FieldValidator | undefined {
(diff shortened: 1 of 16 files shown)


javascript/packages/core/components/form/validation/resolve-validation.ts
```diff
@@ -0,0 +1,41 @@
...
+  validation: FieldValidation | undefined
+): FieldValidator | undefined {
+  if (!validation) return undefined;
+
+  const validators: FieldValidator[] = [];
+
+  if (validation.min !== undefined) validators.push(minValidator(validation.min));
+  if (validation.max !== undefined) validators.push(maxValidator(validation.max));
+  if (validation.minLength !== undefined) validators.push(minLengthValidator(validation.minLength));
+  if (validation.maxLength !== undefined) validators.push(maxLengthValidator(validation.maxLength));
```
donovanlopez, line 24: random thought but what happens if there's an invalid combination of `min`/`max` or `minLength`/`maxLength` like `min > max`?

donovanlopez (approved):

javascript/packages/core/components/form/validation/resolve-validation.ts
craigmarker: my guess is that no user would be able to successfully fill in the form. it may be a nice to have to ... validate ... that the validate configuration is valid 🤣

PR #1673, opened 2026-08-03
fix(core): correct Model list columns for description, environment, type, last updated

sallycr:
## Summary
The Trained Models list config (`packages/core/config/entities/model/list.ts`) referenced proto fields that don't exist on the real `ModelSpec`/`ModelStatus` shape (`spec.environment`, `spec.deployed`, `spec.trainingFramework`, `status.state`) — none of it could have worked. This rewrites the column set against the real `Model` proto and adds two shared CRD-metadata utilities other entity lists can reuse.

- Model name column gains a `spec.description` subtitle (matches the `CellType.DESCRIPTION` sub-item pattern already used in `run/list.ts`).
- New Environment column reads the `michelangelo/environment` label via a new `readEnvironmentLabel()` utility (`packages/core/utils/environment-utils.ts`).
- Type column gets a `MODEL_KIND_TEXT_MAP` covering all 9 OSS `ModelKind` values.
- Last Updated column gets a new `getCrdUpdatedSeconds()` utility (`packages/core/utils/timestamp-utils.ts`) that prefers the `michelangelo/SpecUpdateTimestamp` label and falls back to `creationTimestamp`.

**Out of scope (tracked separately):** Model Family column is carried forward unchanged (already correct, no code change needed). No Performance/`qualityScores` column is added — that renderer is a separate follow-up PR.

## Test plan

- [x] `tsc -p packages/core/tsconfig.json --noEmit` — zero errors
- [x] `eslint packages/core/utils packages/core/config/entities/model` — zero errors/warnings
- [x] `vitest --run packages/core/utils packages/core/config/entities/model` — all tests passing, including a new rendering-level test for the Model list columns
- [x] Scrubbed for Uber-internal identifiers — none found


sallycr:
## Sandbox validation — found & fixed a timestamp unit bug

Validated this against real data in a k3d sandbox (`california-housing` project, models produced by an actual training pipeline run). Found the **Last Updated** column rendering "Invalid date" for every model that had been updated at least once.

### Root cause

`getCrdUpdatedSeconds()` treated the `michelangelo/SpecUpdateTimestamp` label value as epoch **seconds**, but the apiserver writes it via `time.Now().UnixMicro()` (`go/api/handler/handler.go:485`) — **microseconds**. Real sandbox data confirms this:

```json
"labels": {
  "michelangelo/SpecUpdateTimestamp": "1785723475826264"
},
"creationTimestamp": { "seconds": "1785723475", "nanos": 0 }
```

`1785723475826264 / 1785723475 ≈ 1,000,000` — a clean microsecond ratio. Passing the raw label value straight through as "seconds" and multiplying by 1000 in the DATE cell renderer overflows JS's valid `Date` range (±8.64×10¹⁵ ms), producing a silent `Invalid Date` rather than a crash — easy to miss without real (non-empty) demo data.

Fixed in `fa0b5b71`: divide the label by `1_000_000` before returning it as seconds. Updated the existing unit test fixture to use a microsecond-scale value matching real data. Re-verified in sandbox — Last Updated now renders correctly for all four models tested.

### Separate finding (not blocking, filed for awareness)

For models produced by the OSS `california-housing` example pipeline, **Environment**, **Model Family**, and **Type** all render blank — correctly, per this PR's graceful-degradation design (`readEnvironmentLabel()` / missing `typeTextMap` entries return `''`/blank rather than throwing). The underlying reason: `ModelPusherPlugin.register_model()` (`python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py`) never accepts or sets `kind`, `model_family`, or a `michelangelo/environment` label — that data simply doesn't exist on models registered through the current example pipeline. This is a pre-existing gap in the example/registry-client code, not something introduced by or fixable in this PR. Worth a follow-up issue if richer demo data is wanted to exercise these columns end-to-end.

🤖 Generated with sandbox validation harness

sallycr:
Verified the UI in sandbox. 
Screenshot 2026-08-03 at 5 19 22 PM

sallycr:
## Second sandbox finding — Type column showed raw enum numbers instead of labels

Extended sandbox validation: created models with `spec.kind` explicitly set to different `ModelKind` values via the API (`model-env-development` → `MODEL_KIND_REGRESSION`, `model-env-production` → `MODEL_KIND_BINARY_CLASSIFICATION`, `model-env-testing` → `MODEL_KIND_CLUSTERING`) to exercise the Type column end-to-end. Found the Type column rendering raw numbers (`2`, `3`, `5`) instead of the mapped labels.

### Root cause

`MODEL_KIND_TEXT_MAP` was keyed by string enum names:
```ts
export const MODEL_KIND_TEXT_MAP: Record<string, string> = {
  MODEL_KIND_INVALID: 'Unknown',
  MODEL_KIND_REGRESSION: 'Regression',
  ...
};
```

But the generated proto client decodes enum fields to their **numeric discriminant** in the browser — `spec.kind` arrives as `2`, not `"MODEL_KIND_REGRESSION"`. `typeCellToString` does a direct `column.typeTextMap?.[value]` lookup (see `packages/core/components/cell/renderers/type/type-cell-to-string.ts`, which even has an existing comment acknowledging `value` can be numeric), so every lookup missed and fell through to rendering the raw number.

This matches the numeric-keying convention already used by **every other** `typeTextMap`/`stateTextMap` in this codebase — `DEPLOYMENT_STAGE_CELL` in `config/entities/deployment/shared.ts`, the old pipeline type map, etc. `MODEL_KIND_TEXT_MAP` was the only one keyed by string names.

The PR's own `model-page.test.tsx` didn't catch this because its mock injected `kind: 'MODEL_KIND_BINARY_CLASSIFICATION'` directly into the component tree, bypassing the real client's decode step — the test exercised the wire/JSON shape, not the actual runtime shape.

### Fix (pushed in `711c5444`)

- Added a `MODEL_KIND` numeric constant object mirroring the `DEPLOYMENT_STAGE` pattern, re-keyed `MODEL_KIND_TEXT_MAP` off of it (`0`–`8` matching `ModelKind` in `model.proto`)
- Fixed the test mock to use `kind: 3` instead of the string name, so this can't silently regress again
- Re-verified in sandbox: all three test models now render correct labels (Regression / Binary Classification / Clustering)

(For the record — I also considered whether the `key: "model_kind"` index annotation on `spec.kind` in `model.proto` implied a separate field name the UI should read from. Traced it through the compiled descriptor and the MySQL metadata storage layer: it's purely an internal secondary-index key for future server-side filtering, not a field ever present in the API response. `id: 'spec.kind'` is correct as-is.)

🤖 Generated with sandbox validation harness

sallycr:
**Compliance follow-up (per `.claude/skills`):** extracted `MODEL_KIND`/`MODEL_KIND_TEXT_MAP` out of `list.ts` into a dedicated `constants.ts` (matching the `deployment/shared.ts` precedent and the file-directory-structure convention for feature-root constants files), and added unit coverage for `MODEL_KIND_TEXT_MAP` — closing the test gap that let the numeric-vs-string key mismatch fixed in `711c5444` ship undetected in the first place.

Verified: `vitest` (12/12 passing), `eslint` clean, `tsc --noEmit` clean, rebuilt `michelangelo-ui:pr-1673` and re-validated in the k3d sandbox — Type/Environment/Model Family/Last Updated columns all render correctly.

Pushed as bf707d02.

javascript/packages/core/config/entities/model/__tests__/constants.test.ts
```diff
@@ -0,0 +1,13 @@
+import { MODEL_KIND, MODEL_KIND_TEXT_MAP } from '../constants';
+
+describe('MODEL_KIND_TEXT_MAP', () => {
+  test.each(Object.entries(MODEL_KIND))('maps MODEL_KIND.%s to a display label', (_name, value) => {
+    expect(MODEL_KIND_TEXT_MAP[value]).toEqual(expect.any(String));
+  });
```
craigmarker: this test is a no-op. in english, it's just checking that all the keys in the dictionary map to a string.

javascript/packages/core/config/entities/model/list.ts
```diff
@@ -1,64 +1,66 @@
...
+        type: CellType.DESCRIPTION,
+      },
+    ],
   },
   {
-    id: 'metadata.creationTimestamp.seconds',
-    label: 'Created',
-    type: CellType.DATE,
+    // Model.spec has no environment field; environment is CRD-label-only metadata, read via
+    // readEnvironmentLabel().
```
craigmarker: nit: comment is too self-aware about previously broken code

javascript/packages/core/utils/__tests__/environment-utils.test.ts
```diff
@@ -0,0 +1,24 @@
+import { readEnvironmentLabel } from '../environment-utils';
+
+describe('readEnvironmentLabel', () => {
+  test('returns "Development" for ENV_TYPE_DEVELOPMENT', () => {
+    expect(readEnvironmentLabel({ 'michelangelo/environment': 'ENV_TYPE_DEVELOPMENT' })).toEqual(
```
craigmarker: I don't see any controller/pipeline code setting environment label, how are we sure the data format is correct?

craigmarker: `michelangelo/environment` is always prefixed by CRD-type, eg `pipelinerun.michelangelo/environment`. https://github.com/search?q=repo%3Amichelangelo-ai%2Fmichelangelo+michelangelo%2Fenvironment&type=code

I'm not sure it makes sense to extract this into a shared utility until a real reuse case appears. reference: [.claude/skills/file-directory-structure/SKILL.md#code-placement-strategy](https://github.com/michelangelo-ai/michelangelo/blob/main/javascript/.claude/skills/file-directory-structure/SKILL.md#code-placement-strategy)

javascript/packages/core/utils/crd-utils.ts
```diff
@@ -0,0 +1,33 @@
+const UPDATE_TIMESTAMP_LABEL_KEY = 'michelangelo/SpecUpdateTimestamp';
```
craigmarker, line 1: nit: probably makes more sense to have this utility file be named something like 'crd-utils.ts'. there is already a [time-utils.ts](https://github.com/michelangelo-ai/michelangelo/blob/4aae742273bfa1272eea67dfb8145261574283cf/javascript/packages/core/utils/time-utils.ts) for generic time utilities.

craigmarker (dismissed):

javascript/packages/core/utils/__tests__/environment-utils.test.ts
sallycr: @craigmarker it's the label, not setting by the controller.

sallycr: We need this for pipeline run list and model list page. Is it justify to be in util?

javascript/packages/core/config/entities/model/list.ts
sallycr: What do you want to do this? delete these comment is what you mean here?

javascript/packages/core/config/entities/model/__tests__/constants.test.ts
sallycr: @craigmarker it was the issue that I see has the numbers are in UI. Seems to me it's the valid case to prevent that issue again? Do you suggest to delete this?

javascript/packages/core/utils/__tests__/environment-utils.test.ts
craigmarker: will they have the same label name? right now I don't see a `michelangelo/environment` label key within the repository. what am I missing?

craigmarker: where does ENV_TYPE_DEVELOPMENT come from? I don't see that anywhere in services or python code.

sallycr: yes. it should be from the registered pipeline label and pipeline run.

javascript/packages/core/utils/crd-utils.ts
sallycr: Renamed to crd-utils.ts in ab250303, thanks!

javascript/packages/core/utils/__tests__/environment-utils.test.ts
sallycr: Good catch — you were right to flag this. I checked against real fixture/label data and confirmed the `michelangelo/environment` label is always stored as a plain lowercase string (`development`, `production`), never the prefixed `ENV_TYPE_*` form I had assumed. I also found that a `testing` value exists in the broader environment type enum, so it should be handled too.

Pushed a fix: `readEnvironmentLabel()` now matches the real lowercase values (`development` -> `Development`, `production` -> `Production`, `testing` -> `Testing`) and still returns an empty string for anything unrecognized. Updated the unit tests and the rendering test's mocked label to match. Thanks for pushing on this.

javascript/packages/core/config/entities/model/list.ts
```diff
@@ -1,64 +1,64 @@
...
   },
   {
     id: 'spec.modelFamily.name',
     label: 'Model Family',
     type: CellType.TEXT,
   },
   {
-    id: 'spec.environment',
-    label: 'Environment',
+    // See MODEL_KIND_TEXT_MAP above for the full set of ModelKind display labels.
```
donovanlopez: nit: I don't think this comment is necessary because it doesn't really explain anything that requires more elaboration.

javascript/packages/core/config/entities/model/list.ts
```diff
@@ -1,64 +1,64 @@
...
-      4: 'Deployed',
-    },
-    stateColorMap: {
-      0: 'gray',
-      1: 'blue',
-      2: 'green',
-      3: 'red',
-      4: 'green',
+    // Prefers the SpecUpdateTimestamp label, falls back to creationTimestamp; see
+    // getCrdUpdatedSeconds().
```
donovanlopez: nit: similarly, I don't think this comment is necessary.

javascript/packages/core/config/entities/model/list.ts
```diff
@@ -1,64 +1,64 @@
...
+    id: 'metadata',
+    label: 'Last Updated',
+    type: CellType.DATE,
+    accessor: (data: unknown) => {
+      // cast: accessor receives unknown data; narrowing to expected proto shape for property
+      // access; see #1425
+      const row = data as {
+        metadata?: { labels?: Record<string, string>; creationTimestamp?: { seconds: number } };
+      };
+      return getCrdUpdatedSeconds(row);
```
donovanlopez, line 58: Perhaps not in this PR but if this is going to be a recurring column, we should make this reusable

donovanlopez (approved):

javascript/packages/core/config/entities/model/list.ts
sallycr: will see if it's reusable. I do see updated time is needed in other lists e.g. pipeline list / pipeline run list.

PR #1678, opened 2026-08-04
fix(release): mark -rc tags as prereleases in GitHub releases

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
All three softprops/action-gh-release steps now derive the release's prerelease flag from the tag name:
- release.yaml (wheel upload and binaries steps): `prerelease: ${{ contains(github.ref_name, '-rc') }}`
- changelog.yml (release-notes update step): `prerelease: ${{ contains(steps.tag.outputs.tag, '-rc') }}`

**Why?**
RC tags currently publish as full releases. Both v0.6.0-rc.1 and v0.7.0-rc.1 went out with `prerelease: false`, and v0.7.0-rc.1 was served as the repo's **latest release** for ~2 days (Aug 2-3) until the rc release entries were deleted after v0.7.0 final shipped. Deleting the releases fixed the symptom but not the workflow - the next -rc tag will do the same thing. The repo's own tag scheme (vX.Y.Z-rc.N, e.g. the pep440 normalization step in release.yaml matches `-rc\.`) makes the tag name the reliable signal.

**How did you test it?**
- Verified expression semantics: `contains('v0.6.0-rc.1', '-rc')` -> true; `contains('v0.6.0', '-rc')` -> false, so final releases are unaffected.
- Verified all three steps and their tag sources against main (release.yaml uses github.ref_name from the tag push trigger; changelog.yml uses its own resolved steps.tag.outputs.tag).
- Observable on the next -rc tag: the created/updated release should show a "Pre-release" badge. Existing releases are not modified by this PR.

**Potential risks**
On workflow_dispatch of release.yaml, github.ref_name is a branch name; branch names containing "-rc" would be flagged as prerelease. No current branch matches, and dispatch runs of this workflow already have other tag-shaped assumptions.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
RC releases (tags containing "-rc") are now marked as prereleases on GitHub.

**Documentation Changes**
None needed.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -181,6 +181,7 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           tag_name: ${{ steps.tag.outputs.tag }}
+          prerelease: ${{ contains(steps.tag.outputs.tag, '-rc') }}
           body_path: RELEASE_NOTES.md
           append_body: false
         env:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -67,6 +67,7 @@ jobs:
       - uses: softprops/action-gh-release@v2
         with:
           tag_name: ${{ github.ref_name }}
+          prerelease: ${{ contains(github.ref_name, '-rc') }}
           files: python/dist/*.whl
 
   build-go-binaries:
@@ -125,6 +126,7 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           tag_name: ${{ github.ref_name }}
+          prerelease: ${{ contains(github.ref_name, '-rc') }}
           files: binaries/*
 
   helm-publish:


apurvapatkeshwar:
@sallycr could someone approve the workflow runs here and on #1679? 
fork PRs seem to need manual approval on each push 
both are rebased on latest main

apurvapatkeshwar:
Still no CI here - the runs from the Aug 7 rebase are parked in `action_required`, so nothing has actually executed yet
I think that might need the "approve and run workflows" button on the actions tab rather than a review: #1711 has an approval on it and still shows zero checks

@sallycr @paulzim could one of you please kick these off? #1679 is also sitting in the same state

Both still reproduce on current main if that helps - the 3 `action-gh-release` steps (release.yaml:67, release.yaml:125, changelog.yml:181) have no `prerelease` key, and metadata_handler.go:133 still logs the delete error even when it's nil

paulzim (approved):
**Approve** — small, correct, well-verified.

Verified independently rather than trusting the description:

- Diff is exactly the 3 lines described: adds `prerelease: ${{ contains(steps.tag.outputs.tag, '-rc') }}` (changelog.yml) and `prerelease: ${{ contains(github.ref_name, '-rc') }}` (release.yaml, ×2) immediately after each `tag_name:` line.
- Confirmed "all three `softprops/action-gh-release@v2` steps" is the complete set: `grep -n "softprops/action-gh-release"` across both files finds exactly 3 usages, and all 3 now have the new `prerelease:` line — nothing missed, nothing extra.
- `contains()` semantics check out: `contains('v0.6.0-rc.1', '-rc')` → true, `contains('v0.6.0', '-rc')` → false. Matches the pep440 tag pattern used elsewhere in `release.yaml` (`^([0-9]+\.[0-9]+\.[0-9]+)-rc\.([0-9]+)$`), so this correctly flags `-rc.N` tags and only those.
- Both workflow files parse as valid YAML.
- Sanity-checked the PR's own disclosed risk (manual `workflow_dispatch` runs use `github.ref_name`, which is a branch name, not a tag, when dispatched off a branch): `release.yaml` does have `workflow_dispatch: {}` with no tag input, so the risk is real as described. Checked all current branches for anything matching `-rc` — none do, so it's not live today, and the PR is upfront about this being a pre-existing, unrelated tradeoff rather than something this change introduces.

Root cause is real and reproducible on current `main`, not just theoretical: the three `action-gh-release` steps currently have no `prerelease` key at all, so every GitHub Release — including `-rc` tags — is published as a full/latest release. Minimal, targeted fix.

CI: only `license/cla` has run — this is the same fork-PR approval gate seen on the rest of this contributor's queue. Note: apurvapatkeshwar directly asked in the PR comments (and tagged you by name) to approve the queued workflow runs for this PR and #1679, since both are stuck in `action_required` after an Aug 7 rebase onto main. That's a GitHub Actions "Approve and run workflows" click, not something I can do from here — over to you.

apurvapatkeshwar:
CI ran on both this and #1679 - all green
Thanks for kicking them off!

PR #1679, opened 2026-08-04
fix(api): log blob-storage delete failure only when it fails

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
In `handleDelete` (go/api/handler/metadata_handler.go), the call to `log.Error(err, "Failed to delete object in blob storage", ...)` after `DeleteFromBlobStorage` now only happens when the returned error is non-nil. Adds a regression test (table-driven, gomock + funcr log capture) asserting no error line on the happy path and exactly one when the blob delete fails.

**Why?**
`log.Error` was called unconditionally, so every successful delete of a blob-backed object emits a false "Failed to delete object in blob storage" ERROR line with a nil error. That pollutes operator logs and anything alerting on error-level entries. Intended behavior (per the surrounding comment) is unchanged: blob-delete failures are still non-fatal and still logged.

**How did you test it?**
Added `TestHandleDeleteBlobStorageErrorLogging` in go/api/handler/metadata_handler_test.go using the existing storagemocks and a funcr logger to capture log output; covers both the success (0 error logs) and failure (1 error log, nil returned error) paths.

**Potential risks**
Minimal: the only behavior change is the removal of the spurious log line on success. Anyone who (inadvertently) keyed monitoring on the false ERROR line will see it disappear.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
Fixed a spurious error log emitted on every successful blob-storage delete.

**Documentation Changes**
None needed.

### DIFF
--- a/go/api/handler/BUILD.bazel
+++ b/go/api/handler/BUILD.bazel
@@ -47,6 +47,7 @@ go_test(
     srcs = [
         "builder_test.go",
         "handler_test.go",
+        "metadata_handler_test.go",
     ],
     embed = [":go_default_library"],
     deps = [
@@ -56,6 +57,7 @@ go_test(
         "//go/storage/storagemocks:go_default_library",
         "//proto/api:go_default_library",
         "//proto/api/v2:go_default_library",
+        "@com_github_go_logr_logr//funcr:go_default_library",
         "@com_github_go_logr_zapr//:go_default_library",
         "@com_github_golang_mock//gomock:go_default_library",
         "@com_github_stretchr_testify//assert:go_default_library",
--- a/go/api/handler/metadata_handler.go
+++ b/go/api/handler/metadata_handler.go
@@ -129,8 +129,9 @@ func handleDelete(ctx context.Context, log logr.Logger, typeMeta *metav1.TypeMet
 		if getErr == nil {
 			// Failed to delete in blob storage is not a critical failure, as orphan blobs can be deleted by garbage
 			// collector. So, do not return error.
-			err := handler.DeleteFromBlobStorage(ctx, object)
-			log.Error(err, "Failed to delete object in blob storage", "uid", object.GetUID())
+			if err := handler.DeleteFromBlobStorage(ctx, object); err != nil {
+				log.Error(err, "Failed to delete object in blob storage", "uid", object.GetUID())
+			}
 		}
 
 		return nil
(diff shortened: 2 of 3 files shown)


paulzim (approved):
**Approve** — small, correct bug fix with a solid regression test; ran it myself.

Verified rather than took on faith:

- Diff is exactly as described: `handleDelete` in `metadata_handler.go` now only calls `log.Error(...)` when `handler.DeleteFromBlobStorage(...)` actually returns a non-nil error, instead of calling it unconditionally on the result. One-line semantic change, no other logic touched — the surrounding comment ("not a critical failure... do not return error") still holds, `nil` is still swallowed rather than propagated.
- Confirmed the bug is real on current `main`, not hypothetical: read the pre-PR code directly — `err := handler.DeleteFromBlobStorage(...); log.Error(err, ...)` unconditionally logs an ERROR-level line with `err=nil` on every successful blob delete.
- New test (`metadata_handler_test.go`, table-driven, gomock for storage + `funcr` for log capture) checks both branches: 0 log lines on success, 1 on failure. Actually ran it (`go test ./api/handler/... -run TestHandleDeleteBlobStorageErrorLogging -v -vet=off`) — both subtests pass.
- Confirmed the funcr-capture design is sound, not a test that would pass vacuously: `handleDelete`'s only logging call in this code path is the one under test, so a captured-line count of 0 vs 1 directly reflects whether `log.Error` fired — there's no other log statement in the function that could mask a broken assertion.
- `BUILD.bazel` diff matches: adds `metadata_handler_test.go` to `go_test.srcs` and the new `funcr` dep — nothing else changed, no missing deps.
- Base `go test ./api/handler/...` (no `-run` filter) fails to build under default vet due to `non-constant format string` errors in `handler.go` — confirmed via `git show upstream/main:go/api/handler/handler.go` that these predate this PR and live in a file this PR never touches (`status.Errorf(codes.InvalidArgument, err.Error())` at several call sites). Unrelated pre-existing vet debt, not a regression from this change; the new test itself passes cleanly with vet disabled.

CI: only `license/cla` so far — same fork-PR approval gate as the rest of this contributor's queue (see #1678, which has an open ask for you/sallycr to click "Approve and run workflows" on both this PR and #1678).

Minimal, well-tested, no behavior change beyond removing the false error line.

PR #1683, opened 2026-08-05
feat(native_transform): add fitted-statistics layers (PR B3)

sallycr:
## Summary

Continues the native_transform migration (PR A, A2, B1, B2 merged) with PR B3: the fitted-statistics torch layers.

- New `stats_layers.py` with `MinMax`, `Normalization`, `Bucketization`, ported from the internal SDK's `base_layers.py` onto the same `TorchTransformBaseLayer` dict-tensor contract established in B1/B2.
- Constructor params (`dim`, `eps`, `dtype`) are explicit and typed rather than pulled out of `**kwargs` post-hoc, matching this repo's established layer style; `**kwargs` is still forwarded to the base layer so `name=` and future options aren't silently dropped.
- `Bucketization` auto-detects and preserves sentinel positions (NaN for float input, `INT32_SENTINEL` for int input) across bucketization, restoring the sentinel in whichever dtype the layer outputs.
- `MinMax`/`Normalization` now validate `output_cols` has exactly one entry (inputs are concatenated into a single output). **This is a genuine behavioral change from internal, not enforcement of an existing contract**: internal has no such check anywhere and silently uses only `output_cols[0]`, discarding any extra entries. Verified against the internal SDK: every real spec/test construction of these layers uses a single input/output column (1:1 scaling), so no known usage relies on the old silent-drop behavior. Failing loudly on a malformed spec is safer than silently dropping data, so the check is kept, but the PR description previously mischaracterized this as "previously documented" — it wasn't; internal's docstrings never mention a single-output-column requirement.

**Module placement**: these land in a new `stats_layers.py` rather than the existing `base_layers.py`. Fitted-statistics layers carry buffer/state concerns (registered `nn.Module` buffers for their fitted parameters) distinct from the stateless layers `base_layers.py` holds, so this is a natural seam to split on. `base_layers.py`'s module docstring is updated to describe what's actually landed there via PR B1/B2 and to point to the new module.

## Review

Reviewed by the migration team before opening: one real bug was caught and fixed pre-merge — `Bucketization`'s sentinel restore always wrote the integer sentinel even when the output dtype was floating-point, silently dropping `NaN` in a float-in/float-out configuration. Fixed to match the output dtype, with a regression test. A second pass added the `output_cols` arity validation above.

## Test plan

- [x] `pytest` on the full `native_transform` suite: 243 passed, 1 pre-existing unrelated skip
- [x] TorchScript script + save/load round-trip covered for all three new layers (parametrized `TestB3TorchScriptRoundTrip`)

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/stats_layers.py
+++ b/python/michelangelo/lib/native_transform/torch/stats_layers.py
@@ -1,0 +1,307 @@
+"""Fitted-statistics PyTorch native transform layers.
+
+Layers whose behavior is parameterized by fitted statistics (a min/max range,
+a mean/std, or bucket boundaries) computed ahead of time and supplied at
+construction. Like the layers in
+:mod:`~michelangelo.lib.native_transform.torch.base_layers`, every layer here
+subclasses
+:class:`~michelangelo.lib.native_transform.torch.base_layers.TorchTransformBaseLayer`
+and uses the same ``dict[str, torch.Tensor]`` in/out contract, but stores its
+fitted parameters as buffers (via ``register_buffer``) so they move with
+``.to(device)`` and persist through TorchScript save/load.
+"""
+
+from __future__ import annotations
+
+import torch
+
+from michelangelo.lib.constants.sentinel import INT32_SENTINEL
+from michelangelo.lib.native_transform.torch.base_layers import (
+    TorchTransformBaseLayer,
+)
+from michelangelo.lib.native_transform.torch.constants import DEFAULT_EPSILON
+from michelangelo.lib.native_transform.torch.utils import (
+    format_inputs,
+    format_outputs,
+    initialize_dtype,
+)
+
+__all__ = [
+    "Bucketization",
+    "MinMax",
+    "Normalization",
+]
+
+
+class MinMax(TorchTransformBaseLayer):
+    """Min-max scale input columns to a ``[0, 1]`` range.
+
+    Concatenates the input columns along ``dim`` and rescales with
+    ``(x - min) / max(max - min, eps)``.
+
+    Args:
+        input_cols: Column names of the input tensors.
+        output_cols: Column names of the output tensor; must contain exactly
+            one column, since the inputs are concatenated into a single
+            output.
+        min: The per-feature minimum values used for scaling.
+        max: The per-feature maximum values used for scaling.
+        dim: The dimension along which input columns are concatenated.
+            Defaults to ``-1``, since dimension ``0`` is typically the batch
+            dimension.
+        eps: A small value added to the denominator to avoid division by zero
+            when ``min == max``. Defaults to ``1e-7``.
+        **kwargs: Additional base-layer options (e.g. ``name``).
+
+    Raises:
+        ValueError: If ``output_cols`` does not contain exactly one column.
+    """
+
+    def __init__(
+        self,
+        input_cols: list[str],
+        output_cols: list[str],
+        min: float | list[float] | torch.Tensor,
+        max: float | list[float] | torch.Tensor,
+        dim: int = -1,
+        eps: float = DEFAULT_EPSILON,
+        **kwargs,
+    ) -> None:
+        """Initialize the MinMax layer.
+
+        Args:
+            input_cols: Column names of the input tensors.
+            output_cols: Column names of the output tensor; must contain
+                exactly one column.
+            min: The per-feature minimum values used for scaling.
+            max: The per-feature maximum values used for scaling.
+            dim: The dimension along which input columns are concatenated.
+            eps: A small value added to the denominator to avoid division by
+                zero when ``min == max``.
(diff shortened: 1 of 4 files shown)


python/michelangelo/lib/native_transform/torch/stats_layers.py
```diff
@@ -0,0 +1,307 @@
...
+        concatenated_tensor.sub_(mean_val).div_(safe_std)
+        return {self.output_cols[0]: concatenated_tensor}
+
+
+class Bucketization(TorchTransformBaseLayer):
+    """Bucketize each input column using shared boundaries.
+
+    Sentinel positions from upstream ragged-batch collation (``NaN`` for float
+    inputs, ``INT32_SENTINEL`` for integer inputs) are automatically detected
+    and preserved through bucketization (rewritten as ``INT32_SENTINEL`` in
```
kechen2013: Just a note: the code writes `NaN` here when the output dtype is float, so only the int output path actually uses `INT32_SENTINEL`. Reads like the pre-fix behavior?

python/michelangelo/lib/native_transform/torch/stats_layers.py
```diff
@@ -0,0 +1,307 @@
...
+            eps: A small value added to the denominator to avoid division by
+                zero when ``min == max``.
+            **kwargs: Additional base-layer options (e.g. ``name``).
+
+        Raises:
+            ValueError: If ``output_cols`` does not contain exactly one
+                column.
+        """
+        super().__init__(input_cols, output_cols, **kwargs)
+        if len(output_cols) != 1:
```
kechen2013, line 88: Just a note this is a behavior change vs internal, not just enforcement: internal has no check and returns `{output_cols[0]: ...}`, so a 2-output-col MinMax silently produced just the first column. Wondering if any existing spec does this, since the B5 specs have no arity constraint on `output_cols`?

python/michelangelo/lib/native_transform/torch/stats_layers.py
sallycr: Confirmed - stale docstring describing the pre-fix behavior. forward()'s own docstring and the actual code already restore the sentinel matching the *output* dtype (NaN when float, INT32_SENTINEL otherwise) - that was the fix from the earlier review pass. Pushed 2b0ca624 correcting the class docstring to match.

python/michelangelo/lib/native_transform/torch/stats_layers.py
sallycr: You're right, this is a real behavior change, not enforcement of an existing contract - internal has no check here and its docstrings never document a single-output-column requirement either. I checked every real construction of MinMaxLayerSpec/StandardScalerLayerSpec/MinMaxScalerLayerSpec in the internal SDK (prod paths in transform_spec.py's spec-resolution logic, plus tests) - all of them use single-element input_cols/output_cols (1:1 per-feature scaling), so nothing relies on the old silent-drop-extra-columns behavior. Given that, I'd rather keep the validation (failing loudly beats silently dropping data) but I'll fix the PR description - it currently claims this was "previously documented but not enforced," which isn't accurate. Flagging for B5 too: MinMaxLayerSpec/NormalizationLayerSpec currently have no matching arity constraint, so a spec could pass Pydantic validation and only fail later at layer-build time (B6) - worth deciding in B6 whether to validate arity at the spec level as well.

kechen2013: Sounds good on keeping the validation and fixing the description. On arity, wondering if B5 is a better home than B6? A validator there would fail a bad spec at parse time rather than at layer build, and it's only a few lines. Fine to batch it into B6 if you'd rather though.

python/michelangelo/lib/native_transform/torch/stats_layers.py
kechen2013: One thing on 2b0ca624: the new wording says `NaN` even for integer input, but that's backwards. Integer input always writes `INT32_SENTINEL` regardless of output dtype, only float input follows the output dtype. So `int32` in with `float32` out gives `INT32_SENTINEL`, not `NaN`.

python/michelangelo/lib/native_transform/torch/stats_layers.py
sallycr: Good point - agreed, B5 is the better home. It's a small, self-contained addition (a model_validator on MinMaxLayerSpec/NormalizationLayerSpec mirroring the layer's own check), and there's no reason to make a user wait until B6's layer-build step to discover a spec is malformed when pydantic can catch it at parse time instead. Adding it to B5 now.

sallycr: fyi - https://github.com/michelangelo-ai/michelangelo/pull/1685/changes/9cc713559f721d588554797a53ee99a33c991fb3

kechen2013 (approved):

PR #1684, opened 2026-08-05
feat(native_transform): add IDHashTokenizer wrapper layer (PR B4)

sallycr:
## Summary

Continues the native_transform migration (PR A, A2, B1, B2, B3 in flight) with PR B4: the `IDHashTokenizer` wrapper layer, mirroring the A/A2 core/wrapper split.

- Composes the already-ported vocabulary-lookup core (`id_hash_tokenizer.py`, from PR A2/#1548) into the `TorchTransformBaseLayer` dict-tensor contract established in B1/B2, as the internal SDK's own wrapper does.
- Sentinel positions (`NaN` for float input, `INT32_SENTINEL` for int input) are captured **before** the float→int32 cast the core requires — `NaN` does not survive that cast — then restored as `INT32_SENTINEL` in the output regardless of the input's original dtype.
- `**kwargs` forwarded to `super().__init__()` per this repo's constructor-parity convention; validates `len(input_cols) == len(output_cols)`.
- Exposes `vocabulary`/`unk_index`/`output_vocab_size` from the wrapped core.

## ⚠️ Breaking change: export identity

`michelangelo.lib.native_transform.torch.IDHashTokenizer` previously referred to the bare lookup core added in PR #1548 (a plain `nn.Module` taking/returning a single `torch.Tensor`). This PR repoints that top-level name to the new wrapper layer above, which takes/returns `dict[str, torch.Tensor]` instead — a different call contract.

The core is unchanged and still available at its original location: `michelangelo.lib.native_transform.torch.id_hash_tokenizer.IDHashTokenizer`. This repo has no other consumers of the top-level name yet (this PR immediately follows the core's own PR), so the impact window is small, but flagging prominently since it's a same-name/different-class swap.

## Review

Reviewed by the migration team before opening: architect review passed on kwargs-forwarding, sentinel-capture-before-cast ordering, and dtype handling with no bugs found. Dev-advocate review flagged the export-identity change above as needing prominent documentation — addressed via an explicit docstring note on the new class plus this section.

## Test plan

- [x] `pytest` on the full `native_transform` suite: 221 passed, 1 pre-existing unrelated skip
- [x] TorchScript script + save/load round-trip covered (plain lookup + NaN-sentinel case)


kechen2013:
Compared against internal and the logic is identical, just two local renames and the same core composition. The coverage failure looks like infra (`ProtocolError` mid-transfer), so probably just needs a re-run.

Wondering if the export-identity note should also mention that internally the core lives in a different package, so there's no collision there. Having both under `native_transform/torch` is what forces the repoint here.

sallycr:
> Compared against internal and the logic is identical, just two local renames and the same core composition. The coverage failure looks like infra (`ProtocolError` mid-transfer), so probably just needs a re-run.
> 
> Wondering if the export-identity note should also mention that internally the core lives in a different package, so there's no collision there. Having both under `native_transform/torch` is what forces the repoint here.

good point. added in the follow up doc when models/layers are introduced in open source. thx Ke.

sallycr:
Thanks for double-checking parity - glad it lines up. On the coverage failure: confirmed it's infra, not this PR - pulled the raw job log and it's a ProtocolError/IncompleteRead mid-download of a CUDA/triton wheel during dependency install, nothing to do with the diff. Re-ran the job.

Good catch on the docstring gap - you're right that internally there's no collision at all, since the core lives in a completely separate package (models/layers/pretrained_embedding.py, alongside PretrainedEmbedding and IDHash2PretrainedEmbedding, which are a different feature not being migrated). Only IDHashTokenizer itself was ported to OSS, and it landed under native_transform/torch/ since that's its only consumer here - that's specifically what creates the naming collision this PR resolves, where internal never had one. Added that explanation to the wrapper's docstring (e1ce2477).

Also flagged as a longer-term follow-up (not blocking this PR): whether OSS should eventually get a real models/layers-style standalone package like internal's, for when more of that layer library migrates. Filed for later rather than doing it speculatively now for one class.

kechen2013 (approved):

PR #1685, opened 2026-08-05
feat(native_transform): add Pydantic layer specs (PR B5)

sallycr:
## Summary

Continues the native_transform migration (PR A, A2, B1, B2, B3, B4 in flight) with PR B5: the Pydantic layer specs.

- Ports the internal SDK's `transform_layer_spec.py` in full: declarative, serializable pydantic v2 specs describing every native transform layer's configuration, consumed by a `TransformSpec` DAG engine that's added in a follow-up PR (not this one — this PR is spec-models only).
- All 27 classes (25 concrete `*LayerSpec` classes + the base `TorchTransformLayerSpec` + the `TransformerMode` enum) ported field-for-field, using `typing.Optional`/`Union` (not `X | None`/`X | Y`) for dtype/optional fields -- see Review section, this repo supports Python 3.9 and pydantic v2 eagerly evaluates field annotations even with `from __future__ import annotations`.
- Placeholder specs (`StandardScalerLayerSpec`, `MinMaxScalerLayerSpec`, `PercentileBucketizationLayerSpec`) resolve to a concrete spec via `_resolved_layer_type`, matching internal's fit-time resolution scheme.
- `NumericalStandardTransformLayerSpec`'s cross-field `cap_min`/`cap_max`/`default_value` percentile-range validation ported with identical error wording.
- This is the **first real usage of pydantic** anywhere in this repo's `lib/` tree (already a declared dependency, previously unused) — sets the v2 style precedent (`Field`, `model_validator`, `field_validator`, `ClassVar`) for future spec PRs.

Per the migration plan, this is scoped to **all** spec classes rather than just the layers already ported (B1-B4) — these are inert models with no functional dependency on the corresponding layer classes existing, so there's no ordering risk and it avoids a second churn PR when B6-B8 land.

## Review

Reviewed by the migration team before opening: architect review confirmed field-for-field parity with internal across all 27 classes, with no bugs found. Dev-advocate review caught a real type/default mismatch (`is_training_features` annotated as non-optional `list[bool]` but defaulting to `None`, only rescued at runtime by a validator) and several under-documented field descriptions — both fixed before this PR opened.

**Post-open fix (2026-08-06)**: caught by review that `X | Y`-style union annotations (the original port style) crash `python-build` under Python 3.9 -- pydantic v2 evaluates field annotations at class-definition time to build validators, and PEP 604's `|` operator on `type` only exists from 3.10; `from __future__ import annotations` defers the Python-level parse but not pydantic's runtime eval. Verified independently against a real Python 3.9 + pydantic install (reproduced the crash on the original file, confirmed clean on the fix). Converted every union annotation in this file to `Optional`/`Union`, and added a scoped `ruff` per-file-ignore (`UP007`/`UP045`) so `ruff check --fix` can't silently reintroduce the `X | Y` syntax and regress this fix later.

**Post-open addition (2026-08-06)**: added a `model_validator` to `MinMaxLayerSpec` and `NormalizationLayerSpec` requiring exactly one `output_cols` entry, matching the runtime check on the corresponding layers (PR B3, `stats_layers.py`). Both layers concatenate all input columns into a single output, so a spec with more than one `output_cols` entry would previously build successfully but silently drop everything past the first only at layer-build time (B6) -- this fails at spec-parse time instead.

## Test plan

- [x] `pytest` on the full `native_transform` suite: 240 passed, 1 pre-existing unrelated skip
- [x] Manual sanity check of defaulting/validation behavior against the internal source

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/transform_layer_spec.py
+++ b/python/michelangelo/lib/native_transform/torch/transform_layer_spec.py
@@ -18,7 +18,7 @@ class themselves.
 from __future__ import annotations
 
 from enum import Enum
-from typing import ClassVar
+from typing import ClassVar, Optional, Union
 
 import torch
 from pydantic import BaseModel, Field, field_validator, model_validator
@@ -100,14 +100,18 @@ class TorchTransformLayerSpec(BaseModel, arbitrary_types_allowed=True):
     # model data; pydantic v2 checks ClassVar annotations before its
     # leading-underscore private-attribute inference, so this is not
     # swallowed into PrivateAttr handling.
-    _resolved_layer_type: ClassVar[str | None] = None
+    _resolved_layer_type: ClassVar[Optional[str]] = None
 
     name: str = Field(..., description="Name")
     input_cols: list[str] = Field(..., description="Input columns")
     output_cols: list[str] = Field(..., description="Output columns")
-    input_dtype: torch.dtype | str | None = Field(None, description="Input dtype")
-    output_dtype: torch.dtype | str | None = Field(None, description="Output dtype")
-    input_shape: list[int] | None = Field(
+    input_dtype: Optional[Union[torch.dtype, str]] = Field(
+        None, description="Input dtype"
+    )
+    output_dtype: Optional[Union[torch.dtype, str]] = Field(
+        None, description="Output dtype"
+    )
+    input_shape: Optional[list[int]] = Field(
         None,
         description=(
             "Static per-sample input shape (no batch dim) for each input "
@@ -115,7 +119,7 @@ class TorchTransformLayerSpec(BaseModel, arbitrary_types_allowed=True):
             "package the served model."
         ),
     )
-    is_training_features: list[bool] | None = Field(
+    is_training_features: Optional[list[bool]] = Field(
         None, description="Is training features"
     )
     mode: TransformerMode = Field(
@@ -187,7 +191,7 @@ class StandardScalerLayerSpec(TorchTransformLayerSpec):
         with_std: Whether to scale the data by the standard deviation.
     """
 
-    _resolved_layer_type: ClassVar[str | None] = "NormalizationLayerSpec"
+    _resolved_layer_type: ClassVar[Optional[str]] = "NormalizationLayerSpec"
 
     with_mean: bool = Field(
         True, description="Whether to center the data by subtracting the mean"
@@ -200,7 +204,7 @@ class StandardScalerLayerSpec(TorchTransformLayerSpec):
 class MinMaxScalerLayerSpec(TorchTransformLayerSpec):
     """Placeholder resolved to :class:`MinMaxLayerSpec` after fitting."""
 
-    _resolved_layer_type: ClassVar[str | None] = "MinMaxLayerSpec"
+    _resolved_layer_type: ClassVar[Optional[str]] = "MinMaxLayerSpec"
 
 
 class ConcatenateLayerSpec(TorchTransformLayerSpec):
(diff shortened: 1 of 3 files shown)


python/michelangelo/lib/native_transform/torch/transform_layer_spec.py
```diff
@@ -0,0 +1,551 @@
...
+    # Declared as ClassVar (not a pydantic field) so it isn't treated as
+    # model data; pydantic v2 checks ClassVar annotations before its
+    # leading-underscore private-attribute inference, so this is not
+    # swallowed into PrivateAttr handling.
+    _resolved_layer_type: ClassVar[str | None] = None
+
+    name: str = Field(..., description="Name")
+    input_cols: list[str] = Field(..., description="Input columns")
+    output_cols: list[str] = Field(..., description="Output columns")
+    input_dtype: torch.dtype | str | None = Field(None, description="Input dtype")
```
kechen2013: Wondering if this was tested on Python 3.9? `from __future__ import annotations` defers evaluation, but pydantic still evaluates field annotations when it builds the model, so `torch.dtype | str | None` throws `TypeError: unsupported operand type(s) for |` on 3.9. That's what's failing `python-build` here, and since it fails at import it aborts the whole test collection. Switching these back to `Optional`/`Union` like internal fixed it locally, with all 238 tests still passing.

sallycr: Confirmed, and thanks for catching this. Verified end-to-end (not just taking your word for it): ran a real Python 3.9 + real pydantic install against the exact PR file and reproduced the identical TypeError at TorchTransformLayerSpec's class construction (torch.dtype | str | None evaluates via eval() at pydantic's class-definition time, and PEP 604's `|` support on `type` only exists from 3.10 - from __future__ import annotations defers the Python-level parse but not pydantic's runtime eval). Pushed 8fee5849 converting every X | Y / X | None annotation in this file back to Optional/Union, matching internal's style. Re-ran the same 3.9 repro against the fix - imports cleanly now. 238 tests still pass under the normal dev venv, ruff clean.

One extra thing this surfaced: ruff's pyupgrade rules (UP007/UP045) want to convert Optional/Union back to X | Y for any file with `from __future__ import annotations`, regardless of target-version = py39 - it doesn't know pydantic will eagerly eval those annotations. Left alone, the next `ruff check --fix` would silently reintroduce this exact break. Added a per-file ruff ignore for UP007/UP045 scoped to just this file with a comment explaining why, so it can't regress silently.

python/michelangelo/lib/native_transform/torch/transform_layer_spec.py
```diff
@@ -0,0 +1,609 @@
...
+        """
+        if len(self.output_cols) != 1:
+            raise ValueError(
+                "output_cols must contain exactly one column, since the "
+                "input columns are concatenated into a single output."
+            )
+        return self
+
+
+class StandardScalerLayerSpec(TorchTransformLayerSpec):
```
kechen2013, line 254: Wondering if the placeholders should get this too? `StandardScalerLayerSpec`/`MinMaxScalerLayerSpec` subclass `TorchTransformLayerSpec` directly, so they skip the new validator and `StandardScalerLayerSpec(input_cols=["a","b"], output_cols=["c","d"])` still parses fine. Those are the hand-written ones pre-fit, so probably where it matters most.

kechen2013 (approved):

python/michelangelo/lib/native_transform/torch/transform_layer_spec.py
sallycr: Good catch — fixed in 49c5a926. Extracted the check into a shared `_require_single_output_col` helper and applied it to both `StandardScalerLayerSpec` and `MinMaxScalerLayerSpec` (their `PercentileBucketizationLayerSpec` sibling resolves to `BucketizationLayerSpec`, which has no such constraint, so it's untouched). Added regression tests for both placeholders; 30 tests passing in this file, ruff clean.

PR #1688, opened 2026-08-05
fix: make TriggerRun deletion idempotent

zhouxianzx:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

- Treat typed Temporal NotFound errors as successful cleanup when a TriggerRun schedule is already deleted or its workflow has already completed.
- Persist the terminal KILLED state through the Kubernetes status subresource before recording the drain-counted annotation.
- Add regression coverage for repeated Temporal cleanup and persisted TriggerRun status.

**Why?**

TriggerRun cascade deletion can successfully remove its Temporal resources while leaving the Kubernetes status as RUNNING or PAUSED because a normal resource update does not persist the status subresource. The next reconciliation retries deletion, receives NotFound for the already-deleted schedule, and previously treated that completed cleanup as a failure. This can leave the drain finalizer retrying until the safety timeout.

**How did you test it?**

- `bazel test //go/base/workflowclient/temporalclient:go_default_test //go/cascadedelete:go_default_test //go/components/triggerrun:go_default_test`

**Potential risks**

Low. The change ignores only typed Temporal NotFound errors, while all other schedule-deletion and workflow-termination errors still propagate. Status and metadata are persisted separately using their correct Kubernetes endpoints.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Fix TriggerRun cascade deletion so retries complete promptly after Temporal resources have already been removed.

**Documentation Changes**

N/A — internal controller behavior only.


go/components/triggerrun/cascade_test.go
```diff
@@ -62,6 +63,7 @@ func TestTriggerRun_RequestCancel_AtomicTokenAndKill(t *testing.T) {
 	got := &v2pb.TriggerRun{}
 	require.NoError(t, c.Get(context.Background(), types.NamespacedName{Name: "tr1", Namespace: "ns"}, got))
 	assert.Equal(t, "true", got.GetAnnotations()["cascade.michelangelo.uber.com/drain-counted"], "token persisted with the kill")
+	assert.Equal(t, v2pb.TRIGGER_RUN_STATE_KILLED, got.Status.State, "terminal status persisted through /status")
```
ishtoo1, line 67: **1. No test for the actual bug**

The tests cover the two fixes separately, but nothing checks that a drain finishes across two loops. Every test in `drain_test.go` uses `fakeTarget`, which only models metadata — it can't catch a dropped status. That's why this shipped, and collapsing `RequestCancel` back to one `Update` would still pass everything here.

Please add a test that calls `RunDrainStep` twice, re-fetching in between like `Reconcile` does, against a fake client with `WithStatusSubresource`, asserting the finalizer is gone after loop 2. ~50 lines; fails on `main`, passes here.

go/base/workflowclient/temporalclient/temporalclient.go
```diff
@@ -400,13 +402,24 @@ func (c *TemporalClient) UnpauseTrigger(ctx context.Context, workflowID string)
...
+	if err := c.Client.TerminateWorkflow(ctx, workflowID, runID, "trigger killed"); err != nil && !isTemporalNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+// isTemporalNotFound makes trigger deletion idempotent. Temporal reports both an
+// already-deleted schedule and an already-completed workflow as NotFound; neither
+// means there is remaining trigger work to drain.
+func isTemporalNotFound(err error) bool {
```
ishtoo1, line 438: **2. The swallow is silent**

`isTemporalNotFound` drops the error with no log or counter, so "schedule already gone" and "we built the wrong schedule ID" look identical in production. The ID is derived (`ns.name-schedule`), so if that ever breaks, kills silently report success.

Please log or count the swallowed NotFound.

go/cascadedelete/drain.go
```diff
@@ -71,8 +71,8 @@ type DrainState struct {
 // method mutates the child and persists the change via the controller's client;
 // RunDrainStep itself holds no client.
 type DrainTarget interface {
-	// RequestCancel begins cancellation AND stamps MarkDrainCounted in ONE
-	// persisted update.
+	// RequestCancel begins cancellation, persists its terminal status, and stamps
+	// MarkDrainCounted before returning.
```
ishtoo1: **3. New wording is wrong for PipelineRun**

`pipelineRunDrainTarget.RequestCancel` only sets `Spec.Kill = true` — no terminal status at that point, cancellation happens later in `Progress`. The old wording was wrong for TriggerRun, this one is wrong for PipelineRun.

Suggest:
> begins cancellation and stamps MarkDrainCounted, both durably persisted before returning. Kinds whose cancellation is carried in status MUST persist it through the status subresource.

ishtoo1 (commented):
Verified the fix — reproduced the wedge on `main`, confirmed it's gone here. Four comments, none blocking.

---

**4. `go/base/workflowclient/cadenceclient/cadenceclient.go:299` — same gap on Cadence**

(Leaving this here since the file isn't in the diff.)

Cadence avoids this by luck, not design:

```go
if runID == "" {
    return nil      // returns before touching anything
}
```

On a retry there's no open execution, so it exits early. Temporal deletes the schedule *before* that check, which is why only Temporal wedged.

But Cadence has no not-found handling either — if an execution is still open and the workflow finishes between the list and the terminate, you get the same wedge. Narrow race, fine as a follow-up, but worth noting so it isn't assumed handled.

go/components/triggerrun/cascade_test.go
zhouxianzx: Test case added

go/base/workflowclient/temporalclient/temporalclient.go
zhouxianzx: Fixed—swallowed NotFound errors are now logged with the schedule ID or workflow/run IDs for observability.

go/cascadedelete/drain.go
zhouxianzx: Updated the shared wording as suggested so it accurately covers both PipelineRun and TriggerRun cancellation semantics.

ishtoo1 (approved):

PR #1694, opened 2026-08-05
Add google.protobuf.Duration support to protoc-gen-validation

badcount:
Duration fields previously fell through validation silently: the (michelangelo.api.validation) min/max rule only supported numeric and enum fields, and Duration doesn't implement the custom Validate(string) error interface used for recursive message validation, so annotating a Duration field with min/max caused a generator-time compiler error and leaving it unannotated caused no validation to run at all.

- Add isDurationKind to recognize google.protobuf.Duration fields (including repeated and map-value positions) regardless of Go protobuf implementation.
- validateMaxMin now dispatches Duration fields to validateDurationMaxMin, interpreting min/max as seconds, consistent with how they're already interpreted as the target numeric type for int/float fields.
- Every Duration field unconditionally gets a structural validity check (nanos in range, sign matches seconds) regardless of whether a bound annotation is present, since a malformed Duration is never meaningful.
- Add ValidationMsg13 test fixture covering plain/bounded/repeated/map Duration fields.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Breaking Changes**

- [ ] No breaking changes

**Migration guide**

**Release notes**

**Documentation Changes**


go/cmd/kubeproto/protoc-gen-validation/main.go
```diff
@@ -399,6 +419,25 @@ func getScalarKind(field *protogen.Field, target targetType) protoreflect.Kind {
...
+		return false
+	}
+	switch target {
+	case fieldTarget, itemsTarget:
+		return field.Desc.Message().FullName() == durationMessageFullName
+	case valuesTarget:
+		return field.Desc.MapValue().Message().FullName() == durationMessageFullName
+	}
+	return false
+}
```
aritra0227, line 452: Is this how we should classify durations? Is there some evidence which guarantees that having `field.Desc.Message().FullName()` or `field.Desc.MapValue().Message().Fullname()` equal to google.protobuf.Duration is the correct way to classify durations?

From the definition here: https://protobuf.dev/reference/protobuf/google.protobuf/#duration
> A Duration represents a signed, fixed-length span of time represented as a count of seconds and fractions of seconds at nanosecond resolution

go/cmd/kubeproto/protoc-gen-validation/main.go
```diff
@@ -43,6 +43,17 @@ var validateMsg = `
 			}
 		}`
 
+// validateDurationStructural is emitted unconditionally for every google.protobuf.Duration
+// field, regardless of whether a bound annotation is present, since a structurally invalid
+// Duration (out-of-range nanos, or seconds/nanos with mismatched signs) is never meaningful.
+var validateDurationStructural = `
+		if v.GetNanos() <= -1e9 || v.GetNanos() >= 1e9 {
```
aritra0227: can we make constants for `-1e9` and `1e9`?
For example:
```
const MAX_NANOS = 1e9;
const MIN_NANOS = -1e9;
```

Also, from here: https://googleapis.dev/nodejs/pubsub/1.5.0/google.protobuf.html
it looks like, seconds must also fall under -315,576,000,000 to +315,576,000,000 inclusive. Can we have a check for that?

go/cmd/kubeproto/protoc-gen-validation/main.go
badcount: Yes — comparing `FullName()` to `google.protobuf.Duration` is the standard, implementation-agnostic way protobuf tooling recognizes well-known types, and it's exactly how the reference implementation does it: `google.golang.org/protobuf/encoding/protojson` dispatches its own Duration marshal/unmarshal special-casing by switching on the message's short name against `genid.Duration_message_name`, where that constant is defined as `protoreflect.FullName("google.protobuf.Duration")` in `internal/genid/duration_gen.go`.

This works regardless of which Go codegen backend produced the struct (golang/protobuf vs gogo/protobuf) because `FullName()` comes from the compiled `.proto` descriptor — declared once in `google/protobuf/duration.proto` as `package google.protobuf; message Duration` — not from the generated Go type. Protobuf's global file registry also guarantees no other `.proto` file can register a second message under that same full name, so there's no risk of a name collision misclassifying an unrelated message as a Duration.

I don't see a more robust signal available at the `protoreflect` level (there's no separate "is this a well-known type" bit on the descriptor) — this FullName check is the idiomatic approach.

aritra0227 (approved):

PR #1697, opened 2026-08-05
docs(pipelines): clarify architectural difference between remote-run and dev-run

### LINKED ISSUE #1670: docs: pipeline-running-modes.md doesn't explain how remote-run and dev-run differ architecturally
## Context

A community member asked why the docs weren't clear on the practical difference between `remote-run` and `pipeline dev-run` — both are described in [pipeline-running-modes.md](../user-guides/ml-pipelines/pipeline-running-modes.md) purely in terms of dev-lifecycle "why this mode exists" framing (e.g. "my laptop can't handle this dataset" vs. "does my Dockerfile actually work?"), which makes them look like two flavors of the same thing. They also guessed the difference was "remote-run requires a registered Pipeline entity" — which isn't accurate.

## What's missing

The correct distinction is already documented in [cli.md's "Differences between dev-run and remote-run"](../user-guides/reference/cli.md) section, but it isn't surfaced or cross-linked from `pipeline-running-modes.md`, which is the more discoverable entry point for this question. The missing piece is architectural, not just "which dev stage am I in":

- Whether the run is routed through the Michelangelo API server + controller (`dev-run`) vs. bypassing it and submitting straight to Cadence/Temporal (`remote-run`)
- Whether a `PipelineRun` entity gets created (`dev-run`: yes; `remote-run`: no Michelangelo entities of any kind)
- Consequently, whether the run is traceable, resumable (`--resume_from`), shareable by name, and whether it shows up in MA Studio

## Suggested fix

Add a section to `pipeline-running-modes.md`, e.g. right before "Decision tree: which mode should I use?":

```markdown
## Remote Run vs Pipeline Dev Run: what's actually different

From the "why this mode exists" framing above, Remote Run and Pipeline Dev Run can look interchangeable — both let you test against real compute without registering a pipeline. The distinction that actually matters when choosing between them is architectural, not just "which dev stage am I in":

| | Remote Run | Pipeline Dev Run |
| ----- | ----- | ----- |
| Invoked via | `python workflow.py remote-run` (a Uniflow feature, not an `ma` command) | `ma pipeline dev-run -f pipeline.yaml` |
| Routed through the Michelangelo API server + controller? | No — submits directly to Cadence/Temporal | Yes |
| Creates a `PipelineRun` entity? | No | Yes |
| Resumable with `--resume_from`? | No | Yes |
| Visible in MA Studio? | No | No — no parent `Pipeline` entity is created either, so there's nothing to attach it to in Studio's pipeline list. The run itself is still a real, queryable entity via `ma get pipeline_run` / kubectl |
| Requires a prebuilt image? | Yes — you pass `--image` | No — builds and validates the image as part of the run |

:::tip
If you'll need to come back to a run later, share it with a teammate, or resume it from a failed step, use **Pipeline Dev Run** — it's the only one of the two that creates a trackable entity. Reach for **Remote Run** only once you already trust your image and just need real compute for a quick check.
:::

Job scheduling itself is identical either way: both ultimately submit to the same Cadence/Temporal worker, which is what talks to the Ray/Spark compute cluster. The difference is entirely in what happens *before* that — whether the Michelangelo control plane is involved at all.

See the [CLI reference](../reference/cli.md#differences-between-dev-run-and-remote-run) for the full breakdown, including how `--file-sync` behaves differently under each mode.
```

Also worth a one-line back-link from `cli.md`'s "Differences between dev-run and remote-run" heading pointing up to this new section, so the two pages cross-reference each other.

Flagging as an issue rather than a PR since docs/ is your territory — happy to open the PR too if that's easier, just say the word.

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

- Added a new "Remote Run vs Pipeline Dev Run: what's actually different" section to `pipeline-running-modes.md` with a comparison table covering: invocation method, API server routing, `PipelineRun` entity creation, `--resume_from` support, MA Studio visibility, and image requirements. Includes a tip callout and a link to the CLI reference.
- Added a back-link at the top of the "Differences between dev-run and remote-run" section in `cli.md` pointing back to the new section.

**Why?**

A community member was confused about the practical difference between `remote-run` and `pipeline dev-run` because `pipeline-running-modes.md` only framed the distinction in terms of dev lifecycle stage, not architecture. The correct technical explanation existed in `cli.md` but was not surfaced from the more discoverable entry point. Closes #1670.

**How did you test it?**

Verified table content against `cli.md`'s existing "Differences between dev-run and remote-run" section. Docs-only change; no functional code affected.

**Potential risks**

None.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

- `docs/user-guides/ml-pipelines/pipeline-running-modes.md` — new architectural comparison section before the decision tree
- `docs/user-guides/reference/cli.md` — back-link added at "Differences between dev-run and remote-run" heading

### DIFF
--- a/docs/user-guides/ml-pipelines/pipeline-running-modes.md
+++ b/docs/user-guides/ml-pipelines/pipeline-running-modes.md
@@ -189,6 +189,27 @@ ma pipeline run --namespace="my-project" --name="my-pipeline"
 ma pipeline run --namespace="my-project" --name="my-pipeline" --resume_from=previous-run:step
 ```
 
+## Remote Run vs Pipeline Dev Run: what's actually different
+
+From the "why this mode exists" framing above, Remote Run and Pipeline Dev Run can look interchangeable — both let you test against real compute without registering a pipeline. The distinction that actually matters when choosing between them is architectural, not just "which dev stage am I in":
+
+| | Remote Run | Pipeline Dev Run |
+| ----- | ----- | ----- |
+| Invoked via | `python workflow.py remote-run` (a Uniflow feature, not an `ma` command) | `ma pipeline dev-run -f pipeline.yaml` |
+| Routed through the Michelangelo AI API server + controller? | No — submits directly to Cadence/Temporal | Yes |
+| Creates a `PipelineRun` entity? | No | Yes |
+| Resumable with `--resume_from`? | No | Yes |
+| Visible in MA Studio? | No | No — no parent `Pipeline` entity is created either, so there's nothing to attach it to in Studio's pipeline list. The run itself is still a real, queryable entity via `ma get pipeline_run` / kubectl |
+| Requires a prebuilt image? | Yes — you pass `--image` | No — builds and validates the image as part of the run |
+
+:::tip
+If you'll need to come back to a run later, share it with a teammate, or resume it from a failed step, use **Pipeline Dev Run** — it's the only one of the two that creates a trackable entity. Reach for **Remote Run** only once you already trust your image and just need real compute for a quick check.
+:::
+
+Job scheduling itself is identical either way: both ultimately submit to the same Cadence/Temporal worker, which is what talks to the Ray/Spark compute cluster. The difference is entirely in what happens *before* that — whether the Michelangelo AI control plane is involved at all.
+
+See the [CLI reference](../reference/cli.md#differences-between-dev-run-and-remote-run) for the full breakdown, including how `--file-sync` behaves differently under each mode.
+
 ## Decision tree: which mode should I use?
 
 ### Stage-based
--- a/docs/user-guides/reference/cli.md
+++ b/docs/user-guides/reference/cli.md
@@ -299,6 +299,8 @@ ma pipeline dev-run -f "./examples/bert_cola/pipeline.yaml" --file-sync --storag
 
 ##### Differences between dev-run and remote-run
 
+> For an architectural overview of how these two modes differ, see [Remote Run vs Pipeline Dev Run: what's actually different](../ml-pipelines/pipeline-running-modes.md#remote-run-vs-pipeline-dev-run-whats-actually-different).
+
 **1. dev-run: Test Pipeline from Local File**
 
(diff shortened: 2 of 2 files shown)


paulzim (approved):
**Approve** ✅

Verified this against the underlying issue (#1670) and the existing `cli.md` reference text:

- The new "Remote Run vs Pipeline Dev Run" table's claims all cross-check against pre-existing, unmodified content in `cli.md`:
  - "Routed through Michelangelo AI API server?" → matches `cli.md`'s "remote-run ... bypasses the Michelangelo AI API server and submits your workflow directly to Cadence/Temporal"
  - "Creates a `PipelineRun` entity?" → matches "No Michelangelo AI entities are created" for remote-run
  - "Requires a prebuilt image?" (Yes for Remote Run) → matches the existing `--image` flag documented in the Remote Run mode section
  - "Resumable with `--resume_from`?" (No for Remote Run, Yes for Pipeline Dev Run) → matches: `--resume_from` only ever appears alongside `dev-run`/`run` examples, never `remote-run`, elsewhere in these docs
- Both new anchor links resolve correctly against Docusaurus's standard slug generation:
  - `cli.md` → `pipeline-running-modes.md#remote-run-vs-pipeline-dev-run-whats-actually-different`
  - `pipeline-running-modes.md` → `cli.md#differences-between-dev-run-and-remote-run`
- `:::tip` admonition syntax matches existing usage elsewhere in `docs/` (e.g. `notifications.md`, `backfill.md`, `model-registry-guide.md`).
- Branding is consistently "Michelangelo AI" throughout, matching the rest of both files (this PR correctly uses "Michelangelo AI" where my original issue's suggested-fix draft had used plain "Michelangelo").
- Docs-only change (2 files, +23/-0), all CI checks green (Docs Check, API Surface Check, CLA).

Nice fast turnaround — this closes #1670 essentially as originally proposed there, cross-checked and technically sound.

PR #1711, opened 2026-08-07
ci: add least-privilege permissions to workflows missing them

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
Adds a top-level `permissions: contents: read` block to the four workflows that currently declare no permissions at all: `main.yml` (Bazel Test), `helm-lint.yaml`, `docs-check.yml`, and `ui-pre-land.yaml`. Every other workflow in the repo already declares explicit permissions; this brings the last four in line with that convention (placement matches `api-surface-check.yml`).

**Why?**
A workflow with no `permissions` block runs with the repository's default token grant, which can be much broader than what the jobs need. These four only check out code and run tests/lints - none of them write to the repo, comment on PRs, or publish anything - so `contents: read` is sufficient. Scoping the token limits what a compromised action or dependency in these workflows could do, and is the standard hardening recommendation (GitHub's own docs and OpenSSF Scorecard's Token-Permissions check).

**How did you test it?**
Verified each of the four workflows job-by-job: every step is checkout, toolchain setup, cache, build/test, or lint - no step uses `GITHUB_TOKEN` for anything beyond fetching the repo (the `bufbuild/buf-setup-action` step in `main.yml` takes `github_token` only for rate-limited reads). Confirmed no other workflow in `.github/workflows/` is missing a permissions block. YAML validated by parsing all four files.

**Potential risks**
If a future step added to one of these workflows needs a write scope (for example commenting on a PR), it will now fail with a clear permissions error until that scope is added explicitly - which is the intended behavior of least privilege, and the fix is a one-line addition at the job level.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.

### DIFF
--- a/.github/workflows/docs-check.yml
+++ b/.github/workflows/docs-check.yml
@@ -8,6 +8,9 @@ on:
       - 'docs/**'
       - 'website/**'
 
+permissions:
+  contents: read
+
 jobs:
   validate-docs:
     runs-on: ubuntu-latest
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -11,6 +11,9 @@ on:
     paths:
       - 'helm/**'
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     name: Lint and template check
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,9 @@ on:
       - 'website/**'
       - 'javascript/**'
 
+permissions:
+  contents: read
+
 jobs:
   bazel-build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ui-pre-land.yaml
+++ b/.github/workflows/ui-pre-land.yaml
@@ -10,6 +10,9 @@ on:
     paths:
       - 'javascript/**'
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     name: Lint & Typecheck


paulzim (approved):
**Approve** — clean, well-scoped least-privilege hardening.

Verified:
- Diff is exactly what's described: adds a top-level `permissions: contents: read` block to the 4 workflows that had none (`main.yml`, `helm-lint.yaml`, `docs-check.yml`, `ui-pre-land.yaml`) — 4 files, +12/-0, no other scope creep. Placement matches the existing `api-surface-check.yml` convention.
- Read every job in all 4 files: each step is checkout, toolchain setup, cache, build/test, or lint. Nothing pushes to the repo, comments on a PR, or publishes an artifact. `bufbuild/buf-setup-action`'s `github_token` input in `ui-pre-land.yaml` is only used for rate-limited GitHub API reads (release lookups), not writes — `contents: read` covers all of it.
- Checked the "no other workflow is missing a permissions block" claim: 12 other workflow files don't have a *top-level* `permissions:` key, but all of them declare `permissions:` per-job instead (e.g. `release.yaml`, `npm-publish.yml`, `nightly.yml`) — so the claim holds; these 4 were genuinely the only ones with zero declaration at any level.
- All 4 files parse as valid YAML (`yaml.safe_load`).

No functional risk — this only narrows the default `GITHUB_TOKEN` grant for jobs that already only read.

PR #1730, opened 2026-08-08
feat(native_transform): add TransformSpec DAG engine + TorchTransformModule (PR B6)

sallycr:
## Summary

This is **PR B6** of the `native_transform` torch-core migration — the assembler-unblock milestone. It delivers:

- `lib/native_transform/torch/transform_spec.py` — the `TORCH_TRANSFORM_LAYERS_DICT` / `TORCH_TRANSFORM_LAYERS_SPECS_DICT` / `TORCH_TRANSFORM_LAYER_CLASS_NAME_TO_SPEC_CLASS_NAME_DICT` registries (spanning all 21 layer classes from PRs B1-B4 and all spec classes from B5), the `TransformSpec` class (topological sort into dependency levels, the `update_*` placeholder-hydration methods, `get_numerical_statistics_computation_specs`, and JSON/dict ser-de), and
- `lib/native_transform/torch/base_transform_module.py` — `TorchTransformModule` (the TorchScript-able `nn.Module` that runs a DAG of transform layers in topological order) and `get_transform_module` (materializes a level range of a `TransformSpec` into one).

Together these are the four+one symbols the `tabular_assembler` torch fuser needs (`TransformSpec`, `TORCH_TRANSFORM_LAYERS_DICT`, `TorchTransformModule`, `generate_layer_name`, `TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP`) — this PR flips that fuser from gated to functional.

## Design deviation from internal (flagged per migration design doc)

Internal's `TransformSpec` dispatches placeholder hydration and fitted-category detection by parsing a **name-string prefix** off `layer_spec.name` (the lowercased class name from `generate_layer_name()`) — flagged in the OSS design doc as "the single riskiest hidden contract in the whole migration," since it breaks silently if layer naming ever drifts.

This port instead dispatches via **`isinstance()`** against the spec's real pydantic type (e.g. `isinstance(layer_spec, StandardScalerLayerSpec)`), since every spec here is a concrete, statically-known pydantic class rather than an untyped dict-like object. This is strictly more robust — it cannot silently break regardless of naming — and eliminates the hidden contract entirely rather than just preserving it. Verified behaviorally equivalent to internal for every hydration path exercised by the ported tests, and added an explicit test (`test_hydration_dispatches_via_generated_name_placeholder`) that builds a placeholder via a raw spec dict (auto-generating its name through the same `generate_layer_name()` used elsewhere), confirms the generated name format, and confirms hydration still resolves it end to end.

## Testing

- 346 tests passed, 1 skipped (pre-existing skip, unrelated) — full `lib/native_transform/` suite, no regressions.
- 99.27% combined statement+branch coverage on the two new files (99.18% `transform_spec.py`, 100% `base_transform_module.py`).
- Added a `torch.jit.script` + save/load round-trip test for `TorchTransformModule` (internal has no such test for this module).

## Scope notes

- Does not touch the Ray-based batch inference/orchestration code (`lib/native_transform/ray/`, a separate PR), the `TransformSpec` <-> `ModelSchema` bridge (`schema.py`, PR B7), or `scale.py`/`duration.py`/`transform_utils.py` (PR B8, in progress in parallel).

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/transform_spec.py
+++ b/python/michelangelo/lib/native_transform/torch/transform_spec.py
@@ -264,10 +264,9 @@ class TransformSpec:
                 layer_spec.input_cols[0], torch.float32
             )
             # Regenerate the layer spec so any dtype-dependent validators re-run.
-            spec_cls = TORCH_TRANSFORM_LAYERS_SPECS_DICT[
-                layer_spec.__class__.__name__.replace("LayerSpec", "")
-            ]
-            self.transform_specs[layer_name] = spec_cls(**layer_spec.model_dump())
+            self.transform_specs[layer_name] = type(layer_spec)(
+                **layer_spec.model_dump()
+            )
 
     def update_numerical_standard_transform_parameters(
         self, percentile_dict: dict[str, float], target_transform_level: int = 0
--- a/python/michelangelo/lib/native_transform/torch/transform_spec.py
+++ b/python/michelangelo/lib/native_transform/torch/transform_spec.py
@@ -1,0 +1,911 @@
+"""The ``TransformSpec`` DAG engine.
+
+Bridges the two parallel representations of a native transform: a
+declarative, serializable ``*LayerSpec`` (see
+:mod:`~michelangelo.lib.native_transform.torch.transform_layer_spec`) and an
+executable ``nn.Module`` layer (see
+:mod:`~michelangelo.lib.native_transform.torch.base_layers` and
+:mod:`~michelangelo.lib.native_transform.torch.stats_layers`).
+
+:class:`TransformSpec` parses a raw dict (typically loaded from YAML) of
+layer specs, topologically sorts them into dependency "levels" by their
+input/output column references, and exposes per-level accessors used to
+drive fit-then-materialize training: compute the statistics a level's
+placeholder specs need
+(:meth:`TransformSpec.get_numerical_statistics_computation_specs`), hydrate
+those placeholders into concrete specs once the statistics are available
+(the ``update_*`` methods), then materialize the level as executable layers
+(:meth:`TransformSpec.to_transform_layers`).
+
+Dispatch for placeholder hydration and this fitted category detection is by
+``isinstance()`` against the spec's real pydantic type. This is a deliberate
+change from the internal implementation, which dispatched by parsing a
+name-string prefix off ``layer_spec.name`` (the lowercased class name, from
+:func:`~michelangelo.lib.native_transform.torch.utils.generate_layer_name`).
+Since every spec here is a concrete, statically-known pydantic class,
+``isinstance()`` is strictly more robust: it cannot silently break if a
+layer's name or ``generate_layer_name()``'s output format ever changes,
+which the string-prefix approach could.
+"""
+
+from __future__ import annotations
+
+import json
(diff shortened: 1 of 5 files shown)


python/michelangelo/lib/native_transform/torch/transform_spec.py
```diff
@@ -0,0 +1,911 @@
...
+
+Dispatch for placeholder hydration and this fitted category detection is by
+``isinstance()`` against the spec's real pydantic type. This is a deliberate
+change from the internal implementation, which dispatched by parsing a
+name-string prefix off ``layer_spec.name`` (the lowercased class name, from
+:func:`~michelangelo.lib.native_transform.torch.utils.generate_layer_name`).
+Since every spec here is a concrete, statically-known pydantic class,
+``isinstance()`` is strictly more robust: it cannot silently break if a
+layer's name or ``generate_layer_name()``'s output format ever changes,
+which the string-prefix approach could.
```
kechen2013, line 28: Wondering if it's worth pinning a test for a placeholder with an explicit `name`? Looks like the only case where this diverges from internal — the prefix check there misses it, so the placeholder never hydrates and `NumericalStandardTransform` keeps its percentile strings as cap values.

python/michelangelo/lib/native_transform/torch/transform_spec.py
```diff
@@ -0,0 +1,911 @@
...
+                    "data dtype dict"
+                )
+            layer_spec.input_dtype = data_dtype_dict.get(
+                layer_spec.input_cols[0], torch.float32
+            )
+            # Regenerate the layer spec so any dtype-dependent validators re-run.
+            spec_cls = TORCH_TRANSFORM_LAYERS_SPECS_DICT[
+                layer_spec.__class__.__name__.replace("LayerSpec", "")
+            ]
+            self.transform_specs[layer_name] = spec_cls(**layer_spec.model_dump())
```
kechen2013: Wondering if `type(layer_spec)(**layer_spec.model_dump())` would work here instead? The dict lookup only resolves for the placeholders because their `transform_name` keys happen to match the class-derived names, so it's still name-coupled like the prefix check was.

python/michelangelo/lib/native_transform/torch/transform_spec.py
sallycr: Added in cc0ccb84 -- two tests locking in that a placeholder with an explicit, non-generated `name` still hydrates correctly via `isinstance()`, for both `update_standard_scaler_specs` and `update_numerical_standard_transform_parameters` (the NumericalStandardTransform case you flagged). Confirmed these fail against a hypothetical name-prefix-based dispatch and pass against the current isinstance()-based one.

python/michelangelo/lib/native_transform/torch/transform_spec.py
sallycr: Good point -- fixed in cc0ccb84. Replaced the name-keyed `TORCH_TRANSFORM_LAYERS_SPECS_DICT[layer_spec.__class__.__name__.replace("LayerSpec", "")]` lookup with `type(layer_spec)`, which gives the same class directly with no name coupling.

kechen2013 (approved):

Issue #1745, opened 2026-08-08
Bump gazelle from 0.40.0 to a current release compatible with rules_go 0.57.0


sallycr:
## Current state

`MODULE.bazel` currently pins:

```
bazel_dep(name = "rules_go", version = "0.57.0", repo_name = "io_bazel_rules_go")
bazel_dep(name = "gazelle", version = "0.40.0", repo_name = "bazel_gazelle")
```

`gazelle` is pinned at `0.40.0`, which is now well behind both the current `rules_go` pin and the latest `bazel-gazelle` release (`v0.52.2` as of this writing — https://github.com/bazel-contrib/bazel-gazelle/releases). Given this repo generates Go code from proto (`proto/api/v2/*.proto`, consumed via `go_proto_library`/custom `protoc-gen-*` plugins), keeping `gazelle` current with `rules_go` reduces the risk of `BUILD.bazel` generation drifting out of sync with what `rules_go` 0.57.0 expects.

## Proposed change

Bump `gazelle` from `0.40.0` to a `rules_go`-0.57.0-compatible release. Releases `0.41.0` through `0.44.0` don't show documented breaking changes in their release notes. Recommend confirming the current latest stable release at implementation time (`v0.52.2` today) rather than hardcoding a target version in this issue, since gazelle ships frequently.

## Known risk to verify against

bazel-contrib/bazel-gazelle#2252 is an **open, unresolved** report that bumping gazelle from `0.44.0` to `0.45.0` (and even `0.47.0`) broke builds that generate Go code from proto — specifically via changed `proto_library`/`go_proto_library` dependency rewiring and gRPC compiler target references (e.g. `@rules_go//proto:go_grpc` vs `@io_bazel_rules_go//proto:go_grpc_v2`). A related-but-distinct fix (bazel-contrib/bazel-gazelle#2326, landed in `v0.51.0`) addresses compiler label resolution for the `rules_go` Bzlmod apparent name, but #2252 itself remains open, so it isn't clear that upgrading past `v0.51.0` fully avoids the underlying dependency-rewiring risk.

Since this repo has real `proto_library`/`go_proto_library` targets, this bump should not be a blind version-pin change. Before merging:

1. Re-run `bazel run //:gazelle` repo-wide and carefully diff the generated `BUILD.bazel` changes, paying particular attention to any `proto_library`/`go_proto_library` targets — deps and `compilers` attributes especially.
[...]

PR #1567, opened 2026-07-21
fix(build): bump Go SDK to 1.26.3 + rules_go to 0.57.0 (CVE Scan fix, step 1/2)

sallycr:
## Summary

Fixes #1565 (partially — see "Two-step plan" below).

`MODULE.bazel` pinned the actual Bazel-built Go toolchain to `1.23.3`, independently of `go/go.mod`'s `go 1.24.0` directive (a minimum-compat floor, not a toolchain pin). This left 15 known Go stdlib CVEs (1 CRITICAL, 14 HIGH) baked into every Go-built service image (`apiserver`, `controllermgr`, `worker`) — see the failing [`CVE Scan` run](https://github.com/michelangelo-ai/michelangelo/actions/runs/29853464881).

This bumps:
- `go_sdk.download(version = ...)`: `1.23.3` → `1.26.3` (this org's internal Go version standard)
- `rules_go`: `0.50.1` → `0.57.0` (required companion bump — see below)

### Why rules_go also needed a bump

`rules_go` 0.50.1 doesn't build cleanly against Go 1.26.3:
1. It hardcodes `GOEXPERIMENT=coverageredesign`, which Go 1.25+ no longer recognizes as a valid experiment (`go: unknown GOEXPERIMENT coverageredesign`). Fixed in [rules_go#4397](https://github.com/bazel-contrib/rules_go/commit/2af3f27a), first released in `v0.56.0`.
2. It expects a prebuilt `pack` tool under `pkg/tool/<platform>/`, which newer Go SDK releases no longer ship there (`pack` moved to `bin/`). `rules_go` `v0.57.0` fixes this by building `pack` from source instead ([commit](https://github.com/bazel-contrib/rules_go/commit/4d0a821e)).

`v0.57.0` is the first `rules_go` release containing both fixes.

## Two-step plan

This is **step 1 of 2** (see #1565 for full context). `1.26.3` does not close every CVE found in the scan — 3 confirmed CVEs remain open at this version:

| CVE | Needs |
|---|---|
| CVE-2026-27145 | 1.26.4 |
| CVE-2026-42504 | 1.26.4 |
| CVE-2026-39822 | 1.26.5 |

Step 2 (planned ~1 week after this merges, to let this bump soak in CI/nightly builds first) will bump `1.26.3` → `1.26.5`, closing the remaining CVEs and turning the `CVE Scan` workflow fully green.

## Test plan

- [x] `bazel build //go/...` succeeds against the new SDK + `rules_go` versions.
[...]

PR #1747, opened 2026-08-09
build: bump gazelle 0.40.0 to 0.52.2, compatible with rules_go 0.57.0

sallycr:
Part of #1745.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Chore / build tooling

**What changed?**
Bumps `gazelle` from `0.40.0` to `0.52.2` in `MODULE.bazel`, and re-runs `bazel run //:gazelle` to regenerate `BUILD.bazel` files across the repo. `rules_go` was already bumped to `0.57.0` (#1567), and `gazelle` had fallen several releases behind — this repo's Bazel build depends on `rules_go`/`gazelle` staying close together for `BUILD.bazel` generation to stay correct.

Re-running gazelle after the bump produced two categories of change:

1. `proto_library`'s load source switched from `@rules_proto//proto:defs.bzl` to `@com_google_protobuf//bazel:proto_library.bzl` across 8 proto packages. No dependency or `go_grpc` compiler target changes — this is the newer gazelle preferring protobuf's own Bzlmod-native `proto_library` now that `protobuf` is a direct `bazel_dep` in this repo.
2. Gazelle wanted to strip the hand-written `gc_linkopts` static-linking block from `apiserver`/`controllermgr`/`worker`'s `go_binary` rules — that block is required for these binaries to run in the distroless base image (see the inline comment in each `BUILD.bazel`). The newer gazelle version treats `gc_linkopts` as a gazelle-managed attribute for `go_binary` and removes values it didn't generate itself. Fixed with a `# keep` marker above each `gc_linkopts` assignment (gazelle's documented mechanism for protecting a manually-set attribute) — verified by re-running gazelle again and confirming only the `# keep` comment was added, the block itself untouched.

**Why?**
Closes the `gazelle` version gap tracked in #1745, and specifically checks for the `proto_library`/`go_proto_library` dependency-rewiring regression reported in [bazel-contrib/bazel-gazelle#2252](https://github.com/bazel-contrib/bazel-gazelle/issues/2252) (still open/unresolved upstream) before landing — this repo has real proto-generated Go code, so that regression pattern was concretely worth checking for. It did not occur here (the only proto-related change is the `proto_library` load-source switch above, no target/dependency rewiring), but the static-linking-flag stripping was a separate, real regression this bump did trigger, now fixed.

**How did you test it?**
- `bazel build //...` — 5908 actions, full repo build succeeds.
- `bazel test //...` — 94/94 tests pass.
- `gofmt -l .` (scoped to `go/`) — clean.
- `golangci-lint run ./...` — clean (only pre-existing `godox` TODO-tracking findings in files this PR doesn't touch).
- Manually diffed the gazelle-regenerated output twice: once before adding `# keep` (confirmed the `gc_linkopts` stripping issue), once after (confirmed only the `# keep` comment was added, no other changes).

**Potential risks**
Low, but worth a careful review of the `BUILD.bazel` diff specifically: this is a generated-file-heavy change (gazelle can touch any package on a version bump), even though the effective behavior change here is narrow (a `proto_library` load-source swap with no semantic difference, plus a `# keep` marker preserving existing behavior). No source-code changes.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A — build-tooling change only, no functional or schema change.

**Documentation Changes**
N/A.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,7 +5,7 @@ module(name = "michelangelo")
 ###########################################
 bazel_dep(name = "rules_python", version = "0.40.0")
 bazel_dep(name = "rules_go", version = "0.57.0", repo_name = "io_bazel_rules_go")
-bazel_dep(name = "gazelle", version = "0.40.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "gazelle", version = "0.52.2", repo_name = "bazel_gazelle")
 bazel_dep(name = "rules_python_gazelle_plugin", version = "0.40.0")
 bazel_dep(name = "rules_pkg", version = "1.0.1")
 bazel_dep(name = "rules_oci", version = "2.0.1")
--- a/go/cmd/apiserver/BUILD.bazel
+++ b/go/cmd/apiserver/BUILD.bazel
@@ -47,6 +47,7 @@ go_binary(
     env = {
         "CONFIG_DIR": "$(location :config)",
     },
+    # keep
     gc_linkopts = select({
         "@io_bazel_rules_go//go/platform:linux": [
             # The following GC options instruct Bazel to build a static binary for running the service in a static
--- a/go/cmd/controllermgr/BUILD.bazel
+++ b/go/cmd/controllermgr/BUILD.bazel
@@ -56,6 +56,7 @@ go_binary(
     env = {
         "CONFIG_DIR": "$(location :config)",
     },
+    # keep
     gc_linkopts = select({
         "@io_bazel_rules_go//go/platform:linux": [
             # The following GC options instruct Bazel to build a static binary for running the service in a static
--- a/go/cmd/worker/BUILD.bazel
+++ b/go/cmd/worker/BUILD.bazel
@@ -32,6 +32,7 @@ go_binary(
     env = {
         "CONFIG_DIR": "$(location :config)",
     },
+    # keep
     gc_linkopts = select({
         "@io_bazel_rules_go//go/platform:linux": [
             # The following GC options instruct Bazel to build a static binary for running the service in a static
--- a/proto/api/BUILD.bazel
+++ b/proto/api/BUILD.bazel
@@ -1,6 +1,6 @@
+load("@com_google_protobuf//bazel:proto_library.bzl", "proto_library")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
-load("@rules_proto//proto:defs.bzl", "proto_library")
 
 proto_library(
     name = "api_proto",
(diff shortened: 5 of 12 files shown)


sallycr:
## Smoke test: local k3d sandbox, real pipelines end-to-end

**Commit tested:** `cb2eae564aa5f59dd9b07c2fd7b3aceade03de3a` (branch `gazelle-bump-rules-go-compat`)

**What was verified:** that `apiserver`/`controllermgr`/`worker` binaries built from this branch's *generated* BUILD.bazel files — specifically the gazelle-regenerated `proto_library` loads and the `# keep`-preserved `gc_linkopts` static-linking block — actually link, run, and execute real workloads end-to-end, not just that `bazel build`/`bazel test` succeed (which the PR author already confirmed).

**Method:**
- Built `//go/cmd/apiserver`, `//go/cmd/controllermgr`, `//go/cmd/worker` natively on linux/arm64 (the static/cgo linking this PR's `# keep` marker preserves requires a real ELF/glibc target — cross-compiling from macOS isn't viable for this link mode) and packaged them into `docker/service.Dockerfile` images.
- Confirmed each binary boots correctly under the distroless base image (`gcr.io/distroless/base-debian12`) — i.e. static linking survived the gazelle-regenerated BUILD.bazel unchanged.
- Deployed the images into a local k3d sandbox (`ma sandbox`) via `helm upgrade --set images.<svc>.repository/tag=...`, replacing the running `apiserver`/`controllermgr`/`worker` deployments.
- Ran two full example pipelines from `michelangelo-examples` (project `california-housing`) against the sandbox:
  - **`xgb-train`** (Ray + Spark XGBoost): run `run-20260809-030701-f7d54023` → **`PIPELINE_RUN_STATE_SUCCEEDED`**
  - **`pytorch-train`** (Ray + Spark, PyTorch Lightning): run `run-20260809-031040-85dbff8f` → **`PIPELINE_RUN_STATE_SUCCEEDED`**

**Note on one transient failure:** the first `xgb-train` attempt failed on a dataset-fetch step with `SSL: WRONG_VERSION_NUMBER` — this was a local corporate proxy/TLS-interception artifact of my sandbox environment (outbound HTTPS through a transparent proxy), not related to this PR's code or BUILD.bazel changes. A retry with the same images succeeded cleanly.

**What this confirms / doesn't confirm:** this is a smoke test against two example workloads in one local sandbox, not a substitute for CI. It confirms the binaries produced from this branch's regenerated BUILD.bazel — including the preserved static-linking `gc_linkopts` and the switched `proto_library` load source — build correctly and execute real pipeline workloads successfully end-to-end. Combined with the PR author's `bazel build //...` (5908 actions) and `bazel test //...` (94/94) results, I have no reservations about this change.

badcount (approved):

Issue #1748, opened 2026-08-09
Remove core's stale california_housing_xgb example, repoint docs to michelangelo-examples' xgb_train


sallycr:
## Summary

`python/examples/pipelines/california_housing_xgb/` is dead weight in core `michelangelo` — it's been fully ported to `michelangelo-examples` as `xgb_train` (#1440... actually see michelangelo-examples PR history), and every doc in core that still references the old core-hosted example needs to be repointed there instead.

## Background

Core's `california_housing_xgb` example predates `michelangelo-examples` (the dedicated satellite repo for runnable pipeline examples). Once the example was migrated there as `xgb_train`, core's copy became a stale duplicate — worse, several docs (including the main `getting-started.md` tutorial) still walk new users through the old, soon-to-be-removed core copy.

## Scope

- Delete `python/examples/pipelines/california_housing_xgb/` (8 files) and the now-empty parent `pipelines/__init__.py`.
- Repoint 5 doc files to `michelangelo-examples`' `xgb_train` instead of dropping the walkthrough entirely:
  - `python/README.md`
  - `docs/user-guides/examples/index.md`
  - `docs/user-guides/getting-started/getting-started.md` (the main hands-on tutorial — needs a real command-by-command rewrite, not just a link swap)
  - `docs/user-guides/ml-pipelines/workflow-patterns.md`
  - `docs/getting-started/sandbox-setup.md`
- Remove `xgboost`/`xgboost_ray` from `pyproject.toml`'s optional dependencies and `example` extra — a repo-wide import grep confirms these packages have zero remaining consumers once the pipeline is deleted.
- **Not** touched: `python/michelangelo/cli/sandbox/resources/spark.yaml`/`spark-app.yaml` (static illustrative fixtures, no runtime consumer) and `python/examples/config/project.yaml` (still a live dependency — registers the `ma-examples` project used by other, unrelated examples and docs).

## Verification

- Repo-wide `grep -rn "california_housing_xgb"` after the change returns only the two intentionally-kept fixture references above.
- `ruff check`/`ruff format --check` clean on changed Python files.
- Existing `api_client_test.py` suite (25 tests) passes unchanged.

PR #1746, opened 2026-08-09
docs: repoint california_housing examples to michelangelo-examples

sallycr:
Closes https://github.com/michelangelo-ai/michelangelo/issues/1748

**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Deletes the `python/examples/pipelines/california_housing_xgb/` pipeline (8 files) and its now-empty parent `__init__.py`, since it has been fully ported to `michelangelo-examples` as `xgb_train` (see feature 013). Repoints five doc files (`python/README.md`, `docs/user-guides/examples/index.md`, `docs/user-guides/getting-started/getting-started.md`, `docs/user-guides/ml-pipelines/workflow-patterns.md`, `docs/getting-started/sandbox-setup.md`) to reference the `michelangelo-examples` version instead. Removes `xgboost` and `xgboost_ray` from `python/pyproject.toml`'s optional dependencies and `example` extra, since the deleted pipeline was their only consumer. Also swaps a california-housing-specific docstring in `api_client.py` and a test fixture string in `api_client_test.py` for generic placeholders, since they no longer needed to reference this pipeline by name.

Intentionally **not** touched:
- `python/michelangelo/cli/sandbox/resources/spark.yaml` / `spark-app.yaml` — static illustrative fixtures with no runtime consumer.
- `python/examples/config/project.yaml` — still a live dependency for `workflow-patterns.md` and other docs, unrelated to the deleted pipeline.
- `CHANGELOG.md` — historical record, out of scope.

**Why?**

`california_housing_xgb` was ported to `michelangelo-examples` and no longer needs to live in core. Keeping a duplicate pipeline in core alongside the actively-maintained example in `michelangelo-examples` is confusing for new users and creates maintenance drift (e.g. unused `xgboost`/`xgboost_ray` dependencies). This cleanup keeps core docs and dependencies pointed at the single source of truth.

**How did you test it?**

- `grep -rn "california_housing_xgb"` — only the two intentionally-kept fixture references remain (`spark.yaml`, `spark-app.yaml`), plus one historical `CHANGELOG.md` entry.
- `grep -rn "california_housing_lightning"` — zero hits.
- `ruff check` and `ruff format --check` on the two changed Python files — pass, already formatted.
- `python -m pytest python/michelangelo/lib/model_manager/registry/tests/api_client_test.py -x` — 25/25 passed.
[...]

PR #1749, opened 2026-08-09
docs: repoint california_housing doc references to michelangelo-examples

sallycr:
Part of https://github.com/michelangelo-ai/michelangelo/issues/1748

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Repoints four doc files (`docs/getting-started/sandbox-setup.md`, `docs/user-guides/examples/index.md`, `docs/user-guides/getting-started/getting-started.md`, `docs/user-guides/ml-pipelines/workflow-patterns.md`) to reference the `michelangelo-examples` version of the `california_housing_xgb` pipeline (now `xgb_train`) instead of the in-core copy.

This is part 1 of a 3-way split of #1746, which bundled these doc changes with unrelated `python/michelangelo/lib` docstring/test updates and the pipeline deletion itself. Splitting lets each group be reviewed independently. See also:
- chore(lib): update california_housing example references in model_manager registry docstrings/tests
- chore(examples): remove california_housing_xgb pipeline (migrated to michelangelo-examples)

**Why?**

`california_housing_xgb` was ported to `michelangelo-examples` and no longer needs to live in core. Keeping stale doc links pointed at the deleted in-core pipeline would break the docs once the pipeline is removed.

**How did you test it?**

- Manually reviewed each repointed doc link against the `michelangelo-examples` repo structure.
- `grep -rn "california_housing_xgb"` across `docs/` — only the intentionally-repointed references remain, pointed at `michelangelo-examples`.

**Potential risks**

None. This is a docs-only change with no runtime code changes.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Yes — repointed the XGBoost/California-housing example references in the examples gallery, getting-started guide, workflow-patterns guide, and sandbox-setup guide to `michelangelo-examples`' `xgb_train` pipeline.


sallycr:
Ran this repo's doc-review-team skill (`docs/.claude/skills/doc-review-team/SKILL.md`) against this PR's 4 changed files — an engineer pass for technical accuracy and a product-manager pass for audience/journey fit — then applied fixes for the findings.

**Fixed (pushed in bdb68eb7):**
- **Broken command**: `getting-started.md`'s "Option A" local-run invocation (`python -m michelangelo_examples.california_housing.pipelines.xgb_train`) has no `__main__.py` and would fail with `No module named ...xgb_train.__main__`. The entry point only exists in `.pipeline` (confirmed against `michelangelo-examples`' actual package layout) — the remote-run examples elsewhere in the same doc already had this right. Fixed to match.
- **Dropped prerequisite**: the macOS `libomp.dylib`/`brew install libomp` note for XGBoost was deleted with no replacement, but the tutorial's local-run path still trains XGBoost locally via the `california-housing` extra. Restored, scoped to that path.
- **Self-contradicting claim**: `examples/index.md` said "All examples live in `python/examples/`" directly above a table whose first row lives in a different repo. Scoped the claim and added a one-line rationale for the split (heavier Spark/XGBoost-dependency examples live in `michelangelo-examples` to keep core's own deps lean).
- **Misleading framing**: the "Option A / Option B" split presented `bert_cola` as a peer way to run the tutorial's own pipeline — it's an unrelated example with no connection to the `feature_prep`/`train` code the tutorial just walked through. Replaced with a single clear path, kept `bert_cola` as a clearly-labeled aside, and moved the "fastest way to iterate / zero infrastructure setup" line so it no longer directly follows a `pip install` step it now contradicts.

**Verified correct (no change needed)**: the `pip install "michelangelo-examples[california-housing]"` extra name, the embedded `pipeline.yaml` snippet's fields, `--environ`/`--image`/`--storage-url`/`--yes` as real CLI flags, the sandbox credential/endpoint values, and all `michelangelo-examples` GitHub links — all checked directly against `michelangelo-examples`' current `main` and this repo's CLI implementation.

**Flagged, not a doc bug in this PR**: merge-ordering — this PR's docs describe `california_housing_xgb` as gone, but the actual deletion lives on the sibling PR #1751, not yet merged. Not something to fix here, just worth knowing if #1749 merges well before #1751.

Ran a broken-internal-link check across all 4 changed files afterward — all relative links resolve.

sallycr:
## Live sandbox execution report (real k3d cluster, not static review)

Ran the doc's California Housing (`pytorch_train`) flow end-to-end against a real local `michelangelo-sandbox` (k3d + Cadence + MinIO), against the branch tip as of `5a40b42a`.

### 1. Local run (`pip install` + `python -m ...pipeline`) — **BROKEN, blocking**

```
pip install "michelangelo-examples[california-housing]"
python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline
```

Fails immediately on import with:

```
File ".../michelangelo/uniflow/core/utils.py", line 28, in dot_path
    assert "." not in file
AssertionError
```

Root cause: `dot_path()` reconstructs the dotted module path for `__main__` via `os.path.relpath(file, start=os.getcwd())`, then asserts no `.` remains. Any standard venv layout (`.../lib/python3.11/site-packages/...`) has a `.` in `python3.11`, so this assertion fails for **any** pip-installed package run via `python -m pkg.module`, regardless of cwd. Reproduced identically for both `xgb_train.pipeline` and `pytorch_train.pipeline` (i.e. this isn't specific to the new pytorch_train example — it affects the general `python -m <pkg>.pipeline` pattern this guide teaches). The bug lives in `michelangelo/uniflow/core/utils.py::dot_path`, not in this doc or in `michelangelo-examples`, so it isn't fixable by editing `getting-started.md` — flagging here since it directly breaks this guide's headline "Then run it" command as published on PyPI (`michelangelo==0.7.0`).

Confirmed via workaround (`from ...pipeline import train_workflow` + manual `ctx.run(train_workflow)`, which avoids `__main__`-based path resolution) that everything downstream of this bug works correctly — the workflow itself runs to completion successfully.

The doc's separate **bare** invocation (`python -m michelangelo_examples.california_housing.pipelines.pytorch_train`, no `.pipeline` suffix — the lightweight local-only PyTorch Lightning trainer) ran correctly end-to-end, 5 epochs, no errors. That caveat block is accurate.

### 2. Register + run via `ma` CLI — **PASS**

```
ma project apply -f src/michelangelo_examples/california_housing/config/project.yaml
ma pipeline apply -f src/michelangelo_examples/california_housing/pipelines/pytorch_train/pipeline.yaml
ma pipeline run -n california-housing --name pytorch-train
```

All three commands ran exactly as documented. Project and pipeline registered (`PIPELINE_STATE_READY`). Pipeline run `run-20260809-153155-3b052ee3` reached terminal state:

```
"status": { "state": "PIPELINE_RUN_STATE_SUCCEEDED", ... }
```

with all three conditions (`SourcePipeline`, `Image Build`, `Execute Workflow`) `CONDITION_STATUS_TRUE`. This section is accurate as written.

### 3. Docker build + remote-run — **partially blocked by the same bug**

```
docker build -t michelangelo-examples:local .
k3d image import michelangelo-examples:local -c michelangelo-sandbox
```

Both ran cleanly and exactly as documented.

```
python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline remote-run --image ... --yes
```

Fails with the identical `dot_path` `AssertionError` as flow 1 (decoration happens at import time, before the `remote-run` subcommand is even parsed) — so this command as literally written in the doc never reaches the point of dispatching a run.

Using the same import-based workaround, the workflow *did* dispatch correctly to Cadence (`Started Workflow Id: ...train_workflow.t700r`), the Ray cluster/worker pods came up, and the k8s Job for the Ray job submitter reported `Completed`. However, both attempts ultimately closed as `CloseStatus: FAILED` ("Ray job failed after all attempts were exhausted") a few seconds after the k8s Job completed — the pod-readiness check backing `process_terminated_job` in `commons.star` appears to race with normal RayCluster teardown after a successful job. This looks like separate flakiness in the uniflow RayJob completion path, not a doc issue, and I didn't chase it further given the primary `dot_path` bug already blocks the command as written.

### Net: no doc edit was made this pass

Flow 2 is solid and this doc's text for it is accurate. Flows 1 and 3's headline commands are copy-paste-accurate to the underlying `michelangelo-examples` source, but the core `michelangelo` library's `dot_path()` bug means neither actually runs as documented today for a real pip install. There's no doc-only rewording that fixes this (the bug fires on any `python -m pkg.module` pattern for a pip-installed package) — this needs a fix in `michelangelo/uniflow/core/utils.py::dot_path` (or an accompanying `ctx.run()`-style helper that doesn't depend on `__main__` file-path resolution) before this guide's central commands can be verified as working, not just plausible.

paulzim (approved):
**Approve** — thoroughly self-vetted docs migration; one non-blocking suggestion.

This one had unusually deep verification already done by the author (a `doc-review-team` skill pass plus a live k3d sandbox execution of every command in the guide, both posted as PR comments) — I re-verified the underlying facts independently rather than trusting either the description or the comments at face value.

**What actually changed, vs. what the PR description says:** the description frames this as a pure link-repoint ("references the ... pipeline (now `xgb_train`)"). That's accurate for `sandbox-setup.md`, `examples/index.md`, and `workflow-patterns.md`, but `getting-started.md` (the last of the 4 commits, `5a40b42a`) does something bigger: it swaps the tutorial's entire worked example from XGBoost to PyTorch Lightning (`pytorch_train`) — different hyperparameters block, different diagram comment, a removed XGBoost/OpenMP prerequisite, and a new callout explaining the two ways to invoke `pytorch_train`. The content itself is well-executed (see below), but the top-level PR description was never updated to say "and also switches which example the flagship tutorial teaches" — worth a one-line addition for anyone skimming just the description before diving into commits/comments.

**Independently verified against the actual `michelangelo-examples` repo** (not just taken from the PR body):
- `xgb_train/` and `pytorch_train/` both exist at the claimed paths, each with `pipeline.py`/`pipeline.yaml`.
- `pytorch_train/pipeline.yaml` and `config/project.yaml` are byte-identical in namespace/name/image to what the doc now shows (`california-housing` / `pytorch-train` / `ghcr.io/michelangelo-ai/michelangelo-examples:california-housing`).
- `pytorch_train/train.py` really does call `tabular_trainer`'s `train_tabular()`, matching the doc's framing.
- `pytorch_train/README.md`'s remote-run block (image, `--environ` vars, sandbox note) matches the doc's new snippet verbatim.
- `ma pipeline run -n <namespace> --name <name>` is a documented short-flag form elsewhere in this repo's own `cli.md` — not a made-up shortcut.
- `grep -rn california_housing_xgb docs/` returns nothing — the "only intentionally-repointed references remain" claim holds.

**The one real substantive issue, already surfaced by the author's own testing, not by me:** the live-sandbox comment found that this guide's headline "run it locally" and "run it remotely" commands (`python -m michelangelo_examples....pipeline`) currently crash with an `AssertionError` in `michelangelo/uniflow/core/utils.py::dot_path()` for *any* pip-installed package run via `python -m pkg.module` — not specific to this doc or to `pytorch_train`. Root-caused, filed as #1753, with a fix already up in #1754 (both open, unmerged, same author). Only the CLI-registration path (`ma pipeline apply` + `ma pipeline run`) was confirmed working end-to-end on a real sandbox.

Given the bug is pre-existing, unrelated to this doc's content, and already tracked with a fix in flight, I don't think this should block the docs PR — but as it stands, a reader following "Step 3: Run locally" exactly as written today will hit a cryptic assertion with zero indication it's a known issue. **Suggestion (non-blocking):** add a one-line callout near that command noting the known issue/fix-in-progress (#1753/#1754), so readers aren't left thinking they misconfigured something. Fine to land as a fast-follow once #1754 merges instead, if you'd rather not touch this PR again.

CI: `filter`/`license/cla`/`update-release-draft`/`validate-docs` all pass; the three advisory API-surface checks correctly report `skipping` since no `go/`, `proto/`, or `helm/` paths changed.

Solid work — the level of hands-on verification here (including catching and filing a real upstream bug along the way) is well above the bar for a docs PR.

kenns29 (approved):

andriika (approved):

PR #1751, opened 2026-08-09
chore(examples): remove california_housing_xgb pipeline (migrated to michelangelo-examples)

sallycr:
Part of https://github.com/michelangelo-ai/michelangelo/issues/1748

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Deletes the `python/examples/pipelines/california_housing_xgb/` pipeline (8 files) and its now-empty parent `__init__.py`, since it has been fully ported to `michelangelo-examples` as `xgb_train` (see feature 013). Repoints `python/README.md` to reference the `michelangelo-examples` version instead. Removes `xgboost` and `xgboost_ray` from `python/pyproject.toml`'s optional dependencies and `example` extra, since the deleted pipeline was their only consumer.

This is part 3 of a 3-way split of #1746, which bundled this deletion/dependency cleanup with unrelated doc repointing and `python/michelangelo/lib` docstring/test updates. Splitting lets each group be reviewed independently. See also:
- docs: repoint california_housing doc references to michelangelo-examples (#1749)
- chore(lib): update california_housing example references in model_manager registry docstrings/tests (#1750)

Intentionally **not** touched:
- `python/michelangelo/cli/sandbox/resources/spark.yaml` / `spark-app.yaml` — static illustrative fixtures with no runtime consumer.
- `python/examples/config/project.yaml` — still a live dependency for `workflow-patterns.md` and other docs, unrelated to the deleted pipeline.
- `CHANGELOG.md` — historical record, out of scope.

**Why?**

`california_housing_xgb` was ported to `michelangelo-examples` and no longer needs to live in core. Keeping a duplicate pipeline in core alongside the actively-maintained example in `michelangelo-examples` is confusing for new users and creates maintenance drift (e.g. unused `xgboost`/`xgboost_ray` dependencies).

**How did you test it?**

- `grep -rn "california_housing_xgb"` — only the two intentionally-kept fixture references remain (`spark.yaml`, `spark-app.yaml`), plus one historical `CHANGELOG.md` entry.
- `grep -rn "california_housing_lightning"` — zero hits.
- `ruff check` on `python/README.md` — no Python files under that path, no-op as expected.

**Potential risks**

None. This is a deletion/config cleanup with no runtime code changes; the deleted pipeline was example-only and superseded by `michelangelo-examples`.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Yes — repointed the `python/README.md` XGBoost/California-housing example reference to `michelangelo-examples`' `xgb_train` pipeline.


badcount (approved):

paulzim:
## Verdict: Approve

Deletion PRs live or die on dangling references and dependency fallout. I checked both exhaustively, plus the cross-repo dependency question the description doesn't raise, and everything holds. Branch is fully rebased on current `main` (merge-base == `d4182322` == `main` HEAD, 0 behind / 2 ahead).

### Verified

**The deletion is complete and exact.** `git ls-tree` on `main` lists exactly the 8 files the PR removes — no stragglers left in the directory, and `python/examples/pipelines/` is gone entirely.

**Zero dangling references, repo-wide.** A whole-repo grep for `california_housing_xgb` leaves only four hits, all correct to keep: `CHANGELOG.md:689` / `docs/about/changelog.md:389` (historical #1447 entries), and the two `spark.yaml`/`spark-app.yaml` fixtures called out as intentionally kept. Greps for `examples/pipelines`, `examples.pipelines`, and `xgboost_ray` return **nothing** anywhere in the tree.

**The "no runtime consumer" claim on the spark fixtures checks out — I didn't take it on faith.** `sandbox.py` builds its resource list by explicit filename (`resources.append("minio.yaml")` etc.) and applies via `_kube_apply(_dir / "resources" / r)`; neither file is in that list. The only two `glob()` calls target `_dir / "demo" / "pipeline"` and a snapshots directory, not `resources/`. Nothing in `python/`, `helm/`, `go/`, `scripts/`, or `.github/` references either filename. They're inert.

**`xgboost` is correctly *retained* as an optional dependency.** Worth stating explicitly since it's the part most likely to be waved through: `michelangelo.lib.trainer.xgboost` is live in-tree with its own tests, and the `trainer-xgboost` and `dev` extras both need it. The PR keeps `xgboost = {version = "2.1.4", optional = true}` and removes it only from `example` — exactly right. (`xgboost_ray` is genuinely unreferenced; note `xgboost_trainer.py` imports `ray.train.xgboost`, which ships with `ray[train]`, not the `xgboost_ray` package.)

**The lock regeneration is correct, validated independently rather than trusting `dirty-check`.** `poetry check --lock` exits 0. The diff removes the `xgboost-ray` block wholesale — name, description, both wheel hashes, its `xgboost = ">=0.90"` dep — updates `content-hash`, and narrows `xgboost`'s own marker from `extra == "example" or extra == "dev" or extra == "trainer-xgboost"` to `extra == "dev" or extra == "trainer-xgboost"`. The rest is poetry reordering extras within existing markers.

**Cross-repo dependency risk: cleared.** The description doesn't mention this, and it's the one way this could have broken the very example it points users to — if `michelangelo-examples` relied on core's `example` extra for xgboost. It doesn't. Its `california-housing` extra declares `"xgboost>=2.0,<4"` and `"xgboost_ray>=0.1.19,<1"` directly, with a comment stating it deliberately avoids core's `example` extra because that extra "drags in a pile of unrelated deps (xgboost, transformers, mlflow, kaggle, chronon-ai, ...)". This PR moves core *toward* what that comment asks for.

**The port target exists and accounts for all 8 deleted files.** `xgb_train/` has `pipeline.py`, `train.py`, `push.py`, `pipeline.yaml`, `README.md`, `_backend.py`, `__init__.py`. The two without an obvious counterpart — `feature_prep.py` and `preprocess.py` — live at `california_housing/pipelines/libs/tasks/`, shared with `pytorch_train`, and `xgb_train/pipeline.py` imports them from there. Nothing was dropped in the migration.

**Sequencing is right.** #1749 merged 2026-08-19 and #1750 merged 2026-08-10, so this correctly lands last. All five doc files enumerated in #1748's scope now return zero `california_housing_xgb` hits.

**Independent test verification.** `pytest --collect-only` gathers **2951 tests with zero collection errors** — the deletion breaks no imports. CI is fully green. Repo-wide `ruff check` reports 179 pre-existing errors and 4 format diffs, all in files this PR doesn't touch; it adds zero Python lines, so none are attributable to it.

### Non-blocking notes

**1. Worth `Closes #1748` rather than "Part of".** With #1749 and #1750 merged, this completes every item in #1748's scope — the deletion, the five doc repoints, and the dependency cleanup — including the "not touched" carve-outs, which match the issue's text verbatim. Unless it's held open for something else, it'll otherwise sit open after the work is done. (We just hit this with #1867, still open after both halves of its fix merged.)

**2. Small imprecision in the description.** "Removes `xgboost` and `xgboost_ray` from `python/pyproject.toml`'s optional dependencies and `example` extra" reads as though both were removed from both places. Precisely: `xgboost_ray` from both; `xgboost` from the `example` extra only, correctly staying an optional dependency for `dev`/`trainer-xgboost`. The code is right — only the sentence is ambiguous, and it's the sentence a future reader will use to judge whether the cleanup was complete.

**3. If anyone ports the deleted README's "Debugging" section — fix it first, it's wrong.** The migrated README is 133 lines against core's 250, and one dropped section documented `breakpoint=True`, a real `RayTask` field (`plugins/ray/task.py:81`). But the deleted text claimed "UniFlow halts the workflow **before the task runs** and waits for you to attach." The implementation in `plugins/ray/task.star:348-354` does the opposite: it submits and runs the job, *then* prints the cluster coordinates, sleeps 24 hours with the Ray cluster left alive, and hard-fails the task with `internal: breakpoint timeout`. Deleting that paragraph removes an incorrect doc — a point in the PR's favor — just don't copy it across later. The feature's only accurate description now is the one-line docstring plus commented-out usages in `bert_cola`, `llm_prediction`, and `nomic_ai`.

The other dropped section, `DatasetVariable — passing data between tasks`, is well covered elsewhere — 8 files under `docs/` including the glossary, TERMINOLOGY, and the Python SDK reference. No loss there.

**4. One real gap, already tracked.** Core's README had a `## Local Run` section documenting the zero-external-services path ("Without `AWS_ENDPOINT_URL`, `push_step` uses `LocalStorageBackend` … no external services required"). The migrated README is k3d-only; the fact survives in its env-var table but the runnable recipe doesn't. Same local-run-vs-remote-run gap as #1670, so it belongs there rather than here.

**5. Optional follow-up, genuinely out of scope.** The two spark fixtures now reference `examples.california_housing_xgb.preprocess.preprocess`, a module path that exists nowhere in the repo. They're inert and their retention was agreed at issue level, so this shouldn't touch this PR — but they're snapshots with baked-in `s3://` UUIDs and `docker.io/library/examples:latest`, sitting in a directory where every other file *is* applied to the cluster. A one-line "historical snapshot, not applied" comment would save someone the ten minutes I spent proving they're dead.

Clean, correctly sequenced, complete, and the risky parts all verify. The notes above are polish and follow-ups, none of them reasons to hold it.

Issue #1752, opened 2026-08-09
CI: Go API Surface (advisory) check fails repo-wide — unpinned apidiff@latest + actions/setup-go v7's GOTOOLCHAIN=local


sallycr:
## Summary

`.github/workflows/api-surface-check.yml`'s `Go API Surface (advisory)` job now fails on every PR that touches `go/**`, regardless of the PR's actual content. First noticed on #1747 (a gazelle/BUILD.bazel-only change with no Go source diff), but confirmed to be unrelated to that PR — this is a pre-existing, repo-wide CI break.

Failing step: `go install golang.org/x/exp/cmd/apidiff@latest`

```
go: golang.org/x/exp/cmd/apidiff@latest: golang.org/x/exp@v0.0.0-20260727155853-b88d891fe743
requires go >= 1.25.0 (running go 1.24.3; GOTOOLCHAIN=local)
##[error]Process completed with exit code 1.
```

## Root cause

Two independent, pre-existing facts combined:

1. **`apidiff` is installed unpinned** (`@latest`). `golang.org/x/exp` has no tagged releases, so `@latest` floats to the tip of its default branch. A commit on `golang/exp` (`2735e65f0518`, 2026-02-11, "all: upgrade go directive to at least 1.25.0") bumped that module's own `go.mod` to require Go ≥ 1.25 — true for every `@latest` resolution since February.
2. **This only broke CI once `actions/setup-go` was bumped from v5 to v7** (#1743, merged 2026-08-09T00:23:10Z). `actions/setup-go` v6.0.0 changed its default from `GOTOOLCHAIN=auto` to `GOTOOLCHAIN=local` ("Improve toolchain handling to ensure more reliable and consistent toolchain selection and management," a deliberate reproducibility improvement upstream). With `auto`, `go install` silently downloaded a compatible toolchain; with `local`, the mismatch is a hard error.

Confirmed with two real CI runs straddling the #1743 merge, same job, ~1h40m apart:

| Run | PR | `setup-go` | `GOTOOLCHAIN` | Result |
|---|---|---|---|---|
| [31284440857](https://github.com/michelangelo-ai/michelangelo/actions/runs/31284440857/job/93170776987) | #1738 | v5 | `auto` | PASS (auto-downloaded go1.25.12) |
| [31287981035](https://github.com/michelangelo-ai/michelangelo/actions/runs/31287981035/job/93180167080) | #1747 | v7 | `local` | FAIL |

## Impact

[...]

PR #1755, opened 2026-08-09
ci(go): migrate golangci-lint config off deprecated options

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
Moves two deprecated keys in `go/.golangci.yml` to their current spellings: `run.skip-dirs` becomes `issues.exclude-dirs`, and `output.format` becomes `output.formats` (now a list of format/path entries).

**Why?**
golangci-lint v1.64.8, the version pinned in `.github/workflows/go-lint.yml`, logs both on every run:

```
level=warning msg="[config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`."
level=warning msg="[config_reader] The configuration option `output.format` is deprecated, please use `output.formats`"
```

Both keys still work today, so this is warning cleanup rather than a functional fix, but it removes recurring noise from the lint output and means the config will not need attention when these options are eventually removed.

**How did you test it?**
The two warnings were read from the actual job log of a recent Go Lint run on `main`, so this is the real current output rather than an inference from the docs. Those are the only two deprecation warnings the linter emits for this config.

The replacement shapes come from golangci-lint's own reference config at the pinned version (`.golangci.reference.yml` at tag `v1.64.8`), which documents `issues.exclude-dirs` as a list of paths and gives the default for the new output key as `colored-line-number` written to `stdout` - matching the behavior configured here today. The file still parses as valid YAML.

Worth noting the output setting is not what CI actually renders with in any case: the workflow passes `--out-format=github-actions`, and the CLI flag overrides the config file, so this change is warning-only in CI and preserves the existing local behavior.

**Potential risks**
Very low. Both replacements are the linter's own documented equivalents for the same version, and the excluded directories (`gen`, `thirdparty`) are unchanged. If a directory exclusion were somehow dropped, the effect would be extra `godox` findings from generated code, which the existing `exclude-rules` entries for `gen/.*\.go` and `thirdparty/.*\.go` would still suppress.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.


paulzim (approved):
**Approve** — correct, well-verified deprecation cleanup.

Verified against golangci-lint's own reference config, not just the PR's description:

- Diff is exactly the one file described: `go/.golangci.yml`, +7/-4. Moves `run.skip-dirs: [gen, thirdparty]` → `issues.exclude-dirs: [gen, thirdparty]`, and `output.format: colored-line-number` → `output.formats: [{format: colored-line-number, path: stdout}]`. Nothing else touched.
- Pulled `.golangci.reference.yml` at the exact pinned tag (`v1.64.8`, matching `go-lint.yml`'s install version) directly from golangci-lint's repo: `skip-dirs` doesn't appear anywhere in it (fully gone, confirming the deprecation), and `exclude-dirs` is documented under `issues:` as a list of paths — matching the PR's replacement.
- The reference config's own default for the new `output.formats` key is written as:
  ```
  formats:
    - format: colored-line-number
      path: stdout
  ```
  — byte-for-byte the same shape this PR produces. Also confirms in its own comment text: *"The CLI flag (`--out-format`) override the configuration file"* — independently corroborating the PR's claim that `go-lint.yml`'s `--out-format=github-actions` flag makes this output-format change cosmetic/config-only, not a CI-behavior change.
- Confirmed `gen`/`thirdparty` are still excluded (now via `issues.exclude-dirs`) and the `godox` `exclude-rules` for `gen/.*\.go` and `thirdparty/.*\.go` are untouched and present, by parsing the resulting file with `yaml.safe_load` and inspecting the loaded structure directly.
- YAML parses cleanly.
- No CI has run yet on this PR (only `license/cla` shows; no check-runs on the head commit) — looks like a first-time fork contributor pending workflow approval, not a problem with the change itself. Worth kicking off the Actions run manually before merge so the "no functional change" claim gets one real confirmation from `go-lint.yml` itself (should now log zero deprecation warnings).

Low risk, same author/pattern as the other CI-hardening PRs in this batch (#1711, #1756, #1757) — no functional change, just quieter logs.

apurvapatkeshwar:
CI ran - `Go lint with TODO tracking` is green, so the deprecation cleanup is confirmed

The one red is `Go API Surface (advisory)` and it's unrelated to this PR, which only touches `go/.golangci.yml`
That job does `go install golang.org/x/exp/cmd/apidiff@latest`, and x/exp now requires go 1.25 while the runner is on 1.24.3:
```
go: golang.org/x/exp@v0.0.0-20260810151157-a8b543ca52da requires go >= 1.25.0 (running go 1.24.3; GOTOOLCHAIN=local)
```
Same failure on `fix/sandbox-workflow-execution-url-v2`
Happy to send a fix for it separately

apurvapatkeshwar:
`Go API Surface (advisory)` is failing at `Install go-apidiff`, before it reaches any code in this PR -
```
golang.org/x/exp@v0.0.0-20260810151157-a8b543ca52da requires go >= 1.25.0 (running go 1.24.3; GOTOOLCHAIN=local)
```
I think it'll keep hitting every PR while the install is `@latest` - #1781 pins it

paulzim (approved):
**Update (2026-08-13): checked the two follow-up comments, nothing changes here.**

The PR's own diff is unchanged since approval — still the single file, `go/.golangci.yml`, +7/-4, confirmed via `gh pr view --json files`. Both comments are the author explaining an unrelated, pre-existing CI failure, not raising anything about this PR's actual change:

- **Aug 11:** `Go API Surface (advisory)` is red because `go install golang.org/x/exp/cmd/apidiff@latest` pulls a revision of `x/exp` requiring Go ≥1.25, while the runner is on 1.24.3 — a floating-dependency problem in the shared workflow, unrelated to anything `go/.golangci.yml` touches. Author confirmed the same failure independently reproduces on an unrelated branch (`fix/sandbox-workflow-execution-url-v2`).
- **Aug 12:** points at #1781 as the fix (pin `apidiff` to a Go-1.24-compatible pseudo-version instead of `@latest`).

**#1781 merged today** (`f4b92b63`, 2026-08-13T21:39:58Z) with exactly that one-line pin. I checked whether that already clears #1755's red check: the only `Go API Surface (advisory)` run recorded against #1755's head commit (`f9f8e3c9`) completed **2026-08-10T20:29:26Z** — three days before #1781 merged — so the currently-displayed failure is stale, not a fresh re-check against the fix. It hasn't been re-run since. Practically this doesn't block anything: `gh pr view` reports `mergeable: true` on #1755 regardless, and this check is explicitly advisory (same tier as the `Helm Values Key Diff`/`Proto Breaking Changes` checks that "skip" when not applicable) — it was never a merge gate.

One more thing surfaced while reading #1781's own PR body, worth knowing about but **not a #1755 issue**: #1781 flags that even pinned, the `apidiff` invocation itself looks broken — wrong flags (`-export` isn't a real flag; the base/new positional args aren't both supplied) and errors are silently swallowed with `2>/dev/null || true`, so the job apparently has never actually detected a breaking API change regardless of green/red. That's a pre-existing gap tracked separately in #1752, unrelated to golangci-lint config and outside this PR's scope.

**Verdict unchanged: Approve.** Nothing about the actual change is in question; if you want a clean, current confirmation of the advisory check, it just needs a re-run (push/rebase, or trigger manually) now that #1781 is in — but that's optional polish, not a blocker.

PR #1712, opened 2026-08-07
chore: remove dead internal tracker links from code comments

apurvapatkeshwar:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**
Removes two code-comment links to internal Uber trackers, which are not reachable from outside Uber and so are dead links for every reader of this repository:

1. `go/components/utils/proto.go` - a `code.uberinternal.com` ticket link in the comment above `NewValue`. The substance of the note (that the protobuf-go repo is not imported directly due to compatibility issues) is kept as-is.
2. `javascript/packages/core/components/row/types.ts` - a `@see` link to a `t3.uberinternal.com` ticket on the `hideEmpty` prop. The `@example` above it already documents the behavior, so only the tracker reference is dropped.

**Why?**
Dead internal links in a public codebase are noise for external readers and can be confusing - they look like a resource you are expected to be able to consult. These were the only two occurrences of `uberinternal.com` in tracked files; after this change `git grep uberinternal` comes back empty.

**How did you test it?**
Comment-only change, no code touched. `git grep uberinternal` across the repo confirms these were the only two occurrences and that none remain after the change.

**Potential risks**
None - no code changes.

**Breaking Changes**
- [x] No breaking changes
- [ ] API changes (Go exported symbols or function signatures, Python public functions or classes)
- [ ] Proto changes (enum value renumbering, field number changes, field removal, service removal)
- [ ] Helm changes (new required values, renamed or removed keys, changed value semantics)
- [ ] Config/deployment changes (new required env vars, renamed container args, changed ports or mount paths)

**Release notes**
None.

**Documentation Changes**
None needed.

PR #1756, opened 2026-08-09
fix(ci): skip JavaScript coverage PR comment on fork pull requests

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
One line in `.github/workflows/javascript-coverage.yml`: the "Coverage report PR comment" step now carries the same-repository guard that the equivalent step in `go-coverage.yml` already has, so it is skipped on pull requests from forks.

```
- if: github.event_name == 'pull_request'
+ if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
```

**Why?**
Pull requests from forks run with a read-only `GITHUB_TOKEN`, so the comment step cannot post and fails with `Resource not accessible by integration`. That turns the JavaScript Test & Coverage check red on external contributions even when the build and tests pass, which is a confusing first signal for a new contributor and makes a green run impossible to reach from a fork.

The repository already solves this in the two sibling workflows, in two different ways: `go-coverage.yml` guards the identical `sticky-pull-request-comment` step with the condition added here, and `python-coverage.yml` marks its comment step `continue-on-error: true`. `javascript-coverage.yml` has neither, which makes it the only one of the three that reports failure on fork pull requests. This adopts the `go-coverage.yml` approach so the two workflows using the same action stay consistent.

The coverage job itself is untouched and still runs on forks, including the baseline threshold check. Only the comment is skipped, and only when it could not succeed anyway.

**How did you test it?**
Reproduced on a real fork pull request: the JavaScript Test & Coverage check failed on #1712 while every other check passed. Its failing job's annotation is `Resource not accessible by integration - https://docs.github.com/rest/issues/comments#create-an-issue-comment`, and the failing step is "Coverage report PR comment", confirming the comment call rather than the tests. The pull request was merged with that check red.

Compared all three coverage workflows to confirm the inconsistency and that the condition used here is copied verbatim from the working one in `go-coverage.yml`. YAML still parses.

The behavior change is only observable on a fork pull request, which this PR is, so the JavaScript coverage check on this PR should skip the comment step and complete rather than fail.

**Potential risks**
On pull requests from branches in this repository nothing changes: the guard is true and the comment posts as before. On forks the coverage comment will no longer appear, but it never did - the step failed instead. If you would rather keep attempting the comment and just not fail the check, `continue-on-error: true` as used in `python-coverage.yml` is the alternative, happy to switch.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.

### DIFF
--- a/.github/workflows/javascript-coverage.yml
+++ b/.github/workflows/javascript-coverage.yml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Coverage report PR comment
         uses: marocchino/sticky-pull-request-comment@v2
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
         with:
           header: javascript-coverage
           message: |


apurvapatkeshwar:
CI ran on this one and it does what it says - the coverage comment step gets skipped instead of failing, and the job is green:
```
9. Check coverage threshold (baseline: 78.74%) success
10. Coverage report PR comment                 skipped
11. Check diff coverage (90% for new code)     success
```

All 9 checks green or skipped, no failures. Both coverage gates still run on forks, it's only the comment that's skipped. Could someone please take a look when they get a chance?

paulzim (approved):
**Approve** — correct, minimal, self-verifying fix.

Verified:
- Diff is exactly the one line described: adds `&& github.event.pull_request.head.repo.full_name == github.repository` to the `if:` on the "Coverage report PR comment" step in `javascript-coverage.yml`. Nothing else touched.
- Confirmed the condition is copied verbatim from `go-coverage.yml` (line 124, identical string) — this brings all three coverage workflows in line: `go-coverage.yml` uses this guard, `python-coverage.yml` uses `continue-on-error: true` on its comment step, `javascript-coverage.yml` had neither before this PR.
- Confirmed the cited failure on PR #1712: pulled its check-runs directly — `JavaScript Test & Coverage` is the only non-green check (everything else passed or was skipped-advisory), and the job annotation is exactly `Resource not accessible by integration - .../rest/issues/comments#create-an-issue-comment`, matching the PR description precisely. That PR merged with this check red.
- Self-verifying: this PR's own CI (a fork PR, so it exercises the new guard) now shows `JavaScript Test & Coverage` passing, with the coverage-comment step reported as `skipped` rather than `failure` — both coverage gates (baseline threshold, diff coverage) still ran and passed.
- YAML parses fine (`yaml.safe_load`).

No functional change for same-repo PRs (guard evaluates true, comment posts as before). Low risk, well-scoped, addresses a real recurring fork-contributor friction point.

austingreco (approved):

PR #1758, opened 2026-08-10
Add optional onClick to LinkCellConfig and LinkProps

sallycr:
## Summary

Adds an optional `onClick` callback to `LinkCellConfig` (`packages/core/components/cell/renderers/link/`) and `LinkProps` (`packages/core/components/link/`), so a caller can run code when a config-driven link is clicked, before navigation occurs — without either component knowing anything about what that callback does (analytics, logging, or anything else).

- `LinkCellConfig.onClick?: (record: unknown) => void` — invoked with the row's record when the rendered link is clicked.
- `LinkProps.onClick?: () => void` — invoked when the underlying link element is clicked.

Both props are optional and default to a no-op, so this is a non-breaking, additive change: every existing consumer of `LinkCellConfig`/`LinkProps` continues to render and navigate exactly as before.

This PR only adds the generic click-hook capability to the shared, reusable components. It does not wire `onClick` into any specific entity/column configuration — that's intentionally left for a follow-up change in the application that consumes this library, once its own click-tracking callback is ready to be supplied.

## Test plan

- [x] Extended `packages/core/components/link/__tests__/link.test.tsx`: clicking still works with `onClick` omitted (no regression), `onClick` fires on click for both internal and external links.
- [x] Extended `packages/core/components/cell/renderers/link/__tests__/link-cell.tests.tsx`: clicking still navigates normally with `onClick` omitted, `onClick` is called with the row record when provided.
- [x] `npx tsc -p packages/core/tsconfig.json --noEmit` — no errors.
- [x] `npx eslint packages/core/components/link packages/core/components/cell/renderers/link` — no errors/warnings.
- [x] `npx vitest --run packages/core/components/link packages/core/components/cell/renderers/link` — all 22 tests passing.
- [x] `npx prettier --check` on all changed files — clean.

### DIFF
--- a/javascript/packages/core/components/cell/renderers/link/__tests__/link-cell.tests.tsx
+++ b/javascript/packages/core/components/cell/renderers/link/__tests__/link-cell.tests.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import { getIconProviderWrapper } from '#core/test/wrappers/get-icon-provider-wrapper';
 import { LinkCell } from '../link-cell';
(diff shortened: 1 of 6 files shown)


craigmarker (commented):
What will this be used for specifically? I'm not sure I love having side effects configured in this way. 

For click tracking, we could architect a solution that would be inherited by the link cell.

PR #1760, opened 2026-08-10
fix(python): wire up workflowClient.executionUrlFormat for the sandbox

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`_build_helm_set_args()` in `python/michelangelo/cli/sandbox/sandbox.py` now sets `controllermgr.workflowClient.executionUrlFormat` per workflow engine (Cadence or Temporal) when standing up the local sandbox, alongside the existing per-engine `workflow.engine`/`workflow.endpoint` `--set` flags.

**Why?**

`helm/michelangelo/values.yaml`'s comment for `workflowClient.executionUrlFormat` claims "set in values-k3d.yaml for the local sandbox," but `values-k3d.yaml` never actually set it — only `workflow.domain`/`workflow.endpoint` are configured there, which is a separate config block. As a result, a `PipelineRun`'s parent "Execute Workflow" step never got a working `log_url` in the sandbox, even though the code to build one (`ExecuteWorkflowActor.GetWorkflowUrl`) already exists and only needs `executionUrlFormat`/`domain` to be non-empty. This PR wires up a real value for both supported engines, matching the Cadence Web UI URL already referenced elsewhere in this same file (`http://localhost:8088/domains/default/workflows`) and the equivalent Temporal Web pattern.

**How did you test it?**

`python3 -m py_compile python/michelangelo/cli/sandbox/sandbox.py` (no syntax errors). No existing unit tests cover `_build_helm_set_args()`; this is CLI-glue code exercised via `michelangelo sandbox up`, which wasn't run end-to-end as part of this change (no live k3d cluster in this environment) — flagging that as a gap for reviewer verification.

**Potential risks**

Low — this only adds two new `--set-string` Helm flags; it does not change any existing flag's value or the default chart behavior for non-sandbox (production) deployments, which still default `executionUrlFormat` to `""` unless explicitly configured by the operator.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A — sandbox-only developer-experience fix.

**Documentation Changes**

N/A


sallycr:
Closing — this branch was accidentally created off the still-unmerged PR #1733 dependabot branch instead of main, so it pulled in 6 unrelated commits. Replaced by #1761, branched cleanly off main with just the sandbox fix.

PR #1761, opened 2026-08-10
fix(python): wire up workflowClient.executionUrlFormat for the sandbox

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Two related fixes to `PipelineRunStepInfo.log_url` for the "Execute Workflow" step:

1. `ExecuteWorkflowActor.GetWorkflowUrl` now takes a `runID` parameter and exposes it as `{{.RunID}}` in the configured URL template, in addition to the existing `{{.Domain}}`/`{{.ExecutionID}}`. The step's `LogUrl` is now set once, right after `StartWorkflow` returns with the real run ID — not at PENDING-state creation, before the run ID is known.
2. The sandbox's `executionUrlFormat` (`python/michelangelo/cli/sandbox/sandbox.py`) now includes the run ID, matching Cadence Web's actual routing requirements.

**Why?**

`helm/michelangelo/values.yaml`'s comment for `workflowClient.executionUrlFormat` claims "set in values-k3d.yaml for the local sandbox," but it never actually was — only `workflow.domain`/`workflow.endpoint` were set. Wiring up a first-pass value surfaced a second, more fundamental issue confirmed against a live sandbox run: Cadence Web needs domain + workflow ID + run ID together to resolve to a workflow's page (it redirects to the right cluster/summary view on its own from just those three) — a URL built from workflow ID alone (a `{{.ExecutionID}}`-only template, computed before the run even started) never resolves correctly, since the run ID is only available after `StartWorkflow` succeeds.

**How did you test it?**

`go build`/`go test` and `bazel test //go/components/pipelinerun/...` (4/4 targets pass) for the Go change; `gofmt`/`golangci-lint` clean (no new findings). `python3 -m py_compile` for the sandbox change. The exact URL shape (domain + workflow ID + run ID, no extra path segments needed) was confirmed against a real workflow run in a live sandbox instance. The Temporal Web path shape is *not* independently verified against a live Temporal-backed sandbox — flagged in a code comment for follow-up.

**Potential risks**

Low-moderate: `GetWorkflowUrl`'s signature changed (added a parameter) — this is a private/package-internal actor method, not a public API, so no external callers are affected. The parent step's `LogUrl` is now empty for one reconcile longer (from PENDING until `StartWorkflow` succeeds) rather than populated immediately at creation — acceptable since the earlier value could never resolve to a working link anyway.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A — sandbox-only developer-experience fix plus an internal (non-exported) method signature change.

**Documentation Changes**

N/A

### DIFF
--- a/go/components/pipelinerun/actors/executeworkflow.go
+++ b/go/components/pipelinerun/actors/executeworkflow.go
@@ -228,10 +228,16 @@ func (a *ExecuteWorkflowActor) Retrieve(ctx context.Context, resource *v2.Pipeli
 //
 // Parameters:
 //   - name: The workflow execution name (usually the pipeline run name)
+//   - runID: The workflow run ID. Cadence/Temporal Web UIs route a workflow's
+//     detail/summary page by (workflow ID, run ID) together, not by workflow ID
+//     alone, so a format string that only references {{.ExecutionID}} resolves
+//     to a listing rather than a specific execution. Pass "" if the run ID isn't
+//     known yet (e.g. before the workflow has started) — the caller should defer
+//     calling this until it is, rather than publish a URL that won't resolve.
 //
 // Returns a formatted URL string for the workflow monitoring interface, or an
 // empty string if the workflow client configuration cannot be retrieved.
-func (a *ExecuteWorkflowActor) GetWorkflowUrl(name string) string {
+func (a *ExecuteWorkflowActor) GetWorkflowUrl(name string, runID string) string {
 	workflowConfig, getWorkflowClientConfigErr := config.GetWorkflowClientConfig(a.configProvider)
 	if getWorkflowClientConfigErr != nil {
 		return ""
@@ -244,7 +250,7 @@ func (a *ExecuteWorkflowActor) GetWorkflowUrl(name string) string {
 
 	tmpl, _ := template.New("url").Parse(workflowConfig.ExecutionUrlFormat)
 	var buf bytes.Buffer
-	tmpl.Execute(&buf, map[string]string{"Domain": workflowConfig.Domain, "ExecutionID": name})
+	tmpl.Execute(&buf, map[string]string{"Domain": workflowConfig.Domain, "ExecutionID": name, "RunID": runID})
 	return buf.String()
 }
 
@@ -280,7 +286,10 @@ func (a *ExecuteWorkflowActor) Run(ctx context.Context, pipelineRun *v2.Pipeline
 			DisplayName: pipelinerunutils.ExecuteWorkflowStepName,
 			State:       v2.PIPELINE_RUN_STEP_STATE_PENDING,
 			StartTime:   pbtypes.TimestampNow(),
-			LogUrl:      a.GetWorkflowUrl(pipelineRun.Name),
+			// LogUrl is deliberately left unset here: the workflow hasn't
+			// started yet, so there's no run ID to build a resolvable
+			// monitoring link with (see GetWorkflowUrl's runID parameter).
+			// It's populated once, right after StartWorkflow succeeds below.
 		}
 		pipelineRun.Status.Steps = append(pipelineRun.Status.Steps, executeWorkflowStep)
 	}
@@ -370,6 +379,7 @@ func (a *ExecuteWorkflowActor) Run(ctx context.Context, pipelineRun *v2.Pipeline
 		executeWorkflowStep.EndTime = nil
 		pipelineRun.Status.WorkflowRunId = workflowExecution.RunID
 		pipelineRun.Status.WorkflowId = workflowExecution.ID
+		executeWorkflowStep.LogUrl = a.GetWorkflowUrl(workflowExecution.ID, workflowExecution.RunID)
 		return &apipb.Condition{
 			Type:   ExecuteWorkflowType,
 			Status: apipb.CONDITION_STATUS_UNKNOWN,
--- a/go/components/pipelinerun/actors/executeworkflow_test.go
+++ b/go/components/pipelinerun/actors/executeworkflow_test.go
@@ -2947,6 +2947,7 @@ func TestGetWorkflowUrl(t *testing.T) {
 		name        string
 		configSetup map[string]interface{}
 		inputName   string
+		inputRunID  string
 		expectedUrl string
 		expectError bool
 	}{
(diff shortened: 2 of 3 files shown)


sallycr:
## Sandbox verification (local build, latest commit `6d19ed15`)

Built and deployed this branch's `controllermgr` locally (Bazel build → `docker build --platform linux/arm64` → `k3d image import` → `kubectl set image` + `helm upgrade --set-string controllermgr.workflowClient.executionUrlFormat=...`), then ran both `xgb-train` and `pytorch-train` example pipelines end-to-end against a live k3d sandbox (Cadence workflow engine).

**Result: PASS.** The "Execute Workflow" step's `LogUrl` is now populated with a real, run-specific Cadence Web link only after `StartWorkflow` returns (not at PENDING-state creation, matching this PR's fix), e.g.:

```
http://localhost:8088/domains/default/workflows/run-20260810-055510-4e5cbf2c/3e332171-96af-43b1-be68-8d3b74d29d84/
```

Confirmed this resolves to a live, correct Cadence Web summary page (`curl -sL` → final `HTTP 200`) for both pipeline runs. Both runs also completed successfully end-to-end (model + eval report + datasets pushed for `xgb-train`; all associated Ray jobs `SUCCEEDED` for `pytorch-train`).

One unrelated gotcha hit during setup, not a finding on this PR: a stale sandbox needed `ma sandbox sync`/a manual `helm upgrade --set workflow.endpoint=michelangelo-cadence-frontend:7833` to fix, since `values-k3d.yaml`'s own `workflow.endpoint: cadence:7833` default doesn't match the real Cadence service name — `_build_helm_set_args()` already overrides this correctly, this only bit me because I ran a raw `helm upgrade` by hand without going through `ma sandbox sync`.

sallycr:
## Temporal-engine verification

Followed up on the Cadence verification above by also testing against a live **Temporal**-backed sandbox, since this fix touches the workflow-engine-agnostic `GetWorkflowUrl`/`executionUrlFormat` path.

**Setup:** fresh `ma sandbox create --workflow temporal` (clean cluster, no leftover state from the Cadence run), rebuilt `controllermgr` from this branch's latest commit, re-applied the `california-housing` project/pipelines.

**Result:** both example pipelines completed successfully end-to-end, and the "Execute Workflow" step's `logUrl` was correctly deferred until `StartWorkflow` returned a real run ID, then built from the run-specific Temporal Web URL template — exactly as designed:

- `xgb-train` (`run-20260810-133530-a6b68827`): `http://localhost:8080/namespaces/default/workflows/run-20260810-133530-a6b68827/ae78ea35-84a5-4a47-b8a7-c9519f9b05e5/history` → HTTP 200
- `pytorch-train` (`run-20260810-135110-f9ed7416`): `http://localhost:8080/namespaces/default/workflows/run-20260810-135110-f9ed7416/3c3a1c88-55d8-4e96-8ffb-02aa9adfa435/history` → HTTP 200

Both driver logs confirm successful model + dataset pushes (`[ ok ]`). No code changes needed on the Temporal side — confirms the fix works for both engines, not just Cadence.

(Local port note: my sandbox had a host port 8080 conflict unrelated to this PR, so I tested against a locally remapped port and normalized the URLs above back to the default 8080 for readability — the resolved path/template behavior is identical either way.)

sallycr:
## Re-verified Temporal-engine (latest commit)

Re-ran the Temporal verification against this branch's latest commit (`2d1c1dfb`), since two commits landed after the previous Temporal/Cadence verification (`Update sandbox.py` — re-added `/summary` to the Cadence URL — and the new Go regression test for `GetWorkflowUrl`). Fresh `ma sandbox create --workflow temporal`, rebuilt `controllermgr` from this commit.

**Unrelated hiccup along the way:** the fresh sandbox's Temporal `default` namespace wasn't registered before the worker started (a startup-ordering race between the namespace-setup Job and the Temporal frontend becoming ready — not caused by this PR), so the worker crash-looped until I registered it manually and restarted the worker. Not something this PR needs to fix, just flagging in case it's a recurring rough edge worth a follow-up.

**Result:** both example pipelines completed successfully end-to-end on Temporal, with the "Execute Workflow" step's `logUrl` again correctly deferred until `StartWorkflow` returned a real run ID:

- `xgb-train` (`run-20260810-193913-fbc642b8`): `http://localhost:8080/namespaces/default/workflows/run-20260810-193913-fbc642b8/f42b366f-1b37-4825-a09c-1a5ff8f2306f/history` → HTTP 200
- `pytorch-train` (`run-20260810-194300-4e7298b1`): `http://localhost:8080/namespaces/default/workflows/run-20260810-194300-4e7298b1/069d1abe-ee9c-4e87-95ed-0e8c84c04580/history` → HTTP 200

Both driver logs confirm successful model + dataset pushes (`[ ok ]`). (Same local port-8080-conflict caveat as before — tested against a remapped port locally, normalized back to 8080 above; behavior is identical.)

The Temporal-side fix continues to hold on the latest commit.

badcount (approved):

sallycr:
Re-verified against the latest commit (`78fa0e75`, "temporal doesn't need runId") on a fresh Temporal-backed sandbox.

**Change tested:** Temporal's `executionUrlFormat` simplified from `.../workflows/{{.ExecutionID}}/{{.RunID}}/history` to `.../workflows/{{.ExecutionID}}` (no run ID needed).

**Verification approach:** since a plain HTTP 200 against the Temporal Web UI isn't proof by itself (it's a client-rendered SPA that returns 200 for any path via catch-all routing), I validated the claim at the Temporal server/API level as well:

```
tctl --address michelangelo-temporal-frontend:7233 workflow describe -w <workflowId>
```

(no `-r`/runId flag) correctly resolved to the current run for both pipeline runs below — confirming Temporal's workflow-ID-only lookup works as the PR assumes, not just that the URL happens to 200.

**End-to-end pipeline runs**, both reaching `PIPELINE_RUN_STATE_SUCCEEDED` with the simplified URL live via `helm upgrade --set-string controllermgr.workflowClient.executionUrlFormat=...`:

- `xgb-train`: `run-20260810-201446-5cecd002` (RunID `c28e696a-3cfb-4f30-a525-9668037b4641`) — driver pod completed with `[ ok ]`, model + train/validation dataset pushes all succeeded.
- `pytorch-train`: `run-20260810-202100-a066c73e` (RunID `9fb41ed3-f090-4524-907a-abc33fa8e9df`) — driver pod completed with `[ ok ]`, model + train/validation dataset pushes all succeeded.

logUrl format resolved to e.g. `http://localhost:8080/namespaces/default/workflows/run-20260810-202100-a066c73e` and loaded correctly in Temporal Web.

Both the deferred run-specific logUrl fix and this run-ID simplification look good across Cadence and Temporal, for both example pipelines.

PR #1721, opened 2026-08-07
Merge back release/v0.8 to main

sallycr:
Merge-back for the v0.8.0 release. Tracking issue: #1718. Marked ready for review once the release is promoted to final.

PR #1768, opened 2026-08-10
fix(helm): drop placeholder signKey annotation from chart metadata

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Removes the `artifacthub.io/signKey` annotation from `helm/michelangelo/Chart.yaml`. Three lines, no other changes.

```
-  artifacthub.io/signKey: |
-    fingerprint: 0000000000000000000000000000000000000000
-    url: https://github.com/michelangelo-ai/michelangelo/releases
```

**Why?**
The fingerprint is an all-zero placeholder, and the chart is not actually signed. Artifact Hub reads `artifacthub.io/signKey` as a claim that a chart is signed and surfaces it to users deciding whether to trust the package, so a placeholder there is worse than the annotation being absent: it advertises a verification path that does not exist and cannot be followed. Removing it makes the chart metadata honest about its current state.

Deleting rather than filling it in, because there is no key to fill in with. If chart signing gets set up later, the annotation should come back at the same time as a real fingerprint, in the same change.

**How did you test it?**
Checked that the chart is genuinely unsigned rather than assuming it from the placeholder. `helm package helm/michelangelo` is invoked in `release.yaml` and `nightly.yml` with no `--sign`, no keyring, and no `.prov` provenance file produced or published. Searched all of `.github/workflows/` for `--sign`, `gpg`, `keyring`, and `.prov` -- the only provenance in the repo is `npm publish --provenance`, which is unrelated to the Helm chart.

Chart.yaml still parses as valid YAML, and the other six `artifacthub.io/*` annotations (links, maintainers, changelog and so on) are untouched. This is the only placeholder value left in the file.

**Potential risks**
Very low, and reversible. Artifact Hub will stop displaying a signing-key entry for the chart, which is the intended effect since there is no key. No chart behavior, template, or value changes -- this is metadata only, so nothing about an install or upgrade is affected.

**Breaking Changes**
- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.

### DIFF
--- a/helm/michelangelo/Chart.yaml
+++ b/helm/michelangelo/Chart.yaml
@@ -65,6 +65,3 @@ annotations:
   artifacthub.io/maintainers: |
     - name: Michelangelo Maintainers
       email: maintainers@michelangelo-ai.org
-  artifacthub.io/signKey: |
-    fingerprint: 0000000000000000000000000000000000000000
-    url: https://github.com/michelangelo-ai/michelangelo/releases


paulzim (approved):
**Approve** — honest, minimal metadata fix.

Verified rather than took on faith:

- Diff is exactly the 3 lines described: removes the `artifacthub.io/signKey` block (all-zero placeholder fingerprint + releases URL) from `helm/michelangelo/Chart.yaml`. Deletion only, nothing else touched.
- Confirmed the "only six other annotations remain, untouched" claim: `category`, `license`, `prerelease`, `changes`, `links`, `maintainers` — exactly six, all still present and unmodified.
- Confirmed the chart is genuinely unsigned, not just assuming it from the placeholder: `grep -rniE "--sign|gpg|keyring|\.prov" .github/workflows/` returns nothing. Both `helm package` invocations (`release.yaml:162`, `nightly.yml:255`) run with no signing flags and produce no `.prov` file.
- Chart.yaml still parses as valid YAML.
- Scoping note: this PR is based on pre-#1721 `main` (still shows Chart.yaml at `0.7.0`/`ba7c5e8f` as the merge base), so it doesn't carry the pending v0.8.0 version bump — that's expected since #1721 hasn't merged yet, and this diff is otherwise fully self-contained regardless of merge order.

Rationale is sound: an all-zero fingerprint pointing at a real-looking URL is worse than no annotation at all, since Artifact Hub surfaces `signKey` as a trust signal to users. Deleting rather than stubbing is the right call — nothing to fill it with, and re-adding it alongside a real fingerprint if/when signing is set up is the correct sequencing.

Metadata-only, zero runtime/install risk. CI hasn't started yet (only `license/cla`) — same fork-approval gate as the last couple.

PR #1729, opened 2026-08-08
feat(native_transform): add auxiliary transform helpers (PR B8)

sallycr:
## Summary

Ports the `native_transform` "aux island" — `ClipAndScale`, `TimeDuration`, and the imperative `generate_*_transformation` tensor_map helpers — to `lib/native_transform/torch/`. This is PR B8 of the native_transform migration (see `oss_design.md` PR B breakdown).

**Self-contained island, off the assembler-critical path.** `ClipAndScale`/`TimeDuration` are not registered in `TORCH_TRANSFORM_LAYERS_DICT`; `transform_utils.py` composes them (and the already-merged `IDHashTokenizer` wrapper layer from PR B4) imperatively into a `tensor_map`, an alternate usage path distinct from the main `TransformSpec`/`TorchTransformModule` DAG machinery. Verified via grep that nothing else in `native_transform/` imports these three modules, in both this repo and internal.

## Changes

- `lib/native_transform/torch/scale.py` — `ClipAndScale`: clip + scale + optional dtype cast over one or more concatenated input columns.
- `lib/native_transform/torch/duration.py` — `TimeDuration`: floored, unit-scaled duration between two timestamp tensors, with optional log scaling and post-log clipping.
- `lib/native_transform/torch/transform_utils.py` — `generate_numerical_scaled_transformation`, `generate_concatenation_transformation`, `generate_cast_transformation`, `generate_duration_transformation`, `generate_idhash_tokenization_transformation`, `update_output_tensor_map`.
- `lib/native_transform/torch/__init__.py` — export `ClipAndScale` and `TimeDuration` at the package root, alongside the existing layer exports.
- Scrubbed an internal source-attribution URL comment that was present at the top of the internal `scale.py`/`duration.py`/`transform_utils.py` files; behavior is otherwise a faithful port (clipping bounds, duration unit conversion, log-before-clip ordering, empty-vocabulary skip, and other edge cases match internal's test coverage).
- No deliberate behavior changes vs. internal.

## Tests

- New: `tests/test_scale.py`, `tests/test_duration.py`, `tests/test_transform_utils.py` — forward-semantics coverage matching internal's edge cases (clipping, scaling, dtype resolution/errors, 1D unsqueeze, reshaping, log-scale-before-clip ordering, missing-feature no-op, empty-vocabulary skip, custom output columns) plus `torch.jit.script`/`torch.jit.trace` round-trip tests (script+save+load, and trace across dtypes) for the two new layers.
[...]

PR #1772, opened 2026-08-11
feat(native_transform): add schema bridge for fitted transform DAGs (PR B7)

sallycr:
## Context

This is PR B7 of the `native_transform` migration split (B1-B8), and the last sub-PR before PR C (Ray compute lib). B1-B6 and B8 are already merged (#1570, #1603, #1683, #1684, #1685, #1730, #1729).

Once a `TransformSpec` DAG (PR B6) is fitted and materialized into a `TorchTransformModule`, something needs to describe that module's input/output contract as a `ModelSchema` — the shape/dtype metadata Triton config generation and served-model validation consume. This PR adds that bridge.

## Changes

- New `michelangelo/lib/native_transform/torch/schema.py`:
  - `derive_native_transform_schema(transform_spec, transform_module, sample_data=None)`: the public entry point. Filters output columns via the spec's `columns_to_keep`, optionally runs `sample_data` through the module to infer shapes/dtypes, and assembles a `ModelSchema`.
  - Supporting helpers: `_to_batched_tensor` (sample value -> batched tensor), `_get_sample_shapes_and_dtypes` (forward-pass-driven inference), `_collect_input_shape_overrides` / `_validate_input_shape_overrides` (the spec's manual `input_shape` overrides, needed for columns a forward pass can't pin — e.g. a ragged array's fixed max length, or an all-null sample column), `_build_schema_items` / `_filter_output_cols` / `_create_native_transform_schema` (schema assembly with a documented dtype/shape merge priority: spec declarations win over forward-pass inference).
  - Imports `DataType`, `ModelSchema`, `ModelSchemaItem` from the existing `michelangelo.lib.model_manager.schema` package rather than redefining them.
- New `michelangelo/lib/native_transform/torch/tests/test_schema.py`: pytest-style tests covering every function, mirroring the source repo's equivalent test coverage (shape/dtype inference across dtype combinations including int/float promotion, null-column handling, `columns_to_keep` filtering, the `input_shape` override guard and its misplaced-override error path, and full end-to-end `derive_native_transform_schema` runs).

## Design notes

- Preserves the internal implementation's exact dtype/shape merge semantics (a spec-declared `output_dtype` always overrides a forward-pass-derived dtype, even in the one internal-only case where `output_dtype` isn't a real torch dtype — kept as-is for behavioral parity, since no current layer spec in this codebase actually hits that path).
- No changes to `transform_spec.py`, `base_transform_module.py`, or any other B1-B6/B8 file — this PR only adds new files and imports from what's already merged.

## Testing

- `pytest michelangelo/lib/native_transform/` (full suite, not just the new file): 433 passed, 0 failed, no regressions.
- `ruff format --check` and `ruff check`: both clean on the new files.
- Coverage: automated `coverage run` segfaults on this torch-importing target in the sandbox test venv (a known environment limitation, not specific to this PR); coverage was instead verified by manual traceability — every function in `schema.py` has at least one dedicated test class, and every branch (derived vs. spec-declared dtype/shape, override validation, null-sample fallback, `columns_to_keep` filtering with and without exclusions) has an explicit assertion.

## Scope

This PR is self-contained and does not overlap with PR C (Ray compute lib), which starts only after this merges.

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/schema.py
+++ b/python/michelangelo/lib/native_transform/torch/schema.py
@@ -1,0 +1,363 @@
+"""Bridges a fitted ``TransformSpec`` DAG to a ``ModelSchema``.
+
+Once a :class:`~michelangelo.lib.native_transform.torch.transform_spec.TransformSpec`
+has been fitted and materialized into a
+:class:`~michelangelo.lib.native_transform.torch.base_transform_module.TorchTransformModule`,
+:func:`derive_native_transform_schema` describes the module's input/output
+contract as a :class:`~michelangelo.lib.model_manager.schema.ModelSchema` —
+the shape/dtype metadata Triton config generation and served-model validation
+consume.
+
+Shapes and dtypes are resolved from two sources, in priority order:
+
+* A forward pass over ``sample_data`` (when given), which runs the module and
+  reads the resulting tensors' shapes and dtypes directly.
+* The spec's own declarations: a level-0 layer's ``input_dtype`` (populated
+  from the dataset's Arrow schema), a layer's ``output_dtype`` (user-set), and
+  a layer's ``input_shape`` override (needed for columns a forward pass alone
+  cannot pin, such as a ragged array's fixed max length or an all-null sample
+  column).
+
+Spec declarations win over the forward pass wherever both are present, since
+they encode an explicit contract rather than an inference from one sample row.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+import torch
+
+from michelangelo.lib.model_manager.schema import DataType, ModelSchema, ModelSchemaItem
+
+if TYPE_CHECKING:
+    from michelangelo.lib.native_transform.torch.base_transform_module import (
+        TorchTransformModule,
+    )
+    from michelangelo.lib.native_transform.torch.transform_spec import TransformSpec
+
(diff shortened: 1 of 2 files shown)


python/michelangelo/lib/native_transform/torch/schema.py
```diff
@@ -0,0 +1,363 @@
...
+        A tuple of ``(batched_tensor, shape)``, where ``shape`` excludes the
+        batch dimension. Returns a zero-filled fallback tensor and ``None``
+        shape when ``value`` is ``None`` or fails conversion.
+    """
+    fallback = torch.zeros([1, 1], dtype=dtype_hint or torch.float32)
+    if value is None:
+        return fallback, None
+    try:
+        tensor = torch.as_tensor(value)
+    except (TypeError, ValueError):
```
kechen2013: Wondering if this should stay `except Exception` like internal? `torch.as_tensor` raises `RuntimeError` (not `TypeError`/`ValueError`) for a list with a null element, `Decimal`, `datetime`, or a dict, so a `TensorColFillNone` column sampling `[1.0, None, 3.0]` now propagates out of `derive_native_transform_schema` instead of falling back to the zero-filled tensor the docstring promises. The output forward pass just below still catches broadly.

sallycr: thanks for catching this. updated with Exception and adding the tests.

kechen2013 (approved):

PR #1773, opened 2026-08-11
feat(native_transform): add Ray execution adapter (PR C)

sallycr:
## Context

Next PR in the `native_transform` migration after B1–B8 (torch compute core:
`TransformSpec`, `TorchTransformModule`, `TORCH_TRANSFORM_LAYERS_DICT`, all
merged). This PR ports the Ray execution adapter — the piece that runs a
`TransformSpec` DAG over a `ray.data.Dataset` in production (statistics
hydration, level-by-level materialization, batched torch inference).

## What changed

Consolidated 4 internal source files into a single
`uniflow/plugins/ray/native_transform.py`, matching the flat `io.py`/`task.py`
convention already used by sibling plugins in that package:

- **Dataset I/O** (`read_dataset`, `write_dataset`, `filter_columns`)
- **Statistics hydration** (`compute_numerical_statistics`,
  `numerical_statistics_preparation`, `check_stats_exist`,
  `get_numerical_stats_names`)
- **DAG orchestration** (`transform`, `get_transform_torch_inference`,
  `get_torch_dtype`)
- **Torch batch inference** (`DataProcessor`, `DefaultDataProcessor`,
  `TorchBatchPredictor`, `_pad_ragged_arrays`)

### Deviations from internal (intentional, per IMPL-001)

- **Filesystem seam, no custom registration.** Internal registers a
  proprietary filesystem implementation into `fsspec` at import time and lets
  `read_dataset`/`write_dataset` take a `use_fsspec` bool to opt in or out of
  it. That registration doesn't belong in the OSS package (internal consumers
  can still do it in their own entrypoint), so this PR removes it entirely
  and routes both functions through the filesystem-resolution helper already
  used by the sibling `RayDatasetIO` in `uniflow/plugins/ray/io.py`
  (env-driven fsspec-vs-native-PyArrow selection). This drops the
  `use_fsspec` parameter — the seam itself decides.
- **Logging.** Internal's custom logger wrapper is replaced with stdlib
  `logging`, matching every other OSS plugin module.
- **Numpy/pyarrow helpers.** `pyarrow_to_numpy`, `assemble_output_table`,
  `pad_ragged_tensor`, `infer_dtype` now come from the already-migrated
  `lib/_internal/utils/numpy_utils` package (from PR A) instead of being
  reimplemented. `_pad_ragged_arrays` keeps its own validation (empty/
  non-ndarray/non-numeric checks matching internal's contract) but delegates
  the actual padding to the shared `pad_ragged_tensor`, which already handles
  sentinel selection internally — so the local sentinel-computation logic
  internal had is no longer needed.
- **Dropped two unused block-size tuning constants** that internal ships
  alongside this module but never references anywhere in its own source —
  confirmed via search across the internal module. Not ported (CODE-004: no
  dead code).

### Behavioral equivalence (self-review as CanvasFlex Architect)

Verified line-by-line against internal's 4 source files:
- `transform()`'s per-level loop (dtype resolution, stats-hydration
  short-circuit via `check_stats_exist`, `map_batches` kwargs including the
  CPU/GPU dispatch and `num_cpus=1` GIL-bound-inference note) is unchanged.
- `TorchBatchPredictor.__call__`'s zero-copy passthrough contract (only
  `input_columns` reach the model; everything else stays native Arrow) and
  its default overwrite-on-collision behavior for output columns is
  unchanged.
- `DefaultDataProcessor.prepare_inputs`'s numeric-only validation, object-dtype
  stacking, and ragged-array padding fallback chain (`np.stack` ->
  `_pad_ragged_arrays` -> recursive `tolist()` conversion) is unchanged.
- `get_transform_torch_inference`'s `columns_to_keep` union with
  future-level input columns is unchanged.

## Tests

- 68 unit tests (`native_transform_test.py`), fully mocked for Ray-dependent
  paths, real torch/numpy/pyarrow execution for `DefaultDataProcessor`,
  `_pad_ragged_arrays`, and `TorchBatchPredictor` — one test per function and
  per internal branch (non-numeric skip, GPU dispatch, columns_to_keep
  variants, ragged-array edge cases, etc).
- 2 real (non-mocked) Ray integration tests
  (`native_transform_integration_test.py`) running `transform()` end-to-end
  against a local single-process Ray cluster. Following the precedent set by
  trainer PR 4's Ray V2 auto-resume integration tests, these are **skipped in
  CI by default** (a real Ray cluster can be memory-heavy on constrained
  runners) — set `MICHELANGELO_RUN_RAY_INTEGRATION_TESTS=1` to force them on;
  they run by default locally.
- Full regression: `pytest michelangelo/uniflow/plugins/ray/` (103 passed)
  and `pytest michelangelo/lib/native_transform/` (393 passed) — no
  regressions.
- Coverage tooling hit the same torch/numpy C-extension incompatibility
  under `coverage run` noted in PR #1772 (B7) — used the same fallback:
  manual traceability review confirming a dedicated test per function and
  per branch (see the by-function list above under Tests) instead of a
  coverage percentage.
- `ruff format --check` and `ruff check`: clean on all 3 touched files.

## Scope notes

- Depends on PR A (`#1543`, pyarrow helpers) and PR B6 (`#1730`, `TransformSpec`
  + `TorchTransformModule` + `get_transform_module`), both merged.
- No changes to `RayDatasetIO`, `SparkIO`, `RayTask`, `SparkTask`, or the
  `io_registry` — additive only.

### DIFF
--- a/python/michelangelo/uniflow/plugins/ray/native_transform.py
+++ b/python/michelangelo/uniflow/plugins/ray/native_transform.py
@@ -186,11 +186,7 @@ def compute_numerical_statistics(
     if not aggregate_fns:
         return numerical_stats
 
-    batch_count = (
-        len(numerical_statistics_computation_specs)
-        // numerical_statistics_batch_fn_size
-        + 1
-    )
+    batch_count = len(aggregate_fns) // numerical_statistics_batch_fn_size + 1
     for i in range(batch_count):
         batch = aggregate_fns[
             i * numerical_statistics_batch_fn_size : (i + 1)
--- a/python/michelangelo/uniflow/plugins/ray/native_transform.py
+++ b/python/michelangelo/uniflow/plugins/ray/native_transform.py
@@ -1,0 +1,749 @@
+"""Ray execution adapter for the ``native_transform`` torch compute core.
+
+Runs a fitted :class:`~michelangelo.lib.native_transform.torch.TransformSpec`
+DAG over a ``ray.data.Dataset``: hydrates any missing numerical statistics
+(percentiles, min/max/mean/std) level by level, materializes each level as a
+:class:`~michelangelo.lib.native_transform.torch.TorchTransformModule`, and
+runs it as a Ray ``map_batches`` stage via :class:`TorchBatchPredictor`. The
+same ``TransformSpec`` DAG is TorchScript-exportable, so training-time
+transforms here match serving-time transforms exactly.
+
+Filesystem access for dataset I/O goes through the
+:func:`~michelangelo.uniflow.plugins.ray.io._fs_path` seam already used by
+:class:`~michelangelo.uniflow.plugins.ray.io.RayDatasetIO`, so this module has
+no filesystem-registration side effects of its own.
+"""
+
+from __future__ import annotations
+
+import logging
+import time
+from typing import Any
+
+import numpy as np
+import pyarrow as pa
+import ray
+import torch
+
+from michelangelo.lib._internal.utils.numpy_utils import (
+    assemble_output_table,
+    infer_dtype,
+    pad_ragged_tensor,
+    pyarrow_to_numpy,
+)
+from michelangelo.lib.native_transform.torch import (
+    TorchTransformBaseLayer,
+    TransformSpec,
+    get_transform_module,
+)
+from michelangelo.lib.native_transform.torch.constants import (
+    DEFAULT_NUMERICAL_OUTPUT_DTYPE,
+)
+from michelangelo.uniflow.plugins.ray.io import _fs_path
+
+_logger = logging.getLogger(__name__)
+
+DEFAULT_MAP_BATCH_CONCURRENCY = 100
+DEFAULT_NUM_STATS_BATCH_FN_SIZE = 500
+DEFAULT_MAP_BATCH_SIZE = 20000
+
+# Only numeric pyarrow types are supported by the torch transform layers.
+PYARROW_TYPE_TO_TORCH_DATA_TYPE_MAP: dict[pa.DataType, torch.dtype] = {
+    pa.int32(): torch.int32,
+    pa.int64(): torch.int64,
+    pa.float32(): torch.float32,
+    pa.float64(): torch.float64,
+}
+
+__all__ = [
+    "DataProcessor",
+    "DefaultDataProcessor",
+    "TorchBatchPredictor",
+    "check_stats_exist",
+    "compute_numerical_statistics",
+    "filter_columns",
+    "get_numerical_stats_names",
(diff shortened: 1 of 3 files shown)


python/michelangelo/uniflow/plugins/ray/native_transform.py
```diff
@@ -0,0 +1,749 @@
...
+        len(numerical_statistics_computation_specs)
+        // numerical_statistics_batch_fn_size
+        + 1
+    )
+    for i in range(batch_count):
+        batch = aggregate_fns[
+            i * numerical_statistics_batch_fn_size : (i + 1)
+            * numerical_statistics_batch_fn_size
+        ]
+        numerical_stats.update(ray_data_df.aggregate(*batch))
```
kechen2013: Wondering if `batch_count` should come from `len(aggregate_fns)`? It counts specs (columns), but the slices index `aggregate_fns` — 200 columns with mean/std/min/max builds 800 aggregates and only the first 500 ever run. The rest silently fall through to `stats_dict.get(f"{col}_std", 1.0)` in `update_standard_scaler_specs`, so those columns serve unnormalized. Matches internal exactly, so nothing regressed in the port — just looks like the bound is off.

python/michelangelo/uniflow/plugins/ray/native_transform.py
```diff
@@ -0,0 +1,745 @@
...
+        if spec["std"] and f"{input_col}_std" not in existing_numerical_stats:
+            aggregate_fns.append(Std(input_col, alias_name=f"{input_col}_std"))
+        if spec["mean"] and f"{input_col}_mean" not in existing_numerical_stats:
+            aggregate_fns.append(Mean(input_col, alias_name=f"{input_col}_mean"))
+
+    numerical_stats: dict[str, float] = {}
+    if not aggregate_fns:
+        return numerical_stats
+
+    batch_count = len(aggregate_fns) // numerical_statistics_batch_fn_size + 1
```
kechen2013, line 189: Wondering if this wants ceiling division rather than `+ 1`? Now that the bound tracks `aggregate_fns`, `len(aggregate_fns) % batch_fn_size == 0` gives a guaranteed empty trailing slice — 125 columns with max/min/mean/std is exactly 500 — and `ray_data_df.aggregate()` with no aggregators dies inside Ray (`assert self._columns` in `table_block.py`, surfaced as `RayTaskError(AssertionError)`). Repro'd on ray 2.56.1 with 8 aggregates and `batch_fn_size=4`; `-(-len(aggregate_fns) // batch_fn_size)` avoids it. The new test's mock returns `{}` for the empty batch, so it stays green either way.

kechen2013 (approved):

PR #1774, opened 2026-08-11
feat(native_transform): add TransformSpecIO + usage guide (PR E)

sallycr:
## Summary

Last sub-PR of the `native_transform` migration (PR E — "TransformSpecIO + docs/DX polish" from `specs/native_transform/oss_design.md`). Depends only on `TransformSpec` (merged in PR B6, #1730); no dependency on the still-open B7 (#1772) or PR C (#1773).

`TransformSpec` already serializes itself via `to_json`/`load_from_json`. This PR wraps that in the `IO[T]` protocol so a fitted spec round-trips as a first-class Uniflow workflow value the same way a `DataFrame` or `Dataset` does, and adds a user guide for the module.

## Changes

- **`lib/native_transform/torch/io.py`** — new `TransformSpecIO(IO[TransformSpec])`, following the `ProtoIO`/`SparkIO`/`RayDatasetIO` pattern (fsspec-backed read/write, no extra metadata needed since the JSON is self-describing).
  - `read()` bypasses `TransformSpec.__init__` (which requires a yaml path or raw spec dict) via `TransformSpec.__new__` + `load_from_json`, since `load_from_json` already fully replaces every attribute `__init__` would set — this avoids needing a throwaway placeholder spec dict just to satisfy the constructor.
- **`lib/native_transform/torch/__init__.py`** — registers `default_io[TransformSpec] = TransformSpecIO` (class, not instance — matches the existing `BytesIO: BytesIOIO` registration's lazy-instantiation pattern in `IORegistry.__getitem__`). Registered eagerly at package import time since this `__init__.py` already imports torch-dependent modules unconditionally, so there's no laziness to preserve.
- **`docs/user-guides/train-and-deploy-models/native-feature-transforms.md`** — new usage guide: what the module does (spec DAG + fit/materialize/export lifecycle), a full runnable example (define spec → fit stats → hydrate placeholders → persist via `TransformSpecIO` → materialize `TorchTransformModule` → run + TorchScript-export), and how to extend it with a new layer/spec pair or a new `IO[T]` handler. Linked from `docs/user-guides/index.md`'s Quick Navigation table.

## Tests

- New: `lib/native_transform/torch/tests/test_io.py` — 7 tests: write/read round-trip fidelity (`to_dict()` equality), `write()` returning no metadata, on-disk JSON matching `to_json()` exactly, `FileNotFoundError` on missing file, `JSONDecodeError` on malformed JSON, `default_io` registration presence/type, and independence of a restored instance from later mutation of the original.
- Full `pytest lib/native_transform/`: 400 passed.
- Full `pytest uniflow/`: 50 passed (no regressions to existing IO plugins).
- The doc guide's example was run end-to-end against this branch (fit → hydrate → `TransformSpecIO` round-trip → materialize → forward pass → `torch.jit.script`) to confirm it actually works before publishing.

## Lint

`ruff format --check` and `ruff check` on all new/touched files: clean.

## Scope notes

This is the last PR in the `native_transform` migration's current sequence (A, A2, B1–B8, C, E). No behavior change to `TransformSpec` itself — this PR only adds an adapter and documentation around its existing serialization.

### DIFF
--- a/python/michelangelo/lib/native_transform/torch/__init__.py
+++ b/python/michelangelo/lib/native_transform/torch/__init__.py
@@ -54,12 +54,6 @@ from michelangelo.lib.native_transform.torch.transform_utils import (
     update_output_tensor_map,
 )
 from michelangelo.lib.native_transform.torch.utils import generate_layer_name
-from michelangelo.uniflow.core.io_registry import default_io
-
-# Register TransformSpec as a first-class IO-serializable workflow value, so
-# it can be passed between Uniflow tasks the same way a DataFrame or Dataset
-# is (see michelangelo.uniflow.core.io_registry.default_io).
-default_io[TransformSpec] = TransformSpecIO
 
 __all__ = [
     "TORCH_TRANSFORM_LAYERS_DICT",
--- a/python/michelangelo/lib/native_transform/torch/__init__.py
+++ b/python/michelangelo/lib/native_transform/torch/__init__.py
@@ -33,6 +33,7 @@ from michelangelo.lib.native_transform.torch.constants import (
     TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP,
 )
 from michelangelo.lib.native_transform.torch.duration import TimeDuration
+from michelangelo.lib.native_transform.torch.io import TransformSpecIO
 from michelangelo.lib.native_transform.torch.scale import ClipAndScale
 from michelangelo.lib.native_transform.torch.stats_layers import (
     Bucketization,
@@ -53,6 +54,12 @@ from michelangelo.lib.native_transform.torch.transform_utils import (
     update_output_tensor_map,
 )
 from michelangelo.lib.native_transform.torch.utils import generate_layer_name
+from michelangelo.uniflow.core.io_registry import default_io
+
+# Register TransformSpec as a first-class IO-serializable workflow value, so
+# it can be passed between Uniflow tasks the same way a DataFrame or Dataset
+# is (see michelangelo.uniflow.core.io_registry.default_io).
+default_io[TransformSpec] = TransformSpecIO
 
 __all__ = [
     "TORCH_TRANSFORM_LAYERS_DICT",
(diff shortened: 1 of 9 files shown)


python/michelangelo/lib/native_transform/torch/__init__.py
```diff
@@ -53,6 +54,12 @@
     update_output_tensor_map,
 )
 from michelangelo.lib.native_transform.torch.utils import generate_layer_name
+from michelangelo.uniflow.core.io_registry import default_io
+
+# Register TransformSpec as a first-class IO-serializable workflow value, so
+# it can be passed between Uniflow tasks the same way a DataFrame or Dataset
+# is (see michelangelo.uniflow.core.io_registry.default_io).
+default_io[TransformSpec] = TransformSpecIO
```
kechen2013: Wondering if `TransformSpecIO` belongs under `uniflow/plugins/` next to `PandasIO`/`RayDatasetIO`? Registering it here makes `lib.native_transform.torch` import `uniflow.core` (so `context`/`build`/`remote_run` load with it) and mutate `default_io` as an import side effect, where every other IO adapter sits on the uniflow side. Non-blocking, just cheaper to settle before the import path is public.

python/michelangelo/lib/native_transform/torch/__init__.py
```diff
@@ -33,6 +33,7 @@
     TORCH_TYPE_TO_TORCH_DTYPE_CLASS_NAME_MAP,
 )
 from michelangelo.lib.native_transform.torch.duration import TimeDuration
+from michelangelo.lib.native_transform.torch.io import TransformSpecIO
```
kechen2013, line 36: Wondering if the `TransformSpecIO` class itself should move to the new plugin package too? The registration moved, but this import still chains `lib.native_transform.torch` -> `...torch.io` -> `uniflow.core.io_registry`, so importing the layer package keeps pulling in `uniflow.core` (and `context`/`build`/`remote_run` with it) — only the global mutation went away. `RayDatasetIO`/`SparkIO`/`PandasIO` all live in `uniflow/plugins/<x>/io.py`, so moving the class there would leave `lib/` uniflow-free. Non-blocking.

kechen2013 (approved):

PR #1775, opened 2026-08-11
fix(python): sandbox sync kubeconfig context and self-referential compute cluster

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Two independent fixes to `ma sandbox sync`, both root-caused from real CI failures on the `Integration tests` workflow:

1. `_sync()` now runs `k3d kubeconfig merge <cluster> -d --kubeconfig-switch-context` right after `k3d cluster start`. `k3d cluster start` (unlike `k3d cluster create`) never touches kubeconfig, so on a host that's ever had more than one k3d cluster, the ambient `current-context` can silently point at the wrong cluster, and even a stale-but-correctly-named context entry can carry a dead host port from a prior delete+recreate. `k3d kubeconfig merge -d` refreshes the entry (port/certs) and switches the context in one step.
2. `_create_compute_cluster_crd()` now special-cases the self-referential case (a k3d cluster registering itself as a `Cluster` CRD, later read by `apiserver` — a Pod running inside that same cluster — to make outbound Ray/K8s API calls "into" it). It previously always extracted a host-published address (`0.0.0.0:<port>` on Linux) from `k3d kubeconfig get`, which is unreachable from a Pod's own network namespace. It now uses `https://kubernetes.default.svc:443`, which is always resolvable in-cluster and is already a valid SAN on the k3s server's default certificate. The genuinely-separate-cluster path (cluster A registering cluster B) is left as before, plus a best-effort `host.k3d.internal` substitution for loopback-equivalent hosts — noted in-code that this still needs `--tls-san=host.k3d.internal` at that cluster's creation time to pass TLS verification end-to-end; not exercised by this sandbox's own setup (which only ever registers itself).
3. Small fix bundled in: `_helm_dependency_update()` now runs explicitly before any `helm template` call used for introspection (`_helm_delete_services()`, `_helm_adopt_orphaned_resources()`), so a cold chart-dependency cache doesn't pollute `helm template`'s stdout with fetch-status lines ahead of the YAML the callers parse.

**Why?**

CI (`Integration tests` workflow, self-hosted GCP runner) was failing deterministically at the `Sync sandbox` step. Root cause was the runner user's kubeconfig context having drifted to an unrelated k3d cluster on the shared VM (traced via live inspection of the sandbox instance), which every subsequent `kubectl`/`helm` call in `_sync()` silently followed. Fixing that surfaced a second, previously-latent bug: once sync targeted the right cluster, the sandbox's Ray pipeline test (`test_bert_cola_e2e`) still failed because apiserver couldn't reach the compute cluster it had registered for itself, due to the host-published address being unreachable from inside the cluster's own pod network.

**How did you test it?**

Verified live against the real GCP sandbox instance and real CI runs (not just local k3d):
- Manually confirmed and fixed the stale kubeconfig context/entry on the runner user's account, then verified `kubectl get nodes` worked against the intended cluster.
- Live-patched the `Cluster` CRD's `host`/`port` on the running sandbox to `kubernetes.default.svc:443` and confirmed a RayCluster subsequently reached `Launched` with real head/worker pods staying alive and pulling images (vs. instant termination before the fix).
- Triggered the `Integration tests` GitHub Actions workflow against this branch twice (runs `31498187639` and `31501619092`); both completed with the `Sync sandbox` step succeeding cleanly (`helm upgrade ... STATUS: deployed`) and the full workflow concluding `success`.

**Potential risks**

Low. The kubeconfig-merge call is additive and idempotent (safe to run every sync). The compute-cluster host change only affects the self-referential path exercised by this sandbox; the separate-cluster path's behavior is unchanged except for the same loopback-to-`host.k3d.internal` substitution added defensively (not currently exercised in CI).

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -362,7 +362,6 @@ def _sync(ns: argparse.Namespace):
     # Upgrade or install the control plane via Helm.
     # Infrastructure (mysql, cadence, minio, grafana, prometheus) is left running.
 
-    _ensure_mysql_pod_running()
     _refresh_mysql_schema()
 
     _ensure_credentials_secret()
@@ -451,47 +450,6 @@ def _sync(ns: argparse.Namespace):
         )
 
     _helm_wait(ns)
-
-
-def _ensure_mysql_pod_running():
-    """Revive the bare `mysql` Pod if it isn't in the `Running` phase.
-
-    `_sync()`'s fast path leaves infrastructure Pods running rather than
-    recreating them, which assumes the previous sync's `mysql` Pod is still
-    alive. On a long-lived host, a node-level container-runtime restart can
-    leave a bare Pod (no Deployment/ReplicaSet to self-heal it) stuck in a
-    terminal phase (`Succeeded`/`Failed`) even though its containers never
-    exited on their own — `kubectl exec` then fails outright. Detect that
-    and delete+reapply the Pod manifest before `_refresh_mysql_schema()`
-    tries to exec into it.
-    """
-    phase_result = subprocess.run(
-        ["kubectl", "get", "pod", "mysql", "-o", "jsonpath={.status.phase}"],
-        capture_output=True,
-        text=True,
-    )
-    phase = phase_result.stdout.strip()
-
-    if phase_result.returncode == 0 and phase == "Running":
-        return
-
-    print(
-        f"mysql Pod is not Running (phase={phase or 'missing'}) — "
-        "recreating it before refreshing the schema."
-    )
-    subprocess.run(
-        ["kubectl", "delete", "pod", "mysql", "--ignore-not-found=true", "--wait=true"],
-        check=False,
-    )
-    _kube_apply(_dir / "resources" / "mysql.yaml")
-    _exec(
-        "kubectl",
-        "wait",
-        "--for=condition=ready",
-        "pod",
-        "mysql",
-        "--timeout=180s",
-    )
 
 
 def _refresh_mysql_schema():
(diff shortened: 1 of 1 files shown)


python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -331,6 +331,24 @@ def _sync(ns: argparse.Namespace):
...
+    # a delete+recreate) still has the old, now-dead port. `k3d kubeconfig
+    # merge -d` rewrites the entry with the live port/certs and switches to
+    # it in one step (see spec 035 addendum).
+    _exec(
+        "k3d",
+        "kubeconfig",
+        "merge",
+        _michelangelo_sandbox_kube_cluster_name,
+        "-d",
+        "--kubeconfig-switch-context",
```
badcount, line 349: This sounds like a hack, so what's the issue we're facing and trying to solve?

sallycr: @badcount yes. but what happened that it happened in integration test environment as it's using your context. and your context is set for some scalar spark env. this is what we can think about and follow up in integration test env

python/michelangelo/cli/sandbox/sandbox.py
```diff
@@ -1621,27 +1662,48 @@ def _create_compute_cluster_crd(cluster_name: str):
...
-    # Example: "https://host.docker.internal:52910"
-    import re
-
-    match = re.search(r"(https://[^:]+):(\d+)", server_url)
-    if not match:
-        raise ValueError(
-            f"Could not extract cluster host and port from server URL: {server_url}"
-        )
-    host, port = match.groups()
+    if cluster_name == _michelangelo_sandbox_kube_cluster_name:
```
sallycr, line 1665: @badcount this is what we have reported for restricted local host due to the security reason. No longer host.docker.internal is accessible in GCP. (publicly exposed) this is worth to investigate more.

PR #1776, opened 2026-08-11
feat(js): conditional layout with field-reference conditions

craigmarker:
## Summary
Implement condition layout system so form sections can show or hide based on another field's value. Each condition subscribes to exactly one field via `useField`, so only the condition component re-renders when its referenced field changes.

Four comparison operators exposed in the API: exact match, negation (empty fields are treated as "not yet determined" and stay hidden), presence toggle, and array membership. Repeated layout contexts are supported for conditions inside array fields.

## Test plan

I added 5 integration tests covering each operator — `is`, `isNot`, `isEmpty` (both true/false), and `containsAny` — each verifying that conditional items reactively appear and disappear as the referenced field value changes. All 33 config-driven-form tests pass.

### DIFF
--- a/javascript/packages/core/components/form/__tests__/config-driven-form.test.tsx
+++ b/javascript/packages/core/components/form/__tests__/config-driven-form.test.tsx
@@ -491,4 +491,38 @@ describe('ConfigDrivenForm', () => {
       expect(screen.getByLabelText('Advanced Setting')).toBeInTheDocument();
     });
   });
+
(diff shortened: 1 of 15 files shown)


craigmarker:
Closing to restack: this branch has been rewritten to sit on top of #1804 (craig.marker/cdf-condition-is-only) instead of directly on cdf-3-validation, per updated PR-splitting plan. GitHub won't allow an in-place base change for a PR that's part of a native stack, so re-opening as a fresh PR.

PR #1781, opened 2026-08-11
ci: pin apidiff to a Go 1.24-compatible x/exp pseudo-version

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
One line in `.github/workflows/api-surface-check.yml`: `apidiff` is installed from a pinned `golang.org/x/exp` pseudo-version instead of `@latest`.

```
-        run: go install golang.org/x/exp/cmd/apidiff@latest
+        run: go install golang.org/x/exp/cmd/apidiff@v0.0.0-20260209203927-2842357ff358
```

**Why?**
This is the fix proposed in #1752, with the placeholder resolved to a concrete version. Credit for the root cause goes there -- I hit the same failure on #1755 and only found the existing issue afterwards.

Short version: `golang.org/x/exp` has no release tags, so `@latest` floats to the tip of its default branch, and that module's own `go.mod` has required Go >= 1.25 since February. That was harmless while `actions/setup-go` defaulted to `GOTOOLCHAIN=auto`, because `go install` quietly downloaded a newer toolchain. Since the v5 to v7 bump in #1743, `setup-go` sets `GOTOOLCHAIN=local`, so the mismatch is a hard error:

```
go: golang.org/x/exp@v0.0.0-20260811152304-ee035b5b010f requires go >= 1.25.0
(running go 1.24.3; GOTOOLCHAIN=local)
```

`v0.0.0-20260209203927-2842357ff358` is the last x/exp revision before the Go 1.25 bump landed -- it is the parent of `2735e65f0518` ("all: upgrade go directive to at least 1.25.0"), and its `go.mod` still says `go 1.24.0`.

This is the follow-up I offered in #1767, which pinned the golangci-lint installer and left this one alone.

**How did you test it?**
I cannot trigger a workflow run on this repo, so this is verified by inspection rather than by a green check:

- The pinned version resolves: `https://proxy.golang.org/golang.org/x/exp/@v/v0.0.0-20260209203927-2842357ff358.mod` returns HTTP 200 and its go directive is `go 1.24.0`, which the runner's 1.24.3 satisfies.
- `cmd/apidiff` exists at that revision.
- The tool's flag surface is byte-identical between the pinned revision and current tip -- both define exactly `-w`, `-incompatible`, `-allow-internal`, `-m`. The two call sites in this workflow use `-incompatible`, so the pin does not change how the tool is invoked.
- The workflow still parses as valid YAML.
- Confirmed the break is not specific to any one branch: the same failure is present on `fix/mysql-direct-update` and `fix/sandbox-workflow-execution-url-v2`.

**Potential risks**
Low for the change itself. The pin is a fixed pseudo-version, so it cannot drift; when someone wants a newer `apidiff` it becomes an explicit, reviewable bump.

One thing worth flagging separately, because it affects how much this PR is actually worth: **pinning makes the job green again, but the job still does not detect anything.** Both `apidiff` call sites look malformed, and every error is suppressed with `2>/dev/null || true`, so the failures are invisible:

- Line 71 passes `-export`, which is not a flag `apidiff` defines. `flag.Parse` rejects it and the process exits 2. The real flag is `-w FILE IMPORT_PATH`.
- Line 82 passes one positional argument. `apidiff` requires exactly two (`apidiff OLD NEW`) and exits 2 with usage otherwise, so `result` is always empty and `breaking` is always `false`.
- `/tmp/base-api.txt` is written on line 72 and never read anywhere in the repo, so the base checkout and its export step are unused.

I read this out of `cmd/apidiff/main.go` at the pinned revision rather than by running it, so it is worth a second pair of eyes. If that reading is right, the job has never been able to report a breaking change, and #1752's "zero PRs are currently being checked" would remain true after this merges. That is why this says `Refs #1752` rather than `Closes` -- it fixes the install failure and the floating dependency, not the detection gap. Happy to send the invocation fix as a separate PR if you want it, though that one changes the check from "never fires" to "can fire", which felt like a maintainer's call rather than something to slip into a one-line pin.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.

### DIFF
--- a/.github/workflows/api-surface-check.yml
+++ b/.github/workflows/api-surface-check.yml
@@ -62,7 +62,7 @@ jobs:
           cache-dependency-path: head/go/go.sum
 
       - name: Install go-apidiff
-        run: go install golang.org/x/exp/cmd/apidiff@latest
+        run: go install golang.org/x/exp/cmd/apidiff@v0.0.0-20260209203927-2842357ff358
 
       - name: Export base API
         working-directory: base/go


paulzim (approved):
**Approve** — one-line pin, thoroughly self-documented, and I independently re-derived every claim in the PR body rather than taking it on trust. All of it checks out, including the part where the author flags a second, unfixed bug in the same job.

**The pin itself:**
- `golang.org/x/exp/@v/v0.0.0-20260209203927-2842357ff358.mod` → confirmed via the module proxy: `go 1.24.0`, which the repo's own `go/go.mod` (`go 1.24.0`) satisfies.
- `golang.org/x/exp/@v/list` → empty, confirming x/exp has no release tags, so `@latest` was always going to float.
- Current `@latest` (`v0.0.0-20260812173653-3d80eb74bc5b`) → `go 1.25.0`, matching the CI error message quoted in the PR ("requires go >= 1.25.0; running go 1.24.3").
- `actions/setup-go` v6.0.0's release notes list "Improve toolchain handling to ensure more reliable and consistent toolchain selection and management" as a breaking change — the `GOTOOLCHAIN=auto`→`local` default flip the PR describes. This repo's setup-go bump was v5→v7 directly (#1743, confirmed real and merged 2026-08-09), so it inherited that change without a dedicated v6 step.
- Issue #1752 is real and matches the PR's citation precisely, including that it left the version as a placeholder (`@<pinned-pre-2026-02-11-pseudo-version>`) rather than a concrete one — this PR is exactly "resolve the placeholder," which is why it says `Refs #1752` rather than `Closes`.
- #1767 (same author, pins the golangci-lint installer script to a tag instead of `master`) is real and is legitimate precedent for "pin CI tool sources instead of trusting a moving target."

**The secondary claim — verified this one myself since it's the more consequential part of the PR body:** the author flags that even after this fix, the job still can't detect anything, and asks for a second pair of eyes on that reading. I extracted `cmd/apidiff/main.go` straight from the pinned pseudo-version's module zip (not from the author's description) and confirmed both bugs exactly as described:
- Flag surface really is only `-w`, `-incompatible`, `-allow-internal`, `-m` (`flag.String`/`flag.Bool` declarations, lines 22-27) — `-export` (used on workflow line 71) isn't defined anywhere, so `flag.Parse()` rejects it and the process exits 2 before doing anything.
- `main()` hard-requires exactly two positional args (`if len(flag.Args()) != 2 { flag.Usage(); os.Exit(2) }`) — workflow line 82 passes only one, so that call also exits 2.
- Both failures are swallowed by `2>/dev/null || true`, and the resulting `/tmp/base-api.txt` (line 72) is never read anywhere in the repo (grepped the whole tree — one hit, the write itself).

So the job goes from "hard-fails at `go install`" to "passes green but the actual diff logic never runs" — a real improvement (no more spurious red advisory check on every Go PR) but not the fix for #1752's headline concern ("zero PRs are checked for breaking API changes"), and the PR is upfront about that rather than burying it. Correctly scoped as out-of-scope for a pin-only PR — changing a no-op advisory check into one that can actually fire is a maintainer call, per the author's own framing, and would be a good candidate for a follow-up issue if one doesn't exist yet.

CI: full green on this PR itself (`bazel-build`, `dirty-check`, `python-build`, `filter`, `update-release-draft`, `license/cla`; the three advisory checks correctly `skipping` since the diff only touches `.github/workflows/`, not `go/**`/`proto/**`/`helm/**`). No comments on the PR yet.

Correct, minimal, low-risk, and the rare PR where the author already did the adversarial review of their own fix's limits — I just confirmed they were right.

PR #1420, opened 2026-06-30
chore(precommit): add gofmt and go vet hooks for Go files

craigmarker:
**What changed?**

Two hooks added to `.pre-commit-config.yaml`: `gofmt` auto-formats Go files on commit, and `go vet` runs package-level static analysis from the `go/` directory. Both are scoped to `^go/.*\.go$`.

**Why?**

Python and JavaScript both have formatter + linter pre-commit hooks; Go had none. Formatting drift and basic static analysis issues (wrong printf format strings, unreachable code, etc.) could accumulate silently between CI runs.

**How did you test it?**

Verified each hook command directly:

1. **Syntax error blocked by gofmt** — created a file with a missing comma; `gofmt -l -w` exited 2 and blocked the commit.
2. **Formatting auto-fixed by gofmt** — created a file with `x:=1` (missing spaces); `gofmt -l` listed the file as needing changes; `gofmt -w` rewrote it. Pre-commit would see the staged file modified and fail, requiring a re-stage.
3. **go vet violation blocked** — created a file with `fmt.Printf("%d", "not an int")`; `cd go && go vet ./cmd/...` exited 1 with "format %d has arg of wrong type string".

**Breaking Changes**

- [x] No breaking changes

PR #1782, opened 2026-08-11
fix(api): use status.Error for non-format error messages

apurvapatkeshwar:
- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Seven call sites across two files switch from `status.Errorf` to `status.Error`:

```
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
```

- `go/api/handler/handler.go` -- 5 sites
- `go/api/utils/utils.go` -- 2 sites

Nothing else changes. Both files still use `status.Errorf` elsewhere for calls that genuinely format.

**Why?**
`status.Errorf(code, err.Error())` passes a runtime string as the format argument with no arguments after it. In grpc-go v1.71.0 (the version pinned in `go/go.mod`) that forwards to `fmt.Sprintf(format)`:

```go
func Error(c codes.Code, msg string) error { return New(c, msg).Err() }
func Errorf(c codes.Code, format string, a ...any) error { return Error(c, fmt.Sprintf(format, a...)) }
```

So any literal `%` in the underlying error gets read as a verb, and the message the caller receives comes back mangled -- a wrapped error containing something like `100%` or a `%s` from a nested template renders as `%!(NOVERB)` or `%!s(MISSING)`. It is unlikely but not impossible for these paths, and the message is the only thing the client gets.

`status.Error` takes the message directly and cannot do this. It is already the idiom here: `handler.go:285` uses it, as do `go/cmd/kubeproto/protoc-gen-validation/main.go` and several tests.

This also silences `go vet`'s non-constant format string check, which is what the `cd go && go vet ./...` pre-commit hook added in #1420 runs.

**How did you test it?**
By inspection, not by running the suite:

- Checked the substitution is exhaustive rather than partial. A repo-wide scan of all Go files found exactly seven instances of this pattern, and all seven are in this diff. Five are the ones in `handler.go`; the other two are the identical pattern in `utils.go`, which is why this covers both files rather than just the one.
- Confirmed `status.Error(codes.Code, string) error` exists with that exact signature in grpc-go at the pinned v1.71.0, not just on tip.
- The swap is behaviour-preserving for every message without a literal `%`, which is all of them in practice, since `fmt.Sprintf` on a verb-free string returns it unchanged. For messages that do contain a `%`, current output is already wrong and this corrects it. There is no input for which this makes correct output incorrect.
- Both files still reference `status` and `codes`, so no imports are affected.

**Potential risks**
Low. To be clear about what this does and does not do: **CI is green today and stays green either way.** `go/.golangci.yml` sets `disable-all: true` and enables only `godox`, so `govet`/`printf` is not running in the lint job at all, and the Bazel job passes with this code present. This is a lint-hygiene and correctness cleanup that shows up for developers running `go vet` or the pre-commit hook locally, not a fix for anything currently red.

Checked for conflicts against the two open PRs that touch `handler.go` -- #1477 and #896. Neither touches any of these lines, so this should not create work for either.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None needed.

### DIFF
--- a/go/api/handler/handler.go
+++ b/go/api/handler/handler.go
@@ -83,7 +83,7 @@ func (handler *apiHandler) Create(ctx context.Context, obj ctrlRTClient.Object, 
 
 	objMeta, err := meta.Accessor(obj)
 	if err != nil {
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	if metadataErr := handler.handleMetadataStorageForCreate(ctx, obj, objMeta); metadataErr != nil {
@@ -160,7 +160,7 @@ func (handler *apiHandler) Update(ctx context.Context, obj ctrlRTClient.Object, 
 
 	hasSpecChange, err := handler.hasSpecChange(ctx, obj)
 	if err != nil {
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
 	}
 	setUpdateTimestamp(obj, hasSpecChange)
 
@@ -241,7 +241,7 @@ func (handler *apiHandler) Delete(ctx context.Context, obj ctrlRTClient.Object,
 
 	objMeta, err := meta.Accessor(obj)
 	if err != nil {
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	dryRun, err := checkDryRun(opts.DryRun)
@@ -372,13 +372,13 @@ func (handler *apiHandler) DeleteCollection(
 	// List objects from metadata storage
 	gvk, err := ctrlRTApiutil.GVKForObject(objType, scheme.Scheme)
 	if err != nil {
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	listGVK := gvk.GroupVersion().WithKind(gvk.Kind + "List")
 	newObj, err := scheme.Scheme.New(listGVK)
 	if err != nil {
-		return status.Errorf(codes.InvalidArgument, err.Error())
+		return status.Error(codes.InvalidArgument, err.Error())
 	}
 	listObj, ok := newObj.(ctrlRTClient.ObjectList)
 	if !ok {
--- a/go/api/utils/utils.go
+++ b/go/api/utils/utils.go
@@ -21,7 +21,7 @@ import (
 func GetObjectTypeMetaFromList(list runtime.Object, scheme *runtime.Scheme) (*metav1.TypeMeta, error) {
 	listGVK, err := apiutil.GVKForObject(list, scheme)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	kindLen := len(listGVK.Kind)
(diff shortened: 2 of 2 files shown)


paulzim (approved):
**Approve** — small, mechanical fix with a real underlying bug, and I independently confirmed every claim rather than taking the PR body's word for it.

**The bug is real, verified straight from the pinned dependency's source, not just the PR's description of it.** Read `status.go` in the actual `google.golang.org/grpc@v1.71.0` module cache:
```go
func Error(c codes.Code, msg string) error { return New(c, msg).Err() }
func Errorf(c codes.Code, format string, a ...any) error { return Error(c, fmt.Sprintf(format, a...)) }
```
`status.Errorf(code, err.Error())` really does route `err.Error()` through `fmt.Sprintf` as a *format string* with zero args. Any literal `%` in the wrapped error (a percentage, a stray `%s`/`%v` from some nested template, etc.) gets consumed as a verb and comes back mangled or with `%!s(MISSING)` — and since this is a gRPC status message, it's the only thing the client sees. `status.Error` bypasses `Sprintf` entirely, so this is a strict fix, not a stylistic one.

**Exhaustiveness — checked myself, not just trusted the count:** diffed `handler.go`/`utils.go` on `upstream/main` (pre-fix) against the exact anti-pattern `status.Errorf(codes.X, err.Error())` — found exactly 5 hits in `handler.go` (lines 86, 163, 244, 375, 381) and 2 in `utils.go` (lines 24, 46), i.e. exactly the 7 the PR claims, and the diff changes precisely those 7 lines and no others. Re-ran the same grep post-fix (repo-wide, not just these two files): zero remaining hits of the pattern anywhere.

**"Already the idiom here" claim — confirmed:** `handler.go:285` does use `status.Error(codes.Unavailable, "...")` today, pre-existing, unrelated to this diff — so the PR is aligning these 7 call sites with an existing convention in the same file, not inventing a new one.

**"CI is green either way" claim — confirmed via the actual lint config:** `go/.golangci.yml` really does set `disable-all: true` and enable only `godox`; `govet`/`printf` isn't wired into the lint job at all. So this genuinely doesn't move any CI signal from red to green — it's a fix for people running `go vet`/the pre-commit hook locally, exactly as described, not a claimed-but-unverifiable CI fix.

**The one CI check that's currently red on this PR is a red herring, and I confirmed why:** `Go API Surface (advisory)` fails with `golang.org/x/exp/cmd/apidiff@latest: ... requires go >= 1.25.0 (running go 1.24.3; GOTOOLCHAIN=local)` — this is the exact pre-existing, unrelated `apidiff@latest`-floating bug tracked in #1752 and just fixed on the PR I reviewed immediately before this one (#1781, pins the pseudo-version). Nothing to do with this PR's diff; it fails identically on unrelated PRs that touch `go/**`. Every substantive check passes: `Go Test & Coverage (Bazel)`, `Go Lint with TODO Tracking`, `bazel-build`, `python-build`, `dirty-check`, `license/cla`, `update-release-draft`.

**Conflict claim against #1477/#896 — checked at the line level, not just "does the PR touch this file":** both of those open PRs do touch `handler.go`, so a file-level check alone would have been misleading. Pulled both diffs directly: #1477 touches lines ~198-211 (removing a validation block from `UpdateStatus`); #896 touches lines ~122-136 and ~549-557 (`Get`/`handleBlobStorageForGet`) plus `metadata_handler.go`/`handler_test.go`/`BUILD.bazel`, none of it `utils.go`. Neither overlaps this PR's five changed lines (86, 163, 244, 375, 381). The "no conflict" claim holds.

No breaking changes — `status.Error(codes.Code, string) error` exists with that exact signature at the pinned grpc-go version, and both files' `status`/`codes` imports are untouched. No PR comments.

Textbook small fix: real bug, exhaustive and minimal diff, every supporting claim independently reproducible.

Issue #1705, opened 2026-08-07
Docs: Migration guides from kubeflow trainer, kuberay, and sagemaker/vertex


apurvapatkeshwar:
**Is your feature request related to a problem? Please describe.**
Getting started overview already has an ecosystem mapping table that positions the project against tools people currently run, but there is no path shaped documentation for someone arriving with an existing stack: what their PyTorchJob/TrainJob, RayJob/RayCluster, or sagemaker/vertex training job becomes here, which concepts map 1:1, and which have no equivalent yet. For a pre GA project, "how do I get here from what I run today" guides are arguably the highest leverage docs to have.

**Describe the solution you'd like**
3 guides under docs, one per source stack (Kubeflow trainer, Kuberay, Sagemaker/Vertex), each structured as a concept table, a worked before/after example manifest, and an honest gaps section with roadmap links for the missing pieces - extending the overview's own mapping table rather than inventing a new format. 1 PR per guide so each is independently reviewable by whoever knows that ecosystem best.

**Describe alternatives you've considered**
- A single combined migration page (harder to review and to keep current per ecosystem)
- A converter tool (docs are cheaper, and a converter needs the concept mapping written down anyway)

**Component**
Which part of the project does this affect?
- [ ] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [ ] Helm chart / Kubernetes deployment
- [ ] UI (Studio)
- [x] Documentation
- [ ] Other

**Additional context**
Would you accept PRs for these (could start with whichever source stack has most adoption)

PR #1783, opened 2026-08-11
docs: add a migration guide for KubeRay users

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**
Adds a `Migration Guides` section under the user guides, with one guide in it: `docs/user-guides/migration/migrate-from-kuberay.md`.

The guide follows the structure proposed in #1705 -- a concept mapping table, worked before/after manifests, and a gaps section -- and reuses the three-column shape of the overview's ML Workflow Mapping rather than inventing a new one.

**Why?**
#1705 asked for path-shaped docs for people arriving with an existing stack. There are already three places that do concept-level mapping: the overview's ML Workflow Mapping, the glossary's Concept mapping, and the FAQ's "how do I migrate from my current ML stack". None of them map at the resource level, which is the level a KubeRay user actually thinks at -- they have a `RayCluster` YAML in a repo and want to know what it becomes. The guide links to all three instead of restating them.

KubeRay went first because it is the stack where this project's behaviour is most concrete and therefore most verifiable. Ray job launch, persistent Ray clusters via the `RayCluster` resource, and federated dispatch are all listed as available on the roadmap, so most of the page describes shipped behaviour rather than intent.

The framing the guide lands on is that this is not a migration away from KubeRay. `mapRayCluster` builds a real `rayv1.RayCluster` using the upstream operator's own Go types, so the reader keeps their operator and gains a control plane above it. That seemed like the single most useful thing to tell someone evaluating the move, and it is not currently stated anywhere in the docs.

**How did you test it?**
I did not run `bun run build`, so the docs build has not been verified locally -- please treat that as unchecked. Everything else was verified by reading the source:

- All eight internal links were resolved by hand from the new file's location, and both `#anchor` targets were confirmed against real headings (`glossary.md:153`, `overview.md:50`).
- Every field name in the "after" manifests was checked against `proto/api/v2/ray_cluster.proto` and `ray_job.proto`, and cross-checked against the real fixtures at `scripts/ingester/ingester-test-crs/09-raycluster.yaml` and `10-rayjob.yaml`.
- Both Python imports were checked against the real definitions. `RayTask` is re-exported from `michelangelo.uniflow.plugins.ray`, which matches how every example in the repo imports it.
- The `RayTask` field list is the complete dataclass, not a selection.
- Each claim in the gaps section was checked in the negative: `enableInTreeAutoscaling` appears nowhere, `rayservice` appears nowhere, and nothing under `python/` imports Ray Tune.
- The page was then reviewed line by line against `docs/contributing/documentation-guide.md`.

**Potential risks**
Low, but three things are worth a reviewer's judgement rather than mine.

**Placement.** This adds a new section rather than a new page in an existing one. `docs/getting-started/` has `sidebar_position` 1 through 9 already taken, so a guide there lands after Support and Community, which seemed like the wrong home for something #1705 describes as high leverage. `docs/user-guides/` children use 1 through 4, so a new section at 5 needs no renumbering of anything that exists. If you would rather these live somewhere else, moving them is cheap now and gets harder once all three exist.

**One guide, not three.** #1705 proposes one PR per source stack so each can be reviewed by whoever knows that ecosystem. This is the first. Kubeflow Trainer and SageMaker/Vertex would follow the same shape, and I would rather find out this format is wrong now than after writing all three.

**One thing the guide contradicts.** It states there is no Ray Tune integration, which is what the code shows -- nothing under `python/` imports it. But `overview.md:59` lists "Uniflow tasks with Ray Tune" as the hyperparameter tuning equivalent. I left it alone rather than widen a docs PR into someone else's area, but the two cannot both be right, and a reader following the cross-link will hit the disagreement. Happy to fix the overview in this PR, or separately, or to be told the integration exists somewhere I did not find.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
This PR is the documentation change. New section `Migration Guides` under User Guides, with one page for KubeRay.

### DIFF
--- a/docs/user-guides/migration/_category_.json
+++ b/docs/user-guides/migration/_category_.json
@@ -1,0 +1,6 @@
+{
+  "label": "Migration Guides",
+  "position": 5,
+  "collapsed": true,
+  "link": {"type": "generated-index"}
+}
--- a/docs/user-guides/migration/migrate-from-kuberay.md
+++ b/docs/user-guides/migration/migrate-from-kuberay.md
@@ -1,0 +1,206 @@
+---
+sidebar_position: 1
+sidebar_label: "From KubeRay"
+---
+
+# Migrating from KubeRay
+
+This guide is for teams already running Ray on Kubernetes with the KubeRay operator, who are evaluating what changes if they adopt Michelangelo AI.
+
+The short answer is that fewer things change than you might expect. Michelangelo AI does not replace KubeRay — it runs on top of it. The control plane accepts its own `RayCluster` and `RayJob` resources, then creates real `ray.io/v1` resources on a registered compute cluster for the KubeRay operator to reconcile. Your operator, your Ray version, and your cluster topology all survive the move.
+
+:::note
+This page maps **resources and manifests**. For concept-level mapping across MLflow, Kubeflow, Ray, and Airflow, see the [glossary's concept mapping](../../getting-started/glossary.md#concept-mapping). For a broader tool-by-tool view, see [ML Workflow Mapping](../../getting-started/overview.md#ml-workflow-mapping) in the overview.
+:::
+
+## What does not change
+
+- **You still install and run KubeRay yourself.** The chart's README lists KubeRay as an optional cluster operator to install separately from its upstream chart, required only if your pipelines use Ray tasks. It is not a chart dependency — the local sandbox installs it as its own Helm release into a `ray-system` namespace:
+
+  ```bash
+  helm repo add kuberay https://ray-project.github.io/kuberay-helm
+  helm install kuberay-operator kuberay/kuberay-operator --namespace ray-system --create-namespace
+  ```
+- **The resources KubeRay reconciles are still `ray.io/v1`.** Michelangelo AI builds them using the upstream KubeRay Go types, so what lands in your compute cluster is an ordinary `RayCluster` or `RayJob`.
+- **Ray itself is unchanged.** Ray Data and Ray Train are used directly — `ray.data.Dataset` is a first-class type in the pipeline IO system, and the bundled Lightning trainer subclasses Ray Train's `TorchTrainer`.
+
+:::warning
+There is no documented minimum KubeRay version. The control plane currently compiles against the KubeRay operator's Go types at `v1.2.2`, while the local sandbox installs operator `v1.4.2`. Both are in the `ray.io/v1` API, so an existing install in that range is very likely fine, but no version floor is stated anywhere, so confirm against your own operator version before migrating anything you care about.
+:::
+
+## What changes
+
+You stop applying `ray.io/v1` manifests to a cluster directly. Instead you apply a Michelangelo AI resource to the control plane, and it decides which registered compute cluster the workload lands on.
+
+That indirection is the point: it is what enables federated dispatch across multiple compute clusters, and what lets a Ray cluster become one step inside a larger pipeline rather than a standalone object you manage by hand.
+
+## Concept mapping
+
+| Concept | KubeRay | Michelangelo AI |
+|---|---|---|
+| **Cluster resource** | `RayCluster` (`ray.io/v1`) | `RayCluster` (`michelangelo.api/v2`), translated into a `ray.io/v1` `RayCluster` |
+| **Job resource** | `RayJob` (`ray.io/v1`) | `RayJob` (`michelangelo.api/v2`), translated into a `ray.io/v1` `RayJob` |
+| **Head node** | `spec.headGroupSpec` | `spec.head` |
+| **Worker groups** | `spec.workerGroupSpecs[]` | `spec.workers[]` |
+| **Worker group name** | `groupName` | `nodeType`, though the group name on the generated resource is derived from the cluster name rather than from this field |
+| **Worker count** | `replicas`, `minReplicas`, `maxReplicas` | `minInstances`, `maxInstances` |
+| **Ray start params** | `rayStartParams` | `rayStartParams`, on both head and workers |
+| **Ray version** | `spec.rayVersion` | `spec.rayVersion` |
+| **Pod customization** | `template` (a `PodTemplateSpec`) | `pod` (a `PodTemplateSpec`), on both head and workers |
+| **Which cluster a job runs on** | `clusterSelector` or an inline `rayClusterSpec` | `spec.cluster`, a name and namespace reference |
+| **Applying a manifest** | `kubectl apply -f` | `ma ray_cluster apply --file` |
+| **Where the workload lands** | The cluster your `kubectl` context points at | A registered compute cluster, chosen by the control plane |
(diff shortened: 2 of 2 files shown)


apurvapatkeshwar:
@sallycr @paulzim a few more are sitting in `action_required` with no checks yet - #1781, #1782, #1783, #1787, #1788, #1793

paulzim (approved):
**Approve** — this is an unusually careful, deeply technical migration guide, and every checkable claim in it holds up: proto field names, CLI surface, hardcoded values, sidebar positions, internal links and anchors, roadmap citations. I flagged one line as a factual error on first pass and want to walk that back — it turns out to match a pre-existing repo convention, not something this PR invented, so it's not a fair reason to hold this up.

**On the "vLLM and SGLang for LLM serving" line (194):** I initially read this as wrong, since the actual `BackendType` enum (`proto/api/v2/inference_server.proto:71-81`) only defines `TRITON`/`LLM_D`/`DYNAMO`/`TORCHSERVE` — no `VLLM` or `SGLANG` value at all — and only `triton.go` exists as a concrete implementation in `go/components/inferenceserver/backends/`.

But that same "Triton (traditional) / vLLM+SGLang (LLM serving)" framing already exists, nearly verbatim, in `docs/getting-started/glossary.md:73` (merged 2026-08-03 in #1628: *"Michelangelo AI supports Triton Inference Server (for traditional models) and vLLM / SGLang (for LLM serving)"*), and vLLM shows up as an assumed-real backend across half a dozen other already-merged docs (`operator-guides/index.md`, `operator-guides/integrations/mlflow.md`, `operator-guides/serving/index.md`, `contributing/dev/go/key-concepts-and-terms.md`, `user-guides/examples/index.md`) and even a Go source comment (`backends/interface.go:22`). So #1783 is repeating established repo terminology, not fabricating a new claim — and rewriting just this one line would leave it inconsistent with the glossary page it directly cross-links to.

What this actually looks like is a **pre-existing, repo-wide gap between docs and the current OSS proto/Go state** — plausibly because vLLM/SGLang support is real on Michelangelo's internal side and hasn't been ported into this open-source fork's proto enum and backend registry yet (consistent with other internal-only-so-far features). I don't have visibility into anything outside this repo, so I can't confirm that either way — but either way, it's not this PR's bug to carry alone. Worth a separate follow-up issue to reconcile the `BackendType` enum and `backends/` implementations against what the docs collectively claim (add the missing enum values/backends, or walk the docs back to "Triton today, others planned") — happy to file that if useful, but it shouldn't block this guide.

**Everything else checked out, and there's a lot of it — this PR makes an unusual number of specific, falsifiable claims and I verified essentially all of them directly against source rather than trusting the prose:**

- **Concept mapping table** (`proto/api/v2/ray_cluster.proto`, `ray_job.proto`, `options.proto`): field names match exactly — `head`/`workers`/`min_instances`/`max_instances`/`ray_start_params`/`node_type`, `RayJobSpec.cluster` as a `ResourceIdentifier{namespace, name}`.
- **"Worker group name derived from cluster name, not `nodeType`"** — confirmed in `getWorkerGroupSpecs()`, `go/components/jobs/client/k8sengine/ray.go`: `GroupName: RayWorkerNodePrefix + clusterName`.
- **CLI surface is `get`/`apply`/`delete` only, no `run`** — confirmed via `python/michelangelo/cli/mactl/crd.py` (only `apply_func_impl`/`delete_func_impl`/`get_func_impl` exist) and cross-checked against `docs/user-guides/reference/cli.md`'s own resource table.
- **`RayTask` field list** (`python/michelangelo/uniflow/plugins/ray/task.py`) — every field named in the doc (`head_cpu`, `head_memory`, `head_disk`, `head_gpu`, `head_object_store_memory`, the worker equivalents, `worker_instances`, `breakpoint`, `runtime_env`) is present, nothing extra, nothing missing.
- **"Ray version hardcoded to 2.3.1" and "`worker_instances` sets min=max"** — confirmed directly in `task.star`: `"rayVersion": "2.3.1"` and `"minInstances"`/`"maxInstances"` both set to `worker_instances`.
- **"No autoscaling" / "no in-tree autoscaler"** — `enableInTreeAutoscaling` has zero hits anywhere in the repo.
- **"No `RayService` equivalent" (the absence itself)** — `rayservice`/`RayService` has zero hits in `go/` or `python/` outside generated proto stubs and tests.
- **"Bundled Lightning trainer subclasses Ray Train's `TorchTrainer`"** — confirmed: `LightningTrainer(TorchTrainer)` in `lightning_trainer.py`, importing `ray.train.torch.TorchTrainer`.
- **`ray.data.Dataset` as a first-class IO type** — confirmed, `io_registry()[Dataset] = RayDatasetIO`.
- **KubeRay chart install details** (namespace `ray-system`, optional/separate-install, sandbox pins operator `v1.4.2`) — confirmed against `helm/michelangelo/README.md` and `sandbox.py`'s actual `helm upgrade --install kuberay-operator ... --version 1.4.2 --namespace ray-system`.
- **Roadmap citations** ("Ray job launch", "Persistent Ray clusters via RayCluster CRD", "Federated multi-cluster dispatch" as shipped) — all three appear verbatim under `## Available Now` → `**Distributed Training**` in `docs/getting-started/roadmap.md`.
- **All 8 internal links resolve to real files**, and both anchor links resolve to real headings: `glossary.md:153` is `## Concept mapping`, `overview.md:50` is `## ML Workflow Mapping`.
- **Sidebar positioning** — `docs/user-guides/migration/_category_.json` claims position 5; positions 1-4 are already taken by `getting-started`/`ml-pipelines`/`train-and-deploy-models`/`reference`, so 5 doesn't collide or require renumbering anything.
- **CI is clean**: `validate-docs` passes, `license/cla` passes, the three advisory Go/proto/Helm-surface checks correctly skip (nothing in `go/`, `proto/`, or `helm/` changed). The one PR-thread comment is an unrelated bot-style status note about several PRs sitting in `action_required`, not feedback on this PR's content.

**One thing worth surfacing but not blocking:** the PR body itself flags a possible contradiction with `overview.md`'s Ray Tune row. That's already resolved — PR #1788 (merged 2026-08-12) updated `overview.md:59` to read "Uniflow tasks with hand-written sweeps," which matches this guide's "No Ray Tune integration" claim exactly. The author's self-flagged concern is stale; no action needed there.

**The two judgment calls the author flagged (single combined guide vs. issue #1705's "one per stack" ask, and the position-5 placement)** are reasonable as shipped and don't need to block merge — issue #1705 doesn't mandate three separate PRs, just three stacks' worth of coverage eventually, and position 5 doesn't conflict with anything. Your call whether to ask for the KubeRay-specific content to be split now or accept incremental guides later.

**Bottom line:** solid, well-verified guide. The one thing I'd have flagged as a bug turned out to be existing repo convention rather than a new error — worth a follow-up issue to reconcile docs and code on the inference-backend list, but not something to hold this PR on.

zhoward-1 (approved):

PR #1787, opened 2026-08-12
fix(docs,examples): correct Lightning trainer imports and parameters

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fixes the Lightning trainer snippet in two docs and one example -- `docs/user-guides/train-and-deploy-models/train-and-register-a-model.md`, `docs/user-guides/getting-started/prepare-your-data.md`, and `python/examples/gpt_oss_20b_finetune/simple_train.py`.

All three imported `create_scaling_config` and/or `create_run_config` from `michelangelo.lib.trainer.torch.pytorch_lightning.lightning_trainer`, which defines neither. `create_run_config` is real, just imported from the wrong place. `create_scaling_config` isn't real at all outside one unrelated local helper with a different signature in `california_housing_xgb/train.py`.

They also passed `LightningTrainerParam` kwargs that don't match its actual fields: `create_model` instead of `create_model_fn`, `model_kwargs` instead of `create_model_fn_kwargs`, `validation_data` instead of `val_data`.

`simple_train.py` had a second, independent bug: it read the trainer's return value as a Ray Train `Result` object (`result.get_best_checkpoint(...)`), but `LightningTrainer.train()` returns a plain dict, so that line would raise `AttributeError`.

**Why?**
Copying any of these three snippets as written raises `ImportError` before training even starts. `train-and-register-a-model.md` is the primary "how do I train with Ray" doc, so this is the first thing a lot of people will hit.

**How did you test it?**
Not run -- verified by reading source. `create_run_config`'s real location and signature: `python/michelangelo/uniflow/plugins/ray/run_config.py:27`. `LightningTrainerParam`'s real fields: `python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py:79-98`. `train()`'s real return shape: same file, `lightning_trainer.py:301-304`. Grepped for every other `validation_data=` call site in the repo to make sure I wasn't fixing an intentional API elsewhere -- the one hit left (`california_housing_xgb/preprocess.py:86`) is an unrelated dataclass field, not this one.

**Potential risks**
None to running code -- this only touches docs and one example script, no library code changed.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
Corrects the Lightning trainer example in two user guides and one example script to use real imports and real parameter names.

### DIFF
--- a/docs/user-guides/getting-started/prepare-your-data.md
+++ b/docs/user-guides/getting-started/prepare-your-data.md
@@ -137,10 +137,10 @@ def use_prepared_data(datasets: dict):
 
 ```py
 trainer_param = LightningTrainerParam(
-    create_model=create_model_function,
-    model_kwargs=model_config,
+    create_model_fn=create_model_function,
+    create_model_fn_kwargs=model_config,
     train_data=train_dv.value,
-    validation_data=val_dv.value,
+    val_data=val_dv.value,
     batch_size=32,
     num_epochs=10
 )
--- a/docs/user-guides/train-and-deploy-models/train-and-register-a-model.md
+++ b/docs/user-guides/train-and-deploy-models/train-and-register-a-model.md
@@ -86,10 +86,10 @@ To scale training across CPUs/GPUs, wrap your training task using **RayTask**.
 
 ```py
 from michelangelo.lib.trainer.torch.pytorch_lightning.lightning_trainer import (
-    LightningTrainer, LightningTrainerParam, create_run_config, create_scaling_config
+    LightningTrainer, LightningTrainerParam
 )
-from michelangelo.uniflow.plugins.ray import RayTask
-from ray.train import CheckpointConfig
+from michelangelo.uniflow.plugins.ray import RayTask, create_run_config
+from ray.train import CheckpointConfig, ScalingConfig
 
 @uniflow.task(
     config=RayTask(
@@ -119,9 +119,9 @@ def train_distributed_model(
     validation_data = validation_dv.value
 
     # Scaling config
-    scaling_config = create_scaling_config(
-        trainer_cpu=2,
-        cpu_per_worker=4,
+    scaling_config = ScalingConfig(
+        trainer_resources={"CPU": 2},
+        resources_per_worker={"CPU": 4},
         num_workers=num_workers,
         use_gpu=use_gpu,
     )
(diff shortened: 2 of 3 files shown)


paulzim (approved):
**Approve** — small, well-researched docs/example bugfix; every claimed root cause checks out against the actual source.

This is a "the tutorial code doesn't run" fix, so the whole point is whether the *before* was really broken and the *after* is really correct. Verified both independently rather than trusting the PR body's citations.

**Confirmed each of the four distinct bugs is real, not hypothetical:**
1. `create_run_config` was imported from `lightning_trainer.py`, which doesn't define it. Real location: `michelangelo/uniflow/plugins/ray/run_config.py:27`, and it *is* re-exported from `michelangelo.uniflow.plugins.ray`'s `__init__.py` — the PR's new import (`from michelangelo.uniflow.plugins.ray import RayTask, create_run_config`) is the correct, already-public path.
2. `create_scaling_config` doesn't exist anywhere importable. The only hit in the whole repo is a function *nested inside* `california_housing_xgb/train.py`'s local scope (not a module-level export, can't be imported from anywhere), with a different, keyword-only signature (`cpu_per_worker`, `trainer_cpu`) — confirmed by reading it directly. The PR's replacement, plain `ray.train.ScalingConfig(trainer_resources=..., resources_per_worker=..., num_workers=..., use_gpu=...)`, matches Ray's actual constructor signature (checked via `inspect.signature`).
3. `LightningTrainerParam`'s real field names, read straight from the dataclass: `create_model_fn`, `create_model_fn_kwargs`, `val_data` — exactly matching what the PR renames `create_model`/`model_kwargs`/`validation_data` to. Grepped the whole repo for leftover `model_kwargs=`/`create_model=`/`validation_data=` after the fix: zero remaining hits except one unrelated dataclass field in `california_housing_xgb/preprocess.py:86` that has nothing to do with `LightningTrainerParam` — matches the PR's own testing note exactly.
4. `simple_train.py`'s `result.get_best_checkpoint(...)` bug: confirmed `LightningTrainer.train()`'s actual return type is a plain `dict` (`{"checkpoint_path", "path", "metrics"}`), not a Ray `Result` — so `.get_best_checkpoint()` would indeed raise `AttributeError` on a dict. The fix (`result["checkpoint_path"]`) is correct.

**One pre-existing thing worth flagging but not blocking:** after this fix, `checkpoint = result["checkpoint_path"]` is always a plain string (per `LightningTrainer.train()`'s return line, `CHECKPOINT_PATH_KEY: result.checkpoint.path`), so the subsequent `if hasattr(checkpoint, "as_directory")` branch in `simple_train.py` is now dead code — it'll always take the `else` path. Harmless (still produces the right `checkpoint_path` value), just a residual left over from when the code assumed a `Checkpoint` object. Not something this PR needs to clean up, and out of scope for a "fix the broken imports" PR — just flagging in case you want a fast-follow.

**Sanity-checked the whole modified file, not just the diff hunks:** `python3 -m py_compile` and `ruff check` both pass clean on `simple_train.py` post-fix.

CI: only `license/cla` — same fork-PR `action_required` gate as this author's other queued PRs; will likely need the same "Approve and run workflows" click.

Good catch — this is the primary "how do I train with Ray" doc, so a copy-pasted `ImportError` here would be many new users' first bad experience.

zhoward-1 (approved):

Issue #1706, opened 2026-08-07
Xgboost trainer built on Ray Train


apurvapatkeshwar:
**Is your feature request related to a problem? Please describe.**
`python/pyproject.toml` already ships xgboost (2.1.4) and xgboost_ray (0.1.19) as optional dependencies. But the trainer extra is deep learning shaped (ray, torch, lightning, transformers) and there is no supported path for the gradient boosted tree workloads that dominate a lot of tabular ML

**Describe the solution you'd like**
A trainer built on ray train's first class xgboost support, reusing the existing ray provisioning and the already declared deps - distributed data parallel xgboost training, checkpointing to the existing artifact path, an xgboost extra, and an example pipeline, tested against a local ray session so CI needs no cluster

**Describe alternatives you've considered**
- A standalone XGBoost operator or job type (which is new infra that the ray path would make unnecessary)
- LightGBM first (xgboost deps are the ones already declared)

**Component**
Which part of the project does this affect?
- [x] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [ ] Helm chart / Kubernetes deployment
- [ ] UI (Studio)
- [ ] Documentation
- [ ] Other

**Additional context**
I see the trainer area is an active workstream right now (profiler series just landed via #1614, #1702 in flight), so I am checking in first before having a diff out that collides with in-flight refactors. Would you accept an xgboost trainer on ray train, and is there a trainer interface change in flight I should build against rather than current main?

PR #1793, opened 2026-08-12
feat(trainer): add an XGBoost trainer built on Ray Train

apurvapatkeshwar:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Adds `michelangelo.lib.trainer.xgboost` -- a distributed, data-parallel XGBoost trainer built on `ray.train.xgboost.XGBoostTrainer`.

It follows the shape the Lightning trainer already established, so the two read the same way: an `XGBoostTrainerParam` dataclass for configuration, an `XGBoostTrainer` subclassing Ray's trainer, and a `train()` returning `{"checkpoint_path", "path", "metrics"}`. Checkpointing goes through Ray Train's own `RayTrainReportCallback`, so it lands in the run's storage path with no new artifact handling.

Also adds a `trainer-xgboost` extra alongside the existing `trainer-comet` / `trainer-mlflow`, and moves the `xgboost` dependency declaration out of the example-only block, since it now backs library code rather than just an example.

**Why?**
Per #1706. The existing `trainer` extra is deep-learning shaped (ray, torch, lightning, transformers), so there's no supported path for gradient-boosted trees, which cover a lot of tabular ML. Ray Train already has first-class XGBoost support and the Ray provisioning already exists, so this is a wrapper over both rather than new infrastructure.

**How did you test it?**
Unit tests in `tests/test_xgboost_trainer.py` -- dataclass validation, dataset wiring with and without a validation set, the `train()` result contract including the error path, and the worker loop's feature-column inference and eval-set construction. Ray's parent `__init__` and `fit()` are patched throughout, so no cluster is needed. 24 tests, all passing on Python 3.11 / ray 2.56.1 / xgboost 2.1.4.

Also ran it end-to-end against a real local Ray cluster, since mocked tests can't tell you whether the thing actually trains. Five scenarios, all green:

- train + validation across 2 workers -- produces a real checkpoint (`model.ubj`) on disk and reports both `train-rmse` and `validation-rmse`. On synthetic data with a known linear signal, rmse converges to 0.40 against a ~3.6 baseline, so the model is genuinely fitting rather than just running.
- train-only with no validation set -- reports `train-rmse` only, confirming the validation shard is skipped rather than half-wired.
- an explicit `feature_columns` subset.
- single worker.
- a bad `label_column` -- surfaces as a `WorkerGroupError` whose message names the offending column, rather than failing obscurely inside XGBoost.

Note that CI installs only the `dev` extra, so I added `xgboost` there too -- otherwise the tests `importorskip` and never actually run. That does grow the dev install; happy to move them behind a separate extra and a marker instead if you'd rather keep `dev` lean.

**Potential risks**
Low -- this is additive. No existing module imports it, and nothing changes for current trainer users.

Three things worth a reviewer's eye:

**This overlaps #1751.** That PR removes `california_housing_xgb` and, with it, the `xgboost` / `xgboost_ray` dependency declarations. This PR needs `xgboost` to stay declared, so I moved it to its own line rather than leaving it in the example block. If #1751 lands first the two will want a small manual merge -- happy to rebase whichever way you'd prefer, and to sequence behind it.

**No example pipeline.** #1706 proposed one, but since examples are migrating to `michelangelo-examples`, adding one here seemed like it would just need moving again. Glad to write it in whichever repo you'd like.

**`xgboost_ray` is not used.** `ray.train.xgboost` doesn't need it -- the old example imported it only for build-time dependency discovery. This PR doesn't add it to the new extra.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
New optional `trainer-xgboost` extra providing `michelangelo.lib.trainer.xgboost` for distributed XGBoost training on Ray.

**Documentation Changes**
None yet. Happy to add a user guide page alongside the Lightning trainer docs if you'd like this documented before it ships.

### DIFF
--- a/python/michelangelo/lib/trainer/xgboost/__init__.py
+++ b/python/michelangelo/lib/trainer/xgboost/__init__.py
@@ -1,0 +1,20 @@
+"""XGBoost trainer wrapping Ray Train.
+
+Public surface re-exported below:
+
+* :class:`XGBoostTrainer` — Ray ``XGBoostTrainer`` subclass that runs a
+  data-parallel XGBoost training loop and reports checkpoints through Ray
+  Train.
+* :class:`XGBoostTrainerParam` — dataclass holding the training configuration
+  (label column, datasets, booster params, boosting rounds, ...).
+"""
+
+from michelangelo.lib.trainer.xgboost.xgboost_trainer import (
+    XGBoostTrainer,
+    XGBoostTrainerParam,
+)
+
+__all__ = [
+    "XGBoostTrainer",
+    "XGBoostTrainerParam",
+]
--- a/python/michelangelo/lib/trainer/xgboost/tests/__init__.py
+++ b/python/michelangelo/lib/trainer/xgboost/tests/__init__.py
@@ -1,0 +1,1 @@
+"""Tests for the XGBoost trainer library."""
--- a/python/michelangelo/lib/trainer/xgboost/tests/test_xgboost_trainer.py
+++ b/python/michelangelo/lib/trainer/xgboost/tests/test_xgboost_trainer.py
@@ -1,0 +1,340 @@
+"""Tests for the XGBoost trainer module.
+
+Cover the public surface of ``michelangelo.lib.trainer.xgboost.xgboost_trainer``:
+dataclass validation, trainer construction and dataset wiring, the ``train()``
+result contract, and the per-worker training loop. Ray Train's parent
+``__init__`` and ``fit()`` are patched throughout, so nothing here needs a Ray
+cluster.
+"""
+
+from __future__ import annotations
+
+import sys
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+pytest.importorskip("xgboost")
+pytest.importorskip("ray.train.xgboost")
+
+from michelangelo.lib.trainer.xgboost.xgboost_trainer import (  # noqa: E402
+    CHECKPOINT_PATH_KEY,
+    TRAIN_DATASET_KEY,
+    VALIDATION_DATASET_KEY,
+    XGBoostTrainer,
+    XGBoostTrainerParam,
+    _train_loop_per_worker,
+)
+
+_PARENT_INIT = "ray.train.xgboost.XGBoostTrainer.__init__"
+
+
+def _make_param(**overrides) -> XGBoostTrainerParam:
+    """Build a minimally-valid ``XGBoostTrainerParam`` for tests."""
+    defaults = {
+        "label_column": "target",
+        "train_data": MagicMock(name="train_data"),
+    }
+    defaults.update(overrides)
+    return XGBoostTrainerParam(**defaults)
+
+
+# -----------------------------------------------------------------------------
+# XGBoostTrainerParam
+# -----------------------------------------------------------------------------
+
+
+class TestXGBoostTrainerParam:
+    """``XGBoostTrainerParam`` dataclass behavior."""
+
+    def test_defaults(self):
+        param = _make_param()
+        assert param.val_data is None
+        assert param.params == {}
+        assert param.num_boost_round == 10
+        assert param.feature_columns is None
+        assert param.xgboost_train_kwargs == {}
+
+    def test_mutable_defaults_are_not_shared(self):
+        first = _make_param()
+        second = _make_param()
+        first.params["eta"] = 0.1
+        assert second.params == {}
+
+    def test_empty_label_column_rejected(self):
+        with pytest.raises(ValueError, match="label_column"):
+            _make_param(label_column="")
+
+    @pytest.mark.parametrize("rounds", [0, -1])
+    def test_non_positive_boost_rounds_rejected(self, rounds):
+        with pytest.raises(ValueError, match="num_boost_round"):
+            _make_param(num_boost_round=rounds)
+
+    def test_user_callbacks_rejected(self):
+        with pytest.raises(ValueError, match="callbacks"):
+            _make_param(xgboost_train_kwargs={"callbacks": [object()]})
+
+    def test_other_train_kwargs_allowed(self):
+        param = _make_param(xgboost_train_kwargs={"verbose_eval": False})
+        assert param.xgboost_train_kwargs == {"verbose_eval": False}
+
+
+# -----------------------------------------------------------------------------
+# XGBoostTrainer construction
+# -----------------------------------------------------------------------------
+
+
+class TestXGBoostTrainerInit:
+    """Argument wiring handed to Ray's ``XGBoostTrainer``."""
+
+    def test_train_only_dataset_wiring(self):
+        param = _make_param()
+        with patch(_PARENT_INIT, return_value=None) as parent:
+            XGBoostTrainer(param)
+
+        datasets = parent.call_args.kwargs["datasets"]
+        assert datasets == {TRAIN_DATASET_KEY: param.train_data}
+        assert parent.call_args.kwargs["train_loop_config"]["has_validation"] is False
+
+    def test_validation_dataset_wiring(self):
+        val = MagicMock(name="val_data")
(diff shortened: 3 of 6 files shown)


paulzim (approved):
**Approve** — clean, additive, well-tested new trainer; verified the code, the tests (ran them), and both dependency-declaration edges the author flagged.

**What it adds:** `michelangelo/lib/trainer/xgboost/` — an `XGBoostTrainer(RayXGBoostTrainer)` + `XGBoostTrainerParam` dataclass wrapping `ray.train.xgboost`, deliberately shaped to match the existing `lib/trainer/torch/pytorch_lightning` trainer (same `CHECKPOINT_PATH_KEY = "checkpoint_path"` constant, same `train()` → `{"checkpoint_path", "path", "metrics"}` result-dict contract, same "small param dataclass + thin train() wrapper" shape). Read both files in full to confirm the comparison is real, not just asserted in the docstring — it holds.

**Code read in full, not skimmed:**
- `XGBoostTrainerParam.__post_init__` rejects empty `label_column`, non-positive `num_boost_round`, and a user-supplied `callbacks` key (which would silently disable Ray's checkpoint-reporting callback) — all three are exercised by tests.
- `_train_loop_per_worker` defers `import xgboost` / `from ray.train.xgboost import RayTrainReportCallback` to worker-side, so importing the module doesn't require `xgboost` on the driver — consistent with the PR's CI-footprint argument.
- Validation-set handling is driven by an explicit `has_validation` flag computed at construction time rather than probing `get_dataset_shard()` for `None`/an exception — a sane, explicit choice the code comments call out.
- `XGBoostTrainer.__init__` builds `train_loop_config` field-by-field instead of `dataclasses.asdict(trainer_param)`, specifically to avoid deep-copying the Ray `Dataset` objects — correct concern, `asdict()` would recursively copy attributes and Ray Datasets aren't meant to be copied that way.

**Tests — ran them myself, didn't just read the file:**
```
poetry run pytest michelangelo/lib/trainer/xgboost/tests/test_xgboost_trainer.py -v
======================== 24 passed in 1.76s ========================
```
24/24 as claimed (7 dataclass-validation + 6 constructor/dataset-wiring + 5 `train()`-contract + 6 worker-loop tests, matching the PR's own breakdown). Confirmed `xgboost` and `ray.train.xgboost` really are installed in this env (not skipped via the `pytest.importorskip` guards at the top of the file) — all 24 genuinely exercised, not vacuously skipped. Ray's `__init__`/`fit()` are patched throughout as described; no cluster needed, consistent with the "CI only installs `dev`" constraint.

**Additive-only claim, verified:** grepped for any existing import of `michelangelo.lib.trainer.xgboost` outside the new package itself — none found. Nothing else in the tree references this code yet, so there's no behavior-change blast radius; worst case if something's wrong here is the new module itself.

**The two dependency-declaration items the author flagged — checked both, not just taken on faith:**
1. **Overlap with #1751** (also open/unmerged): confirmed real. #1751 deletes both `xgboost_ray` and `xgboost` lines outright (plus drops them from the `example` extra), on the assumption nothing consumes them anymore post-migration to `michelangelo-examples`. #1793 touches the *same two lines* — moves `xgboost` to its own annotated line, leaves `xgboost_ray` alone, and adds `xgboost` to `dev` + a new `trainer-xgboost` extra. Whichever merges second will need a rebase (git will flag it as a real conflict, not silently misresolve it — checked both merge orderings). Since #1793 makes `xgboost` a real, live dependency again (this trainer is exactly the "trainer interface" issue #1706 asked for), the sane sequencing is **#1793 first, then #1751 rebased to keep the `xgboost` declaration and only drop `xgboost_ray`** (which #1793's own PR body confirms is unneeded by `ray.train.xgboost`). Worth a quick sync between the two authors/reviewers before merging either, but not a reason to block this PR.
2. **`dev` extra growing to include `xgboost`:** confirmed accurate — CI's `dev` install is the only thing that exercises these tests, so without this the new tests would silently skip in CI via `importorskip`, not fail. Adding `xgboost` (already a declared, versioned dependency elsewhere) to `dev` is the correct call over letting a real test suite go unexercised in CI.

**Traced back to its source:** issue #1706 (open) asked for exactly this — a Ray-Train-based XGBoost trainer, checkpointing reuse, no-cluster-needed testing — and explicitly asked for an example pipeline. This PR's body explains the omission (examples are actively migrating to `michelangelo-examples`, offered to add one wherever preferred) — a reasonable, disclosed scope trim rather than a silent gap.

CI: only `license/cla` has run — same fork-PR `action_required` gate seen on this author's other queued PRs (#1678/#1679); no comments on this PR yet asking for a workflow-approval click, but you'll likely need to do the same "Approve and run workflows" step here too.

Solid, self-contained addition. My only ask is coordinating merge order with #1751 — not a change to this PR's own diff.

paulzim (approved):
**Update (2026-08-13): two follow-up commits added since the original approval, re-reviewed the delta only (`0bdfa70d..c8588572`).**

1. **`ce59bac5` — refresh `poetry.lock` for the `trainer-xgboost` extra.** Diffed it directly: only the `[extras]` table (adds `xgboost` to `dev`, adds the new `trainer-xgboost` entry) and the `content-hash` changed — zero package version pins moved. `pyproject.toml` itself has zero diff between the two commits, so this is exactly the lockfile-freshness fix CI would otherwise reject `poetry install` for. Confirmed `poetry check --lock` passes clean on the final head (deprecation warnings only, no lock-mismatch error).

2. **`c8588572` — satisfy ruff on the new files.** Three independent fixes, all verified:
   - **`TC004`** — moves `import ray.data` out of a `TYPE_CHECKING` block in `xgboost_trainer.py`, since `ray.data.Dataset` is used in two `XGBoostTrainerParam` field annotations outside type-checking. I chased down whether this is a real reachable bug or a lint-only concern, since the file has `from __future__ import annotations` (PEP 563 — annotations are stored as unevaluated strings). Conclusion: **it's a hygiene fix, not a live-bug fix, in this codebase as it stands today** — `XGBoostTrainerParam` is a plain `@dataclass` (not a `JSONData`/pydantic subclass, confirmed by comparison to the sibling `LightningTrainerParam`, also a plain dataclass), so nothing calls `typing.get_type_hints()` on it (that call exists in exactly one place repo-wide, `json_data.py`, and doesn't touch this class). The only `dataclasses.fields()` call sites in `uniflow/core/` only read `f.name`, never `f.type`. So under `from __future__ import annotations`, the string `"ray.data.Dataset"` is never actually resolved at runtime anywhere in this call path — but ruff is still right to flag it as a static-analysis correctness issue (a linter or a future `get_type_hints()` caller would need the real import), so the fix is correct regardless, just not fixing an active bug.
   - **`SIM117`/`RUF100` in the test file** — collapsed two nested `with patch(...): with pytest.raises(...):` blocks into single parenthesized multi-context-manager statements, and removed a stale `# noqa: E402` that no longer applies. Diffed in full: no assertions or test logic changed, purely mechanical.
   - **`D102`** — added one-line docstrings to all ~18 test methods, consistent "It <does X>" style matching the rest of the suite.

3. **Re-verified independently rather than trusting the commit message:** re-ran the full suite on the final head — `24 passed` (same count, no new/removed tests, just docstrings added). `ruff check` and `ruff format --check` both clean on `michelangelo/lib/trainer/xgboost/`.

4. **CI:** `Build (linux/amd64)` and `Build (linux/arm64)` both fail, but with `denied: installation not allowed to Write organization package` when pushing to `ghcr.io/michelangelo-ai/examples` — this is GitHub blocking a fork PR's default token from package-write access, not a build/code failure. Confirmed it's a repo-wide gate unrelated to this PR by checking #1751 (another external-contributor Python PR): identical failure, same job, same error. Everything else (bazel-build, python-build, ruff, coverage, license/cla) passes. No new PR comments since the original review.

Nothing here changes the verdict — still **Approve**. The two new commits are exactly what they claim to be: a lockfile fix and a lint cleanup, both verified correct and behaviorally inert.

PR #1797, opened 2026-08-13
fix(go): pipeline revision resolve and latest

mouyuanyap:
- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Two related fixes so Pipeline revisioning behaves correctly end-to-end: (1) PipelineRuns can pin a Revision into an inline `pipelineSpec` at create time, and (2) `Pipeline.status.latestRevision` only advances for default-branch commits.

### 1) PipelineRun create hook — resolve Revision → inline `pipelineSpec`
(`go/components/pipelinerun/apihook`)

On `BeforeCreate`, when a PipelineRun references a Pipeline:

1. **Explicit `spec.revision`** → load the Revision CR, unmarshal the snapshotted Pipeline, set `spec.pipelineSpec`, backfill `spec.pipeline` / annotations as needed. **Hard-fail** if the Revision is missing, not a Pipeline snapshot, empty, or conflicts with an existing inline `pipelineSpec`.
2. **No `spec.revision`** → if the live Pipeline has `status.latestRevision`, pin and resolve it the same way.
3. **Soft fallback to live Pipeline** when latestRevision is unset, the Pipeline Get soft-fails, or the Revision CR is not found (clear the tentative pin so SourcePipelineActor still runs against the live Pipeline).
4. Existing inline `spec.pipelineSpec` (dev-run) wins — do not override with latestRevision.
5. OwnerRef stamping + notification inheritance still run after resolution (and observe any backfilled `spec.pipeline`).

### 2) Pipeline controller — gate `status.latestRevision` to `main` / `master`
(`go/components/pipeline`)

When revisioning is enabled, every commit still upserts a Revision CR, but **`status.latestRevision` is only updated when `spec.commit.branch` is `main` or `master`**. Feature-branch applies no longer rewrite what “latest” means for create-hook consumers.

**Why?**

Without (1), omitting `spec.revision` always meant “use the mutable live Pipeline,” even when the platform had already published a stable latestRevision. Explicit revision also needed to be materialized into the same shape as a dev-run (`pipelineSpec`) so the execution path stays revision-agnostic.

Without (2), applying a Pipeline from a feature branch would advance `status.latestRevision` to that branch’s snapshot. Subsequent PipelineRuns that omit `spec.revision` would then pin an unintended (non-default) revision.

Together: default-branch publishes “latest”; create-time resolution prefers that pin when present, with a safe live fallback when it is not.

### Behaviour overview

```mermaid
flowchart TD
  A[CreatePipelineRun BeforeCreate] --> B{spec.revision set?}
  B -->|yes| C[resolveRevision]
  C -->|ok| D[Set spec.pipelineSpec<br/>backfill pipeline + annotations]
  C -->|error| E[Reject create]
  B -->|no| F{inline pipelineSpec already set?}
  F -->|yes| G[Skip pin — keep dev-run]
  F -->|no| H{Pipeline.status.latestRevision?}
  H -->|empty / no Pipeline| I[Live Pipeline path]
  H -->|present| J[Pin Spec.Revision = latestRevision]
  J --> K[resolveRevision]
  K -->|ok| D
  K -->|Revision not found| L[Clear pin → live Pipeline]
  K -->|other error| E
  D --> M[Notifications + ownerRef stamp]
  G --> M
  I --> M
  L --> M
```

```mermaid
flowchart LR
  subgraph pipeline_reconcile[Pipeline Reconcile — revisioningEnabled]
    P[Pipeline with Spec.Commit] --> U[Upsert Revision CR]
    U --> BR{branch == main or master?}
    BR -->|yes| LR[Set status.latestRevision]
    BR -->|no| SKIP[Do not advance latestRevision\nRevision CR still exists]
  end
```

```mermaid
sequenceDiagram
  participant Client
  participant APIHook as PipelineRun BeforeCreate
  participant Pipeline
  participant Revision
  participant SPA as SourcePipelineActor

  Client->>APIHook: CreatePipelineRun (pipeline ref, optional revision)
  APIHook->>Pipeline: Get live Pipeline
  alt explicit revision
    APIHook->>Revision: Get Revision
    Revision-->>APIHook: snapshotted Pipeline Any
    APIHook->>APIHook: pipelineSpec = snapshot.Spec
  else no revision, latestRevision present
    APIHook->>Revision: Get status.latestRevision
    alt found
      APIHook->>APIHook: pin + pipelineSpec
    else not found
      APIHook->>APIHook: fall back to live Pipeline
    end
  else no latestRevision
    APIHook->>APIHook: live Pipeline path
  end
  APIHook-->>Client: Create accepted
  Note over SPA: Execution uses inline pipelineSpec if present, otherwise the live Pipeline
```

**How did you test it?**

- Unit tests in `go/components/pipelinerun/apihook/apihook_test.go`:
  - explicit revision → inline `pipelineSpec`
  - missing explicit revision → create rejected
  - no revision + `status.latestRevision` → pinned
  - latestRevision CR missing → live fallback
  - existing inline `pipelineSpec` not overridden
  - non-Pipeline Revision kind rejected
- Unit tests in `go/components/pipeline/controller_test.go`:
  - `main` / `master` advance `status.latestRevision`
  - other branches upsert Revision but do not advance latestRevision
- Local: `go test ./components/pipelinerun/apihook/ ./components/pipeline/ -count=1`

**Potential risks**

- **Behaviour change for create without `spec.revision`:** when `status.latestRevision` is set and the Revision CR exists, runs now pin that snapshot instead of always using the live Pipeline. Intended; operators who relied on “omit revision = always tip of live Pipeline” should either clear `latestRevision` or pass an explicit revision / inline `pipelineSpec`.
- **Behaviour change for non-default-branch Pipeline applies:** feature-branch reconciles no longer move `status.latestRevision`. Existing latestRevision is left as-is until a main/master apply updates it.
- Soft Get failures on the live Pipeline still skip ownerRef/notifications (pre-existing soft path, preserved).

**Breaking Changes**

- [x] No breaking changes

> Behavioural semantics change for revision-aware creates / latestRevision publishing, but no API/proto/helm/config contract break. Callers that omit `spec.revision` intentionally for “always live” should be aware of the latestRevision preference above.

**Release notes**

When revisioning is enabled:
- PipelineRuns without `spec.revision` prefer `Pipeline.status.latestRevision` (resolved to inline `pipelineSpec`) when available, with live-Pipeline fallback if the Revision CR is missing.
- Explicit `spec.revision` is required to resolve successfully (no silent live fallback).
- Only `main`/`master` Pipeline commits advance `status.latestRevision`; other branches still create Revision CRs.

**Documentation Changes**

N/A for this PR. Follow-up welcome: wiki / user docs on “PipelineRun revision resolution order” and “when latestRevision advances.”

### DIFF
--- a/go/components/pipelinerun/apihook/apihook.go
+++ b/go/components/pipelinerun/apihook/apihook.go
@@ -2,16 +2,27 @@ package apihook
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	"go.uber.org/zap"
+
+	pbtypes "github.com/gogo/protobuf/types"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/michelangelo-ai/michelangelo/go/api"
 	"github.com/michelangelo-ai/michelangelo/go/api/utils"
 	"github.com/michelangelo-ai/michelangelo/go/cascadedelete"
+	apipb "github.com/michelangelo-ai/michelangelo/proto-go/api"
 	v2 "github.com/michelangelo-ai/michelangelo/proto-go/api/v2"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 )
+
+// pipelineKind is the Revision.Spec.BaseType.Kind value for Pipeline snapshots.
+// Revisions also snapshot other resource kinds (models, deployments), which are
+// not valid targets for a PipelineRun.
+const pipelineKind = "Pipeline"
 
 // RegisterPipelineRunAPIHook registers the API hook that stamps the owning
 // Pipeline as the controller ownerReference on PipelineRuns at creation, so a run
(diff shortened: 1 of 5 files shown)


go/components/pipelinerun/apihook/apihook.go
```diff
@@ -34,24 +45,231 @@ type apiHook struct {
...
 	}
-	namespace := pipelineRef.GetNamespace()
-	if namespace == "" {
-		namespace = request.PipelineRun.GetNamespace()
+
+	if len(request.PipelineRun.Spec.Notifications) == 0 && len(pipeline.Spec.Notifications) > 0 {
+		request.PipelineRun.Spec.Notifications = pipeline.Spec.Notifications
+		a.logger.Info("BeforeCreate: copied notifications from Pipeline to PipelineRun",
+			zap.Int("count", len(pipeline.Spec.Notifications)))
 	}
```
craigmarker, line 87: I'm not sure if this is the right user experience—consider reverting the notification copying for further discussion.

For context, I'm concerned that inserting notification setting without user expressly indicating they want it could lead to noisiness that users would not want.

go/components/pipelinerun/apihook/apihook.go
```diff
@@ -34,24 +45,231 @@ type apiHook struct {
...
+func resourceNamespace(ref *apipb.ResourceIdentifier, fallback string) string {
+	if ns := ref.GetNamespace(); ns != "" {
+		return ns
+	}
+	return fallback
+}
+
+// isNotFoundError reports whether err (or any wrapped cause) is a not-found
+// error. resolveRevision wraps apiHandler.Get failures with fmt.Errorf %w.
+func isNotFoundError(err error) bool {
```
craigmarker: nit: it's unfortunate to need a copy/wrapper of utils.isNotFoundError. I wonder if there is a way to rely only on utils.isNotFoundError?

craigmarker (commented):
i think using pipeline_run.spec.pipeline_spec for revision usecase is interesting. Technically, runnable pipeline revisions (in ready state, with image built) should be immutable.

CC @badcount @npuppala111 can you take a look at this?

go/components/pipelinerun/apihook/apihook.go
mouyuanyap: Hi @badcount, you have added this particular logic in Grab's MA repo. Should this be synced back to OSS MA? 

MR [fix: add slack-go and golang-jwt dependencies for notification system](https://gitlab.myteksi.net/grab/entities/services/aeroflow/michelangelo/-/merge_requests/77#top) on 14th July 2026

go/components/pipelinerun/apihook/apihook.go
mouyuanyap: Good catch — removed it. The wrapper only existed because resolveRevision was wrapping the Get error with %w, which defeats utils.IsNotFoundError (it inspects the error you hand it via status.FromError and doesn't walk the chain).

Turned out the wrapping wasn't earning anything either: apiHandler.Get routes every error through surfaceGrpcError, so the message already carries failed to get API object. namespace: <ns>, name: <name> — my resolve revision %s/%s: prefix was just restating it. So the Get error is now returned unwrapped and the caller calls utils.IsNotFoundError on it directly. No sentinel, no local helper.

go/components/pipelinerun/apihook/apihook.go
```diff
@@ -34,24 +44,223 @@ type apiHook struct {
...
+	pipeline := &v2.Pipeline{}
+	if err := pbtypes.UnmarshalAny(rev.Spec.Content, pipeline); err != nil {
+		return fmt.Errorf("unmarshal snapshotted pipeline from revision %s/%s: %w", namespace, revisionRef.GetName(), err)
+	}
+
+	request.PipelineRun.Spec.PipelineSpec = &pipeline.Spec
+
+	if request.PipelineRun.Spec.GetPipeline().GetName() == "" {
+		request.PipelineRun.Spec.Pipeline = rev.Spec.GetBaseResource()
+	}
```
zhouxianzx, line 252: Should we validate that an existing spec.pipeline matches rev.spec.baseResource? Currently a request can reference Pipeline A with a Revision belonging to Pipeline B. We would execute B’s snapshotted pipelineSpec while retaining A for owner references and notification inheritance. I think we should reject the request when the names or namespaces differ.

go/components/pipelinerun/apihook/apihook.go
badcount: This change is import back from Grab

badcount (approved):

go/components/pipelinerun/apihook/apihook.go
mouyuanyap: Good catch! 
Empty spec.pipeline still backfills from baseResource. 
When spec.pipeline is set, we now reject if name or namespace differs from revision.spec.baseResource, before materializing pipelineSpec.

PR #1798, opened 2026-08-13
feat(form): strip stale values for `is`-hidden condition fields on submit

craigmarker:
## Summary
FormCondition only ever controlled rendering — a value entered while an
`is` condition was true stayed in form state (and got submitted) even
after the user changed the triggering field so the condition no longer
holds. filterHiddenConditionFields strips those fields from the payload
at submit time via ConfigDrivenForm, using final-form's own getIn/setIn
for an immutable strip.

Scoped to the `is` operator only for this first slice; isNot/isEmpty/
containsAny-triggered conditions are left untouched for now.


craigmarker:
Superseded — this content has been folded into #1776's stack as a new base branch (craig.marker/cdf-condition-is-only) per updated PR-splitting plan.

PR #1801, opened 2026-08-13
fix: plumb RayTask gpu, disk and object store memory into the cluster spec

### LINKED ISSUE #1794: RayTask accepts GPU, disk and object store settings that never reach the pod
**Describe the bug**
`RayTask` accepts `head_gpu`, `worker_gpu`, `head_disk`, `worker_disk`, `head_object_store_memory`, and `worker_object_store_memory`, but none of them reach the generated pod spec. A task that asks for GPUs gets a CPU only cluster, with no error and nothing in the logs

**To Reproduce**
1. Declare a task with a GPU request:

```py
@uniflow.task(config=RayTask(head_cpu=2, head_memory="4Gi", worker_cpu=2, worker_memory="4Gi", worker_gpu=1, worker_instances=2))
def train(): ...
```

2. Run it and inspect the generated `RayCluster`
3. The worker container's `resources.requests` contains only `cpu` and `memory`

**Expected behavior**
Either the requested gpu/disk/object-store resources land on the pod, or the task fails fast with a clear message saying they aren't supported on this path

**Actual behavior**
They're silently dropped. `plugins/ray/task.star:139-157` reads all four of `_head_disk`, `_head_gpu`, `_worker_disk`, `_worker_gpu` from env overrides and casts the GPU values to `int`, and then the call at `task.star:170-177` passes only `cpu` and `memory`:

```py
head_resource = resource_dict(
    cpu = _head_cpu,
    memory = _head_memory,
),
```

The helper supports more than that - `commons.star:65` is `def resource_dict(cpu, memory, disk = None, gpu = None, gpu_sku = "")`.

The Spark plugin calls the same helper and does pass them (`plugins/spark/task.star:112-124`):

```py
driver_resource = resource_dict(
    cpu = _driver_cpu,
    memory = _driver_memory,
    disk = _driver_disk,
    gpu = _driver_gpu,
),
```

so this reads like an omission on the ray path rather than a deliberate limitation

Two other things point the same way. `task.star:484` computes `support_gpu = head_resource.get("gpu", 0) + worker_resource.get("gpu", 0) * worker_instances > 0`, which can never be true, because `"gpu"` is never a key in either dict -- and the variable is never read anywhere. And `head_object_store_memory`/`worker_object_store_memory` are accepted as parameters (`task.star:94`, `task.star:99`) and threaded through the config dict, but never referenced in the task body at all

**One caveat on the fix**

Adding `gpu =`/`disk =` to that call isn't sufficient on its own, and I think that's the part worth a maintainer's judgement

`resource_dict` emits the keys `"gpu"` and `"diskSize"`. On the Ray path, that dict is spliced straight into a Kubernetes pod (`task.star:521-522`):

```py
"resources": {
    "requests": head_resource,
},
```

and the go side consumes it as a literal `PodTemplateSpec` with no renaming (`getHeadGroupSpec`, `go/components/jobs/client/k8sengine/ray.go:160-165`). So the keys have to already be valid Kubernetes resource names -- `nvidia.com/gpu` (there's a constant at `go/components/jobs/common/constants/constants.go:142`), `ephemeral-storage`. `"gpu"` and `"diskSize"` are neither

Spark gets away with the same helper because its path goes through a `ResourceSpec` proto and `utils.ConvertToResourceList` (`go/components/jobs/scheduler/framework/job.go:305`), which does the renaming. Ray has no equivalent step.

So a correct fix needs a few calls that aren't mine to make:
- which GPU resource name to emit, and whether `gpu_sku` is meant to select it (the dead `support_gpu` hook suggests vendor-specific naming was intended at some point)
- whether `head_disk`/`worker_disk` should become an `ephemeral-storage` request, or resize the existing `/tmp/ray` hostPath volume
- whether object-store memory belongs in `rayStartParams` as Ray's native `--object-store-memory` rather than as a Kubernetes resource at all

Happy to send a PR once there's a preferred direction

**Environment**
- Applies to `main` as of `f86b91c3`; found by reading source, not by running a cluster.

**Component**
Which part of the project is affected?
- [x] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [ ] UI (Studio)
- [ ] Documentation
- [ ] Other

**Additional context**
Same shape as #1604 / #1605 -- a caller-supplied value accepted by `RayTask` and then dropped before it reaches the cluster spec, in the same file.

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Fixes #1794. `RayTask` accepts `head_gpu`/`worker_gpu`, `head_disk`/`worker_disk` and `head_object_store_memory`/`worker_object_store_memory` (plus `RAY_OVERRIDE_*` env overrides for gpu and disk), but none of them ever reached the cluster: `ray_cluster_spec` was only handed cpu and memory, built via commons.star `resource_dict`, whose proto-JSON key names (`gpu`, `diskSize`) aren't real Kubernetes resource names anyway.

The fix, all in `python/michelangelo/uniflow/plugins/ray/task.star`:

- new `container_resources` helper builds a real `ResourceRequirements` dict with real k8s resource names, because unlike the Spark path (typed `ResourceSpec` through `utils.ConvertToResourceList`, `go/components/jobs/client/k8sengine/helper.go:217-255`) the Ray pod spec is spliced into a literal `PodTemplateSpec` with no renaming step (`go/components/jobs/client/k8sengine/ray.go:159-165`)
- gpu goes to `nvidia.com/gpu` in requests **and** limits (extended resources require limit == request), matching `constants.ResourceNvidiaGPU` used by the Spark path
- disk goes to `ephemeral-storage` in requests
- object store memory goes to `rayStartParams["object-store-memory"]`, which KubeRay passes verbatim to `ray start` (`ray.go:162,175`)
- removes the dead `support_gpu` computation (assigned, never read) and the now-unused `resource_dict` import

One deliberate choice: disk is only forwarded when it differs from `RAY_DEFAULT_HEAD_DISK`/`RAY_DEFAULT_WORKER_DISK` (512Gi). That default has never reached the pod spec, so honoring it unconditionally would suddenly attach a 512Gi ephemeral-storage request to every existing RayTask and likely make many unschedulable. Explicit parameters and env overrides are always forwarded. The side effect is that explicitly passing exactly "512Gi" is treated as the default and not forwarded; happy to change the mechanism if you'd rather distinguish "unset" with a `None` default.

`gpu_sku` and `zone` remain unwired, as the issue notes: they need a scheduling design decision (nodeSelector vs tolerations vs something at the Job Controller layer), which seems like a maintainer call rather than something to guess at in this PR. Same for the workers' hardcoded `objectStoreMemoryRatio: 0.0`, which is untouched.

**Why?**
Today a user who sets `head_gpu=1` on a RayTask gets a pod with no GPU at all and no error anywhere -- the parameter is accepted, cast to int, and dropped. Same for disk and object store memory. Silent parameter drops are worse than unsupported parameters because the workload runs and produces slow or OOM-killed results instead of failing loudly.

**How did you test it?**
- extended `ray_cluster_spec_test.py` from 2 to 9 tests: gpu request+limit shape, gpu=0 emitting nothing, disk mapping to `ephemeral-storage`, object store memory landing in both head and worker `rayStartParams` as a string, absence of the flag when unset, and the full `ResourceRequirements` dicts passing through to the containers verbatim; all 9 pass (python 3.11, and the file compiles under 3.9)
- the test loads the real functions out of task.star by ast, same harness as before, so it exercises the shipped Starlark, not a copy
- `go test ./...` in `go/worker/plugins/ray` passes (its testdata uses its own test.star, so this confirms no interface drift on the Go plugin side)
- grepped the repo for other `ray_cluster_spec`/`head_resource`/`worker_resource` callers: none outside this file and its test
- ruff format + ruff check clean on the changed test file

**Potential risks**
For any RayTask that never set gpu/disk/object-store-memory (which is all of them that worked correctly so far), the generated spec is byte-identical to before except that nothing changes at all -- requests still contain exactly cpu and memory. The new behavior only activates when a caller explicitly asked for the resources that were previously dropped. Two adjacent open PRs touch the same file: #1698 (draft PoC around resource value sources; its own discussion notes this forwarding bug but the draft doesn't fix it) and #896 (adds a parameter on adjacent lines) -- either may need a trivial rebase depending on merge order.

**Breaking Changes**

- [x] No breaking changes

`ray_cluster_spec`'s keyword names changed (`head_resource` -> `head_resources` etc.), but it has no callers outside task.star itself and is not exported through any Python module.

**Release notes**
RayTask gpu, disk and object store memory settings now reach the Ray cluster pods instead of being silently dropped.

**Documentation Changes**
None -- the existing parameter docstrings in task.star already describe this as the behavior; now it's true.

### DIFF
--- a/python/michelangelo/uniflow/plugins/ray/task.star
+++ b/python/michelangelo/uniflow/plugins/ray/task.star
@@ -1,5 +1,5 @@
 load("@plugin", "atexit", "json", "os", "ray", "time")
-load("../../commons.star", "CACHE_OPERATION_GET", "CACHE_OPERATION_PUT", "DEFAULT_RETRY_ATTEMPTS", "TASK_STATE_FAILED", "TASK_STATE_KILLED", "TASK_STATE_PENDING", "TASK_STATE_RUNNING", "TASK_STATE_SKIPPED", "TASK_STATE_SUCCEEDED", "TIME_FOMART", "create_cached_output", "get_cache_enabled", "get_cache_keys", "get_cached_output", "get_pythonpath", "get_result_url", "get_task_image", "get_task_name", "io_read_json", "process_terminated_job", "report_progress", "resource_dict", COMMONS_ENV = "ENV")
+load("../../commons.star", "CACHE_OPERATION_GET", "CACHE_OPERATION_PUT", "DEFAULT_RETRY_ATTEMPTS", "TASK_STATE_FAILED", "TASK_STATE_KILLED", "TASK_STATE_PENDING", "TASK_STATE_RUNNING", "TASK_STATE_SKIPPED", "TASK_STATE_SUCCEEDED", "TIME_FOMART", "create_cached_output", "get_cache_enabled", "get_cache_keys", "get_cached_output", "get_pythonpath", "get_result_url", "get_task_image", "get_task_name", "io_read_json", "process_terminated_job", "report_progress", COMMONS_ENV = "ENV")
 
 DEFAULT_CREATE_CLUSTER_TIMEOUT_SECONDS = 60 * 30  # Timeout duration for cluster creation in seconds.
 RAY_ENV = {
@@ -164,18 +164,31 @@ def task(
         cluster_image = get_task_image(task_name)
         print("ray | create cluster:", "ns:", cluster_namespace, "image:", cluster_image, "task_name:", task_name)
 
+        # Forward disk only when it differs from the shipped default. The
+        # default (512Gi) has never reached the pod spec, so honoring it now
+        # would suddenly attach a large ephemeral-storage request to every
+        # existing RayTask; an explicit parameter or env override still wins.
+        _head_disk_request = _head_disk if _head_disk != RAY_DEFAULT_HEAD_DISK else None
+        _worker_disk_request = _worker_disk if _worker_disk != RAY_DEFAULT_WORKER_DISK else None
+
         cluster = ray_cluster_spec(
             namespace = cluster_namespace,
             image = cluster_image,
-            head_resource = resource_dict(
+            head_resources = container_resources(
                 cpu = _head_cpu,
                 memory = _head_memory,
+                disk = _head_disk_request,
+                gpu = _head_gpu,
             ),
-            worker_resource = resource_dict(
+            worker_resources = container_resources(
                 cpu = _worker_cpu,
                 memory = _worker_memory,
+                disk = _worker_disk_request,
+                gpu = _worker_gpu,
             ),
             worker_instances = _worker_instances,
+            head_object_store_memory = head_object_store_memory,
+            worker_object_store_memory = worker_object_store_memory,
             debug_enabled = breakpoint,
             runtime_env = runtime_env,
         )
@@ -427,6 +440,29 @@ def ray_job_entrypoint(task_path, result_url, args = None, kwargs = None):
 
     return "python3 -m michelangelo.uniflow.core.run_task --task '" + task_path + "' --args '" + args + "' --kwargs '" + kwargs + "' --result-url '" + result_url + "'"
 
+# Builds the k8s.io.api.core.v1.ResourceRequirements dict for a Ray container.
+#
+# Unlike the Spark path, which sends a typed ResourceSpec through
+# utils.ConvertToResourceList on the Go side, the Ray pod spec is consumed as a
+# literal PodTemplateSpec with no renaming step. The keys here must therefore
+# be real Kubernetes resource names ("ephemeral-storage", "nvidia.com/gpu"),
+# not the proto JSON names produced by resource_dict in commons.star
+# ("diskSize", "gpu"), which the scheduler would silently ignore.
+def container_resources(cpu, memory, disk = None, gpu = None):
+    requests = {
+        "cpu": cpu,
+        "memory": memory,
+    }
+    if disk:
+        requests["ephemeral-storage"] = disk
+    resources = {"requests": requests}
+    if gpu:
+        # nvidia.com/gpu is an extended resource: Kubernetes requires the
+        # limit to be present and equal to the request, so it is set in both.
+        requests["nvidia.com/gpu"] = gpu
+        resources["limits"] = {"nvidia.com/gpu": gpu}
+    return resources
+
 # Constructs a Unified API resource for provisioning a Ray Cluster.
 # This function generates a RayJob Custom Resource Definition (CRD) that defines the specifications for a Ray cluster.
 # Refer to the RayJob CRD: https://github.com/michelangelo-ai/michelangelo/blob/main/proto/api/v2/ray_job.proto
@@ -440,17 +476,24 @@ def ray_job_entrypoint(task_path, result_url, args = None, kwargs = None):
 #         - The Docker image containing Ray, application code, and dependencies.
 #         - Example: "127.0.0.1:5055/uber-usi/uber-one-michelangelo-sandbox:bkt1-produ-1719018451-45448"
 #
-#     head_resource (dict):
-#         - Resource configuration for the Ray **head node**.
-#         - Reference: `resource_dict` function in commons.star.
+#     head_resources (dict):
+#         - Kubernetes ResourceRequirements for the Ray **head node** container.
+#         - Reference: `container_resources` in this file.
 #
-#     worker_resource (dict):
-#         - Resource configuration for the Ray **worker nodes**.
-#         - Reference: `resource_dict` function in commons.star.
+#     worker_resources (dict):
+#         - Kubernetes ResourceRequirements for the Ray **worker node** containers.
+#         - Reference: `container_resources` in this file.
 #
 #     worker_instances (int):
 #         - Number of Ray worker instances to launch.
 #         - Must be a non-negative integer.
+#
+#     head_object_store_memory (int, optional):
+#         - Object store memory for the head node, in bytes, passed verbatim to
+#           `ray start --object-store-memory` via rayStartParams.
+#
+#     worker_object_store_memory (int, optional):
+#         - Object store memory per worker node, in bytes, same mechanism.
 #
 #     debug_enabled (bool, optional):
 #         - Enables debugging tools if set to True.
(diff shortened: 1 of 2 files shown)


paulzim (approved):
## Review: #1801 — fix: plumb RayTask gpu, disk and object store memory into the cluster spec

**Verdict: Approve**

Fixes #1794: `RayTask`'s `head_gpu`/`worker_gpu`, `head_disk`/`worker_disk`, and `head_object_store_memory`/`worker_object_store_memory` were accepted, cast, and then silently dropped — the generated pod only ever got `cpu`/`memory`. Scoped tightly to one file (`python/michelangelo/uniflow/plugins/ray/task.star`) plus its test, across two commits (the fix, then a follow-up adding `TestTaskResourcePlumbing` to satisfy the 90% diff-coverage gate).

**Verified the root-cause claim against the actual Go code, not just the PR's description of it:**
- `getHeadGroupSpec`/`getWorkerGroupSpecs` in `go/components/jobs/client/k8sengine/ray.go:159-165` do `Template: k8sptr.Deref(head.GetPod(), corev1.PodTemplateSpec{})` — the Ray pod spec is deserialized straight into a `corev1.PodTemplateSpec`, so `resources.requests`/`resources.limits` keys must already be real Kubernetes resource names. Confirmed `constants.ResourceNvidiaGPU = "nvidia.com/gpu"` (`go/components/jobs/common/constants/constants.go:142`), matching what the new `container_resources` emits.
- Confirmed `resource_dict` in `commons.star` really does produce `diskSize`/`gpu` (proto-JSON names for the Spark `ResourceSpec` path), which the k8s scheduler would silently ignore on a literal pod spec — so the bug and fix both check out mechanically, not just by description.
- Confirmed `resource_dict` is still used (and still needed) by `plugins/spark/task.star`, so dropping its now-unused import from the Ray file doesn't touch dead code elsewhere.
- Grepped the whole repo for `ray_cluster_spec(`, `container_resources(`, `head_resource`/`worker_resource`: no callers outside this file, matching the PR's claim.

**Ran the tests myself, including a control test:**
- `pytest .../ray_cluster_spec_test.py`: 13/13 pass (python 3.11 here; description also notes 3.9 compiles).
- Control test — reverted `container_resources` to drop gpu/disk handling (simulating the pre-fix bug) and reran: 4 tests fail exactly where expected (`test_disk_maps_to_ephemeral_storage`, `test_gpu_sets_request_and_limit`, `test_env_override_gpu_reaches_the_pod`, `test_explicit_resources_reach_the_pods`), confirming the suite actually exercises the regression surface rather than passing vacuously. Restored the fix afterward; all 13 pass again.
- `ruff check`/`ruff format --check` on the test file: clean.
- `go test ./worker/plugins/ray/...`: passes, confirming no interface drift on the Go plugin side.
- Pulled CI on the live PR: all checks green (`Check Coverage (Python 3.9)`, ruff, bazel-build, python-build, dirty-check all pass).
- Confirmed issue #1794's repro and root-cause citation (`task.star:139-157`/`170-177` on old `main`) match what's actually in the pre-fix file.

**Design choices hold up:**
- gpu uses Starlark truthiness (`if gpu:`) so `gpu=0` (the resolved default) correctly emits nothing — verified by `test_gpu_zero_adds_nothing` and the control test.
- The GPU request/limit pairing (`nvidia.com/gpu` in both, since it's an extended resource) is correct and matches the Spark path's `constants.ResourceNvidiaGPU` usage.
- The disk-default-suppression heuristic (only forward `head_disk`/`worker_disk` when they differ from `RAY_DEFAULT_HEAD_DISK`/`RAY_DEFAULT_WORKER_DISK`) is a reasonable, explicitly-disclosed trade-off to avoid silently attaching a 512Gi `ephemeral-storage` request to every existing RayTask. The PR is upfront about the one edge case this creates (explicitly passing exactly "512Gi" is indistinguishable from "unset").
- `gpu_sku`/`zone` remaining unwired, and the workers' hardcoded `objectStoreMemoryRatio: 0.0` being untouched, are both correctly identified as out-of-scope scheduling-design decisions rather than something to guess at here.
- Checked the two disclosed adjacent PRs for rebase risk: #1698 (draft PoC) and #896 both do in fact touch `commons.star`/`plugins/ray/task.star`, so the disclosed "trivial rebase" risk is accurate, not hand-waved.

**One thing not mentioned in the write-up, non-blocking:** unlike `head_gpu`/`head_disk` (which get `RAY_OVERRIDE_HEAD_GPU.<task_path>`/`RAY_OVERRIDE_HEAD_DISK.<task_path>` env-var overrides, pre-existing and now finally forwarded), `head_object_store_memory`/`worker_object_store_memory` have never had an env-override mechanism and still don't after this PR — only the direct parameter is honored. That's consistent with existing behavior and not a regression, but since the PR is careful to call out the other two known gaps (`gpu_sku`/`zone`), this one could've used the same one-line callout for anyone who goes looking for a `RAY_OVERRIDE_*_OBJECT_STORE_MEMORY` knob later.

No other concerns. Precise root-cause diagnosis, a fix that's minimal and traceable back to real Kubernetes/Go semantics, and a test suite that's verifiably not just for show.

python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -164,18 +164,31 @@ def task(
         cluster_image = get_task_image(task_name)
         print("ray | create cluster:", "ns:", cluster_namespace, "image:", cluster_image, "task_name:", task_name)
 
+        # Forward disk only when it differs from the shipped default. The
+        # default (512Gi) has never reached the pod spec, so honoring it now
+        # would suddenly attach a large ephemeral-storage request to every
+        # existing RayTask; an explicit parameter or env override still wins.
+        _head_disk_request = _head_disk if _head_disk != RAY_DEFAULT_HEAD_DISK else None
```
KyleKoon, line 171: Users should be able to request 512Gi.

KyleKoon, line 171: Users should be able to request 512Gi.

apurvapatkeshwar: Fixed in #1927 - the default is now empty so any explicit value forwards, 512Gi included

PR #1802, opened 2026-08-13
ci: build examples image without pushing on fork PRs

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
`.github/workflows/build-examples-image.yaml` now skips the GHCR push (and the manifest merge job) for pull requests coming from forks, while still building the image to validate the Dockerfile and dependency changes.

- the build job gets a `PUSH_OK` env flag: true for pushes to main, workflow_dispatch, and same-repo PRs; false for fork PRs
- `Log in to GHCR`, `Build and push by digest`, `Export digest` and `Upload digest artifact` are gated on `PUSH_OK == 'true'`
- the build step always runs, with `push=${{ env.PUSH_OK }}` templated into its `outputs:` string, so fork PRs build the image without pushing (single step, per review feedback)
- the `merge` job gets the equivalent job-level `if:` (the env context isn't available at job level, so the expression is repeated there)

The push path for same-repo refs is byte-identical to before.

**Why?**
Fork PRs run with a read-only `GITHUB_TOKEN` that cannot write org packages, so any fork PR touching `python/examples/**`, `pyproject.toml` or `poetry.lock` currently fails both `Build (linux/amd64)` and `Build (linux/arm64)` at the GHCR push. The failure is literally `denied: installation not allowed to Write organization package` after a successful image build -- see #1787 and #1793 for live examples. The red X looks like broken code but is pure permissions, which costs maintainers a triage round-trip on every affected external PR.

**How did you test it?**
- `actionlint` on the changed workflow: no new findings (the only two reported are pre-existing SC2046 shellcheck warnings on the untouched `imagetools create` step)
- verified the gate expression against the documented event payloads: `github.event.pull_request.head.repo.full_name == github.repository` is the standard same-repo test, and short-circuits to true for non-PR events via the `github.event_name != 'pull_request'` clause
- verified from primary sources that `push-by-digest=true` with `push=false` is well-defined: buildkit's image exporter only consults push-by-digest inside the push path (moby/buildkit exporter/containerimage/export.go), so with push=false it is inert; docker/build-push-action#1009 confirms `outputs.digest` is simply unpopulated when push=false, and the digest-consuming steps here are gated off on forks anyway

I can't exercise the fork path end-to-end from here -- ironically, this PR itself (from a fork) touches `.github/workflows/**` only, so it won't trigger the workflow. The first fork PR touching the examples inputs after merge will be the real test.

**Potential risks**
If the gate expression were wrong in the false direction, same-repo builds would silently skip the push; the expression is the widely used canonical form and `workflow_dispatch`/`push` events bypass it entirely. Fork PRs lose the "image actually pushed" signal, but they never had it -- they had a guaranteed red X instead.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None -- CI-only change.

**Documentation Changes**
None.

### DIFF
--- a/.github/workflows/build-examples-image.yaml
+++ b/.github/workflows/build-examples-image.yaml
@@ -77,7 +77,6 @@ jobs:
 
       - name: Build and push by digest
         id: build
-        if: env.PUSH_OK == 'true'
         uses: docker/build-push-action@v5
         with:
           context: ./python
@@ -88,19 +87,10 @@ jobs:
           # overwrite each other.
           cache-from: type=gha,scope=${{ env.PLATFORM_SLUG }}
           cache-to: type=gha,mode=max,scope=${{ env.PLATFORM_SLUG }}
-          outputs: type=image,name=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }},push-by-digest=true,name-canonical=true,push=true
-
-      - name: Build without pushing (fork PR)
-        if: env.PUSH_OK != 'true'
-        uses: docker/build-push-action@v5
-        with:
-          context: ./python
-          file: ./python/examples/Dockerfile
-          platforms: ${{ matrix.platform }}
-          labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=gha,scope=${{ env.PLATFORM_SLUG }}
-          cache-to: type=gha,mode=max,scope=${{ env.PLATFORM_SLUG }}
-          push: false
+          # push-by-digest is only consulted on the push path inside buildkit's
+          # image exporter, so with push=false it is inert and this builds
+          # without touching the registry.
+          outputs: type=image,name=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }},push-by-digest=true,name-canonical=true,push=${{ env.PUSH_OK }}
 
       - name: Export digest
         if: env.PUSH_OK == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-examples-image.yaml
```diff
@@ -82,13 +90,27 @@ jobs:
           cache-to: type=gha,mode=max,scope=${{ env.PLATFORM_SLUG }}
           outputs: type=image,name=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }},push-by-digest=true,name-canonical=true,push=true
 
+      - name: Build without pushing (fork PR)
```
craigmarker: does it make sense to factor the building into its own step so it can be reused? this new step duplicates part of `Build and push by digest`.

what am I missing?

apurvapatkeshwar: Yup, it was unnecessary duplication. I didn't want to template push= into the outputs string without knowing `push-by-digest=true,push=false` was well-defined
buildkit's image exporter only consults push-by-digest inside the push path, so with push=false it's inert
`moby/buildkitexporter/containerimage/export.go; build-push-action#1009` confirms outputs.digest just
comes back empty, which the digest steps never read on forks
Collapsed to a single step with `push=${{ env.PUSH_OK }}`
Thanks!

craigmarker (approved):

apurvapatkeshwar:
FYI the 2 red checks on #1787 are exactly this failure - both are the examples image build dying at the ghcr push on a fork PR
Merging this and re-running #1787 should clear them

PR #1545, opened 2026-07-17
Add LiteLLM gateway Helm chart

austingreco:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

- Added a standalone, independently versioned `michelangelo-llm-gateway` Helm chart for LiteLLM v1.85.1 with multi-architecture image digest pins.
- Added existing Secret and ConfigMap contracts, fail-closed Helm and Argo CD migration modes, externally managed migrations, authenticated Helm tests, and optional Ingress, NetworkPolicy, ServiceMonitor, HPA, PodDisruptionBudget, and topology spread resources.
- Added strict values validation, a GCP example, compatibility coverage, packaging checks, and a guarded GHCR publication workflow.

**Why?**

Provide the reusable Milestone 1 data-plane package for running LiteLLM alongside Michelangelo without coupling inference traffic to the Michelangelo control plane. Milestone 0 runtime validation, project reconciliation, caller credential delivery, provider infrastructure, and production platform controls remain separate work.

**How did you test it?**

- Ran `bash .github/scripts/validate-llm-gateway-chart.sh` with Helm 3.12.0 and Helm 4.1.0, including strict lint, every supported render profile, invalid-value rejection, private-registry rendering, and chart packaging.
- Ran `shellcheck .github/scripts/validate-llm-gateway-chart.sh` and `git diff --check`.
- Verified the LiteLLM and curl image tags, digests, and amd64/arm64 manifests with `docker buildx imagetools inspect`.
- Real GKE, Cloud SQL, External Secrets Operator, Prometheus Operator, provider traffic, migration rollback, and uninstall validation are still required before the first production release.

**Potential risks**

- The upstream LiteLLM database image runs as root, so production requires a validated hardened image or an explicit policy exception.
- Database migrations are not reversed by Helm rollback and require backward-compatible changes or a tested Cloud SQL recovery path.
- The chart depends on externally provisioned PostgreSQL, Secrets, provider access, private networking, and environment-specific policy resources.

**Breaking Changes**

- [x] No breaking changes
- [ ] API changes (Go exported symbols or function signatures, Python public functions or classes)
- [ ] Proto changes (enum value renumbering, field number changes, field removal, service removal)
- [ ] Helm changes (new required values, renamed or removed keys, changed value semantics)
- [ ] Config/deployment changes (new required env vars, renamed container args, changed ports or mount paths)

**Release notes**

[...]

PR #1808, opened 2026-08-13
feat(helm): add LiteLLM gateway wrapper chart

austingreco:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

- Added `michelangelo-llm-gateway`, a thin wrapper around the pinned official `litellm-helm` 1.85.1 chart.
- Added fail-closed migrations, NetworkPolicy, authenticated ServiceMonitor, digest-pinned tests, strict values validation, GCP examples, and chart CI.
- Right-sized gateway and migration memory defaults and added exact render assertions for their requests and limits.
- Extends the standard Michelangelo release so that, beginning with the first release after this PR merges, the gateway and control-plane charts share a version while remaining separate Helm releases and runtime lifecycles.

**Why?**

This supersedes #1545's duplicated LiteLLM templates with a thin upstream wrapper. Shared artifact versioning gives operators and integration tests one compatibility coordinate without coupling gateway scaling, migrations, upgrades, or rollback to the Michelangelo control plane.

**How did you test it?**

- Ran `.github/scripts/validate-llm-gateway-chart.sh` with Helm 3.17.0 and 4.1.0 across supported profiles, invalid-value cases, dependency checksum verification, private-registry rendering, kubeconform, and package-content assertions.
- Ran `actionlint` 1.7.12, `shellcheck`, `scripts/version-bump.sh --check`, `git diff --check`, and packaged both charts at version 0.10.0.
- In a local k3d sandbox, completed a fresh PostgreSQL-backed install and upgrade. The fresh migration applied 122 migrations, peaked at 1,376 MiB with a 1 GiB request and 2 GiB limit, and completed without a restart.
- Passed the authenticated wrapper Helm test and routed both mock-provider and real-provider chat completions through the local k3d gateway; the deployed gateway remained Ready with zero restarts.
- Real GKE, Cloud SQL, External Secrets, Prometheus Operator, HA/load, and production-environment qualification remain staging work.

**Potential risks**

- This is a prerelease chart. Local k3d qualification, including real-provider connectivity, does not replace production GKE, Cloud SQL, External Secrets, Prometheus Operator, HA/load, or production-environment qualification.
- Before first publication, maintainers must configure the `llm-gateway-chart-release` repository environment with the intended protection policy; GitHub otherwise creates an unprotected environment on first use.
- Operators need enough spare cluster capacity for the pre-upgrade migration Job and one rolling-update surge pod.
- LiteLLM 1.85.1 ships unconditional upstream test hooks with mutable public images; operators must use the documented filtered wrapper test.
- Database migrations are not reversed by Helm rollback. The coordinated publisher validates all versions before publishing and treats duplicate release runs as idempotent only when package contents match.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Adds the prerelease `michelangelo-llm-gateway` OCI chart. Because `v0.10.0` predates this chart, its first provenance-correct OCI package will be published alongside the control-plane chart by the first Michelangelo release after this PR merges.

**Documentation Changes**

Added the chart operations guide, GCP values examples, release and integration-test guidance, and links from the operator and contributor documentation indexes.


.github/workflows/llm-gateway-chart-ci.yml
```diff
@@ -0,0 +1,77 @@
+name: LLM gateway chart
```
sallycr, line 1: @austingreco have we tested this CI? I wonder if it's working.

.github/workflows/llm-gateway-chart-release.yml
```diff
@@ -0,0 +1,186 @@
+name: Publish LLM gateway chart
```
sallycr, line 1: I think we should have the release together with michelangelo release.

helm/michelangelo-llm-gateway/README.md
```diff
@@ -0,0 +1,244 @@
+# Michelangelo LLM Gateway Helm Chart
+
+This chart installs LiteLLM as Michelangelo's independently operated LLM gateway data plane. It wraps the official `litellm-helm` chart instead of maintaining a second implementation of the LiteLLM Deployment and supporting resources.
+
+The gateway is installed, upgraded, and rolled back separately from the Michelangelo control-plane chart. This keeps inference traffic, database migrations, scaling, and availability outside the control-plane release lifecycle.
```
sallycr: if we have separately released, wouldn't it be an issue for the future integration test? what do you think?

.github/workflows/llm-gateway-chart-release.yml
austingreco: Agreed. Addressed in `dd753b1a`: the gateway publisher is now a reusable job in the standard Michelangelo release workflow, gated by the same `vX.Y.Z` tag and version preflight. The release bump keeps both chart versions aligned, and duplicate release runs skip the gateway push only when the existing package content matches. The gateway remains a separate Helm release so its scaling, migrations, upgrades, and rollback are not coupled to the control plane.

.github/workflows/llm-gateway-chart-ci.yml
austingreco: Yes. The workflow has run successfully on this PR, and the fresh run for `dd753b1a` passed both Helm 3.17.0 and Helm 4.1.0: https://github.com/michelangelo-ai/michelangelo/actions/runs/32765912522. It exercises the full chart validation script, kubeconform, shellcheck, release-version alignment, and actionlint. I also reran the same chart matrix locally.

helm/michelangelo-llm-gateway/README.md
austingreco: Good concern. The shared release version now gives future integration tests one compatibility coordinate: install both OCI charts at `X.Y.Z`, then explicitly test independent gateway or control-plane upgrades when needed. The README documents this. They remain separate Helm releases because combining runtime lifecycles would couple gateway database migrations and availability to control-plane rollback.

Issue #678, opened 2025-12-08
SNYK Security Vulnerability in vllm@0.8.1 - Fix: update to vllm@0.8.5


Melda-s:
Introduced through vllm@0.8.1
Fixed in vllm@0.8.5

**Detailed paths and remediation**
Introduced through: michelangelo-ai@0.1.0 › vllm@0.8.1
Fix: Upgrade vllm to version 0.8.5 

**Overview**
[vllm](https://pypi.org/project/vllm) is an A high-throughput and memory-efficient inference and serving engine for LLMs

Affected versions of this package are vulnerable to Deserialization of Untrusted Data through the PyNcclPipe service if it is in use with the V0 engine. An attacker can execute arbitrary code on the server by sending malicious serialized data. This is only exploitable if the service is configured to listen on a publicly accessible port or if the network security settings are improperly configured.

sallycr:
**Current state (May 2026):** pyproject.toml has vllm = {version = "0.8.1", optional = true} — pinned to the vulnerable version exactly. Update to 0.8.5 or later.

paulzim:
`vllm` is still pinned at `0.8.1` in `pyproject.toml`. The fix is straightforward — bump to `>=0.8.5`:

```toml
vllm = { version = ">=0.8.5", optional = true }
```

This remains an open security vulnerability.

PR #1812, opened 2026-08-14
deps: bump vllm to 0.9.2 to shed known CVEs

apurvapatkeshwar:
## Summary

Bumps the optional `vllm` extra from 0.8.1 to 0.9.2, closing #678 (Snyk: deserialization of untrusted data in the PyNcclPipe service, fixed in 0.8.5) and also shedding the tool-schema DoS fixed in 0.9.0 (CVE-2025-48944, affects >=0.8.0,<0.9.0).

## Why 0.9.2 and not the 0.8.5 the issue asks for

0.8.5 is unresolvable in this dependency tree: it pins `opentelemetry-exporter-otlp` to 1.26.x, and that otel release caps `protobuf < 5`, while michelangelo requires `protobuf >= 5.27.1`. vllm 0.9.0 lifted the otel caps, so 0.9.x is the lowest CVE-fixing line that resolves against this project. I verified this by attempting the 0.8.5 lock directly — poetry fails on the protobuf conflict.

Worth stating plainly: vllm has accumulated many more advisories since, and the fully-patched line today requires python >=3.10, torch 2.11, and transformers 5.x — a platform-level migration this PR does not attempt. 0.9.2 is strictly better than 0.8.1 (every advisory affecting 0.9.2 also affects 0.8.1, and 0.9.2 closes at least two more), so this is an incremental risk reduction, not a clean bill. Happy to file the larger migration as a separate issue if there is appetite.

## Companion bumps the resolver forces (all in optional extras)

- `torch` 2.6.0 -> 2.7.0 — vllm 0.9.x pins `torch==2.7.0` exactly
- `transformers` 4.48.2 -> 4.51.3 — vllm needs >=4.51.1
- `setuptools` 74.1.1 -> 79.0.1 — vllm needs >=77.0.3,<80; this pin exists only for the vllm extra (the "For VLLM" block in pyproject)
- the `vllm` dependency now carries `python = "<3.13"` — 0.8.1 declared no upper python bound, 0.8.5+ declare `<3.13`, so poetry requires the marker; on python 3.13+ the extra simply omits vllm

## Blast radius

- `vllm` itself is imported only by the `llm_prediction` examples; CI never installs the vllm extra.
- `torch`/`transformers` are installed by the dev/example/trainer extras, so CI and the examples image pick up 2.7.0/4.51.3. The only library-code use of transformers is a TYPE_CHECKING-only import; pytorch-lightning's `>=2.2,<3` range covers torch 2.7.
- This PR touches `poetry.lock`, so the examples-image workflow will run; on a fork PR its push step fails against ghcr with a permissions denial (see #1802, which fixes that) — those two red X's are expected until #1802 lands.

## Testing

- `poetry lock` resolves cleanly on top of current main (including the xgboost trainer's lockfile changes from #1793); `poetry check --lock` passes.
- Full pytest suite under the dev extra with torch 2.7.0 + transformers 4.51.3 on python 3.9 (the version CI's coverage job uses): 2643 passed, 0 failed.
- Old vs new lock spot-checks: protobuf (6.33.6) and ray (2.51.2) are unchanged; the lock diff is the vllm 0.9.2 dependency tree plus the four pins above.

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -67,7 +67,7 @@ description = "Happy Eyeballs for asyncio"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
-markers = "python_version == \"3.9\" and (extra == \"dev\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\") or python_version != \"3.9\" and python_version < \"3.10\" and (platform_system == \"Linux\" or platform_system == \"Windows\") and (extra == \"dev\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\") and (platform_machine == \"aarch64\" or platform_system == \"Windows\")"
+markers = "python_version == \"3.9\" and (extra == \"dev\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"trainer-xgboost\")"
 files = [
     {file = "aiohappyeyeballs-2.6.1-py3-none-any.whl", hash = "sha256:f349ba8f4b75cb25c99c5c2d84e997e485204d2902a9597802b0371f09331fb8"},
     {file = "aiohappyeyeballs-2.6.1.tar.gz", hash = "sha256:c3f9d0113123803ccadfdf3f0faa505bc78e6a72d1cc4806cbd719826e943558"},
@@ -80,7 +80,7 @@ description = "Happy Eyeballs for asyncio"
 optional = false
 python-versions = ">=3.10"
 groups = ["main"]
-markers = "python_version >= \"3.10\" and (extra == \"dev\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\")"
+markers = "python_version >= \"3.10\" and (extra == \"dev\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\" or extra == \"trainer-xgboost\")"
 files = [
     {file = "aiohappyeyeballs-2.6.2-py3-none-any.whl", hash = "sha256:4708045e2d7a6c6bdf8aafa8ed39649eaf926a4543b54560659129e3365953c4"},
     {file = "aiohappyeyeballs-2.6.2.tar.gz", hash = "sha256:e202810ee718bd01fc6ef49e8ea53d023d5cb6b581076d7925aa499fa55dbe64"},
@@ -93,7 +93,7 @@ description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
-markers = "python_version == \"3.9\" and (extra == \"dev\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\") or python_version != \"3.9\" and python_version < \"3.10\" and (platform_system == \"Linux\" or platform_system == \"Windows\") and (extra == \"dev\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\") and (platform_machine == \"aarch64\" or platform_system == \"Windows\")"
+markers = "python_version == \"3.9\" and (extra == \"dev\" or extra == \"example\" or extra == \"vllm\" or extra == \"trainer\" or extra == \"plugin\" or extra == \"ray-polars\" or extra == \"trainer-xgboost\")"
 files = [
     {file = "aiohttp-3.13.5-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:02222e7e233295f40e011c1b00e3b0bd451f22cf853a0304c3595633ee47da4b"},
     {file = "aiohttp-3.13.5-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:bace460460ed20614fa6bc8cb09966c0b8517b8c58ad8046828c6078d25333b5"},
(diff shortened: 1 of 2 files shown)


paulzim (approved):
## Review: #1812 — deps: bump vllm to 0.9.2 to shed known CVEs

**Verdict: Approve**

Bumps the optional `vllm` extra 0.8.1 → 0.9.2 (closing #678, and separately fixing CVE-2025-48944), plus the four companion pins vllm's own dependency tree forces: `torch` 2.6.0→2.7.0, `transformers` 4.48.2→4.51.3, `setuptools` 74.1.1→79.0.1, and a new `python = "<3.13"` marker on `vllm` itself. Pure dependency-bump PR — 2 files, `pyproject.toml` (8 lines) + `poetry.lock` (regenerated).

**Verified independently, not just taken from the PR description:**
- Diffed every package version in `poetry.lock` old vs. new (parsed both trees, not just skimmed the udiff): all 36 changed entries are either vllm itself, a transitive dependency pulled in by vllm's 0.8.1→0.9.2 bump (the `nvidia-cu12` stack, `torch`/`torchvision`/`torchaudio`, `triton`, `xformers`, `xgrammar`, `sympy`, `numba`/`llvmlite`, `mistral-common`, `gguf`, `compressed-tensors`, the new `mlx*`/`llguidance`/`pybase64` entries), or one of the four pins called out above. `protobuf` and `ray` are confirmed unchanged, matching the description.
- Confirmed the reason 0.8.5 (the version the linked issue asks for) isn't viable: vllm 0.8.5 pins `opentelemetry-exporter-otlp` to 1.26.x, which caps `protobuf < 5`, conflicting with this project's `protobuf >= 5.27.1` — a real, structural resolver conflict, not a preference call.
- `pytorch-lightning`'s pyproject pin (`>=2.2.0,<3.0.0`, unchanged by this PR) resolves to 2.6.0/2.6.5 in the lock, both declaring `torch = ">=2.1.0"` — confirmed it covers 2.7.0 with no pin bump needed.
- Grepped all `transformers`/`vllm` imports in `python/`: the only non-example, non-test library-code use of `transformers` is a `TYPE_CHECKING`-only import in `michelangelo/lib/trainer/torch/utils.py`; `vllm` is imported only by `examples/llm_prediction/vllm_predict.py`. No CI workflow references `vllm` anywhere — confirms the claimed blast radius (dev/example/trainer extras pick up torch/transformers; nothing in CI touches the vllm extra itself).
- Ran `poetry check --lock` directly in the worktree: passes (exit 0), confirming the lock is consistent with the new `pyproject.toml` independent of the PR's own reported `poetry lock`/`poetry check --lock` run.
- Pulled the actual CI failure logs for both red X's (`Build (linux/amd64)`/`Build (linux/arm64)`): both fail at `ERROR: failed to push ghcr.io/michelangelo-ai/examples: denied: installation not allowed to Write organization package` — a fork-PR `GITHUB_TOKEN` permissions limitation, not a build break from this bump. Confirmed open PR #1802 (unmerged) targets exactly this failure signature. `Check Coverage (Python 3.9)` — the real exercise of the new torch/transformers versions — passes.

**One nit not mentioned in the PR description:** `openai` drops from `2.44.0` to `1.90.0` — a major-version downgrade, and the only entry in the whole 36-package diff that moves backward rather than forward. Traced it: vllm 0.9.2 pins `openai = ">=1.52.0,<=1.90.0"` in its own lock entry, so the resolver had no choice — this is forced, not accidental. No repo code imports `openai` directly (it's pulled in solely as a `vllm`-extra transitive dependency), so it's inert today. Worth a one-line mention in the description next time, since a major-version downgrade is the kind of line that looks like a mistake at a glance even when it's correct — but not blocking.

No other concerns. The description's dependency accounting is otherwise unusually thorough and everything in it checks out against the actual lock diff and CI logs.

apurvapatkeshwar:
Heads up, we should consider landing this soon -
`xgrammar` 0.1.16  (which `vllm` 0.8.1 pins exactly and main's `poetry.lock` references) has been deleted from `pypi` (releases now jump 0.1.13 -> 0.1.17)
So today, any change touching `python/pyproject.toml` fails to relock ("package xgrammar (0.1.16) not found") and a fresh install with the vllm extra can't fetch the wheel
This PR's lock resolves to `xgrammar` 0.1.19 and escapes all of that
It probably also explains the silent dependabot pip failures in #1829

paulzim (approved):
## Re-check (2026-08-18): CI-rerun update

**Verdict: Approve stands.**

The only change since the last pass is a new empty commit, `chore: rerun CI now that the examples-image fork fix is on main` (`f82cd6d2`) — no content changes to `pyproject.toml`/`poetry.lock`, so nothing from the original review needed re-verifying on the code side. This was specifically to check whether #1802 (merged to main since the last pass) actually fixed the ghcr fork-permission denial both `Build` jobs hit previously.

- **`Build (linux/amd64)`: now passes (35m30s).** Confirms #1802 fixed the `denied: installation not allowed to Write organization package` failure this same job hit before.
- **`Build (linux/arm64)`: still red, but for an unrelated reason.** New failure is `curl: (28) Failed to connect to archive.apache.org port 443 after 135569 ms: Connection timed out` inside the Dockerfile's Spark-download step (`stage-0 8/19`) — happens before the build ever reaches the ghcr push step, so it's not a recurrence of the #1802 issue. Confirmed this is pre-existing CI flakiness, not something this PR triggers: the identical `Failed to connect to archive.apache.org` signature on the identical curl command hit the arm64 job on the unrelated `feat/xgboost-ray-trainer` branch on 2026-08-12 — six days before this PR's rerun. This PR touches zero Dockerfile/build-infra files (`pyproject.toml` + `poetry.lock` only), so it has no mechanism to cause an arm64 QEMU build to time out fetching Spark from Apache's archive.
- All substantive checks remain green: `bazel-build`, `python-build`, `Check Coverage (Python 3.9)`, ruff, `dirty-check`, `license/cla`.

Net: the update did what it was meant to do (unblock the fork-permission issue), and the remaining red X is flaky infra unrelated to this diff. Recommend merging once the arm64 job is re-run/green, not blocking on it as a reflection of this PR's correctness.

PR #1816, opened 2026-08-14
Add buildIndexedFieldId for resolving repeated-layout field ids

craigmarker:
## Summary
Repeated layouts render one entity-relative config against every row of an array, so a literal field id isn't enough on its own — the current row's index needs to be spliced into the right spot in the path. This adds that path-building step as a standalone, independently tested utility, stacked on top of the base condition operator and ahead of wiring it into anything that consumes it.

## Test Plan
I ran the utility's unit tests directly, covering index insertion, an exact-match root path, multiple existing indices in a nested repeated root, and index 0.

## Issues


craigmarker:
Consolidating into #1817 — decided not to separate the helper from its only caller.

Issue #1471, opened 2026-07-08
Auto-generate grpc_json_transcoder services list from proto descriptors


craigmarker:
## Context

The `grpc_json_transcoder` filter in `helm/michelangelo/templates/core/envoy-configmap.yaml` requires an explicit `services` whitelist. This list must be manually kept in sync with `javascript/packages/rpc/services.ts` — adding a new service to the JS client without updating the Envoy config silently breaks transcoding for that service.

## Proposal

Generate the services list at build time from the same `descriptors.pb` that the transcoder already consumes. Options:

1. **`buf` + `jq` in `gen-grpc-client.sh`**: extract fully-qualified service names from the FileDescriptorSet and write them to a Helm values file (e.g., `values-generated.yaml`) that the template reads via `.Values.envoy.transcoderServices`.
2. **Helm `lookup` or post-renderer**: less portable, ties generation to deploy time rather than build time.

Either way, the `services:` block in the Envoy ConfigMap template would iterate over a list rather than hardcoding names.

## Current state

15 services manually listed in `envoy-configmap.yaml`, mirroring the service imports in `services.ts`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1818, opened 2026-08-14
fix(rpc): preserve string enum names from Envoy through the RPC client

craigmarker:
> **Stacked on #1814 — merge #1814 first.** This branch is rebased on top of `craig/gh-1471-generate-transcoder-services`; the diff shown here is only this PR's own change.

## Summary

Protobuf enum fields sometimes reached the UI as raw numbers (e.g. `2`) instead of string names (e.g. `PHASE_COMPLETE`). Envoy's `grpc_json_transcoder` filter is the only JSON-to-proto conversion path in this repo — no grpc-gateway, no Go-side `protojson` marshaling — and it already defaults `always_print_enums_as_ints` to `false`, so it emits string names for any enum value it recognizes. But it can only resolve names for values present in its own proto descriptor set, and that set had drifted 5+ weeks stale with nothing in CI tying it to `.proto` changes, so any enum value added during that window rendered as a raw number until someone regenerated the descriptor by hand.

#1814 fixes that staleness problem structurally, tying descriptor regeneration to the proto build itself. This PR sets `always_print_enums_as_ints: false` explicitly in the transcoder's `print_options`, so the correct behavior is documented rather than relying on it implicitly surviving a future Envoy upgrade.

That only fixes half the pipeline, though. Even with Envoy emitting the right string, the frontend's RPC client was decoding every response through protobuf-es's `fromJson()`, which normalizes enums back to their numeric discriminant on the way into a typed message. Every config entity that displayed or branched on an enum carried a hand-maintained numeric map shadowing the proto definition to undo that normalization — a second copy of the enum that silently drifted whenever a proto enum changed. This PR removes that round-trip: the RPC client now returns the raw JSON response as-is, and every `typeTextMap`/`stateTextMap`/`stateColorMap` and enum switch/comparison across the pipeline, run, trigger, deployment, target, and model configs is re-keyed to the proto's own string constants. The frontend's defensive numeric-enum guards are left in place on purpose — they guard the deploy-skew window where a backend has already started returning a new enum value before Envoy's descriptor configmap has finished rolling out, a case no marshaling config can eliminate.

## Test plan

I deployed this branch's Envoy config plus a regenerated `descriptors.pb` into a running sandbox and confirmed the descriptor-freshness fix over real RPC calls (`ProjectService.CreateProject`/`UpdateProject`/`GetProject`, enum field `spec.retentionConfig.notificationControl`) — full before/after trace posted as [a PR comment](https://github.com/michelangelo-ai/michelangelo/pull/1818#issuecomment-5296790450):

- With a current descriptor, in sync with proto: `"notificationControl": "NOTIFICATION_CONTROL_PROJECT_OWNER_ONLY"`.
- Reproducing the original failure mode — rebuilt a `descriptors.pb` from a proto copy with that enum value deleted (simulating a descriptor that's drifted stale relative to a newer `.proto`), swapped it into the running Envoy configmap, and re-ran the same call against the same stored data: `"notificationControl": 2`. Restoring the current descriptor flipped it back to the string name immediately, with no other change.

Separately, I stood up the full sandbox (apiserver, Envoy, controller manager, and a branch-built frontend pointed at the real Envoy over the JSON path this PR changes) and walked through every entity config touched by the re-keying:

- Pipelines list: Type and State columns render "Train"/"Evaluation" and "Ready" from the wire values `PIPELINE_TYPE_TRAIN`/`PIPELINE_TYPE_EVAL` and `PIPELINE_STATE_READY`.
- Pipeline Runs list and run detail: State renders "Failed", and the Steps tab renders per-step status ("Success", "Pending") from the string values the `stateBuilder` switch now matches on.
- Trained Models list: Type renders "LLM Chat" from `MODEL_KIND_LLM_CHAT_COMPLETION`.
- Targets list: Type renders "Inference Server" from `TARGET_TYPE_INFERENCE_SERVER`.

No cell showed a raw enum string in place of its label anywhere I could load data. Deployments, Trigger Runs, and the Targets State column weren't exercised — the sandbox had no data reaching those states in the time available (Deployments needs a full Revision chain, no trigger had fired yet, and the InferenceServer controller never populated `.status.state` in this sandbox environment for either the demo resource or a freshly created one).

This PR has no independent CI repro of its own for the Envoy half: `always_print_enums_as_ints: false` is already Envoy's default, so lint/template checks can't distinguish it from a no-op. What actually prevents the drift this PR's setting depends on is #1814's CI enforcement, independently verified there by its own throwaway PRs ([#1842](https://github.com/michelangelo-ai/michelangelo/pull/1842), [#1844](https://github.com/michelangelo-ai/michelangelo/pull/1844)) reproducing a stale-descriptor merge and confirming CI blocks it. The frontend half is covered by unit tests re-asserting each re-keyed map and switch against the proto's string constants.


craigmarker:
## Sandbox verification

Deployed this branch's Envoy config + regenerated `descriptors.pb` into a running sandbox (`k3d-michelangelo-sandbox`) and confirmed end-to-end over real RPC calls (`ProjectService.CreateProject`/`UpdateProject`/`GetProject`, enum field `spec.retentionConfig.notificationControl`):

**With the fix (current descriptor, in sync with proto):**
```json
"retentionConfig": { "notificationControl": "NOTIFICATION_CONTROL_PROJECT_OWNER_ONLY" }
```

**Reproduced the exact failure mode this fix prevents** — rebuilt a `descriptors.pb` from a proto copy with enum value `2` deleted from `NotificationControl` (simulating a descriptor that's gone stale relative to a newer `.proto`), swapped it into the running Envoy configmap, and re-ran the same `GetProject` call against the *same* stored data:
```json
"retentionConfig": { "notificationControl": 2 }
```

Restoring the fixed/current descriptor flipped it back to the string name immediately, no other changes. This confirms the root cause: Envoy's transcoder only prints an enum's string name for values present in its own `descriptors.pb`; anything added to `.proto` after that file was last regenerated silently degrades to a raw number until it's refreshed. Sandbox left running with the fixed config; test project was created and cleaned up.

PR #1828, opened 2026-08-17
fix: create TriggerRun schedules paused

zhouxianzx:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

- Added a start-paused workflow option and use it when a TriggerRun is created with `spec.action: PAUSE`.
- Create new Temporal schedules paused and pause existing Temporal schedules immediately.
- Consume the initial PAUSE action during the first reconciliation and reject unsupported paused creation on Cadence.

**Why?**

A TriggerRun created with a PAUSE action could briefly create an active schedule because the action was not applied until a later reconciliation. Creating the schedule paused prevents an execution from starting during that window.

**How did you test it?**

- `tools/bazel test //go/components/triggerrun:go_default_test //go/base/workflowclient/temporalclient:go_default_test //go/base/workflowclient/cadenceclient:go_default_test --test_output=errors`

**Potential risks**

The change affects initial Temporal schedule state and the existing-schedule path. Regression tests cover both paths, controller action consumption, and the Cadence rejection behavior.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

TriggerRuns created with `spec.action: PAUSE` now create Temporal schedules in the paused state immediately.

**Documentation Changes**

None. This is a backward-compatible behavior fix.

### DIFF
--- a/go/base/workflowclient/temporalclient/temporalclient.go
+++ b/go/base/workflowclient/temporalclient/temporalclient.go
@@ -103,6 +103,12 @@ func (c *TemporalClient) createScheduleForCron(ctx context.Context, options clie
 	if scheduleHandle != nil {
 		_, err := scheduleHandle.Describe(ctx)
 		if err == nil {
+			if options.StartPaused {
+				paused := true
+				if err := c.UpdateTrigger(ctx, options.ID, "", &paused, nil); err != nil {
+					return nil, fmt.Errorf("failed to pause existing Temporal schedule: %w", err)
+				}
+			}
 			// Schedule already exists, return success
 			return &clientInterface.WorkflowExecution{
 				ID:    scheduleID,
@@ -125,6 +131,10 @@ func (c *TemporalClient) createScheduleForCron(ctx context.Context, options clie
 		},
 		Overlap:        overlapPolicy, // Use extracted policy based on maxConcurrency
 		PauseOnFailure: false,
+		Paused:         options.StartPaused,
+	}
+	if options.StartPaused {
+		scheduleOptions.Note = "paused by michelangelo"
 	}
 
 	// Set workflow timeout if provided
--- a/go/components/triggerrun/controller.go
+++ b/go/components/triggerrun/controller.go
@@ -247,6 +247,10 @@ StateMachine:
 			"state", status.State,
 			"execution_workflow_id", status.ExecutionWorkflowId)
 		triggerRun.Status = status
+		if triggerRun.Spec.Action == v2pb.TRIGGER_RUN_ACTION_PAUSE &&
+			status.State == v2pb.TRIGGER_RUN_STATE_PAUSED {
+			triggerRun.Spec.Action = v2pb.TRIGGER_RUN_ACTION_NO_ACTION
+		}
 	case v2pb.TRIGGER_RUN_STATE_RUNNING:
 		log.Info("TRIGGER_RUN_STATE_RUNNING")
 
--- a/go/components/triggerrun/cron_trigger.go
+++ b/go/components/triggerrun/cron_trigger.go
@@ -53,8 +53,8 @@ func NewCronTrigger(log logr.Logger, workflowClient clientInterface.WorkflowClie
 //   - DecisionTaskStartToCloseTimeout: 30 seconds
 //   - CronSchedule: From TriggerRun spec
 //
-// Returns State=RUNNING if workflow starts successfully or is already running,
-// State=FAILED if workflow start fails.
+// Returns State=RUNNING when an active schedule starts successfully, State=PAUSED when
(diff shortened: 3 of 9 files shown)


go/components/triggerrun/controller.go
```diff
@@ -247,6 +247,10 @@ StateMachine:
 			"state", status.State,
 			"execution_workflow_id", status.ExecutionWorkflowId)
 		triggerRun.Status = status
+		if triggerRun.Spec.Action == v2pb.TRIGGER_RUN_ACTION_PAUSE &&
```
badcount, line 250: Can the trigger run start with TRIGGER_RUN_STATE_PAUSED? so it's paused to start?

go/components/triggerrun/cron_trigger.go
```diff
@@ -89,10 +90,21 @@ func (r *cronTrigger) Run(ctx context.Context, triggerRun *v2pb.TriggerRun) (v2p
...
 		// reconciliation.
-		return v2pb.TriggerRunStatus{
+		status := v2pb.TriggerRunStatus{
 			State:  v2pb.TRIGGER_RUN_STATE_RUNNING,
 			LogUrl: getWorkflowURL(wid, r.WorkflowClient.GetProvider()),
-		}, nil
+		}
+		if opt.StartPaused {
+			paused := true
+			if updateErr := r.WorkflowClient.UpdateTrigger(ctx, wid, "", &paused, nil); updateErr != nil {
```
badcount, line 99: On Cadence, this silently reports `PAUSED` without actually pausing anything.

`WorkflowClient.UpdateTrigger` is an unconditional no-op on Cadence that returns `nil`:

```go
// go/base/workflowclient/cadenceclient/cadenceclient.go
func (c *CadenceClient) UpdateTrigger(_ context.Context, _ string, _ string, _ *bool, _ []interface{}) error {
	return nil
}
```

So if this branch (existing schedule + `Action=PAUSE`) ever runs against a Cadence-backed trigger, `status.State` gets set to `PAUSED` on line 105 even though the underlying Cadence cron workflow keeps firing on schedule — a silently wrong status.

That's inconsistent with how pause is handled everywhere else in this PR/codebase, which fails loud on Cadence instead:
- The new `CadenceClient.StartWorkflow` guard: `"starting a paused workflow is not supported by Cadence"`
- `pauseWorkflow()` → `WorkflowClient.PauseTrigger`: `"PauseTrigger not supported by Cadence provider"`

Suggest routing this branch through `PauseTrigger` (or the same not-supported check `StartWorkflow` uses) instead of `UpdateTrigger`, so Cadence fails the same way here as it does in the fresh-create path.

Worth noting: `TestRunPausesExistingScheduleImmediately` mocks `UpdateTrigger` to unconditionally return `nil` — which is exactly real Cadence's behavior — so that test actually demonstrates this gap rather than ruling it out.

go/base/workflowclient/temporalclient/temporalclient.go
```diff
@@ -103,6 +103,12 @@ func (c *TemporalClient) createScheduleForCron(ctx context.Context, options clie
 	if scheduleHandle != nil {
 		_, err := scheduleHandle.Describe(ctx)
 		if err == nil {
+			if options.StartPaused {
+				paused := true
+				if err := c.UpdateTrigger(ctx, options.ID, "", &paused, nil); err != nil {
+					return nil, fmt.Errorf("failed to pause existing Temporal schedule: %w", err)
+				}
+			}
```
badcount, line 111: Why do we need to update Trigger intead of just set the status during creation

zhouxianzx: We do. A new schedule is created paused directly. 

UpdateTrigger handles retries during initial TriggerRun creation, where the Temporal schedule was created but the TriggerRun status was not persisted. It idempotently ensures the existing schedule is paused before returning success.

go/components/triggerrun/controller.go
zhouxianzx: Yes, Temporal can create a paused Schedule

zhouxianzx: It needs to initially be INVALID so the reconciler creates the schedule. Temporal creates the schedule paused directly, then the TriggerRun transitions to PAUSED in the same reconciliation. Starting with PAUSED would skip schedule creation.

go/components/triggerrun/cron_trigger.go
zhouxianzx: Good catch. The StartWorkflow guard only covers fresh creation; the existing-workflow retry path calls UpdateTrigger directly. I updated Cadence’s UpdateTrigger to reject pause/resume requests and added regression tests.

badcount (approved):

Issue #1227, opened 2026-05-21
feat: [CanvasFlex Pusher] MLflowRegistryClient + pyproject.toml extras


sallycr:
## CanvasFlex Pusher — Registry Client Implementations + pyproject.toml

Concrete MLflow and Vertex AI registry clients as optional extras. Independent of plugin logic — can be reviewed by registry/infra specialists separately.

### Scope
- `pusher/implementations/mlflow_client.py` — `MLflowRegistryClient`
- `pyproject.toml` — new optional deps and extras: `pusher`, `mlflow`
- Tests for both clients (via `unittest.mock.patch`)

### Key design
- Lazy import pattern: `try: import mlflow except ImportError as exc: raise ImportError(...) from exc`
- MLflow: deployable URI stored as model-version tag

**Depends on:** #1220 (PR6)
**Part of:** CanvasFlex Pusher Phase 1

apurvapatkeshwar:
With #1220 closed this looks unblocked - is the phase 1 scope still current, and would you take an external PR for the MLflow half? The spec in the issue body looks concrete enough to implement directly: `pusher/implementations/mlflow_client.py` implementing the `ModelRegistryClient` ABC (including `RegisteredModel` aliasing its docstring already prescribes), lazy import guard, deployable URI as a model-version tag, mlflow as an optional dep behind the pusher/mlflow extras, and tests with unittest.mock only (no live MLflow server in CI)

1 scope question: MLflow and Vertex together as the title groups them, or MLflow first as its own reviewable unit? Best to start MLflow only to keep the PR small. And this stays an opt-in client implementation of the public seam; i.e., no interception of user MLflow calls, nothing imported unless the extra is installed

PR #1836, opened 2026-08-17
feat(pusher): add MLflowRegistryClient behind a pusher-mlflow extra

apurvapatkeshwar:
Implements the MLflow half of #1227 (MLflow first as its own reviewable unit, per the scope question raised in the issue comments; Vertex AI is tracked separately under the RFC's later phases).

## What this does

Adds `MLflowRegistryClient`, a `ModelRegistryClient` implementation that registers pushed models in an MLflow Model Registry. It is an opt-in client for the existing public seam -- pass it to `push(..., registry_client=...)` or list it in `ModelPluginConfig.registry_clients`; no plugin logic changes, nothing intercepts user MLflow calls, and nothing imports mlflow unless the client is actually used.

```python
from michelangelo.workflow.tasks.pusher import push, MLflowRegistryClient

result = push(
    config=pusher_config,
    artifacts=artifacts,
    registry_client=MLflowRegistryClient(
        tracking_uri="http://mlflow.example.com:5000",
    ),
)
```

Design choices, following the issue's spec and the ABC's documented contracts:

- **Lazy import with an install hint.** mlflow imports on first use; without it, use raises `ImportError` pointing at `pip install 'michelangelo[pusher-mlflow]'`. Importing the module (or the pusher package) never requires mlflow.
- **Extra named `pusher-mlflow`.** The issue text proposed bare `pusher`/`mlflow` extras, but the repo already established a `<subsystem>-<backend>` convention (`trainer-comet`, `trainer-mlflow`, ...), so this follows it -- happy to rename if the flat names are preferred. The optional `mlflow` dependency itself already existed in pyproject; no new pin.
- **Field mapping.** `labels` map to plain MLflow model-version tags (the ABC documents labels == MLflow tags). Fields MLflow has no native column for travel as reserved `michelangelo_*` tags: the deployable artifact URI (per the issue: "deployable URI stored as model-version tag"), `kind`, and JSON-encoded `metadata`. `metadata["run_id"]` is extracted and passed to MLflow's native run linkage instead, per the ABC's documented convention, and is reconstructed from the model version's `run_id` on `get_model`. `schema` is accepted and silently ignored (no MLflow schema field), also per the ABC contract.
- **MLflow's own `RegisteredModel` collision** is handled with exactly the alias the ABC docstring prescribes.
- **Error contracts.** MLflow's not-found maps to `KeyError` and invalid-parameter to `ValueError`, matching the ABC's documented raises; other registry errors propagate untouched.
- **Filter quoting.** Latest-version lookup goes through `search_model_versions`, whose filter grammar has no in-literal quote escaping -- literals pick the quote style the name doesn't contain, and a (pathological) name containing both quote styles fails loudly with guidance to pass an explicit version instead of silently matching nothing.

## Test plan

- 25 new unit tests against a faked `MlflowClient` via `unittest.mock.patch` -- no live MLflow server in CI, per the issue spec: registration/version-increment semantics, tag mapping, run_id native linkage, reserved-tag collisions, schema-ignored, quote-style selection, both error contracts, lazy-import failure message, constructor forwarding, picklability. 100% line and branch coverage on the new module.
- Smoke-tested against a real local `mlflow server` (sqlite backend, mlflow 2.22.5): full-field round-trip, latest-version lookup through the real search grammar (`order_by=["version_number DESC"]`), name isolation, quote-bearing names through the real filter parser, native run linkage verified with a real run, and the KeyError/ValueError contracts.
- Full pusher test suite passes (147 tests); adjacent registry-package suite passes (56 tests); `ruff format` + `ruff check` clean; pyproject parses with the new extra.


sallycr:
**Automated review notes** (advisory only — not a maintainer approval/request-changes, just an independent pass while the PR is under human review)

Read through `mlflow_client.py`, its ABC (`ModelRegistryClient`), and the 25-test suite, and ran the tests locally (`25 passed`). The field mapping, error contracts (`KeyError`/`ValueError`), lazy-import + `pusher-mlflow` extra wiring, and the filter-quoting workaround for `search_model_versions` all check out against the ABC contract and the sibling `APIRegistryClient` implementation.

One correctness issue found that looks like it would block real usage, plus a docs gap:

**`register_model()` is missing the `source_pipeline_run` parameter that the ABC declares** ([`client.py:135-147`](https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/lib/model_manager/registry/client.py)), and that `ModelPusherPlugin` always passes as a keyword argument on every real `push()` call ([`model_plugin.py:489-502`](https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/workflow/tasks/pusher/plugins/model_plugin.py)). Reproduced directly:

```python
>>> client = MLflowRegistryClient()
>>> client.register_model(name="m", artifact_uri="s3://b", source_pipeline_run=None)
TypeError: register_model() got an unexpected keyword argument 'source_pipeline_run'
```

Because Python's `ABCMeta` only checks that a method with the right *name* exists (not its signature), `MLflowRegistryClient` instantiates and passes `isinstance(client, ModelRegistryClient)` fine — but any call through `push(..., registry_client=MLflowRegistryClient(...))` will raise this `TypeError` today. The sibling `APIRegistryClient` shows the right pattern: accept the param and silently ignore it (per the ABC's own documented convention for fields a backend doesn't support, e.g. `schema`).

This also explains why the test suite is green: none of the 25 tests call `register_model()` with `source_pipeline_run` set, or drive the client through `ModelPusherPlugin`/`push()` — they all call `register_model()` directly against the truncated signature the class actually has. Worth adding a test that exercises the real call path (or just passes `source_pipeline_run=`) so this class of bug fails loudly next time.

Smaller note: `docs/operator-guides/integrations/mlflow.md` has a whole section ("MLflow Model Registry vs Michelangelo AI Model Registry") describing the two registries as independent/log-to-both-separately — worth a short update once the above is fixed, since this PR adds an actual bridge between the pusher and MLflow's registry that operators/users won't discover from that page.

Everything else (tag mapping for `labels`/`kind`/`deployable_artifact_uri`/`metadata`, `run_id` native-linkage handling, quote-escaping fallback to `KeyError` when a name can't be filtered, extras scoping in `pyproject.toml`) looked correct and well-tested.

PR #1837, opened 2026-08-17
feat(blobstore): add a native GCS backend

### LINKED ISSUE #1693: Native GCS backend for the blob storage layer
GCS storage support is on the roadmap under planned (infrastructure & compute), and the shipped section currently scopes storage to any s3 compatible object store. GCS does have an s3 interoperability mode but it needs HMAC keys and diverges on multipart semantics and IAM. Teams on native GCS service accounts and workload identity can't use the storage layer cleanly today.

Native GCS backend behind the existing blob storage interface - config plumbing plus helm values mirroring the s3 backend's shape, auth via Application Default Credentials so workload identity works w/o static keys, and unit tests against a GCS fake rather than a live bucket. Additive change: no behavior change for s3 users, backend selected by config

Documenting the s3 interop mode with HMAC keys (works but keeps static credentials around and breaks on multipart edge cases), or running an s3 translation gateway in front of GCS (extra infra every operator has to run)

**Component**
Which part of the project does this affect?
- [ ] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [x] API server / gRPC
- [x] Helm chart / Kubernetes deployment
- [ ] UI (Studio)
- [ ] Documentation
- [ ] Other

**Additional context**
Since this is a roadmap planned line, I want to confirm before writing code
Would you accept PRs for a native GCS backend, or is this being covered by an existing design?

apurvapatkeshwar:
## Summary

Closes #1693.

Adds a Google Cloud Storage backend behind the existing `blobstore.BlobStoreClient` interface, registered for the `gs://` scheme in both the worker and the controller manager (the same wiring `minio.Module` has). Auth defaults to Application Default Credentials, so GKE workload identity works with zero configuration and no static keys; a service account key file, a custom endpoint, and anonymous access are available as config overrides:

```yaml
# gcs:
#   credentialsFile: ""  # optional: path to a service account JSON key; leave empty for ADC
#   endpoint: ""         # optional: private Google access or a local emulator
#   anonymous: false     # optional: public buckets, emulators
```

The package mirrors the minio backend's shape: a `Config` populated from the `gcs` YAML key, a client implementing `Get`/`Scheme`, and an fx module contributing to the `blobstore_clients` group.

## Design notes

- **Lazy client construction.** Unlike `minio.New`, `storage.NewClient` resolves credentials eagerly, so constructing it at fx-provide time would fail startup on any deployment that has no GCP credentials and never reads `gs://` URIs. The storage client is therefore built on the first `gs://` read (`sync.Once`); registering the scheme is free for s3-only users, and the rendered behavior for them is unchanged. This is the "additive change: no behavior change for s3 users" requirement from #1693.
- **Scheme registration follows the interface docs**: `blobstore.go`'s `Scheme()` docstring already names `"gs"` as an expected value.
- Error wrapping and the cross-scheme rejection message follow the existing backends (`scheme %s is not supported by gcs client`), matching the contract minio's tests pin down.

## Dependencies

Adds `cloud.google.com/go/storage v1.64.0` and `google.golang.org/api v0.287.1` (direct), which also moves a few shared indirects forward (notably `google.golang.org/grpc` v1.71.0 -> v1.82.0 and `google.golang.org/protobuf` v1.36.6 -> v1.36.11). `go build ./...` and the full test suite pass at the new versions (details below). MODULE.bazel `use_repo` and BUILD files were regenerated with `bazel mod tidy` + gazelle.

## Testing

- New unit tests, all offline (no GCP credentials, no network beyond localhost):
  - config: YAML-provider round-trips (success / missing key / partial), same shape as the azure and minio config tests.
  - client: scheme, invalid-URL, wrong-scheme (exact error string), missing bucket/object path.
  - credential selection: table-driven over anonymous / anonymous+endpoint / explicit key file (a syntactically valid throwaway service account key generated in the test) / missing key file.
  - **`Get()` round-trips against a local `httptest` fake GCS server** (success and 404), which is the "unit tests against a GCS fake rather than a live bucket" ask in #1693 -- construction uses `option.WithoutAuthentication()` + `option.WithEndpoint`.
- Package coverage: 92.9% of statements.
- `go build ./...` clean; `go test ./...` passes 89 packages. The only failures are 5 pre-existing `[setup failed]` packages that import a Bazel-generated mock path (`mock/github.com/...`) and fail identically on an unmodified main checkout -- they are Bazel-only tests, unrelated to this change.
- Bazel: `bazel test //go/base/blobstore/...` all 4 targets pass (including the new gcs target, under nogo); `bazel build` of the worker and controller manager libraries succeeds.
- `gofmt` clean; no TODOs introduced.


paulzim (approved):
## Review: #1837 — feat(blobstore): add a native GCS backend

**Verdict: Approve**

Adds a `gs://` backend behind the existing `blobstore.BlobStoreClient` interface, closing #1693. Scoped tightly — 14 files, +649/-71, touching only `go/base/blobstore/gcs/`, the `worker`/`controllermgr` fx wiring, `go.mod`/`go.sum`, and `MODULE.bazel`.

**Verified by actually building and running it, not just reading the diff:**
- `go build ./base/blobstore/... ./worker/... ./controllermgr/...` — clean.
- `go test ./base/blobstore/...` — all packages pass, including the new `gcs` package (13 tests: scheme, invalid-URL, wrong-scheme with exact error-string match, missing bucket/object, credential-selection table test, construction-error propagation, and two round-trips against a local `httptest` fake GCS server for success/404).
- `go vet ./base/blobstore/gcs/...` and `gofmt -l` — clean.
- Confirmed no scheme collision: `abfss` (azure), `s3` (minio), `gs` (this PR) are all distinct in the `blobstore_clients` fx group.

**Design is sound and consistent with the existing backends:**
- `Config`/`newConfig`/module shape mirrors `minio` and `azure` exactly (YAML key `gcs`, same fx.Provide pattern).
- Lazy client construction (`sync.Once` in `ensureClient`) is the right call and clearly commented — `storage.NewClient` resolves credentials eagerly, so building it at fx-provide time would break startup for any deployment with no GCP credentials that never reads `gs://`. This keeps registering the module free for S3-only users.
- Credential precedence (explicit key file → anonymous → ADC) and the cross-scheme rejection message match the pattern the minio tests already pin down.
- `base.yaml` documents the new `gcs:` block the same way the existing commented-out `azure:`/`minio:` blocks are documented.

**One non-blocking nit:** the new `cloud.google.com/go/storage`/`google.golang.org/api` deps bump `go/go.mod`'s language version from `go 1.24.0` (+ `toolchain go1.24.3`) to `go 1.25.0` — not mentioned in the PR description's Dependencies section. `.github/workflows/go-lint.yml` (`go-version: '1.24.3'`) and `.github/workflows/release.yaml` (`go-version: '1.24'`) still hard-pin the old version, so CI will silently auto-download a newer toolchain via Go's default `GOTOOLCHAIN=auto` rather than running the version those workflows declare. Not a functional blocker — `api-surface-check.yml` already reads the version from `go.mod` directly and everything builds/tests fine either way — but worth a follow-up to bump those two pins for consistency, especially given the hardening work we've already been doing on CI reproducibility.

No other concerns. Straightforward, well-tested, additive change.

go/base/blobstore/gcs/client.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func newGcsBlobClient(config Config) *gcsBlobClient {
+	return &gcsBlobClient{config: config, scheme: "gs"}
+}
+
+// ensureClient creates the storage client on first use. The client is
+// created with a background context because it outlives the request that
+// triggers its construction.
+func (c *gcsBlobClient) ensureClient() (*storage.Client, error) {
```
uzbq, line 43: why don't we create the client in newGcsBlobClient while injecting the deps? it will catch failure early at the service deployment time.

apurvapatkeshwar: Eager would break startup for every non GCS deployment - `storage.NewClient` resolves ADC at construction and this module loads unconditionally in both binaries
Lazy keeps the gs scheme registered at zero cost for non GCS users
A middle path is "eager when configured" i.e., construct at startup only when the gcs config section is set, stay lazy otherwise - happy to send a small follow up PR

PR #1840, opened 2026-08-17
[throwaway/ci-repro] services.ts drift without regen

craigmarker:
Throwaway PR to verify `transcoder-services-check.yaml` fails when `services.ts` references a new service without regenerating `transcoder-services.json`. Will be closed after CI confirms.


craigmarker:
Closing: opened from a stale local branch ref, diff got polluted with unrelated already-merged history. Re-running this test correctly.

PR #1841, opened 2026-08-17
[throwaway/ci-repro] descriptors.pb drift without regen

craigmarker:
Throwaway PR to verify main.yml's `dirty-check` job fails when a .proto change isn't accompanied by a regenerated `descriptors.pb`/`transcoder-services.json`. Will be closed after CI confirms.


craigmarker:
Closing: opened from a stale local branch ref, diff got polluted with unrelated already-merged history. Re-running this test correctly.

PR #1842, opened 2026-08-17
test: reproduce transcoder-services-check drift (throwaway, will close)

craigmarker:
Throwaway PR to verify `transcoder-services-check.yaml` fails when `services.ts` drifts from `helm/michelangelo/files/transcoder-services.json`. Not for review; will be closed after CI confirms the check fails.


craigmarker:
Confirmed transcoder-services-check.yaml fails on this drift as expected. Closing throwaway PR.

PR #1843, opened 2026-08-17
test: reproduce descriptor-freshness drift (throwaway, will close)

craigmarker:
Throwaway PR to verify main.yml's `dirty-check` job fails when `proto/api/v2/cluster.proto` drifts from `helm/michelangelo/files/descriptors.pb`. Not for review; will be closed after CI confirms the check fails.


craigmarker:
main.yml's dirty-check job never triggered on this PR (base branch restriction) so this repro is inconclusive; see follow-up report. Closing throwaway PR.

PR #543, opened 2025-11-15
Configure baseline code coverage with 90% requirement for new code

sallycr:
## Summary

Implements baseline code coverage thresholds based on current measured coverage and enforces **90% coverage** for all new/changed code across Python, Go, and JavaScript.

Fixes #542

## Coverage Baselines (Existing Code)

| Language | Baseline | Enforcement |
|----------|----------|-------------|
| Python | 34.94% | Must not decrease below baseline |
| Go | 53.1% | Must not decrease below baseline |
| JavaScript | 78.74% | Must not decrease below baseline |

## New Code Requirements

**Coverage Required:** ≥90% for all languages (updated from 85%)
**Enforcement:** Strictly enforced via diff-cover tool in CI
**Scope:** Applies only to new/changed lines in PR

### GitHub Actions Workflows

- Python coverage: Baseline 34%, New code 90%
- Go coverage: Baseline 53%, New code 90%
- JavaScript coverage: Baseline 78%, New code 90%
- Combined coverage: Updated policy display
- Python tests: Updated documentation

### Configuration Files

- Codecov config: Patch coverage target 90%
- Python pyproject.toml: Updated fail_under from 70 to 34
- JavaScript package.json: Vitest v3.2.4 with coverage-v8
- Lock files updated

## How It Works

When you create a PR, CI will:
1. Run all tests for languages with changes
2. Measure overall coverage and verify it meets baseline
3. Calculate diff coverage for new/changed lines only
4. Enforce 90% coverage on new/changed code
5. Comment on PR with detailed coverage report

## Latest Updates

**2025-11-17**: Increased new code coverage requirement from 85% to 90% to align with Python Code Quality improvement initiative and ensure higher quality standards.

## Testing

All workflow configurations tested:
- Python: 64/64 tests passing, 34.94% coverage
- Go: 53/53 tests passing, 53.1% coverage
- JavaScript: 857/857 tests passing, 78.74% coverage

PR #1850, opened 2026-08-18
ci(codecov): mark project and patch status checks informational

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

**What changed?**
Adds `informational: true` to the two existing Codecov status blocks in `.github/codecov.yml` (project and patch). Both checks keep posting their numbers on every PR, but they can no longer turn a PR red.

**Why?**
The status gates in codecov.yml (project: auto baseline with 1% threshold, patch: 90% target) have existed since #543, but no workflow uploaded coverage until #1834 merged - so they have been dormant for their entire life and were never calibrated against real data. Now that uploads flow, they go live on every PR with a strict, never-exercised bar. Marking them informational surfaces the same numbers without gating merges; once there is enough history to trust the thresholds, flipping them back is a 2-line change.

**How did you test it?**
- YAML parses cleanly.
- Config accepted by Codecov's own validator (`curl --data-binary @.github/codecov.yml https://codecov.io/validate` returns Valid).

**Potential risks**
Coverage regressions will not block merges while the checks are informational - that is the intended trade-off until the thresholds are calibrated.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
None.

**Documentation Changes**
None.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -9,6 +9,7 @@ coverage:
         target: auto          # Use baseline from main branch
         threshold: 1%         # Allow 1% drop from baseline
         if_ci_failed: error
+        informational: true   # Report-only until the check has real history
 
     # Coverage for changes in PR (strictly enforced)
     patch:
@@ -16,6 +17,7 @@ coverage:
         target: 90%           # New code must have 90% coverage
         threshold: 0%         # No tolerance for drops in new code
         if_ci_failed: error
+        informational: true   # Report-only until the 90% bar is calibrated
 
   precision: 2
   round: down


paulzim (approved):
## Review: #1850 — ci(codecov): mark project and patch status checks informational

**Verdict: Approve**

Minimal, well-scoped config change — verified the diff is exactly what the description claims and nothing more:

```diff
         target: auto          # Use baseline from main branch
         threshold: 1%         # Allow 1% drop from baseline
         if_ci_failed: error
+        informational: true   # Report-only until the check has real history
...
         target: 90%           # New code must have 90% coverage
         threshold: 0%         # No tolerance for drops in new code
         if_ci_failed: error
+        informational: true   # Report-only until the 90% bar is calibrated
```

Confirmed:
- `.github/codecov.yml` parses as valid YAML on the PR head commit.
- The `project`/`patch` thresholds being disabled here are real — `go-coverage.yml`, `python-coverage.yml`, and `javascript-coverage.yml` all upload to Codecov via `codecov/codecov-action`, so this config is live and does gate real PRs, not dead config.
- No other file in the repo duplicates or overrides these thresholds.
- This lines up with what I found independently while re-checking #1837 (native GCS backend) yesterday: the `codecov/patch` check was already reporting `SUCCESS` regardless of the comment's `:x:` icon, and there was no `codecov/project` gate active yet. This PR is the deliberate reason why — it keeps both checks non-blocking until there's enough history on `main` to trust the 90%/1% bars, rather than leaving them to go live un-calibrated (which is exactly the gap #1834 opened up). Good proactive catch.
- CI green on the head commit (`57bdacf9`), including `dirty-check` and `license/cla`.

No concerns — this is exactly the kind of low-risk, easily-reversible ("flip back is a 2-line change") config change it's described as.

PR #1852, opened 2026-08-18
fix(auth): enforce authn/authz on generated Get and List handlers

### LINKED ISSUE #1851: Generated Get/List handlers never call the Auth interface
**Describe the bug**

The `kubeyarpc` codegen template (`go/kubeproto/templates/crd_svc.tmpl`) emits an authN + authZ block in every generated Create/Update/Delete/DeleteCollection handler, but the Get and List handler templates contain no auth calls at all. This looks like an omission rather than a decision: the `Action` enum in `go/auth/auth.go` has always defined `Get` and `List` as first class values alongside the four enforced actions - they just never got wired into the template.

**To Reproduce**

1. `grep -c "UserAuthorized" proto-go/api/v2/*.pb.kubeyarpc.go` - every one of the 15 generated services shows exactly 4 call sites (four mutating handlers), never in `Get*`/`List*`
2. Inspect the `Get{{.KindName}}` / `List{{.KindName}}` functions in `go/kubeproto/templates/crd_svc.tmpl` - no `UserAuthenticated`/`UserAuthorized` calls
3. Wire a real `Auth` implementation (anything other than `DummyAuth`) and call any Get/List endpoint: the request is served without any authZ check

**Expected behavior**

Every generated handler consults `Auth` interface; Get and List use `authapi.Get` / `authapi.List` as the action, the way the 4 mutating handlers use theirs

**Actual behavior**

Generated Get/List handlers never call `UserAuthenticated` or `UserAuthorized` - mutating endpoints get enforced while every read endpoint silently bypasses authorization, with no signal short of reading generated code

**Environment**
- OS: N/A (code-level, all platforms)
- Applies to current `main` and every release since the template was introduced

**Component**
Which part of the project is affected?
- [ ] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [x] API server / gRPC
- [ ] UI (Studio)
- [ ] Documentation
- [x] Other - kubeproto codegen (`go/kubeproto`) and the committed generated services in `proto-go`

**Additional context**

Impact in the default deployment is none: the only shipped `Auth` implementation is `DummyAuth`, which returns true for every request, so as configured today no endpoint enforces anything and Get/List behave exactly like the rest. The gap matters for any operator who plugs in a real `Auth` implementation (E.g., RBAC, OIDC etc.) expecting it to be consulted for every action the enum defines.

Proposed fix: Add the same two call auth block to Get/List template functions (`authapi.Get` / `authapi.List`, `request.Namespace` as project arg, matching Delete handlers) and regenerate the committed `proto-go` snapshot with `tools/gen-proto-go.sh`
Behavior preserving as shipped, since `DummyAuth` ignores the `Action` argument
1 open semantic question for real implementations: A cluster-wide List (empty namespace) passes an empty project string to `UserAuthorized` - same contract DeleteCollection already has, so the fix keeps that consistent rather than inventing new semantics, but it is worth a note in the Auth interface docs?

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

- `go/kubeproto/templates/crd_svc.tmpl`: the generated `Get*` and `List*` handlers now carry the same authentication + authorization block as Create/Update/Delete/DeleteCollection, using `authapi.Get` / `authapi.List` as the action and `request.Namespace` as the project (matching the Delete handlers, which also authorize from the request's namespace).
- The audit-log `defer` in Get/List moves ahead of the auth checks, so denied requests are audit-logged -- the ordering the four mutating handlers already have.
- `proto-go/api/v2/*.pb.kubeyarpc.go`: regenerated with `tools/gen-proto-go.sh` -- all 15 generated services change uniformly, no hand edits. Every service now has 6 auth call-sites (was 4).

**Why?**

Fixes #1851. The codegen template emits an authn + authz block in every mutating handler, but Get and List handlers were generated with no auth calls at all -- even though the `Action` enum in `go/auth/auth.go` has always defined `Get` and `List` as first-class values. With the shipped `DummyAuth` this is inert; for any operator wiring a real `Auth` implementation, every read endpoint silently bypassed authorization.

**How did you test it?**

- `tools/gen-proto-go.sh` run end-to-end: the diff is exactly the template plus the 15 generated service files, uniformly, gofmt-formatted by the real pipeline -- confirming no other drift against the templates.
- `bazel build //go/cmd/apiserver:apiserver` succeeds with the regenerated handlers; `bazel test //go/kubeproto/... //go/auth/...` 8/8 pass; `bazel test //go/cmd/kubeproto/...` (the generator's own tests) 4/4 pass.
- Re-validated after rebasing onto current main: `go build ./...` in `proto-go`, `go build ./cmd/apiserver`, and `go test ./kubeproto/... ./auth/... ./cmd/kubeproto/...` all pass. No codegen inputs (protos, templates, generator, `tools/gen-proto-go.sh`) changed upstream since the snapshot was regenerated, so it is exact at HEAD.
- Grep verification: `UserAuthenticated`/`UserAuthorized` now appear in all six handler kinds in every generated service, and only there.

**Potential risks**

Behavior-preserving as shipped: `DummyAuth.UserAuthorized` ignores its `Action` argument and returns true, so responses are unchanged until an operator supplies a real `Auth` implementation. Once a real implementation is wired, read endpoints start enforcing authorization -- that is the point of the fix, but an implementation that never expected `Get`/`List` actions will now receive them. Semantics note: a cluster-wide List (empty `request.Namespace`) passes an empty project string to `UserAuthorized` -- the same contract DeleteCollection already has; this PR keeps that consistent rather than inventing new semantics.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Generated Get/List handlers now consult the `Auth` interface (`authapi.Get` / `authapi.List`). No behavior change under the default `DummyAuth`.

**Documentation Changes**

None -- the `Auth` interface itself is unchanged; a note on empty-project semantics for cluster-wide List could go in the interface docs as a follow-up.

### DIFF
--- a/go/kubeproto/templates/crd_svc.tmpl
+++ b/go/kubeproto/templates/crd_svc.tmpl
@@ -236,14 +236,6 @@ func (c {{.LowerKindName}}ServiceHandler) Get{{.KindName}}(
 		logging.EntityNameTag:   request.Name,
 	})
 
-	getOptions := &metav1.GetOptions{}
-	if request.GetOptions != nil {
-		getOptions = request.GetOptions
-	}
-
-	result := &{{.KindName}}{}
-	err = c.apiHandler.Get(ctx, request.Namespace, request.Name, getOptions, result)
-
 	defer c.auditLogEmitter.Emit(ctx, c.build{{.KindName}}AuditLogEventForGet(
 												ctx,
 												request,
@@ -251,6 +243,32 @@ func (c {{.LowerKindName}}ServiceHandler) Get{{.KindName}}(
 												err,
 												),
 				)
+
+	// Authentication
+	userAuthenticated, err := c.auth.UserAuthenticated(ctx)
+	if !userAuthenticated {
+		logger.Error(err, "User is not authenticated")
+		metric.Counter("unauthenticated").Inc(1)
+		return nil, err
+	}
+
+	// Authorization
+	projectName := request.Namespace
+	userAuthorized, err := c.auth.UserAuthorized(ctx, projectName, authapi.Get, "{{.KindName}}")
+
+	if !userAuthorized {
+		logger.Error(err, "User is not authorized")
+		metric.Counter("unauthorized").Inc(1)
+		return nil, err
+	}
+
+	getOptions := &metav1.GetOptions{}
+	if request.GetOptions != nil {
+		getOptions = request.GetOptions
+	}
+
+	result := &{{.KindName}}{}
+	err = c.apiHandler.Get(ctx, request.Namespace, request.Name, getOptions, result)
 
 	if err != nil {
 		logger.Error(err, "Cannot get {{.KindName}} request info from k8s/ETCD", "error", err, "api_msg_tag", "Get{{.KindName}}")
@@ -516,6 +534,32 @@ func (c {{.LowerKindName}}ServiceHandler) List{{.KindName}}(
 		logging.NamespaceTag:    request.Namespace,
 	})
 
+	defer c.auditLogEmitter.Emit(ctx, c.build{{.KindName}}AuditLogEventForList(
+												ctx,
+												request,
+												resp,
+												err,
+												),
+				)
+
+	// Authentication
+	userAuthenticated, err := c.auth.UserAuthenticated(ctx)
+	if !userAuthenticated {
+		logger.Error(err, "User is not authenticated")
+		metric.Counter("unauthenticated").Inc(1)
+		return nil, err
+	}
+
+	// Authorization
+	projectName := request.Namespace
+	userAuthorized, err := c.auth.UserAuthorized(ctx, projectName, authapi.List, "{{.KindName}}")
+
+	if !userAuthorized {
+		logger.Error(err, "User is not authorized")
+		metric.Counter("unauthorized").Inc(1)
+		return nil, err
+	}
+
 	result := &{{.KindName}}List{}
 	listOptions := &metav1.ListOptions{}
 	if request.ListOptions != nil {
@@ -526,14 +570,6 @@ func (c {{.LowerKindName}}ServiceHandler) List{{.KindName}}(
 		listOptionsExt = request.ListOptionsExt
 	}
 	err = c.apiHandler.List(ctx, request.Namespace, listOptions, listOptionsExt, result)
-
-	defer c.auditLogEmitter.Emit(ctx, c.build{{.KindName}}AuditLogEventForList(
-												ctx,
-												request,
-												resp,
-												err,
-												),
-				)
 
 	if err != nil {
 		logger.Error(err, "API Handler failed to List {{.KindName}}", "error", err, "api_msg_tag", "List{{.KindName}}")
--- a/go/kubeproto/templates/templates_test.go
+++ b/go/kubeproto/templates/templates_test.go
@@ -83,3 +83,62 @@ func TestTemplates(t *testing.T) {
 	assert.NotContains(t, crdSvcHandlerCode, "UnifiedLogInfo")
 	assert.NotContains(t, crdSvcHandlerCode, "UnifiedLogError")
 }
+
+// TestCrdSvcHandlerAuthEnforcement locks the auth contract of the generated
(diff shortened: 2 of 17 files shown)


paulzim (approved):
## Review: #1852 — fix(auth): enforce authn/authz on generated Get and List handlers

**Verdict: Approve**

This is a real, correctly-diagnosed bug fix (#1851): the `kubeyarpc` codegen template wired `UserAuthenticated`/`UserAuthorized` into Create/Update/Delete/DeleteCollection but never into Get/List, even though `Get`/`List` have always been first-class values in the `Action` enum (`go/auth/auth.go`). Under the shipped `DummyAuth` this was inert, but any operator plugging in a real `Auth` implementation would have every read endpoint silently skip authorization.

Verified beyond the description:

- **Template diff is exactly the claimed change** — `go/kubeproto/templates/crd_svc.tmpl`'s Get/List functions now carry the same authn/authz block as Delete (`authapi.Get`/`authapi.List`, `request.Namespace` as project), and the audit-log `defer` moves ahead of the auth checks. I compared this against the existing `DeletePipeline` handler on `main` and confirmed the new Get/List ordering (audit-log defer → authn → authz → operation) is the exact same shape Delete has had all along — this isn't a new pattern, it's Get/List catching up to it.
- **Uniform across all 15 services, nothing missed or double-applied.** `grep -c UserAuthorized proto-go/api/v2/*.pb.kubeyarpc.go` shows exactly 6 call sites (was 4) in all 15 CRUD handler files (`cached_output`, `cluster`, `deployment`, `evaluation_report`, `inference_server`, `model`, `model_family`, `pipeline`, `pipeline_run`, `project`, `ray_cluster`, `ray_job`, `revision`, `spark_job`, `trigger_run`), and 0 in the `_svc.pb.kubeyarpc.go` registration files and non-CRUD message files (`chart`, `job`, `pod`, `user`, etc.), which never had handler bodies to begin with — matches the PR's claimed scope exactly, no stragglers.
- **Enum reuse, not a new surface.** `Get`/`List` were already declared in `go/auth/auth.go`'s `Action` enum; this PR wires existing values into existing template functions, no new API.
- **Builds and tests pass locally**, not just per the PR's own report: `go build ./cmd/apiserver/...` succeeds, and `go test ./kubeproto/... ./auth/... ./cmd/kubeproto/...` is green.
- **Empty-namespace semantics for cluster-wide List** are called out honestly in the description as matching DeleteCollection's existing (also empty-project) contract rather than inventing something new — confirmed this is consistent with what `DeleteCollection`'s handler already does.
- CI green on the head commit (`4e033381`), including `dirty-check`, which would have caught any drift between the regenerated snapshot and a live `tools/gen-proto-go.sh` run.

**One minor, non-blocking observation:** `go/kubeproto/templates/templates_test.go`'s `TestTemplates` only asserts that `GetTestName`/`ListTestName` exist in the rendered template output, not that they contain the auth block. Given the bug being fixed here is specifically "codegen silently omits an auth call with no way to notice short of reading generated code" (the issue's own words), adding one assertion like `assert.Contains(t, crdSvcHandlerCode, "UserAuthorized(ctx, projectName, authapi.Get")` would close the exact gap that let this slip through originally and catch a future regression automatically. Worth a fast-follow, not a blocker for this fix.

apurvapatkeshwar:
@sallycr @paulzim Would appreciate fork CI clicks when you get a chance on these PRs -
#1836, #1852, #1920, #1927, #1977, #1979, #1981, #1982, #1983, #1984, #1985, #1988, #1989, #1990, #1991, #1992

Several got fresh pushes today (#1852, #1920, #1836, #1984, #1985 fixed, #1927 fixed coverage gap), so the list grew rather than shrank

Also #1978 and #1980 are green and approved, just need a merge click, thanks!

apurvapatkeshwar:
@paulzim your approval from two weeks ago still stands, but a later commit means the CI run needs a fresh approve-and-run
It also gates the 5 PR auth stack behind it (#1988–#1992)

PR #1853, opened 2026-08-18
docs: add a migration guide for Kubeflow Trainer users

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Adds `docs/user-guides/migration/migrate-from-kubeflow-trainer.md`, the second guide in the migration section started by #1783 (both were asked for in #1705). It is aimed at teams running `PyTorchJob`/`TFJob`/`TrainJob` who want to know exactly what changes: training-job CRDs are replaced by Uniflow tasks with a `RayTask` resource declaration, and distribution comes from the Ray Train-based trainers (`LightningTrainer`, `XGBoostTrainer`). Follows the same skeleton the KubeRay guide established: concept-mapping table, a real before/after (PyTorchJob YAML to a pipeline task), what you gain, and an explicit "what does not map yet" section.

**Why?**

#1705 asked for migration guides from Kubeflow Trainer and KubeRay. #1783 covers KubeRay; this covers Kubeflow Trainer. Teams evaluating a move currently have to reverse-engineer the trainer library and the Ray plugin to learn that (for example) there is no TensorFlow trainer and no gang scheduling -- this states it plainly.

**How did you test it?**

Docs-only change; testing here means accuracy and link integrity:

- Every capability claim was verified against the code before writing, and re-verified against current main before posting: the example uses the actual `LightningTrainerParam` field names (`create_model_fn`, `create_model_fn_kwargs`, `train_data`, `val_data`, `batch_size`) rather than the incorrect names some existing docs carry (see #1787); the gaps section (no training-job CRD equivalent, no TensorFlow trainer, fixed-size clusters on both paths, no gang scheduling, Ray pinned to 2.3.1 on the task path) matches the plugin and jobs-client behavior at HEAD.
- All relative links and heading anchors resolve against this branch's docs tree (the KubeRay guide link resolves because of the stack).
- Filename, frontmatter, admonitions, and link style follow `docs/contributing/documentation-guide.md`.
- The docs build was not run locally (no bun on this machine) -- flagging so reviewers know to lean on the docs CI job.

**Potential risks**

None at runtime -- documentation only. The main risk is factual drift if the trainer library or Ray plugin changes; the gaps section states current behavior plainly rather than implying anything is scheduled, so it fails safe.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Docs only; nothing operational.

**Documentation Changes**

This PR is the documentation change: one new page under `docs/user-guides/migration/`.

### DIFF
--- a/docs/user-guides/migration/migrate-from-kubeflow-trainer.md
+++ b/docs/user-guides/migration/migrate-from-kubeflow-trainer.md
@@ -1,0 +1,138 @@
+---
+sidebar_position: 2
+sidebar_label: "From Kubeflow Trainer"
+---
+
+# Migrating from Kubeflow Trainer
+
+This guide is for teams running distributed training through Kubeflow Trainer — `PyTorchJob`, `TFJob`, or the v2 `TrainJob` — who are evaluating what changes if they adopt Michelangelo AI.
+
+Unlike the [KubeRay migration](./migrate-from-kuberay.md), this one is a replacement rather than a re-layering. Michelangelo AI has no training-job CRD: there is no resource where you declare per-role replica specs and hand the operator a container to run. Training here is a Python function inside a pipeline, and distribution comes from Ray Train rather than from a framework-specific operator. That is a real authoring change, and this page tries to be precise about what you gain and what you give up.
+
+:::note
+This page maps **training jobs**. Kubeflow Pipelines, Katib, and KServe are a different surface — for those, see the [glossary's concept mapping](../../getting-started/glossary.md#concept-mapping) and [ML Workflow Mapping](../../getting-started/overview.md#ml-workflow-mapping) in the overview.
+:::
+
+## What changes
+
+A `PyTorchJob` bundles three decisions into one manifest: what code runs, how many replicas of each role run it, and what pod each replica gets. Here those decisions move:
+
+- **What runs** becomes a Uniflow task — a Python function in a pipeline.
+- **How it distributes** becomes a Ray Train `ScalingConfig` passed to the bundled Lightning trainer, which subclasses Ray Train's `TorchTrainer`.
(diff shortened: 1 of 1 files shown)


docs/user-guides/migration/migrate-from-kubeflow-trainer.md
```diff
@@ -0,0 +1,138 @@
...
+
+There is no `Master`/`Worker` role split anywhere in that. Ray Train coordinates workers internally, and rank 0 is a runtime concept rather than a differently-shaped Kubernetes object.
+
+## Concept mapping
+
+| Concept | Kubeflow Trainer | Michelangelo AI |
+|---|---|---|
+| **Training job resource** | `PyTorchJob` / `TFJob` / `TrainJob` (CRDs) | None — a Uniflow task with a `RayTask` config |
+| **Distributed PyTorch** | `PyTorchJob` replica specs | `LightningTrainer`, a Ray Train `TorchTrainer` subclass |
+| **Distributed XGBoost** | `XGBoostJob` | `XGBoostTrainer`, a Ray Train `XGBoostTrainer` subclass |
```
andriika, line 32: I believe Michelangelo doesn't currently provide an `XGBoostTrainer` (a subclass of Ray Train's `XGBoostTrainer`), unless I'm missing something.

Could you please double-check that?

apurvapatkeshwar: Double checked - it does exist;  it only landed last week in #1793 so easy to miss 
`XGBoostTrainer` in `michelangelo.lib.trainer.xgboost` (subclass of `ray.train.xgboost.XGBoostTrainer`)
Ref: https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/lib/trainer/xgboost/xgboost_trainer.py#L157

PR #1858, opened 2026-08-18
feat(ui): support pipeline run resume from + retry

npuppala111:
## Summary

Two related additions to the pipeline run UI, both built on the same `spec.resume` mechanism.

**1. Resume from a chosen step** — a collapsible "Select run to resume from" group in the Start new pipeline run dialog. Picking a source run loads that run's DAG tasks into a multi-select whose rows show step name, start, end, duration and state.

- The run picker is filtered **client-side** to finished runs of the same pipeline. The label selector the pipeline detail page uses (`pipeline.michelangelo/name`) is written by nothing in the platform, and `ListOptionsExt` field criteria are silently dropped when metadata storage is disabled — either would widen the picker to other pipelines' runs rather than narrowing it.
- `resumeFrom` submits sub-step **`displayName`**, never `name`. The platform keys its task cache by display name, so submitting the task path (`examples.bert_cola.data.load_data` instead of `load_data`) would silently reuse the cache for the very step the user asked to re-run. There is a test guarding this specifically.
- `resumeUpTo` and `forceResume` exist on the proto but are read by nothing in Go, the Python bindings, or mactl, so they are deliberately not surfaced.
- Adds a `getOptionContent` prop to `SelectField` for rich dropdown rows; selected values still render the plain label so multi-select tags stay compact.

**2. Retry a run from its detail page** — a primary Retry button in the run detail header (and a row action in the runs list, which entity actions get for free).

Retry is resume-shaped rather than a reset: instead of rewinding the original workflow in place the way `spec.retryInfo` and the per-step Retry button do, it creates a **new** run whose `spec.resume` points at the run being retried, so every already-succeeded task is served from cache and execution picks up at the first one that did not. Confirm dialog → new run → toast with a **See new run** link.

- Middleware reshapes the source run into the create payload. Ordering matters: `spec.resume` is written from `metadata.name` *before* that field is replaced, and `metadata` is rebuilt wholesale because the API rejects a create carrying `uid`/`resourceVersion`. An inherited `resumeFrom` is dropped, so retrying an already-resumed run reuses its cached successes instead of re-forcing those steps.
- Gated to failed and killed runs — a succeeded run has everything cached, so resuming it would do no work. Disabled state carries a tooltip.

**Incidental fix:** the merge resolution on this branch had kept the wrong side of `select/types.ts`, deleting `SelectFieldConfig` and its Single/Multi variants (imported by `schema-select-field.tsx` and `form/types/config-types.ts`) and dropping `getOptionContent`, leaving only an unused `ReactNode` import. `packages/core` did not typecheck. Restored in its own commit.

## Test plan

- [x] `yarn typecheck`, `yarn lint --quiet`, `yarn format` clean for `packages/core`
- [x] `yarn test` — 147 files, 1359 tests pass
- [x] Resume: step picker offers only finished runs of the same pipeline; steps populate from `Execute Workflow` sub-steps; `resumeFrom` submits `displayName` (mutation-tested — swapping to `name` fails the suite)
- [x] Retry: payload verified on the wire — fresh `metadata`, `spec.resume.pipelineRun` → source run, no `resumeFrom`, no `status`/`typeMeta`/`uid`/`resourceVersion`; ordering and the identity-transformation requirement both mutation-tested
- [x] End-to-end in the sandbox: resumed run re-ran `load_data` (90s) and continued to `train`, reaching Succeeded; Retry produced a new run and **See new run** navigated to it; Retry disabled with tooltip on a succeeded run; 0 console errors (http://localhost:5173/ma-dev-test/train/runs/run-20260818-220117-e34dbcd0/steps)

Known gap: the **Pipeline Runs tab on a pipeline's detail page** still shows "No data" — that is the pre-existing `pipeline.michelangelo/name` label-selector bug described above, present before this PR and not fixed here.

Resume from:
Screenshot 2026-08-18 at 1 20 10 PM
Screenshot 2026-08-18 at 1 20 24 PM

Screenshot 2026-08-18 at 1 20 31 PM
Screenshot 2026-08-18 at 1 20 34 PM

Retry: 
Screenshot 2026-08-18 at 3 00 57 PM
Screenshot 2026-08-18 at 3 01 05 PM
Screenshot 2026-08-18 at 3 01 18 PM


javascript/packages/core/config/entities/pipeline/resume-run-fields.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const handleResumeGroupToggle = (isExpanded: boolean) => setExpanded(isExpanded);
+
+  const { input: sourceRunInput } = useField<string>(SOURCE_RUN_FIELD);
+  const sourceRunName = sourceRunInput.value;
+
+  const { data: runListData, isLoading: isLoadingRuns } = useStudioQuery<ListPipelineRunResponse>({
+    queryName: 'ListPipelineRun',
+    serviceOptions: {},
+    clientOptions: { enabled: expanded },
+  });
```
donovanlopez, line 90: Since we don't provide any filters here, we end up fetching all pipeline runs. Can we add a criterion operation  to match the `pipeline_name` like:

```
const CRITERION_OPERATOR_EQUAL = 1;

function buildPipelineCriterion(pipelineName: string) {
  return {
    listOptionsExt: {
      operation: {
        criterion: [
          {
            fieldName: 'pipeline_run.pipeline_name',
            operator: CRITERION_OPERATOR_EQUAL,
            matchValue: anyPack(
              StringValueSchema,
              create(StringValueSchema, { value: pipelineName })
            ),
          },
        ],
      },
    },
  };
}




(in ResumeRunFields)

const { data: runListData, isLoading: isLoadingRuns } = useStudioQuery<ListPipelineRunResponse>({
  queryName: 'ListPipelineRun',
  serviceOptions: buildPipelineCriterion(pipelineName),
  clientOptions: { enabled: expanded },
});



```

donovanlopez (approved):

Issue #1692, opened 2026-08-05
Observability starter pack: grafana dashboards, prometheus alert rules, per job scrape config


apurvapatkeshwar:
**Is your feature request related to a problem? Please describe.**
Roadmap lists several planned items that together amount to an out of the box observability layer:
- Dashboard management via OSS Grafana operator and Prometheus based alerting for decommission gating (alerting & monitoring)
- Prometheus ConfigMap auto-creation per job (distributed training)
- Automatic rollback on alert firing (model deployment)

The last one needs alert rules to exist before anything can gate on them. Today an operator gets metrics endpoints but has to hand build every dashboard and alert rule. Sandbox already runs grafana (#1623), but that's dev tooling. It isn't purely a content gap - `FederatedClient.CreatePromConfigMap` in `go/components/jobs/client/client.go` already builds a per-job prometheus `ConfigMap`, but it has no prod callers and a hardcoded target, so item 3 is "wire up or replace" rather than greenfield

**Describe the solution you'd like**
`docs/operator-guides/operations/monitoring.md` (from #1040) already specifies a `ServiceMonitor`, five `PrometheusRule` alerts and a Grafana panel layout in prose. The intent below is to turn that into installable artifacts rather than invent new alert content, so items 1, 2 and 4 would follow that doc's thresholds and panel list.

 `docs/operator-guides/operations/monitoring.md` (from #1040) already specs a `ServiceMonitor`, 5 alert rules and a grafana panel layout in prose - the idea below is to turn that into installable artifacts, not invent new content. I was thinking of contributing this as 3-4 small PRs -
1. `GrafanaDashboard` CRs (grafana operator format) for control plane and job/deployment views, behind a helm toggle that defaults off
2. `PrometheusRule` starter pack for job failures, reconcile errors and serving health, sized so the future rollback gate can consume it
3. Per job prometheus scrape ConfigMap generation named on the roadmap
4. Operator docs tying it together

Everything optional and CRD gated (skipped when the grafana/prometheus operators are absent), vendor neutral.

**Describe alternatives you've considered**
Shipping raw dashboard json instead of operator CRs (works w/o the operator but drifts from the roadmap's stated direction), or leaving dashboards to each operator to hand build (status quo, high setup cost per install)

**Component**
Which part of the project does this affect?
- [ ] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [x] Helm chart / Kubernetes deployment
[...]

PR #1865, opened 2026-08-19
helm: add optional CRD-gated ServiceMonitor for the controller manager

apurvapatkeshwar:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

First slice of the observability starter pack (#1692): a `monitoring.enabled` toggle (default off) that creates a `ServiceMonitor` for the controller manager's `/metrics` endpoint -- the only component that exposes one today.

```yaml
monitoring:
  enabled: true
  serviceMonitor:
    interval: 30s
    additionalLabels: {}   # e.g. to match a Prometheus serviceMonitorSelector
```

The resource is gated on `.Capabilities.APIVersions.Has "monitoring.coreos.com/v1"`, so enabling the toggle on a cluster without the Prometheus Operator is a safe no-op -- matching the issue's "skipped when the operators are absent" intent. A validation guard fails fast if `monitoring.enabled=true` with `controllermgr.enabled=false`.

Two drive-by fixes in the monitoring guide, since this PR ships the exact resource it describes:

- the ServiceMonitor example used an `app: michelangelo-controllermgr` label that doesn't match the chart's real selector labels (`app.kubernetes.io/name` / `instance` / `component`) -- corrected, and the guide now leads with the chart-managed path
- the health-probe port is `8081` (chart default `controllermgr.healthPort`), not `8083`

Follow-up PRs in this series: PrometheusRule starter alerts, GrafanaDashboard CRs, and a docs pass. The per-job scrape ConfigMap named in the issue (item 3) is deferred -- it needs `FederatedClient.CreatePromConfigMap` call-site work, which is a code change rather than installable YAML.

**Why?**

#1692 lays out the gap: the chart deploys the platform with zero scrape configuration, alerts, or dashboards, and the monitoring guide asks operators to hand-write resources (with examples that do not match the chart's labels). This starts closing that gap with the smallest installable piece, wired the same way other optional chart features are.

**How did you test it?**

- `helm lint` clean; all checks re-run on top of current main before posting.
- Default render is byte-identical to main (verified by diffing full `helm template -f values-k3d.yaml` output against a clean main worktree) -- strictly additive.
- Rendered ServiceMonitor's `selector.matchLabels` verified to exactly match the controllermgr Service's labels, and the `metrics` endpoint port name matches the Service port.
- New CI checks in `helm-lint.yaml`: monitoring-enabled render with `--api-versions monitoring.coreos.com/v1` asserts the resource renders; without it, asserts it is skipped; negative test for the controllermgr guard. All three verified locally with the exact CI invocations.

**Potential risks**

- None for existing deployments: the toggle defaults to off and the default render is byte-identical to main.
- The CRD gate uses `.Capabilities.APIVersions`, which is populated from the cluster on real installs but empty under bare `helm template`; anyone rendering manifests offline for a cluster that has the Prometheus Operator needs `--api-versions monitoring.coreos.com/v1`. The values comment says so.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Added an optional `monitoring.enabled` chart toggle that creates a ServiceMonitor for the controller manager when the Prometheus Operator CRDs are present. Default off; no action needed for existing deployments. Also corrected the label selector and health-probe port in the monitoring guide's examples.

**Documentation Changes**

`docs/operator-guides/operations/monitoring.md` now leads with the chart-managed ServiceMonitor path and its examples match the chart's real labels and ports. `helm/michelangelo/README.md` documents the new values.

### DIFF
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -111,6 +111,30 @@ jobs:
             --set temporal.server.config.persistence.visibility.sql.password=root \
             2>/dev/null
 
+      - name: Template — monitoring enabled (ServiceMonitor renders with CRDs present)
+        run: |
+          helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --api-versions monitoring.coreos.com/v1 \
+            --debug \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Template — monitoring enabled without CRDs (must skip gracefully)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Validate fail-fast guard (monitoring without controllermgr must fail)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --set controllermgr.enabled=false \
+            2>/dev/null
+
       - name: Template — Ingress enabled (UI + Envoy, hosts[] syntax)
         run: |
           helm template michelangelo helm/michelangelo \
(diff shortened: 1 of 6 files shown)


apurvapatkeshwar:
closing - the whole monitoring stack landed on main via #1922's squash, the ServiceMonitor template here is already there byte-for-byte

PR #1868, opened 2026-08-19
Add pipeline-run notification setup to Run Pipeline dialog

yangfan0328:
## Summary
- Adds a **"Set Up Notifications (Optional)"** section to `CreatePipelineRunForm`, the dialog opened by the "Run Pipeline" button — reworked to match the internal reference implementation ([`uber-code/web-code`](https://github.com/uber-code/web-code)) instead of the original per-notification type/event-type picker.
- The section opens with a single toggle: **"Do you want to receive notifications when pipeline run completed?"** It's a form-only field (never sent to the backend), mirroring the `virtual: true` marker on the equivalent field internally.
- Once enabled, two lists appear: **Emails** and **Slack Channels or Users**. Either or both can be filled in.
- There is no event-type picker. Every notification the form produces covers **every implemented pipeline-run event** (Started, Succeeded, Failed, Killed, Skipped) — matching the requirement that this be a fixed default rather than a user choice.
- On submit, the form builds up to two `Notification` payloads — one `NOTIFICATION_TYPE_EMAIL` if the Emails list has any non-empty entries, one `NOTIFICATION_TYPE_SLACK` if the Slack list does — since the API's `Notification` message (`PipelineRunSpec.notifications`, field 11) is still per-type. `resource_type` is fixed to `RESOURCE_TYPE_PIPELINE_RUN` and isn't shown as a field.
- Backed entirely by the existing `PipelineRunSpec.notifications` proto field — no backend changes needed.
- Toggle starts off and nothing is sent unless a contributor opts in.

## Test plan
- [x] `tsc -b packages/core` — clean
- [x] `eslint --quiet` on touched files — clean
- [x] `prettier` — clean
- [x] `vitest run packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx` — 7/7 passing, covering: toggle-off sends no notifications, email-only submit covers every event type, email+Slack both submit as separate notifications, and a destination type left blank is omitted even with the toggle on
- [x] `yarn test:core` — full suite, 1344/1344 passing
- [x] start local sandbox and confirm "set up notification" section created successfully 
Screenshot 2026-08-22 at 7 20 27 PM

Screenshot 2026-08-22 at 7 19 56 PM

https://github.com/user-attachments/assets/6986a6d4-7d83-4711-95b1-7f038bdf1bed


donovanlopez:
Updated PR to use multi string component for email and slack form fields:

https://github.com/user-attachments/assets/be2ab613-da07-4848-a337-c00ac811d9dc

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -54,6 +82,74 @@ export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<
...
+        <Field name="notifyOnCompletion" subscription={{ value: true }}>
+          {({ input }) =>
+            input.value ? (
+              <>
+                <StringField
+                  name="notificationEmails"
+                  label="Emails"
+                  multi
+                  validate={validateEmails}
+                  placeholder="name@example.com"
```
craigmarker:
```suggestion
                  placeholder="e.g., name@example.com"
```

javascript/packages/core/config/entities/pipeline/create-pipeline-run-form.tsx
```diff
@@ -54,6 +82,74 @@ export const CreatePipelineRunForm = ({ record, onClose }: ActionComponentProps<
...
+                />
+
+                <StringField
+                  name="notificationSlackDestinations"
+                  label="Slack Channels or Users"
+                  multi
+                  placeholder="#channel or @user"
```
craigmarker:
```suggestion
                  placeholder="e.g., #channel or @user"
```

craigmarker (changes requested):
need to fix placeholders to use `e.g.,` to delineate between "this is filled in" and "this is an example"

craigmarker (approved):
need to fix placeholders to use `e.g.,` to delineate between "this is filled in" and "this is an example"

PR #1870, opened 2026-08-19
Schedule pipelines from triggers, and link pipelines, triggers, and their runs together

yangfan0328:
## User can schedule a pipeline to run regularly

- Adds a **"Run trigger"** action on Pipeline (next to Run/Delete) that opens a form to start one of the pipeline's manifest triggers as a `TriggerRun`. The trigger dropdown is loaded from `GetPipeline` (manifest triggers only exist there, not on list rows), labelled with a human-readable schedule (`cron ...`, `every N minutes`, `batch rerun`).
- Wired up `CreateTriggerRun`: the RPC existed at the proto/API level but the TypeScript RPC layer never registered a handler for it — added it, mirroring the actor-stamping pattern `CreatePipelineRun` already uses.
- The selected trigger's full schedule is copied into `TriggerRunSpec.trigger` on submit, not just `sourceTriggerName` — traced through `go/components/triggerrun/util.go`'s `GetTriggerType` and the cron/interval runners: the reconciler reads `spec.trigger` directly, and nothing on the backend resolves `sourceTriggerName` back into a schedule.
- Expanded the form into a full "Run trigger" flow: optional backfill mode with a start/end execution window, parameter-ID selection (scoped to the trigger's declared parameters), and a max-concurrency override for that run only.
- Shows an "Automatically switch to the latest revision once changes are applied?" choice, but keeps it disabled with a "(Coming soon)" label — that feature hasn't launched — so the run always submits `autoFlip: false`.
- Greys out the "Run trigger" button on the pipeline detail page once the manifest has loaded and declares no triggers, with a tooltip explaining why ("No triggers defined for this pipeline"). List rows carry no manifest, so the button stays enabled there — an absent manifest means "unknown," not "no triggers."

## How can I filter ListPipelineRun by a specific trigger run?

**You already can, and no backend change was needed.** Every trigger-spawned run is stamped with `pipelinerun.michelangelo/triggered-by = <triggerRun.Name>` (`go/worker/workflows/trigger/cron_trigger_workflows.go:305`), labels are persisted to a queryable `pipeline_run_labels` table, and `listOptions.labelSelector` is compiled into real SQL (`go/storage/mysql/mysql.go:858`) — or passed to the K8s API server on the etcd path:

```
ListPipelineRun(namespace, listOptions: { labelSelector: "pipelinerun.michelangelo/triggered-by=<trigger-name>" })
```

Since the label is written at creation time, this works for historical runs too — nothing to backfill.

> ⚠️ **Sharp edge:** `mysql.go:320` only parses the selector strings when `listOptionsExt.operation` is nil. Setting both a `labelSelector` and a Criterion silently drops the label filter — no error, just unfiltered results.

The label key now has a single definition (`TRIGGERED_BY_LABEL`) shared by every consumer, instead of being duplicated as a magic string.

## User can see pipeline runs generated by a trigger

- Adds a **"Triggered by"** cell to `SHARED_RUN_CELL_CONFIG`, so it shows as a runs-list column and in the run detail header, linking to the originating trigger. Manually started runs carry no label, so the cell renders empty with no dead link.
- The trigger detail page gets a **Recent Runs** tab, scoped by the `triggered-by` label selector above. It omits the "Triggered by" column — every row there shares the trigger being viewed, so it would just repeat the page's own name and link back to itself.
- Adds the mirror image on the pipeline side: a **Triggers** tab on the pipeline detail page listing every `TriggerRun` created for that pipeline (name linking to the trigger detail page, created date, started-by, cron/interval schedule, state).
  - `TriggerRun` turned out to carry no `pipeline.michelangelo/name` label — the apiserver only records the relationship via an ownerReference and `spec.pipeline` (`TriggerRunSpec.pipeline` in `proto/api/v2/trigger_run.proto`), so a `labelSelector` can't express this scoping. Added a small, generic client-side `filter` option to `TableDetailPageConfig` (dot-path field match against the fetched rows) and used it here instead.

## Test plan

- [x] `tsc -b packages/core` — clean
- [x] `eslint --quiet` on touched files — clean
- [x] `prettier` — clean
- [x] `yarn test:core` — **1367/1367 passing** (150 files)
- [x] New/updated tests cover: schedule dropdown and formatted schedules, full run-trigger submission (backfill window, parameter filtering, max-concurrency override), disabled "Coming soon" autoFlip choice, disabled "Run trigger" button with tooltip when no triggers exist (and that it stays enabled on list rows where manifest data is absent), the Triggers tab listing and linking trigger runs while excluding runs belonging to other pipelines, "Triggered by" cell/link behavior on runs, and the trigger's Recent Runs tab/label-selector.
- [x] `cd javascript && yarn --ignore-engines dev` start [local server](http://localhost:5173/), and check all features work successfully
Screenshot 2026-08-22 at 9 25 42 PM
Screenshot 2026-08-22 at 9 26 32 PM
Screenshot 2026-08-22 at 9 26 39 PM
Screenshot 2026-08-22 at 9 20 01 PM
Screenshot 2026-08-22 at 9 24 00 PM
Screenshot 2026-08-22 at 10 14 51 PM
Screenshot 2026-08-22 at 9 34 20 PM

https://github.com/user-attachments/assets/1bb0c5d7-20c6-45f7-a581-6860e9888396


javascript/packages/core/components/views/detail-view/components/detail-view-page-renderer/pages/detail-view-table-page/types.ts
```diff
@@ -1,10 +1,14 @@
+import type { TableDetailPageFilter } from '#core/components/views/detail-view/types/detail-view-schema-types';
 import type { TableConfig } from '#core/components/views/types';
 import type { QueryConfig } from '#core/types/query-types';
 
 export interface DetailViewTablePageProps<T extends object = object> {
   /** Query configuration for fetching data to display in the table */
   queryConfig: QueryConfig;
 
+  /** Narrows the fetched rows client-side */
+  filter?: TableDetailPageFilter;
```
donovanlopez: rather than filtering client-side, we should be filtering when we call the List endpoints using `listOptions`

javascript/packages/core/config/entities/pipeline/__tests__/pipeline.test.tsx
```diff
@@ -184,3 +185,164 @@ describe('PIPELINE_ENTITY_CONFIG: delete action', () => {
...
+      expect(await screen.findByRole('dialog', { name: 'Run trigger' })).toBeInTheDocument();
+    });
+  });
+
+  describe('list view', () => {
+    /**
+     * List rows come from `ListPipeline`, which carries no `spec.manifest` (only a
+     * `GetPipeline` response does — see run-trigger-form.tsx). An absent manifest means
+     * "unknown," not "no triggers," so the action must not be disabled here.
+     */
```
craigmarker: I'm not sure this is correct, I tested in sandbox and ListPipeline includes spec.manifest

Image

javascript/packages/core/config/entities/pipeline/run-trigger-form.tsx
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    const sourceTrigger = triggerMap[values.sourceTriggerName];
+    if (!sourceTrigger) {
+      // Only reachable if the manifest changed under an open dialog. Surfacing it beats
+      // sending a TriggerRun with no schedule, which the reconciler would leave inert.
+      throw new Error(
+        `Trigger "${values.sourceTriggerName}" is no longer defined on this pipeline.`
+      );
+    }
```
craigmarker, line 64: was this verified during testing? my sense is that this will throw outside of the form error ecosystem. IMO, just better to let the backend respond with an error. (the error should ideally include remediation steps: please refresh and try again or something)

javascript/packages/core/config/entities/pipeline/types.ts
```diff
@@ -9,6 +10,14 @@ export interface Pipeline {
     owner: {
       name: string;
     };
+    /**
+     * Returned by both `GetPipeline` and `ListPipeline`; optional because a pipeline can be
```
craigmarker: there's a lot of this "returned by both GetPipeline and ListPipeline" in comments throughout this PR. a little noisy and self-aware of a bug introduced within this PR in my opinion

javascript/packages/core/config/entities/trigger/types.ts
```diff
@@ -16,6 +16,52 @@ export type Trigger = {
...
+ * The whole object is copied (and, for a backfill run, partially overridden) into
+ * {@link RunTriggerPayload.spec.trigger} on submit, so fields not read directly by the UI
+ * still survive the round trip via object spread — they just aren't typed here individually
+ * unless a form field needs to read or override them (see `parametersMap`, `maxConcurrency`).
+ */
+export type ManifestTrigger = {
+  triggerType?:
+    | { case: 'cronSchedule'; value: { cron?: string } }
+    | { case: 'intervalSchedule'; value: { interval?: { seconds?: bigint | number | string } } }
+    | { case: 'batchRerun'; value: unknown };
```
craigmarker: these feel wrong -- why is batchRerun value unknown?

javascript/packages/core/config/entities/pipeline/detail.ts
```diff
@@ -39,5 +40,37 @@ export const PIPELINE_DETAIL_CONFIG: DetailViewConfig = {
...
+            url: '/${studio.projectId}/${studio.phase}/triggers/${row.metadata.name}',
+          },
+          { id: 'metadata.creationTimestamp.seconds', label: 'Created', type: CellType.DATE },
+          { id: 'spec.actor.name', label: 'Started by', type: CellType.TEXT },
+          { id: 'spec.trigger.triggerType.value.cron', label: 'Cron', type: CellType.TEXT },
+          {
+            id: 'spec.trigger.triggerType.value.interval.seconds',
+            label: 'Interval seconds',
+            type: CellType.TEXT,
+          },
```
craigmarker: Since these two are oneofs, does it make sense to reuse the formatTriggerSchedule and have a single column called "Schedule" or something?

craigmarker (approved):

Issue #1869, opened 2026-08-19
trainer: MLflow backed ExperimentStore for auto resume


apurvapatkeshwar:
**Is your feature request related to a problem? Please describe.**

#1571 made the `pytorch_lightning` trainer's auto resume pluggable behind the `ExperimentStore` protocol, with `FsspecExperimentStore` as the filesystem default: the resume marker lives as a small file under the run's `storage_path`

That works, but for teams already running an MLflow tracking server the filesystem marker has two rough edges:

- resume state is invisible - you can't see from a dashboard which runs have a resumable checkpoint without listing the storage bucket
- some storage backends make the marker file pattern awkward (object stores with restricted list permissions, short lived scratch volumes)

**Describe the solution you'd like**

`MlflowExperimentStore`, a second `ExperimentStore` implementation that records each run's resume marker as a tagged, immediately terminated run in a dedicated MLflow experiment:

- `track()` (worker rank 0) creates one marker run tagged with the `(storage_path, run_name)` identity and the experiment directory to resume from
- `locate_resumable()` (driver) searches for the newest marker with that exact identity and returns its recorded path
- both methods keep the seam's never-raise contract: any failure (server down, mlflow not installed, auth) is logged and swallowed, so marker bookkeeping can never fail a training run
- mlflow imports lazily and installs via the existing `trainer-mlflow` extra; server/credentials follow MLflow's native env vars (`MLFLOW_TRACKING_URI`, `MLFLOW_TRACKING_USERNAME`/`PASSWORD`/`TOKEN`) -- no secrets in the pickled store

This stays entirely inside the existing seam (no new dependency, no API or CRD change)

Scope notes, to avoid confusion with nearby work:

- this is not #1227 (a model registry client abstraction) - resume markers are experiment tracking state, not registry entries
- it is also distinct from the tracker/logger config work that already merged (#1432/#1434) - those configure what the lightning loggers report during training; this covers where the resume marker lives

**Describe alternatives you've considered**

- key val store or object store native marker: Adds a new infra dependency; MLflow is already a supported integration with an operator guide
- Extending `FsspecExperimentStore` with an fsspec MLflow shim: MLflow has no fsspec filesystem; forcing one would be more code than a direct client

**Component**
Which part of the project does this affect?
- [x] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [ ] Helm chart / Kubernetes deployment
- [ ] UI (Studio)
[...]

PR #1872, opened 2026-08-20
feat(trainer): add an MLflow-backed ExperimentStore for auto-resume

apurvapatkeshwar:
Implements #1869.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds `MlflowExperimentStore`, a second implementation of the `ExperimentStore` auto-resume seam from #1571. Instead of a marker file under the run's `storage_path` (the `FsspecExperimentStore` default), each training attempt records a tagged, immediately-terminated run in a dedicated MLflow experiment, and resume looks up the newest marker for the exact `(storage_path, run_name)` identity.

```python
param = LightningTrainerParam(
    ...,
    experiment_store=MlflowExperimentStore(
        tracking_uri="http://mlflow.example.com:5000",
        experiment_name="team-x-resume",
    ),
)
```

Design choices:

- **Never-raise contract kept.** `track()` logs and swallows any failure; `locate_resumable()` returns `None` on anything unexpected. A down MLflow server means a fresh start, never a failed run.
- **One marker run per attempt, most recent wins.** MLflow has no atomic overwrite or KV lookup, so markers accumulate in the dedicated experiment (default `michelangelo-resume`) and lookup orders by start time. Deliberate v1 tradeoff; `track()` only runs on worker rank 0 so the volume is one run per training attempt.
- **Filter quoting handled per value.** MLflow's search filter grammar has no in-literal quote escaping, so each literal picks the quote style the value doesn't contain; an identity containing both quote styles degrades to "nothing to resume". A client-side identity re-check on the returned runs backstops the filter either way.
- **No secrets in the store.** The store is pickled into the (logged) train loop config, so it only carries `tracking_uri` and `experiment_name`; credentials flow through MLflow's native `MLFLOW_TRACKING_USERNAME`/`PASSWORD`/`TOKEN` env vars. mlflow imports lazily via the existing `trainer-mlflow` extra.

**Why?**

#1869 lays out the gap: with the filesystem marker, resume state is invisible from any dashboard, and some storage backends make the marker-file pattern awkward. Teams already running an MLflow tracking server get resume markers as queryable, visible tracking state instead -- with zero behavior change for anyone not opting in (the filesystem store remains the default).

**How did you test it?**

- 23 new unit tests against a faked `MlflowClient` (no live server): round-trip, most-recent-wins, identity isolation, quote-style selection and the both-quotes degrade path, never-raise on server errors and on mlflow-not-installed, create-experiment race, picklability, protocol conformance. 100% line and branch coverage on the new module; re-run on top of current main before posting.
- Smoke-tested against a real local `mlflow server` (sqlite backend, mlflow 2.22.5): plain round-trip, most-recent-wins, identity isolation, single-quote and double-quote identities resolve through the real filter parser, both-quotes identity returns `None` without raising, unknown identity returns `None`. This caught (and now guards) a real bug: SQL-style `''` quote doubling is rejected by MLflow's actual parser, which the mocked tests could not see.
- Full `pytorch_lightning` package suite passes; `ruff format` + `ruff check` clean.

**Potential risks**

- Opt-in only: nothing changes unless a caller passes `experiment_store=MlflowExperimentStore(...)`. The never-raise contract means the worst failure mode is "no resume, fresh start", same as today's seam guarantees.
- Marker runs accumulate in the dedicated experiment over time (one per training attempt); harmless, but operators who care can point the store at a purpose-made experiment and apply MLflow's own retention.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Added `MlflowExperimentStore`, an opt-in MLflow-backed implementation of the trainer's auto-resume `ExperimentStore` seam. Installs via the existing `trainer-mlflow` extra; the filesystem store remains the default.

**Documentation Changes**

The MLflow operator guide (`docs/operator-guides/integrations/mlflow.md`) gains a section documenting the store, its configuration, and the credentials flow.

---

**Testing addendum (post-review, re: dkurra's run-link question)**

Verified the full train -> stop -> resume loop against a real local Ray cluster and a real local (file-backed) MLflow tracking store -- see the review thread for the run log evidence. A literal same-identity re-run resumes correctly but can't by itself distinguish the store's own `locate_resumable()` from Ray Train V2's native run-directory restoration; confirmed the store is genuinely load-bearing by physically deleting the native checkpoint directory and re-pointing the store at a relocated copy, then observing `LightningTrainer` log that it seeded the resume from that relocated path.

A permanent regression test reproducing this scenario is ready as a follow-up, stacked on this PR so it doesn't require re-review of anything already approved here.

### DIFF
--- a/docs/operator-guides/integrations/mlflow.md
+++ b/docs/operator-guides/integrations/mlflow.md
@@ -182,6 +182,39 @@ Users are responsible for:
 
 ---
 
+## Auto-Resume with MlflowExperimentStore
+
+The [`michelangelo.lib.trainer.torch.pytorch_lightning`](https://github.com/michelangelo-ai/michelangelo/tree/main/python/michelangelo/lib/trainer/torch/pytorch_lightning) trainer supports automatic resumption of interrupted training runs through a pluggable `ExperimentStore`. By default, resume markers are written to the run's storage filesystem (`FsspecExperimentStore`). If your organization already runs an MLflow Tracking Server, `MlflowExperimentStore` records those markers as tagged runs on the MLflow server instead, which makes resume state centrally visible and works even when the storage path is not conveniently listable.
+
+Requires the `mlflow` package in the training image (the `trainer-mlflow` extra: `pip install 'michelangelo[trainer-mlflow]'`).
+
+```python
+from michelangelo.lib.trainer.torch.pytorch_lightning import (
+    LightningTrainerParam,
+    MlflowExperimentStore,
+)
+
+param = LightningTrainerParam(
+    create_model_fn=my_model_factory,
+    create_model_fn_kwargs={},
+    train_data=train_ds,
+    val_data=val_ds,
+    experiment_store=MlflowExperimentStore(
+        tracking_uri="http://mlflow.example.com:5000",
+        experiment_name="team-x-resume",
+    ),
+)
+```
+
+Behavior and configuration notes:
+
+- **Server and credentials follow MLflow's native environment contract.** If `tracking_uri` is omitted, the client resolves `MLFLOW_TRACKING_URI` from the environment. Authentication uses `MLFLOW_TRACKING_USERNAME` / `MLFLOW_TRACKING_PASSWORD` / `MLFLOW_TRACKING_TOKEN`, exactly as in Step 3 above. Credentials are deliberately not constructor arguments: the store is shipped to Ray workers inside the (logged) training loop config, so secrets must stay in the environment.
+- **Best-effort by design.** A marker write or lookup failure (unreachable server, missing `mlflow` package, auth error) is logged and swallowed. It never fails a training run; the run simply starts fresh instead of resuming.
+- **One marker run per training attempt.** Each attempt records a new, immediately terminated MLflow run tagged with the training run's identity; resume uses the most recent marker. Markers accumulate in the dedicated experiment (default name `michelangelo-resume`).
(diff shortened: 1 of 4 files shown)


apurvapatkeshwar:
@KyleKoon @sishilong @dkurra PTAL
This implements #1869 - MLflow-backed ExperimentStore for auto-resume

dkurra (approved):

dkurra:
@apurvapatkeshwar do you have run link (local or remote) showing auto resume working?

apurvapatkeshwar:
@dkurra Confirmed with a real run. Trained 1 epoch, then removed the native Ray checkpoint directory entirely so the only way to resume was through `MlflowExperimentStore`. Re-running the same identity for 3 more epochs resumed correctly (epoch 0 -> 2), and the logs confirm the store drove it end to end:

Auto-resume: will seed from checkpoint .../relocated/resume_run_copy/checkpoint_...
(RayTrainWorker) No native Ray checkpoint found; seeding auto-resume from store-located checkpoint: ...
(RayTrainWorker) Resuming from checkpoint.path=...
(RayTrainWorker) Restored all states from the checkpoint ...

Since the native checkpoint was gone, that restore could only have come from `MlflowExperimentStore.track()`/`locate_resumable()` round-tripping through a real MLflow backend (local file:// tracking URI, no server needed)

Will add a permanent regression test for this exact scenario as a follow-up stacked on this PR

PR #1874, opened 2026-08-20
helm: add PrometheusRule starter alerts

apurvapatkeshwar:
> **Stacked on #1865.** This adds the PrometheusRule under the `monitoring` toggle that PR introduces, so the branch is based on that PR's branch rather than main. The first commit here is #1865's, unchanged; only the last commit (`helm: add PrometheusRule starter alerts`) is new. Once #1865 merges I will rebase and the diff collapses to just this change -- happy to hold review until then if that is easier.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Second slice of the observability starter pack (#1692): the alert rules already specified in `docs/operator-guides/operations/monitoring.md`, shipped as an installable `PrometheusRule` under the `monitoring` toggle (CRD-gated like the ServiceMonitor).

The default group carries the four alerts backed by metrics the controller manager exposes out of the box:

- `PipelineRunFailureRateHigh`
- `PipelineRunDurationHigh` (P99 > 1h)
- `ControllerReconcileErrorRate`
- `CascadeChildDrainTimeout`

The two Envoy-based inference alerts (`InferenceLatencyHigh`, `InferenceErrorRateHigh`) sit behind `monitoring.prometheusRule.envoyAlerts.enabled` (default off): the chart does not enable the Envoy admin interface, so by default those metrics don't exist -- shipping alerts that can never fire seemed worse than shipping fewer that all work. Flip the toggle once you've enabled the admin interface and a scrape job for it.

Thresholds and expressions are taken from the monitoring guide unchanged. Two deliberate deviations in the annotation text: kubectl hints use the release namespace and the generated deployment name instead of the doc's hardcoded `ma-system`/`michelangelo-controllermgr`, and the pipeline-run hint uses `-A` since runs live in job namespaces.

**Why?**

#1692 item 2: the monitoring guide tells operators to hand-copy alert rules from documentation into their own PrometheusRule resources, which drifts (the guide's examples already carry labels and namespaces that don't match the chart). Shipping the same rules as chart-managed resources makes them installable, versioned with the chart, and correct by construction.

**How did you test it?**

- `helm lint` clean; default render byte-identical to the base branch -- the resource only appears under the `monitoring` toggle with the Prometheus Operator CRDs present.
- Rendered rule groups extracted and validated with `promtool check rules` (both the 4-alert default and the 6-alert Envoy-enabled variant pass) -- this also proves the `{{ $value }}`-style Prometheus template escapes survive helm rendering intact. Re-run on the rebased branch before posting.
- CRD-absent render verified to skip both ServiceMonitor and PrometheusRule.

**Potential risks**

- Default off and CRD-gated: no change for existing deployments, and enabling on a cluster without the Prometheus Operator is a no-op.
- Alert thresholds are starter values from the guide; operators with different SLOs will tune them via `additionalLabels`/their own rules. Deliberately conservative rather than exhaustive.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Added chart-managed PrometheusRule starter alerts (pipeline-run failures, long-running pipeline runs, controller reconcile errors, cascade drain timeouts) under the `monitoring` toggle, with optional Envoy-based inference alerts behind `monitoring.prometheusRule.envoyAlerts.enabled`.

**Documentation Changes**

None in this PR; the docs pass repointing the monitoring guide at the chart-managed path comes later in this series.

### DIFF
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -111,6 +111,30 @@ jobs:
             --set temporal.server.config.persistence.visibility.sql.password=root \
             2>/dev/null
 
+      - name: Template — monitoring enabled (ServiceMonitor renders with CRDs present)
+        run: |
+          helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --api-versions monitoring.coreos.com/v1 \
+            --debug \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Template — monitoring enabled without CRDs (must skip gracefully)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Validate fail-fast guard (monitoring without controllermgr must fail)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --set controllermgr.enabled=false \
+            2>/dev/null
+
       - name: Template — Ingress enabled (UI + Envoy, hosts[] syntax)
         run: |
           helm template michelangelo helm/michelangelo \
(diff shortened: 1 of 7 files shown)


apurvapatkeshwar:
closing - landed via #1922's squash along with the rest of the monitoring stack

PR #1876, opened 2026-08-20
helm: add Grafana dashboards for the control plane and pipeline runs

apurvapatkeshwar:
> **Stacked on #1874** (which stacks on #1865). This adds the Grafana dashboards on top of the monitoring toggle and alert rules those PRs introduce, so the branch is based on #1874's branch rather than main. The first two commits here are #1865's and #1874's, unchanged; only the last commit (`helm: add GrafanaDashboard CRs for control plane and pipeline runs`) is new -- [this compare view](https://github.com/apurvapatkeshwar/michelangelo/compare/feat/monitoring-prometheusrules...feat/monitoring-grafana-dashboards) shows just this PR's diff. Once the parents merge I will rebase and the diff here collapses to the same thing -- happy to hold review until then if that is easier.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Third slice of the observability starter pack (#1692): two dashboards shipped as [Grafana Operator](https://github.com/grafana/grafana-operator) `GrafanaDashboard` CRs (`grafana.integreatly.org/v1beta1`), CRD-gated like the other monitoring resources and toggled by `monitoring.grafanaDashboards.enabled`.

- **Control Plane**: pipeline run results and readiness, reconcile error rates and P99 latency by controller, work queue depth, CR reflector errors, and the cascade delete family (active drains, drain duration P99, timeouts, ownerReference backfills).
- **Pipeline Runs**: duration P50/P99, failure rate by reason, results by state and pipeline type, successes vs failures, step completions.

`monitoring.grafanaDashboards.instanceSelector` picks which Grafana instance(s) get the dashboards (default matches `dashboards: grafana`, the operator's common convention).

Two implementation notes:

- The dashboard JSON lives under `files/dashboards/` as plain JSON loaded with `.Files.Get`, so Grafana's `{{label}}` legend syntax never collides with chart templating and the files can be linted or imported directly.
- The panels were built from the metric names in the Go source, not adapted from existing templates -- every query targets a metric the controller manager actually registers today (`pipelinerun_*`, `pipeline_*`, `cascade_*`, `cr_reflector_errors_total`, plus standard controller-runtime/workqueue metrics). Notably the sandbox grafana config references a `crd_unmarshal_errors_*` metric name that doesn't exist in the codebase; these dashboards use the real `cr_reflector_errors_total` instead. No Envoy panels, since those metrics are absent by default.

**Why?**

#1692 item 3: operators wiring up the chart today have to build dashboards from scratch or adapt the sandbox's grafana config, which references metrics that don't exist. Shipping operator-managed dashboards versioned with the chart gives every install the same starting point, correct by construction against the metrics the code actually exposes.

**How did you test it?**

- `helm lint` clean; default render byte-identical to the base branch -- the dashboards only appear under the `monitoring` toggle with the Grafana Operator CRD present.
- Each rendered CR's embedded JSON round-trips through a JSON parser after helm templating (12-panel and 7-panel dashboards, `instanceSelector` present on both). Re-run on the current stack tip before posting.
- Gating matrix verified: Grafana-CRD-absent render skips the dashboards while keeping the Prometheus resources; `monitoring.grafanaDashboards.enabled=false` does the same; `monitoring` off renders none of the three kinds.
- Every PromQL expression in every panel checked against the list of metrics registered in the Go source -- zero references to non-exposed metrics.
- CI monitoring check extended with the Grafana API version and now asserts all three resource kinds render (both the repeated-flag and comma-joined `--api-versions` forms verified against helm v3.16).

**Potential risks**

- Default-off and CRD-gated: no change for existing deployments, and enabling on a cluster without the Grafana Operator is a no-op.
- Dashboards are starter content: operators with existing dashboards can leave the toggle off or use `instanceSelector` to scope them to a sandbox Grafana.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Added chart-managed Grafana dashboards (control plane health and pipeline runs) as Grafana Operator `GrafanaDashboard` resources under the `monitoring` toggle, gated by `monitoring.grafanaDashboards.enabled`.

**Documentation Changes**

None in this PR; the docs pass repointing the monitoring guide at the chart-managed dashboards comes last in this series.

### DIFF
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -111,6 +111,30 @@ jobs:
             --set temporal.server.config.persistence.visibility.sql.password=root \
             2>/dev/null
 
+      - name: Template — monitoring enabled (ServiceMonitor renders with CRDs present)
+        run: |
+          helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --api-versions monitoring.coreos.com/v1 \
+            --debug \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Template — monitoring enabled without CRDs (must skip gracefully)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            | grep 'kind: ServiceMonitor'
+
+      - name: Validate fail-fast guard (monitoring without controllermgr must fail)
+        run: |
+          ! helm template michelangelo helm/michelangelo \
+            -f helm/michelangelo/values-k3d.yaml \
+            --set monitoring.enabled=true \
+            --set controllermgr.enabled=false \
+            2>/dev/null
+
       - name: Template — Ingress enabled (UI + Envoy, hosts[] syntax)
         run: |
           helm template michelangelo helm/michelangelo \
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -118,14 +118,14 @@ jobs:
             --set monitoring.enabled=true \
             --api-versions monitoring.coreos.com/v1 \
             --debug \
-            | grep 'kind: ServiceMonitor'
+            | grep -e 'kind: ServiceMonitor' -e 'kind: PrometheusRule'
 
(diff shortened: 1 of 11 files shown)


apurvapatkeshwar:
closing - the dashboards landed via #1922's squash; merging this now would actually revert the guide wording that #1922 updated

PR #1877, opened 2026-08-20
feat(train): add FSDP2 strategy to the Lightning trainer

KyleKoon:
## Summary

- Adds `strategy: fsdp2` (FSDP2 / `ModelParallelStrategy`) as a training strategy option in the PyTorch Lightning trainer SDK
- New `RayModelParallelStrategy` bridges Lightning's `ModelParallelStrategy` with Ray Train's device and rank wiring
- FSDP2 checkpoint consolidation (sharded DCP → full state_dict) in `LightningTrainerWithStateDict` for serving
- Extracts `_is_deepspeed_strategy` / `_is_model_parallel_strategy` as shared guard functions (previously a private method on the class)

## Usage

Select the strategy in your pipeline config:

```yaml
lightning_trainer_kwargs:
  strategy: fsdp2
```

The user's `LightningModule` must implement `configure_model()` and apply `fully_shard` over the strategy-provided device mesh.

## Design notes

- **Sharded (DCP) checkpointing is enforced** — `save_distributed_checkpoint=True` is always set
- **TP is unsupported** — tensor_parallel_size defaults to 1; an explicit value raises
- **DP is not user-configurable** — data_parallel_size always equals world_size
- **Requires pytorch-lightning >= 2.3** — lazy import with clear error on older versions

## Test plan

- [x] Unit tests for strategy dispatch, guard functions, `_prepare_trainer_for_ray`, per-node callback routing
- [x] Unit tests for FSDP2 checkpoint loading path (success + empty state_dict error)
- [x] Full test suite passes (374 tests)
- [ ] E2E validation on GPU cluster with `strategy: fsdp2` pipeline

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/model_parallel.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/model_parallel.py
@@ -1,0 +1,81 @@
+"""Ray-aware FSDP2 (``ModelParallelStrategy``) glue.
+
+``ModelParallelStrategy`` requires pytorch-lightning >= 2.3, so ``util.py`` imports this module
+lazily (only once FSDP2 is requested) rather than at top level.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import ray
+import torch
+from pytorch_lightning.strategies import ModelParallelStrategy
+
+_logger = logging.getLogger(__name__)
+
+
+class RayModelParallelStrategy(ModelParallelStrategy):
+    """Ray glue for FSDP2.
+
+    Ray Train ships RayDDPStrategy/RayFSDPStrategy/RayDeepSpeedStrategy but no
+    equivalent for ModelParallelStrategy, so we supply the Ray device and rank
+    wiring ourselves. Remove once Ray adds native ModelParallelStrategy support.
+    """
+
+    def __init__(self, *args, **kwargs):
+        if kwargs.get("save_distributed_checkpoint") is False:
+            _logger.warning(
+                "save_distributed_checkpoint=False is not supported; FSDP2 always saves sharded (DCP) "
+                "checkpoints. Forcing save_distributed_checkpoint=True."
+            )
+        kwargs["save_distributed_checkpoint"] = True
+
+        if kwargs.get("tensor_parallel_size") is not None:
+            raise ValueError(
+                "Tensor parallelism is not currently supported. Leave tensor_parallel_size unset. "
+                f"Got tensor_parallel_size: {kwargs.get('tensor_parallel_size')}"
+            )
+
+        if kwargs.get("data_parallel_size") is not None:
+            raise ValueError(
+                "Configuring data parallelism is not supported; FSDP2 always shards across the "
+                f"full world (data_parallel_size = world_size). Leave data_parallel_size unset. "
+                f"Got data_parallel_size: {kwargs.get('data_parallel_size')}"
+            )
+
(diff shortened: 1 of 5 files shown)


sallycr (approved):
LGTM, also trigger the agent code review for dev experience check points.

sallycr (commented):
Reviewed by a two-agent pass (correctness + OSS developer-experience).

**Correctness:** one bug worth fixing before merge — see inline comment on `_private/model_parallel.py`. `RayModelParallelStrategy` rejects `tensor_parallel_size`/`data_parallel_size` whenever they're not `None`, but pytorch-lightning's own `ModelParallelStrategy.__init__` defaults both to the string `"auto"`, not `None`. So a caller passing `"auto"` explicitly (e.g. copying Lightning's documented default, or a config that round-trips a resolved value) hits the `ValueError` even though `"auto"` means exactly the same thing as leaving the kwarg unset.

Everything else checked out well:
- The `_is_deepspeed_strategy` refactor (instance method → module-level helper) has no other call sites depending on the old shape.
- `_prepare_trainer_for_ray`'s FSDP2 bypass was checked against Ray's own `prepare_trainer` source and matches the one real check it performs (cluster-environment class).
- The `"state_dict"` top-level key used for DCP load matches what `ModelParallelStrategy.save_checkpoint` actually writes.
- Error messages for misconfiguration (tensor/data parallelism, process group backend/timeout, unsupported pytorch-lightning version) are all actionable — no raw stack traces or bare internal codes.
- `save_distributed_checkpoint=False` misconfiguration logs at WARN and auto-corrects rather than panicking.
- No breaking changes — purely additive/opt-in `strategy="fsdp2"`.
- Test coverage for strategy resolution, checkpoint load happy/failure paths, and Ray cluster-env branches looks solid.

**Minor (non-blocking):** the new `"fsdp2"` value isn't mentioned in `docs/user-guides/`, but that's consistent with the existing `"ddp"`/`"deepspeed"`/`"fsdp"` strategies, which are equally undocumented outside code — a pre-existing gap, not a regression from this PR.

python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/model_parallel.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, *args, **kwargs):
+        if kwargs.get("save_distributed_checkpoint") is False:
+            _logger.warning(
+                "save_distributed_checkpoint=False is not supported; FSDP2 always saves sharded (DCP) "
+                "checkpoints. Forcing save_distributed_checkpoint=True."
+            )
+        kwargs["save_distributed_checkpoint"] = True
+
+        if kwargs.get("tensor_parallel_size") is not None:
```
sallycr, line 37: This (and the identical check on `data_parallel_size` below) rejects `"auto"`, but `pytorch_lightning.strategies.ModelParallelStrategy.__init__` defaults `tensor_parallel_size`/`data_parallel_size` to the string `"auto"`, not `None`.

Concretely: `RayModelParallelStrategy(tensor_parallel_size="auto")` raises `ValueError("Tensor parallelism is not currently supported...")` even though `"auto"` means exactly the same thing as leaving the kwarg unset — a caller who copies Lightning's own documented default (or round-trips a resolved config value) hits this unnecessarily.

Suggest guarding against both sentinels, e.g. `if kwargs.get("tensor_parallel_size") not in (None, "auto"):` (and the same for `data_parallel_size`).

Issue #1718, opened 2026-08-07
Release v0.8.0


sallycr:
## Pre-Release

- [ ] All CI checks pass on the release branch
- [ ] Version numbers updated across all components:
  - [ ] `python/pyproject.toml`
  - [ ] `javascript/packages/core/package.json`
  - [ ] `javascript/packages/rpc/package.json`
  - [ ] `website/package.json`
  - [ ] `helm/michelangelo/Chart.yaml` (both `version` and `appVersion`)
- [ ] `CHANGELOG.md` updated via `git cliff`
- [ ] Release notes drafted (follows three-layer template: summary, categorized changes, compatibility matrix)
- [ ] Breaking changes reviewed and documented:
  - [ ] All `BREAKING CHANGE:` commits identified in changelog
  - [ ] Migration guide written for each breaking change
  - [ ] `UPGRADING.md` updated with migration steps
  - [ ] Deprecation warnings added for items being removed (see [Deprecation Policy](../../CONTRIBUTING.md#deprecation-policy))
- [ ] Compatibility matrix updated in release notes

## Release Candidate

- [ ] Release branch created: `release/vX.Y`
- [ ] RC tag pushed (e.g. `v0.3.0-rc.1`)
- [ ] RC artifacts published and accessible:
  - [ ] Python wheel on PyPI (PEP 440: `0.3.0rc1`)
  - [ ] Go service containers on ghcr.io
  - [ ] UI container on ghcr.io
  - [ ] npm packages (`@michelangelo-ai/core`, `@michelangelo-ai/rpc`)
  - [ ] Helm OCI chart on ghcr.io
- [ ] RC announcement posted (GitHub Discussions or relevant channel)
- [ ] Soak period complete (minimum 1 week for minor releases)
- [ ] No P0/P1 issues reported against RC

## Final Release

- [ ] Release tag pushed (e.g. `v0.3.0`)
- [ ] All artifacts published and accessible:
  - [ ] Python wheel on PyPI
  - [ ] Go service containers on ghcr.io (tagged `v0.3.0` + `latest`)
  - [ ] UI container on ghcr.io
  - [ ] npm packages (latest dist-tag)
  - [ ] Helm OCI chart on ghcr.io
- [ ] GitHub Release created with generated release notes
- [ ] Announcement posted
- [ ] Release branch merged back to main (if diverged)
- [ ] Next development version bumped on main (if applicable)

---

[...]

Issue #1658, opened 2026-08-02
Release v0.7.0


sallycr:
## Pre-Release

- [ ] All CI checks pass on the release branch
- [x] Version numbers updated across all components:
  - [x] `python/pyproject.toml`
  - [x] `javascript/packages/core/package.json`
  - [x] `javascript/packages/rpc/package.json`
  - [x] `javascript/app/package.json`
  - [x] `helm/michelangelo/Chart.yaml` (both `version` and `appVersion`)
- [x] `CHANGELOG.md` updated via `git cliff` — [PR #1659](https://github.com/michelangelo-ai/michelangelo/pull/1659), CI green, pending review/merge
- [ ] Release notes drafted (follows three-layer template: summary, categorized changes, compatibility matrix)
- [ ] Breaking changes reviewed and documented:
  - [ ] All `BREAKING CHANGE:` commits identified in changelog
  - [ ] Migration guide written for each breaking change
  - [ ] `UPGRADING.md` updated with migration steps
  - [ ] Deprecation warnings added for items being removed (see [Deprecation Policy](../../CONTRIBUTING.md#deprecation-policy))
- [ ] Compatibility matrix updated in release notes

**Note on how this cut happened**: `release-cut.yml` was dispatched (`version=0.7.0`, `rc_number=1`, `base_ref=main`) but failed at its "Bump versions to RC" step — [run 30674238765](https://github.com/michelangelo-ai/michelangelo/actions/runs/30674238765/job/91298063813), `GH013` rule violation pushing directly to the just-created `release/v0.7` branch. Confirmed via run history this is a **regression**: the identical rc_number=1 scenario succeeded on 2026-07-17 (run 29616938626) — the branch ruleset appears to have been tightened since to block this second push unconditionally, the same `bypass_actors: []` class of issue already tracked for `release-promote.yml` in `REL-005`/[michelangelo#1608](https://github.com/michelangelo-ai/michelangelo/issues/1608). `release/v0.7` was created (first push, exempt as ref-creation) but left un-bumped at `main`'s tip. Version bump completed manually via [PR #1657](https://github.com/michelangelo-ai/michelangelo/pull/1657) instead. A follow-up issue for `release-cut.yml` itself is still needed.

## Release Candidate

- [x] Release branch created: `release/v0.7`
- [x] RC tag pushed (`v0.7.0-rc.1`)
- [x] RC artifacts published and accessible:
[...]

PR #1896, opened 2026-08-21
fix(release): close tracking issue on final promotion

sallycr:
## Summary
- `release-promote.yml`'s tracking-issue step now closes the release tracking issue (with a comment linking the release and artifact-publish run) instead of just commenting and leaving it open.

## Why
The release-manager process has been closing these tracking issues by hand after every final promotion (v0.7.0 #1658, v0.8.0 #1718) since the workflow never did it itself. Automating it removes a manual step from every future release.

## Test plan
- [ ] CI green
- [ ] Verify on the next `release-promote.yml` run (v0.9.0's eventual promotion) that the tracking issue closes automatically

### DIFF
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -113,15 +113,15 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Comment on tracking issue
+      - name: Comment on and close tracking issue
         run: |
           ISSUE_NUMBER=$(gh issue list --search "in:title \"Release v${{ steps.vars.outputs.version }}\"" --state open --json number,title \
             --jq '.[] | select(.title == "Release v${{ steps.vars.outputs.version }}") | .number' | head -1)
           if [ -z "$ISSUE_NUMBER" ]; then
             echo "WARNING: no open tracking issue found for Release v${{ steps.vars.outputs.version }}"
           else
-            gh issue comment "$ISSUE_NUMBER" \
-              --body "Promoted to final: ${{ github.server_url }}/${{ github.repository }}/releases/tag/${{ steps.vars.outputs.tag }}. Artifact publish run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            gh issue close "$ISSUE_NUMBER" \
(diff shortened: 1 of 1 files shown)


paulzim (approved):
## Review: #1896 — fix(release): close tracking issue on final promotion

**Verdict: Approve**

Single-purpose change: in `.github/workflows/release-promote.yml`'s "Comment on tracking issue" step, swaps `gh issue comment "$ISSUE_NUMBER" --body "..."` for `gh issue close "$ISSUE_NUMBER" --comment "..."`, so the tracking issue closes automatically instead of just getting a link comment. Matches the stated goal of removing the manual close step the release manager has been doing by hand after v0.7.0/v0.8.0.

Verified:
- `gh issue close --help` confirms `-c/--comment` is a real, current flag (leaves a closing comment as part of the close call).
- Ran a control test against a scratch issue in my own fork (`paulzim/michelangelo-pit-stop#1`, created and deleted for this test) with the exact command shape used in the workflow: `gh issue close <n> --comment "Promoted to final: ...link..."`. Confirmed via `gh issue view --json state,comments` that it closes the issue *and* posts the comment body in one call — same net effect as the old comment-then-leave-open behavior, minus the leave-open part.
- `issues: write` permission is already granted at the job level (line 18, unchanged) — nothing else needs updating for this to work.
- YAML parses cleanly; diff is otherwise a no-op (search/lookup logic for `ISSUE_NUMBER` is untouched).

**One non-blocking note:** this step runs inside the `promote` job, before the separate `publish-npm` job (a reusable-workflow call that does the actual npm publish) and before the async-triggered `release.yaml`/`changelog.yml`/`ui-release.yml`/`dev-release.yml` runs have reported back. If `publish-npm` fails afterward, `notify-on-failure` still pings Slack — but the tracking issue will already be closed with a "Promoted to final" comment, which reads as more final than a plain comment would have. That ordering isn't new (the old code posted the same "Promoted to final" text at the same point, before those steps were confirmed done), but closing the issue raises the stakes of the gap slightly versus just commenting. Worth a quick gut-check that the Slack failure notification is an acceptable safety net for that edge case — not something I'd hold the PR for.

Test plan checkboxes are unchecked, which is expected here: real verification only happens on the next actual `release-promote.yml` run (v0.9.0's promotion) — can't be exercised at review time beyond what's above.

PR #1907, opened 2026-08-22
feat(deployment): resolve model storage path from the Model CR

aritra0227:
## Summary

The deployment controller derived each model's artifact location from a hardcoded `s3://deploy-models/{modelName}/` convention, so a model could only ever be served from a path matching its name. This resolves the location from the `Model` CR referenced by `spec.desiredRevision` instead, so a model can live wherever the pipeline that produced it wrote to.

- Adds `common.FetchModel` / `ResolveModelStoragePath`, which read the Model CR, validate the artifact scheme, and surface typed failure reasons (`ModelNotFound`, `ModelPackageTypeMismatch`, `NoDeployableArtifact`, `UnsupportedArtifactScheme`) as deployment conditions.
- `AssetPreparationActor` becomes a real validation gate instead of a stub, and `RollingRolloutActor` publishes the resolved path to the model-list ConfigMap.
- Adds `Model` CRs to both sandbox inference demos and registers them before the `InferenceServer` / `Deployment`.

Two details worth calling out for review:

**Reads go through `api.Handler`, not `client.Client`.** `Model` is declared `immutable: true`, so the ingester moves it out of etcd into metadata storage shortly after creation. A direct etcd read succeeds only inside a brief window and then starts returning `NotFound`. The handler tries etcd first and falls back to metadata storage, which is the same pattern the internal controller uses. NotFound detection keys off `codes.NotFound`, since the handler surfaces gRPC status errors rather than `k8serrors`.

**`package_type` is treated as Triton when unset.** No registration path populates it today — `_build_model_proto` in the registry client sets the artifact URIs, description, and kind, but never `package_type` — so a strict check would reject every model the pusher registers. An explicitly non-Triton type is still rejected.

**Tar artifacts pass through untouched.** `MinioStorageBackend.upload` tar-archives directory artifacts and stores them at `…/__dir__.tar`, and the Triton packagers return directories, so this is the shape a pipeline-pushed Triton model actually has. The resolver skips its usual trailing-slash normalization for `.tar` URIs, and the model-sync sidecar downloads and unpacks them instead of running a recursive sync. Extraction is guarded against path traversal, with a manual member check on the sidecar's Python 3.11 runtime since `filter="data"` needs 3.12.

## Test plan

- [x] Unit tests for the resolver, asset preparation, and rolling rollout actors, including all four failure reasons and a deliberately non-convention storage path to catch regressions
- [x] End to end in the sandbox: model staged at a custom S3 path unreachable by name convention, confirmed resolve -> ConfigMap -> sidecar sync -> Triton load
- [x] End to end with a pipeline-shaped artifact: `.tar` at `models/{name}/{push_id}/deployable/{name}/__dir__.tar` with `packageType` omitted, confirmed `AssetsPrepared=TRUE`, extraction to `config.pbtxt` + `1/model.pt`, and `RollingRolloutComplete=TRUE`


badcount (approved):

go/components/deployment/plugins/common/model.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// FetchModel loads the Model named by the Deployment's desired revision. The reference's
+// namespace is optional and defaults to the Deployment's own namespace, matching how the
+// Deployment's other references are resolved.
+//
+// The read goes through the API handler rather than a Kubernetes client because Model is
+// an immutable kind: the ingester deletes immutable objects from etcd once they have been
+// written to metadata storage, so a direct etcd read would report them as missing.
+func FetchModel(ctx context.Context, apiHandler goapi.Handler, deployment *v2pb.Deployment) (*v2pb.Model, error) {
```
badcount, line 50: The code doesn't look like oss/plugin specific, suggest to move it to a more "common" folder

aritra0227: updated.

Issue #1908, opened 2026-08-22
Sandbox MySQL has no data volume, so any restart destroys the Cadence schema and all metadata


NolanFoster:
**Describe the bug**

`python/michelangelo/cli/sandbox/resources/mysql.yaml` declares the sandbox database as a bare `kind: Pod` running `mysql:8.0` with **no volumes section at all** — no PVC, no `hostPath`, no `emptyDir`. All data lives in the container's writable layer, so **any** MySQL pod restart destroys the `cadence` database and the `michelangelo` metadata database, including all pipeline run history.

MinIO, defined alongside it, does this correctly — `minio.yaml:22` mounts `hostPath: /shared/minio-data`, and its contents survived every restart in my testing.

Two separate consequences follow from the same manifest:

1. **No data volume** — the sandbox cannot survive a restart, even though `ma sandbox stop` / `ma sandbox start` are documented lifecycle commands.
2. **Bare Pod rather than a Deployment/StatefulSet** — nothing owns it, so if the pod is ever deleted (including by `kubectl delete pod --force` while clearing pods stuck in `Unknown` after a host VM restart) it never comes back, and the sandbox has no database at all until `ma sandbox sync` recreates it.

**To Reproduce**
1. `ma sandbox create`, confirm a pipeline runs.
2. Restart the underlying VM or run `k3d cluster stop && k3d cluster start` (equivalently, anything that restarts the MySQL pod).
3. `kubectl get pods` — every Cadence pod is stuck in `Init` on `wait-for-schema`.
4. `kubectl logs <cadence-pod> -c wait-for-schema`:
```
ERROR 1049 (42000): Unknown database 'cadence'
Waiting for MySQL schema to be ready...
```

**Expected behavior**
The sandbox database survives a restart, so `ma sandbox stop` / `start` return a working environment.

**Actual behavior**
The schema is gone. `michelangelo-cadence-schema-server` is already `Complete` and never re-runs on its own, so the Cadence tier stays wedged indefinitely. Recovery requires `ma sandbox sync`, and then a second manual step because sync does not re-register the Cadence domain (#1296) — so the next run fails with `EntityNotExistsError{Message: Domain default does not exist.}`.

Hit three times in one session: after a Colima restart, after `k3d cluster stop/start`, and after a VM resize.

**Suggested fix**

Give MySQL a volume, matching what MinIO already does — a `hostPath` is consistent with the existing sandbox style and sufficient for local development:

```yaml
[...]

Issue #1902, opened 2026-08-21
createCluster discards the real error when TerminateCluster succeeds, surfacing 'nil (not None)' instead


NolanFoster:
**Describe the bug**

In `(*module).createCluster` (`go/worker/plugins/ray/starlark_module.go`), the cleanup call at L110 **reassigns** the `err` variable that holds the real failure. When cleanup succeeds, `err` becomes `nil` and the builtin returns `(nil, nil)`. starlark-go rejects a nil `starlark.Value` with no error, so the genuine cause is replaced by:

```
internal error: nil (not None) returned from <built-in method create_cluster of ray value>
```

```go
for !sensorResponse.Ready {
    if err := workflow.ExecuteActivity(sensorCtx, ray.Activities.SensorRayClusterReadiness, sensorRequest).Get(sensorCtx, &sensorResponse); err != nil {  // L103 — real error
        logger.Error("builtin-error", ext.ZapError(err)...)                                                                                              // L104 — logged (only surviving copy)
        reason := err.Error()
        ...
        if err = workflow.ExecuteActivity(ctx, ray.Activities.TerminateCluster, ray.TerminateClusterRequest{                                             // L110 — `=` not `:=`
            ...
        }).Get(ctx, nil); err != nil {                                                                                                                   // L115 — nil when cleanup succeeds
            logger.Error("builtin-error", ext.ZapError(err)...)
        }
        return nil, err                                                                                                                                  // L118 — returns (nil, nil)
    }
```

So the better cleanup behaves, the worse the diagnostics: a successful `TerminateCluster` guarantees the real error is discarded.

**To Reproduce**
1. Cause `SensorRayClusterReadiness` to fail for any reason (I hit it via an unreachable compute-cluster endpoint, filed separately; #1473 reaches the same line via `HeadPodNotFound`).
2. Run a pipeline with a Ray task.
3. The task fails with the `nil (not None)` message and no indication of the underlying cause.

**Expected behavior**
The builtin returns the original sensor error, so the workflow failure names the actual problem.

**Actual behavior**
The original error is discarded and replaced with a starlark-go internal error about a malformed return value. The only surviving trace is the `builtin-error` log line at L104 in the worker pod — invisible to anyone reading the pipeline run status, the UI, or the CLI.

**Suggested fix**

Use a distinct variable for the cleanup error:

```go
[...]

Issue #1901, opened 2026-08-21
Sandbox Cluster CRD records unreachable API address (0.0.0.0), so no KubeRay RayCluster is ever created


NolanFoster:
**Describe the bug**

`_create_compute_cluster_crd()` (`python/michelangelo/cli/sandbox/sandbox.py:1619`) copies the API server address straight out of `k3d kubeconfig get` into the `Cluster` CRD. On a k3d sandbox that address is the **host-side** endpoint — here `https://0.0.0.0:59896`. The controllermgr runs *inside* the cluster, where `0.0.0.0` is a bind address, not a routable destination, so every attempt to create a KubeRay `RayCluster` fails with `connection refused` and all Ray-backed pipeline tasks fail.

The code documents the address it expects:

```python
# Extract host and port from server URL
# Example: "https://host.docker.internal:52910"
```

but `host.docker.internal` does not resolve from inside a pod either (confirmed below), so this path is broken in both the observed and the intended case. The port is also re-randomised on every `k3d cluster create` (58132, then 59896 across two creates in one session), so it can never be stable.

**To Reproduce**
1. `ma sandbox create` on a k3d sandbox (macOS/Colima, Apple Silicon).
2. `kubectl get clusters.michelangelo.api michelangelo-sandbox -n ma-system -o jsonpath='{.spec.kubernetes.rest.host}{":"}{.spec.kubernetes.rest.port}'` → `https://0.0.0.0:59896`
3. Run any pipeline with a Ray task.
4. `kubectl get rayclusters.michelangelo.api -A` shows CRs; `kubectl get rayclusters.ray.io -A` shows **none**.

**Expected behavior**
The `Cluster` CRD records an endpoint reachable from inside the cluster, and a `rayclusters.ray.io` object is created.

**Actual behavior**

The michelangelo `RayCluster` never becomes a KubeRay one. Status condition:

```
type: Succeeded
status: CONDITION_STATUS_FALSE
reason: ClusterCreationFailed
message: create ray cluster err:Post "https://0.0.0.0:59896/apis/ray.io/v1/namespaces/default/rayclusters":
         dial tcp 0.0.0.0:59896: connect: connection refused
state: RAY_CLUSTER_STATE_TERMINATED
```

Reachability from inside a pod (busybox, same cluster):

| Address | Result |
|---|---|
| `0.0.0.0:59896` | connection refused |
| `host.docker.internal` | NXDOMAIN |
[...]

PR #1909, opened 2026-08-22
docs: correct sandbox setup prerequisites and resource sizing

NolanFoster:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Nine corrections to the getting-started path, each traced to a specific failure hit while following the documented setup:

| Area | Change |
|---|---|
| Prerequisites | Add `docker buildx` (the kuberay script calls `docker buildx build`, but `brew install docker` ships the CLI alone) |
| Prerequisites | Note the CA-certificate fix for the Poetry install command on python.org builds |
| Colima sizing | Raise to `--cpu 6 --memory 12 --disk 100`; state memory in GiB |
| Colima sizing | Note `--vz-rosetta` for Apple Silicon |
| Quick start | Explain that the kuberay import step runs before the clusters exist, and may need a re-run |
| Troubleshooting | Add architecture and memory triage to `CrashLoopBackOff`, ahead of the delete-and-recreate escalation |
| custom-docker-images | Document `fatal error: fault` alongside `exec format error` |
| README | Add the missing kuberay build step to the quick start |
| Examples | State `poetry install -E example` where the examples are actually run |

**Why?**

Following the documented instructions from a clean macOS machine does not currently reach the guide's stated end state ("a running sandbox cluster and a successful demo pipeline run"). Several steps fail outright, and the errors they produce are not searchable in the docs.

The most consequential is the memory sizing. The guide prints `colima start --cpu 4 --memory 8 --disk 60`, which yields `Total Memory: 7.737GiB` — already below its own stated 8 GB minimum, because the flag is GiB. At that size the `bert_cola` example the repository ships cannot finish: its data-prep task succeeds, then the training task is killed by Ray's memory monitor with

```
Task was killed due to the node running low on memory.
Memory on the node ... was 7.35GB / 7.74GB (0.950357), which exceeds
the memory usage threshold of 0.95.
```

**How did you test it?**

Followed the corrected instructions end to end on macOS 15 / Apple Silicon (M1 Pro, 16 GB) with Colima, from a clean toolchain through to a pipeline run that reaches `SUCCEEDED`.

The memory change is a controlled comparison — same machine, same commit, same pipeline, same uploaded tar, with VM memory as the only variable:

| VM memory | `load_data` | `train` | Run |
|---|---|---|---|
| 7.74 GB (documented) | `[ ok ]` | OOM-killed at 7.35 GB | FAILED |
| 11.65 GB (`--memory 12`) | `[ ok ]` | `[ ok ]` | **SUCCEEDED** |

Both task outcomes were confirmed in the Ray driver logs, captured live before the cluster tears down. Every other note reproduces the exact error string quoted.

**Potential risks**

Documentation only — no code paths, no build or deployment changes.

Two judgement calls worth a reviewer's attention:

1. **Raising the stated minimum from 8 to 12 GiB will turn away users on 8 GB hosts.** That is deliberate: the sandbox cannot finish the bundled example at 8 GiB. Since the VM needs 12 GiB and the host OS needs the rest, a 16 GB machine is effectively the floor. If the project would rather keep a lower entry point, the alternative is to keep 8 GiB documented and state plainly which examples will not run there.
2. **The Minimum and Recommended memory columns now both read 12.** Only 12 was verified as working and 8 as failing; putting a higher figure under "Recommended" would be untested. Happy to adjust.

The buildx install note is Homebrew-specific and uses the Apple Silicon prefix (`/opt/homebrew/...`); Intel Macs need `/usr/local/...`.

**Breaking Changes**

- [x] No breaking changes

---

**Not included, deliberately**

- The required Python version is stated three different ways — `pyproject.toml:28` (`python = "^3.9.0"`), `.envrc:30` (asserts 3.9), and this guide (3.11 or 3.12 recommended, 3.13+ may fail). A contributor on 3.9 passes both machine-enforced checks and only discovers the real constraint when `poetry install` fails. Fixing that means tightening dependency constraints, which does not belong in a docs change.
- `ma sandbox create` never imports `ghcr.io/michelangelo-ai/examples:main`, yet the sandbox pins `pullPolicy: Never`, so the task image cannot be obtained by any documented step. Left alone pending confirmation of whether that import is intended to be part of `create`.

**Related issues**

Three defects found during the same exercise are filed separately, since they are code rather than documentation: #1901 (Cluster CRD records an unreachable `0.0.0.0` address), #1902 (`createCluster` discards the real error, surfacing `nil (not None)`), #1908 (sandbox MySQL has no data volume).

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,9 @@ cd michelangelo/python
 poetry install
 source .venv/bin/activate
 
+# Build the local-only images the sandbox needs
+cd .. && bash scripts/kuberay/build-kuberay-images.sh
+
 # Spin up a local sandbox cluster
 ma sandbox create
 
@@ -69,6 +72,8 @@ def train(learning_rate: float = 0.01) -> str:
 def my_pipeline(learning_rate: float = 0.01):
     model = train(learning_rate=learning_rate)
 ```
+
+See the [Sandbox Setup](https://michelangelo-ai.org/docs/getting-started/sandbox-setup/) guide for prerequisites (including `docker buildx`) and platform-specific notes before running the above.
 
 For a full walkthrough, see the [Getting Started with ML Pipelines](https://michelangelo-ai.org/docs/user-guides/getting-started/getting-started) guide.
 
--- a/docs/contributing/custom-docker-images.md
+++ b/docs/contributing/custom-docker-images.md
@@ -134,6 +134,13 @@ kubectl rollout status  deployment/michelangelo-worker
 
 - **Pod stuck in `ImagePullBackOff`**: The default `pullPolicy` is `IfNotPresent`. If k3d can't find the image, make sure you ran `k3d image import` after building. Check with `k3d image ls -c michelangelo-sandbox | grep local-dev`.
 - **`exec format error`**: Architecture mismatch — you built for `arm64` but the k3d node is `amd64` (or vice versa). Rebuild with the correct `ARCH`.
+- **`fatal error: fault` or a `runtime.sigpanic` stack trace**: also an architecture mismatch, but one that emulation lets start before it dies. When binfmt emulation is registered (as it is under Colima), a mismatched binary launches and then faults inside the Go runtime instead of failing with `exec format error`. Verify by extracting the image's entrypoint and checking it against the node's architecture:
+
+  ```bash
+  cid=$(docker create <image>); docker cp "$cid:/app" ./app-bin; docker rm -f "$cid"
+  file ./app-bin    # want aarch64 on Apple Silicon, x86-64 on Intel/Linux
+  kubectl get nodes -o jsonpath='{.items[*].status.nodeInfo.architecture}'
+  ```
 - **Bazel Go binary not found**: Run any `./tools/bazel build` target first to trigger the SDK download, then re-run the `find` command.
 
 ### Cleanup
--- a/docs/getting-started/sandbox-setup.md
+++ b/docs/getting-started/sandbox-setup.md
@@ -32,6 +32,7 @@ Before you begin, make sure you have the following installed. Install commands b
 | Tool | Install (macOS) | Install (Linux) | Verify |
 |------|-----------------|-----------------|--------|
 | **Docker** | [Docker Desktop](https://docs.docker.com/get-started/get-docker) or [Colima](https://github.com/abiosoft/colima) | [Docker Engine](https://docs.docker.com/engine/install/) | `docker info` |
+| **docker buildx** | `brew install docker-buildx` (see note below) | usually bundled with Docker Engine | `docker buildx version` |
 | **kubectl** | `brew install kubectl` | [official guide](https://kubernetes.io/docs/tasks/tools/#kubectl) | `kubectl version --client` |
 | **k3d** | `brew install k3d` | [official guide](https://k3d.io/#installation) | `k3d --version` |
 | **Helm** | `brew install helm` | [official guide](https://helm.sh/docs/intro/install/) | `helm version` |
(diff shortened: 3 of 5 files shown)


paulzim (approved):
## Review: #1909 — docs: correct sandbox setup prerequisites and resource sizing

**Verdict: Approve**

A single commit (`6273b299`) across five files (+47/-5), all docs. The author (Nolan Foster) ran the documented sandbox setup end-to-end on a clean Apple Silicon Mac and is correcting it against what actually happened, including a controlled before/after comparison (7.74GB Colima → `bert_cola` train task OOM-killed by Ray; 12GiB Colima → same pipeline succeeds). Two judgment calls are called out explicitly by the author for reviewer attention, and I agree with both as written.

### What was verified

- **`docker buildx` prerequisite.** `scripts/kuberay/build-kuberay-images.sh` does call `docker buildx build --platform "$BUILD_PLATFORM" ...` for both the collector and historyserver images — confirmed by reading the script. `brew install docker` really does ship the CLI without the buildx plugin, so the new prerequisites-table row and Homebrew-linking callout are accurate.
- **"Note on step 2" (import runs before clusters exist).** Read the script's k3d-import loop directly:
  ```bash
  if k3d cluster list "$cluster" >/dev/null 2>&1; then
    k3d image import ... -c "$cluster"
  else
    echo "k3d cluster '$cluster' not found — skipping (run \`ma sandbox create\` first)"
  fi
  ```
  This exactly matches the PR's claim — the skip message is expected pre-`ma sandbox create`, and a second run is needed afterward to actually import. This is also a nuance our own local sandbox notes don't spell out, despite already knowing the historyserver-must-be-built-locally requirement.
- **README quick start was missing the build step entirely.** Confirmed against current `README.md` — before this PR, `ma sandbox create` is invoked with no prior mention of `build-kuberay-images.sh` anywhere in the file. The added step closes a real gap for a first-time follow-along.
- **`bert_cola` → `datasets` import chain.** `python/examples/bert_cola/bert_cola.py` imports `from examples.bert_cola.data import load_data`, and `data.py` does `import datasets`, `import ray`, `import transformers` directly. `datasets` is only present in the `example` extra in `pyproject.toml` (confirmed current extras list), not in the base install — so "plain `poetry install` → `No module named 'datasets'`" is accurate.
- **`-E example` vs. a narrower extra for `train-and-register-a-model.md`.** This doc's own code samples use both `sklearn.ensemble.RandomForestRegressor` (Simple Training Example) and `LightningTrainer`/`pytorch_lightning` (distributed example). I checked whether the narrower `trainer` extra (`ray, torch, pytorch_lightning, transformers, numpy`) would be a tighter fit — it isn't, because `trainer` omits `scikit-learn`, which the first code sample needs. `example` is a strict superset that covers both; there's no existing narrower extra that would. So the PR's choice here is correct, not just convenient.
- **Version-mismatch and `ghcr.io/.../examples:main` image-import scope calls.** Both flagged explicitly by the author as deliberately out of scope for a docs-only PR (dependency-constraint fix, and unclear intended behavior needing confirmation) rather than silently ignored. Reasonable to leave for separate issues, especially since three related code-defect issues (#1901, #1902, #1908) are already filed for adjacent problems.
- **Colima memory math and `--vz-rosetta`/QEMU emulation claims** can't be independently reproduced from here (no Colima/Docker on this side), but they're internally consistent with the reported before/after test and with well-understood Colima VM-overhead and Rosetta-vs-QEMU behavior. The exact Ray OOM message quoted (`... 7.35GB / 7.74GB (0.950357), which exceeds the memory usage threshold of 0.95`) is Ray's real memory-monitor log format.

### Non-blocking

- Raising the stated floor from 8→12 GiB (and setting both Minimum/Recommended to 12) will exclude 8GB-host readers entirely for the bundled example — the author flags this as a deliberate tradeoff since the documented example genuinely doesn't finish at the old setting. No better option here short of also shrinking the example's resource footprint, which is out of scope for a docs PR.

### CI

All green, including `validate-docs`: builds (amd64/arm64), bazel-build, python-build, ruff, dirty-check, license/cla. `isDraft: false`, `mergeable: MERGEABLE`.

### Summary

Every checkable claim held up against the actual scripts, source imports, and dependency manifests. The PR is transparent about its own scope limits and judgment calls rather than glossing over them. Good to merge.

PR #1728, opened 2026-08-08
feat(model): add Model API-hook wiring skeleton (Phase 0)

sallycr:
Part of #1720 — Phase 0 of the `michelangelo/environment` label defaulting/inheritance migration. This issue tracks all phases; do not auto-close it from this PR (or any single-phase PR) — close it manually only once all planned phases have landed.

**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Adds a `Model` API-hook wiring skeleton: a new `go/components/model/apihook` package with a `RegisterModelAPIHook` constructor and an `apiHook` struct embedding `v2.NoopModelAPIHook`, mirroring the existing `pipelinerun`/`triggerrun` apihook packages. `BeforeCreate`/`BeforeUpdate` are explicit pass-throughs to the embedded Noop with zero behavior change. Wires the hook into `go/cmd/apiserver/main.go` via one new `fx.Invoke(modelapihook.RegisterModelAPIHook)` line.

This is Phase 0 of a larger environment-label defaulting/inheritance feature (migrated from an internal apiserver feature). It is a pure wiring skeleton — no label logic, no proto/CRD changes, no config plumbing. Real `BeforeCreate`/`BeforeUpdate` defaulting logic lands in a follow-up commit on this same PR/branch (Phase 1), once the preceding label-constant and config plumbing phases are in place, per the design in tracking issue #1720.

**Why?**
Establishes the `Model` API-hook extension point so subsequent commits can add environment-label defaulting/inheritance behavior without needing to first touch the wiring/constructor shape. Splitting the wiring skeleton out as its own commit keeps each diff reviewable and isolates the (zero-risk) fx wiring change from the (behavior-affecting) label logic that follows.

**How did you test it?**
- `gofmt -l .` — clean
- `go vet ./components/model/... ./cmd/apiserver/...` — clean
- `go build ./components/model/... ./cmd/apiserver/...` — passes
- `go test ./components/model/...` — new `TestBeforeCreate_PassesThroughToNoop`/`TestBeforeUpdate_PassesThroughToNoop` smoke tests pass
- `go test ./cmd/apiserver/...` — existing `TestDependenciesAreSatisfied` (`fx.ValidateApp(opts())`) passes, confirming the new `fx.Invoke` line resolves against the existing dependency graph with no new providers needed
[...]

PR #1912, opened 2026-08-22
feat(model): default and inherit the environment label on Models

sallycr:
**Note:** a direct commit to this branch briefly changed `helm/michelangelo/values.yaml`'s `apiserver.pipelineRunDefaults.environment` default from `""` to `"development"`, which contradicted the comment immediately above it in the same file and undid this feature's core design point (unconfigured installs must get the write-time sentinel `"unspecified"`, never a guessed real environment). This has been reverted back to `""` — flagging here so it isn't silently reintroduced. Separately, `helm/michelangelo/values-k3d.yaml` (the local sandbox overrides file) now sets `apiserver.pipelineRunDefaults.environment: "development"` explicitly — that's a deliberate, scoped sandbox-only override (a sandbox genuinely is a development environment), not a change to the shared chart default.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Promotes `EnvironmentLabel` (`pipelinerun.michelangelo/environment`) into a single canonical `go/api.EnvironmentLabel` constant, replacing the two independent literals in `go/worker/workflows/trigger/cron_trigger_workflows.go` and `go/components/triggerrun/schedule_input.go`, and the raw string in `go/components/pipelinerun/controller.go`'s `getEnvironment()`. Mechanical rename, no behavior change.
- Adds `go/api/handler.Config`, sourced from a new `apiserver.pipelineRunDefaults.environment` Helm value / apiserver ConfigMap block, so PipelineRun/Model create paths can share one operator-configured environment-label default.
- Implements `BeforeCreate`/`BeforeUpdate` in `go/components/model/apihook` (previously a Phase-0 no-op skeleton from #1728): sets `api.EnvironmentLabel` to the configured default (or the new write-time sentinel `api.UnspecifiedEnvironment = "unspecified"` when unconfigured) if absent, then overrides it with the source `PipelineRun`'s value when `Spec.SourcePipelineRun` is set, resolves, and itself carries the label. Wires `go/cmd/apiserver/main.go` to provide the new `Config` and thread its value into `RegisterModelAPIHook`'s new `defaultEnv` parameter.
- Sets `apiserver.pipelineRunDefaults.environment: "development"` in `helm/michelangelo/values-k3d.yaml` so the local k3d sandbox gets a real, deliberate value instead of the `"unspecified"` sentinel — the shared `values.yaml` default stays `""`.
- Scope is intentionally label-only: description-length validation, pipeline-type/owner/revision label copy, LDAP validation, deployment-annotation stripping, and any label-length truncation guard are **not** implemented (K8s API-server-side label-length validation applies instead of a silent truncation).

**Why?**
`go/api` had no `Model`-scoped environment-label defaulting/inheritance at all, and the environment-label literal was duplicated across two packages with no shared constant. This closes that parity gap so downstream consumers building on `go/api`'s `Model` create/update path get environment-label defaulting/inheritance, using one canonical constant and one operator-configured default shared with the `PipelineRun` path.

**Called-out design decisions (see PR discussion for sign-off):**
1. **Get-error handling:** a `Get` failure resolving the source `PipelineRun` (not-found *or* any other error) is logged and swallowed, not propagated — this is a best-effort label-inheritance lookup that must never block Model creation, matching the sibling `pipelinerun/apihook.go` pattern.
2. **No context timeout override:** the caller's `ctx` is threaded straight through with no explicit deadline, matching the OSS sibling `pipelinerun/apihook.go` pattern.
3. **No label-length truncation guard**: an overlong label from a source `PipelineRun` is written as-is and will be rejected by the K8s API server's own label-length validation, rather than silently dropped/truncated.
4. **`BeforeUpdate` re-derives the label from `Spec.SourcePipelineRun` on every update** (not just at create), the same as `BeforeCreate` — the source `PipelineRun`'s label is treated as the source of truth, so a manually-edited `Model` label is overwritten again on the next update if a source is still set. Raised in review and **confirmed intentional**, not a bug.

**How did you test it?**
- `go build ./...` — clean.
- `go test ./components/model/... ./api/... ./cmd/apiserver/... ./components/triggerrun/... ./worker/workflows/trigger/... ./components/pipelinerun/...` — all pass, including 10 new table-driven cases in `go/components/model/apihook/apihook_test.go` (default-when-absent, default-to-unspecified-when-unconfigured, inherit-from-source, preserve-explicit, source-not-found, source-Get-error, source-has-no-label, BeforeUpdate parity, and hook registration).
- `gofmt -l .` — clean.
- `golangci-lint run ./...` — clean (no new findings; pre-existing `godox` TODO warnings elsewhere in the repo are unrelated to this diff).
- `bazel run //:gazelle -- update ...` to regenerate the touched `BUILD.bazel` files.
- `python3 -c "import yaml; yaml.safe_load(...)"` to validate `values-k3d.yaml` syntax (no `helm` binary available in this environment to run `helm lint`).

**Potential risks**
- Any `Model` created without an explicit environment label will now get one (`"unspecified"` by default, or the operator's configured value) where previously no label was ever set — purely additive, since no prior OSS behavior existed for `Model`.
- The shared `apiserver.pipelineRunDefaults.environment` Helm value defaults to `""`, so existing installs see no behavior change until they set it. The k3d sandbox overrides this to `"development"` in `values-k3d.yaml` only — that override is intentionally scoped to local sandbox use and does not affect the shared chart default.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
Adds an optional `apiserver.pipelineRunDefaults.environment` Helm value (default `""`) and a new `go/components/model/apihook` package that defaults/inherits the `pipelinerun.michelangelo/environment` label on `Model` create/update. The local k3d sandbox defaults this value to `"development"`.

**Documentation Changes**
N/A — `helm/michelangelo/README.md#values-reference` documentation for the new value is left to a follow-up (this PR is scoped to Model-side label defaulting; the `PipelineRun` create-time defaulting and trigger-fire propagation logic that would consume the same config value are a separate, later change). Note: that Values Reference table is already non-exhaustive for other existing values, so this isn't a new gap introduced by this PR.

### DIFF
--- a/go/api/api.go
+++ b/go/api/api.go
@@ -55,7 +55,7 @@ const (
 	// a PipelineRun or Model belongs to. This is the single canonical
 	// definition, shared by go/worker's trigger-fire path, go/components/triggerrun,
 	// go/components/pipelinerun, and go/components/model/apihook.
-	EnvironmentLabel = "pipelinerun.michelangelo/environment"
+	EnvironmentLabel = "michelangelo/environment"
 
 	// UnspecifiedEnvironment is the write-time sentinel written to
 	// EnvironmentLabel by create/propagation logic when the operator has
--- a/go/api/api.go
+++ b/go/api/api.go
@@ -50,6 +50,21 @@ const (
 	// UpdateTimestampLabel is used to record the last time the object is updated.
 	// The time is stored in Unix microseconds.
 	UpdateTimestampLabel = "michelangelo/UpdateTimestamp"
+
+	// EnvironmentLabel records the environment (e.g. "staging", "production")
+	// a PipelineRun or Model belongs to. This is the single canonical
+	// definition, shared by go/worker's trigger-fire path, go/components/triggerrun,
+	// go/components/pipelinerun, and go/components/model/apihook.
+	EnvironmentLabel = "pipelinerun.michelangelo/environment"
+
+	// UnspecifiedEnvironment is the write-time sentinel written to
+	// EnvironmentLabel by create/propagation logic when the operator has
+	// configured no default value. It is distinct from the pre-existing,
+	// read-time "unknown" fallback used elsewhere (e.g.
+	// go/components/pipelinerun/controller.go's getEnvironment) for objects
+	// whose label is genuinely absent when read — that fallback means "could
+	// not be determined"; this one means "no default was configured."
+	UnspecifiedEnvironment = "unspecified"
 )
 
 // DefaultContextTimeout defines the default timeout for the context
--- a/go/api/handler/BUILD.bazel
+++ b/go/api/handler/BUILD.bazel
@@ -5,6 +5,7 @@ go_library(
     srcs = [
         "blob_handler.go",
         "builder.go",
+        "config.go",
         "handler.go",
         "interfaces.go",
         "k8s_handler.go",
@@ -37,6 +38,7 @@ go_library(
         "@io_k8s_sigs_controller_runtime//pkg/controller/controllerutil:go_default_library",
         "@org_golang_google_grpc//codes:go_default_library",
         "@org_golang_google_grpc//status:go_default_library",
+        "@org_uber_go_config//:go_default_library",
         "@org_uber_go_fx//:go_default_library",
         "@org_uber_go_zap//:go_default_library",
     ],
--- a/go/api/handler/config.go
+++ b/go/api/handler/config.go
@@ -1,0 +1,25 @@
+package handler
+
+import "go.uber.org/config"
+
+// configKey is the config-provider path for Config, matching the apiserver
+// ConfigMap's `apiserver.pipelineRunDefaults` YAML block.
+const configKey = "apiserver.pipelineRunDefaults"
+
+// Config holds apiserver-wide default values applied when creating
+// PipelineRun/Model objects, configurable via the Helm value
+// apiserver.pipelineRunDefaults.environment.
+type Config struct {
+	// PipelineRunDefaultEnvironment is the environment label value used when
+	// a PipelineRun or Model is created without one. Empty means the operator
+	// configured no default; callers fall back to api.UnspecifiedEnvironment
+	// rather than treating empty as a valid label value.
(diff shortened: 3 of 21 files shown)


kenns29 (approved):

sallycr:
Ran a sandbox validation (k3d + Temporal) of this hook against both currently-existing `michelangelo-examples` pipelines, run sequentially (not simultaneously):

**xgb_train** — PASS. Pushed `model-20260824-234115-ed19dd6d`, labeled `pipelinerun.michelangelo/environment: "development"` (the configured sandbox default), `spec.source_pipeline_run` absent, no create-call errors.

**pytorch_train** — PASS. Pushed `model-20260824-234713-09628403`, same result: default-only environment label, no `source_pipeline_run`, no create-call errors.

For both pipelines the inheritance branch (`applyEnvironmentLabel`'s lookup of `Spec.SourcePipelineRun`) never fires, since neither pipeline's pusher ever sets that field today — so in practice both `Model`s get the label purely from the `setIfAbsent` default path. No regression observed in either run.

(Side note, unrelated to this PR's correctness: while tracing task-pod env vars during this validation, found `MA_PIPELINE_RUN_NAME`/`MA_NAMESPACE` are already injected into every task pod via `executeworkflow.go`, which is useful context for anyone looking at wiring up `source_pipeline_run` population in the future.)

sallycr:
Re-ran the sandbox validation against the latest push (the `michelangelo/environment` label rename in 74868c2). Both pipelines still PASS:

**xgb_train** — pushed `model-20260825-045536-83f6e045`, labeled `michelangelo/environment: "development"`.

**pytorch_train** — pushed `model-20260825-045939-6d3ed2bd`, labeled `michelangelo/environment: "development"`.

Same as before: default-only labeling via `setIfAbsent`, no `source_pipeline_run` set by either pipeline so the inheritance branch doesn't fire, no create-call errors from the hook. The rename is reflected correctly end-to-end.

PR #1913, opened 2026-08-22
feat(model): add Information and Performance tabs to the Model detail page

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Trained models previously had only a list view — clicking into one had nowhere to go. This adds
a detail view for the Model entity with two tabs and a filled-out header row:

- **Header row** — source pipeline run (linked), owner, creation time, last updated, and type.
- **Information tab** — a link to the pipeline run that produced the model, and the model's
  free-text description.
- **Performance tab** — resolves the model's referenced evaluation report (`GetEvaluationReport`)
  and lists the titles of the charts it contains.

Also adds a top margin to the shared `DetailView` layout so page content isn't flush against the
sticky navigation bar above it — a pre-existing gap that applies to every entity's detail page,
not just Model's, fixed here since this PR is what surfaced it.

Known limitations, intentionally left as follow-up work:
- The Information tab does not yet show which deployments are currently running this model —
  that needs its own cross-entity query against the deployment service with its own
  loading/empty/table-row states.
- The Performance tab renders chart *titles* only. Rendering a chart's actual content (as a
  table, line chart, histogram, etc., depending on its type) needs a per-chart-type rendering
  approach and, for non-table chart types, a charting library — neither of which exist in this
  codebase yet. This PR proves out the fetch/loading/empty-state plumbing ahead of that work.

**Why?**

Trained models are a core Studio entity, but there was no way to drill into one. Landing the
detail-view registration, header metadata, the Information tab, and the evaluation-report fetch
as one reviewable unit unblocks further tabs (a deployments table, richer chart rendering) as
smaller, independently reviewable follow-ups.

**How did you test it?**

- Added integration-style tests to `packages/core/config/entities/model/__tests__/model-page.test.tsx`
  that render through `EntityDetailRoute` and mock only the RPC boundary (`GetModel`,
  `GetEvaluationReport`), covering: the header metadata row (source-run link, owner, type),
  description rendering (present/absent), the source-pipeline-run link, "no performance report
  configured", chart titles listed from a fetched report, and "report has no charts".
- `yarn typecheck`, `yarn test` (full `packages/core` suite — 146 files / 1347 tests), and
  `yarn lint` all pass for the changed package; `yarn format` reports no diff.

**Potential risks**

Low — this adds a new `views` entry to `MODEL_ENTITY_CONFIG`, two new files, and a `marginTop`
on the shared `DetailView` wrapper (affects spacing on every entity's detail page, verified
against existing tests with no failures). No existing behavior otherwise changes.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/javascript/packages/core/components/views/detail-view/detail-view.tsx
+++ b/javascript/packages/core/components/views/detail-view/detail-view.tsx
@@ -18,7 +18,14 @@ export function DetailView({
   const [css, theme] = useStyletron();
 
   return (
-    <div className={css({ display: 'flex', flexDirection: 'column', gap: theme.sizing.scale800 })}>
+    <div
+      className={css({
+        display: 'flex',
+        flexDirection: 'column',
+        gap: theme.sizing.scale800,
+        marginTop: theme.sizing.scale600,
+      })}
+    >
       <DetailViewHeader
         title={title}
         subtitle={subtitle}
--- a/javascript/packages/core/config/entities/model/__tests__/model-page.test.tsx
+++ b/javascript/packages/core/config/entities/model/__tests__/model-page.test.tsx
@@ -2,11 +2,15 @@ import { render, screen } from '@testing-library/react';
 import { vi } from 'vitest';
 
 import { TRAIN_PHASE } from '#core/config/phases/train';
+import { EntityDetailRoute } from '#core/router/entity-detail-route';
 import { PhaseListRoute } from '#core/router/phase-list-route';
 import { buildWrapper } from '#core/test/wrappers/build-wrapper';
 import { getErrorProviderWrapper } from '#core/test/wrappers/get-error-provider-wrapper';
 import { getRouterWrapper } from '#core/test/wrappers/get-router-wrapper';
-import { getServiceProviderWrapper } from '#core/test/wrappers/get-service-provider-wrapper';
+import {
+  createQueryMockRouter,
+  getServiceProviderWrapper,
+} from '#core/test/wrappers/get-service-provider-wrapper';
 
 describe('Model list page', () => {
   it('renders the column headers', async () => {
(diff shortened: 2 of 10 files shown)


sallycr:
While sandbox-testing this PR, found that navigating to a real model's detail page (e.g. `.../models/model-20260823-034255-3df4ae85`) rendered "Entity not found" even though the model genuinely exists in the backend.

**Root cause:** Model was list-only before this PR, so `packages/rpc/handlers.ts`'s handler map never had a `GetModel` entry (only `ListModel`). The new Information/Performance tabs call `useStudioQuery({ queryName: 'GetModel', ... })` and `GetEvaluationReport`, both of which look up their handler by name — since neither was registered, the call threw, and the query layer surfaced that as the generic "Entity not found" state regardless of whether the entity actually exists. This wasn't caught by this PR's own tests because they mock the RPC boundary directly (`createQueryMockRouter`), bypassing `handlers.ts` entirely — only real sandbox navigation exercises the actual handler registration.

**Fix (pushed as 8591c1335):** registered `GetModel` (`ModelService.getModel`) and `GetEvaluationReport` (new `EvaluationReportService`, which also wasn't wired into `services.ts`/`types.ts` at all — added it there too, following the exact same pattern as every other service).

Verified: `vitest` (26/26 across `packages/rpc` + the model-page tests), `tsc --noEmit`, `eslint`, `prettier --check` all clean. Rebuilt `michelangelo-ui:pr-1913` from a clean (`--no-cache`) build and confirmed via the deployed JS bundle that both handler registrations are present.

sallycr:
Screenshot 2026-08-23 at 12 34 04 PM

PR #1918, opened 2026-08-23
fix(native_transform): use proper ceiling division for stats batch_count

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`compute_numerical_statistics`'s `batch_count` now uses proper ceiling division (`(n + s - 1) // s`) instead of `n // s + 1`, which added a guaranteed-empty trailing batch whenever `len(aggregate_fns)` was an exact multiple of the batch size. Also folded a duplicate test's empty-batch guard into the existing exact-multiple-case test instead of keeping two nearly-identical tests.

**Why?**

`ray_data_df.aggregate()` crashes inside Ray when called with zero aggregators (`assert self._columns` in `table_block.py`, surfaced as `RayTaskError(AssertionError)`). E.g. 125 columns each needing max/min/mean/std is exactly 500 aggregate functions against a default batch size of 500 -- the old formula added a 3rd, empty batch and crashed. This was a regression from #1773's review cycle, which switched `batch_count` to derive from `len(aggregate_fns)` instead of the column-spec count; the spec-based divisor almost never divided evenly, so the exact-multiple case never surfaced until now.

**How did you test it?**

- `pytest python/michelangelo/uniflow/plugins/ray/` (excluding real-Ray-cluster integration tests): 114 passed.
- Added a regression test asserting `aggregate()` is never called with zero aggregators, using 8 aggregate functions with a batch size of 2/4 (the exact-multiple case) -- confirmed it fails against the old `// + 1` formula and passes against the fix.
- Fixed a pre-existing test (`test_batches_aggregate_calls`) that was inadvertently asserting the buggy behavior (2 calls for what should be 1, with the second an empty call that only avoided crashing because it used a mock instead of real Ray).
- `ruff format --check` / `ruff check` on touched files: clean.

**Potential risks**

None -- this only affects batch sizing for internal aggregation calls; output values are unchanged, and the fix strictly removes an invalid empty batch rather than altering any real batch's contents.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/python/michelangelo/uniflow/plugins/ray/native_transform.py
+++ b/python/michelangelo/uniflow/plugins/ray/native_transform.py
@@ -191,7 +191,9 @@ def compute_numerical_statistics(
     # ray_data_df.aggregate() with zero aggregators crashes inside Ray
     # (assert self._columns in table_block.py, surfaced as
     # RayTaskError(AssertionError)).
-    batch_count = -(-len(aggregate_fns) // numerical_statistics_batch_fn_size)
+    batch_count = (
+        len(aggregate_fns) + numerical_statistics_batch_fn_size - 1
+    ) // numerical_statistics_batch_fn_size
     for i in range(batch_count):
         batch = aggregate_fns[
             i * numerical_statistics_batch_fn_size : (i + 1)
--- a/python/michelangelo/uniflow/plugins/ray/native_transform.py
+++ b/python/michelangelo/uniflow/plugins/ray/native_transform.py
@@ -186,11 +186,11 @@ def compute_numerical_statistics(
     if not aggregate_fns:
         return numerical_stats
 
-    # Ceiling division: `// + 1` would add a guaranteed-empty trailing batch
-    # whenever len(aggregate_fns) is an exact multiple of the batch size, and
-    # ray_data_df.aggregate() with zero aggregators crashes inside Ray
-    # (assert self._columns in table_block.py, surfaced as
-    # RayTaskError(AssertionError)).
+    # `// + 1` adds a guaranteed-empty trailing batch whenever len(aggregate_fns)
+    # is an exact multiple of the batch size, and ray_data_df.aggregate() with
+    # zero aggregators crashes inside Ray (assert self._columns in
+    # table_block.py, surfaced as RayTaskError(AssertionError)) -- must be
+    # exact ceiling division, not // + 1.
     batch_count = (
         len(aggregate_fns) + numerical_statistics_batch_fn_size - 1
     ) // numerical_statistics_batch_fn_size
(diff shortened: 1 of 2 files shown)


python/michelangelo/uniflow/plugins/ray/native_transform.py
```diff
@@ -186,7 +186,12 @@ def compute_numerical_statistics(
     if not aggregate_fns:
         return numerical_stats
 
-    batch_count = len(aggregate_fns) // numerical_statistics_batch_fn_size + 1
+    # Ceiling division: `// + 1` would add a guaranteed-empty trailing batch
+    # whenever len(aggregate_fns) is an exact multiple of the batch size, and
+    # ray_data_df.aggregate() with zero aggregators crashes inside Ray
+    # (assert self._columns in table_block.py, surfaced as
+    # RayTaskError(AssertionError)).
+    batch_count = -(-len(aggregate_fns) // numerical_statistics_batch_fn_size)
```
sallycr: nit: -(-n // s) is correct but reads as a puzzle. (n + s - 1) // s or math.ceil

sallycr: Good nit, fixed — switched to `(n + s - 1) // s` in 39ef7a0f, which is exact for ints and reads more directly than the negate-negate trick.

python/michelangelo/uniflow/plugins/ray/tests/native_transform_test.py
```diff
@@ -325,6 +333,66 @@ def __init__(self, input_col, alias_name=None, **_kwargs):
         }
         self.assertEqual(set(result.keys()), expected_keys)
 
+    def test_batch_count_does_not_add_empty_trailing_batch_on_exact_multiple(self):
```
kechen2013: Wondering if this could just be the guard added to `test_batch_count_covers_all_aggregate_fns_not_just_specs` above? That one is already the exact-multiple case — same 2 columns × 4 stats = 8 fns, at `batch_fn_size=2` — and only misses the regression because its `side_effect` lambda returns `{}` for the empty call. Dropping `assert fns` into it fails on `// + 1` and passes on the fix, so the duplicated `_FakeAggFn`/specs/`expected_keys` block is buying just the `call_count` assertion. Left as is, that test keeps exercising the empty `aggregate()` call silently.

sallycr: Good catch, done — folded the guard (and a call_count assertion) into test_batch_count_covers_all_aggregate_fns_not_just_specs in 072ffe92, and dropped the duplicate test entirely.

kechen2013 (approved):

PR #1919, opened 2026-08-23
feat(native_transform): add tabular_native_transform workflow task

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds the tabular native transform workflow task -- the final piece of the native_transform migration. Applies PyTorch-native transform layers (normalization, scaling, bucketization, concatenation, custom numerical transforms) to tabular datasets via Ray Data batch inference, producing transformed datasets plus a portable `TorchTransformModule` usable at both training and serving time.

- New `workflow/tasks/tabular_native_transform/` task: entry point, incremental-training selective-refit support, sampling/conversion helpers, and full test coverage.
- New `workflow/schema/tabular_native_transform.py` config types, plus `RayDataContextConfig`/`WriteConfig` additions to `ray_data_io.py`.
- New standalone Ray Data helper modules (`uniflow/plugins/ray/data_context.py`, `uniflow/plugins/ray/parquet_io.py`) -- general-purpose, not task-specific, so factored out rather than inlined.
- Minor additive extension to `DatasetVariable`/`RayDatasetIO` to forward read/write kwargs, and new `DatasetMetadata`/`NativeTransformResult` types.

Two intentional scope simplifications vs. the internal implementation:
- `IncrementalTrainingConfig` is scoped to only the fields this task uses, omitting fields relevant only to a not-yet-migrated model-initializer consumer.
- A direct artifact URI is used in place of a model-registry lookup, following the existing `StorageBackend`-URI convention already established by the assembler/pusher tasks, rather than adding a new `ModelRegistryClient` dependency to this task.

Post-review fixes (applied after initial open):
- Fixed a misleading `ConfigurationError` message in `merge_specs_for_selective_refit` that hardcoded `mode=REUSE` regardless of the layer's actual mode -- now reports the real mode.
- Clarified `_stats_keys_for_refit`'s docstring: it invalidates stats per transform *level*, not per individual REFIT layer (a REUSE layer sharing a level with a REFIT layer is also recomputed) -- this matches the internal SDK's existing behavior, kept as-is, just documented accurately.
- Fixed `set_ray_data_context` to dedupe `retried_io_errors` against Ray's process-global `DataContext`, so repeated calls in one driver process no longer accumulate duplicate patterns.
- Replaced a couple of docstring `Example:` blocks that used literal `...` as fake doctest output (would fail under `--doctest-modules`) with plain prose.
- Rebased onto current `main`.

**Why?**

This is the last of the 13 planned native_transform migration PRs -- it's the workflow-task join point that wires together the already-merged pieces (TransformSpec DAG + layers, the Ray execution adapter, and TransformSpecIO) into a runnable pipeline task, closing out the native_transform module.

**How did you test it?**

- 83 new unit tests, all passing; 100% line coverage on every new/modified module.
- Full `workflow`, `lib/native_transform`, and `uniflow` regression suites pass (two pre-existing, unrelated Ray-cluster integration test failures reproduced identically on a clean `main` checkout).
- `ruff format`/`ruff check` clean.
- Reviewed by both the migration team (compared against the internal reference implementation for parity) and an independent blind reviewer (no internal context, reviewing purely as an external contributor); findings above were fixed and independently re-verified.

**Potential risks**

Low -- this is a new, self-contained workflow task; nothing existing calls into it yet. The one shared-surface change (`DatasetVariable`/`RayDatasetIO` gaining optional `**read_kwargs`/`**write_kwargs`) is purely additive with defaults matching prior behavior, so existing callers are unaffected.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

New `tabular_native_transform` workflow task available for pipelines needing PyTorch-native tabular transforms at both train and serve time.

**Documentation Changes**

Docstrings added throughout (Args/Returns/Raises) for all new public classes/functions; no wiki update in this PR.

### DIFF
--- a/python/michelangelo/uniflow/plugins/ray/parquet_io.py
+++ b/python/michelangelo/uniflow/plugins/ray/parquet_io.py
@@ -1,0 +1,88 @@
+"""Generic conversion of a Parquet-read config object into ``read_parquet`` kwargs.
+
+Not specific to any single workflow task — any task-level "parquet read
+config" dataclass (or pydantic model) can be converted, which is why this
+accepts a loosely-typed object rather than a single concrete config class.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import logging
+from typing import Any
+
+import ray
+from packaging import version
+
+_logger = logging.getLogger(__name__)
+
+__all__ = ["parquet_read_config_to_kwargs"]
+
+
+def _config_to_dict(prc: Any) -> dict[str, Any]:
+    """Flatten a parquet-read config object into a plain dict, dropping ``None``.
+
+    Supports plain ``dataclasses.dataclass`` instances (the convention used
+    by this repo's own workflow config schemas) and pydantic models (via
+    ``model_dump``), so callers are not forced onto one particular config
+    style.
+
+    Raises:
+        TypeError: If ``prc`` is neither a dataclass instance nor an object
+            exposing ``model_dump``.
+    """
+    if dataclasses.is_dataclass(prc) and not isinstance(prc, type):
+        return {k: v for k, v in dataclasses.asdict(prc).items() if v is not None}
+    if hasattr(prc, "model_dump"):
+        return prc.model_dump(exclude_none=True)
+    raise TypeError(
+        f"Cannot convert {type(prc).__name__} to read_parquet kwargs: expected a "
+        "dataclass instance or a pydantic model with model_dump()."
+    )
+
+
+def parquet_read_config_to_kwargs(
+    prc: Any | None, dataset_name: str | None = None
+) -> dict[str, Any]:
+    """Convert a parquet-read config object to ``ray.data.read_parquet`` kwargs.
+
+    Handles the Ray 2.50 API change where ``num_cpus``, ``num_gpus``, and
+    ``memory`` moved from ``ray_remote_args`` to top-level ``read_parquet``
+    kwargs, and flattens an ``arrow_parquet_args`` mapping into top-level keys
+    for the PyArrow reader.
+
+    Args:
+        prc: A parquet-read config object (dataclass instance or pydantic
+            model). ``None`` preserves Ray's defaults (an empty kwargs dict
+            is returned).
+        dataset_name: Selects the entry to use from an optional
+            ``override_num_blocks_per_dataset`` mapping field on ``prc``, if
+            present. Ignored when ``prc`` has no such field.
+
+    Returns:
+        A kwargs dict ready to splat into ``ray.data.read_parquet(**kwargs)``.
+    """
+    read_kwargs = _config_to_dict(prc) if prc is not None else {}
+
+    per_dataset_overrides = read_kwargs.pop("override_num_blocks_per_dataset", None)
+    if per_dataset_overrides is not None:
+        override = (
+            per_dataset_overrides.get(dataset_name)
+            if dataset_name is not None
+            else None
+        )
+        if override is not None:
+            read_kwargs["override_num_blocks"] = override
+        else:
+            read_kwargs.pop("override_num_blocks", None)
+
+    if version.parse(ray.__version__) < version.parse("2.50"):
+        ray_remote_args = read_kwargs.pop("ray_remote_args", None) or {}
+        for key in ("num_cpus", "num_gpus", "memory"):
+            if key in read_kwargs:
+                ray_remote_args[key] = read_kwargs.pop(key)
+        if ray_remote_args:
+            read_kwargs["ray_remote_args"] = ray_remote_args
+
+    read_kwargs.update(read_kwargs.pop("arrow_parquet_args", None) or {})
+    return read_kwargs
--- a/python/michelangelo/workflow/schema/native_transform.py
+++ b/python/michelangelo/workflow/schema/native_transform.py
@@ -8,8 +8,11 @@ Plain ``@dataclass`` with no Pydantic dependency, matching
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from enum import Enum
 
+from michelangelo.workflow.schema.common import (
+    IncrementalTrainingConfig,
+    TrainingTypeConfig,
+)
 from michelangelo.workflow.schema.ray_data_io import (
     ParquetReadConfig,
     RayDataContextConfig,
@@ -25,76 +28,6 @@ __all__ = [
     "TrainingTypeConfig",
     "WriteConfig",
 ]
-
-
-class TrainingTypeConfig(str, Enum):
-    """Incremental-training mode for ``tabular_native_transform``.
-
-    A config-layer value, parsed directly from pipeline configuration —
-    distinct from (and not interchangeable with) any runtime-layer
-    incremental-training enum a trainer task might construct from it. Trainer
-    tasks that build their own runtime training state are expected to derive
-    it from their own config, not from this type.
-
-    Attributes:
-        INVALID: Not set; treated the same as a fresh (non-incremental) run.
-        BASE: A base run whose fitted transform can later be reused/refit by
-            an ``INCREMENTAL`` run.
-        INCREMENTAL: Reuse (and optionally selectively refit) a base run's
-            fitted transform spec and feature statistics.
-
-    Example:
-        >>> TrainingTypeConfig.INCREMENTAL.value
-        'INCREMENTAL'
-    """
-
-    INVALID = "INVALID"
-    BASE = "BASE"
-    INCREMENTAL = "INCREMENTAL"
-
-
-@dataclass
-class IncrementalTrainingConfig:
-    """Incremental-training configuration for ``tabular_native_transform``.
-
-    A config-layer type read directly from pipeline configuration for this
-    task specifically. It is intentionally distinct from any runtime-layer
-    incremental-training spec a trainer task might build for its own model
-    training — the two live at different layers (parsed configuration vs.
-    constructed runtime state) and are not meant to be unified. Scoped to
-    only the fields this task uses: a shared, more general equivalent
-    configuration (as consumed by a model-initializer-style task elsewhere)
-    would also carry fields like ``load_optimizer_weights`` and
-    ``fused_model_submodule`` that are meaningless here — this config
-    intentionally omits them rather than porting unused surface area.
-
-    Attributes:
-        training_type: Whether this run is a base run, an incremental run,
-            or neither (``INVALID``).
-        baseline_model_uri: URI of the base run's raw model package, as
-            returned by ``StorageBackend.upload()`` — passed directly to
-            ``StorageBackend.download()`` to retrieve
-            ``transform_spec.yaml``/``transform_feature_stats.yaml`` from its
-            metadata directory. Required when ``training_type ==
-            TrainingTypeConfig.INCREMENTAL``.
-        enforce_full_reuse: When ``True`` and ``training_type ==
-            TrainingTypeConfig.INCREMENTAL``, every layer in the (optional)
-            inlined ``transform_spec`` must use
-            :attr:`~michelangelo.lib.native_transform.torch.transform_layer_spec.TransformerMode.REUSE`
-            (or the default ``INVALID``, which behaves as ``REUSE``) — no
-            refitting allowed. This is the only supported setting for now.
-
-    Example:
-        >>> IncrementalTrainingConfig(
-        ...     training_type=TrainingTypeConfig.INCREMENTAL,
-        ...     baseline_model_uri="s3://bucket/models/base-run/",
-        ... ).enforce_full_reuse
-        True
-    """
-
-    training_type: TrainingTypeConfig = TrainingTypeConfig.INVALID
-    baseline_model_uri: str | None = None
-    enforce_full_reuse: bool = True
 
 
 @dataclass
--- a/python/michelangelo/workflow/schema/ray_data_io.py
+++ b/python/michelangelo/workflow/schema/ray_data_io.py
@@ -13,6 +13,8 @@ __all__ = [
     "BatchIterConfig",
     "DataloadingConfig",
     "ParquetReadConfig",
+    "RayDataContextConfig",
+    "WriteConfig",
 ]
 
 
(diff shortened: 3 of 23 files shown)


python/michelangelo/workflow/tasks/tabular_native_transform/task.py
```diff
@@ -0,0 +1,400 @@
...
+    Raises:
+        ConfigurationError: If ``datasets`` is empty or every dataset is
+            empty, if ``config.transform_spec`` is not a dict or string
+            path, if incremental mode is active without a
+            ``storage_backend``, or if an incremental merge/reuse
+            constraint is violated (see
+            :mod:`~michelangelo.workflow.tasks.tabular_native_transform._private.incremental_training`).
+    """
+    rc = config.ray_data_context
+    set_ray_data_context(
```
kechen2013, line 110: Wondering if `object_store_memory_limit` and `wait_for_min_actors_s` should be forwarded here too? `RayDataContextConfig` exposes five knobs and this task is its only consumer in the repo, so setting either one is currently a no-op — after this call `execution_options.resource_limits.object_store_memory` is still `inf` and `wait_for_min_actors_s` still `-1`, while passing all five applies them fine. Internal gets away with the same three-of-five because `tabular_inference`/`lightning_train` share that config class and forward the rest; here nothing does. Both knobs guard the reader-vs-actor-warmup OOM, which is the shape of the `map_batches(..., concurrency=, num_gpus=)` stage this task runs. `test_forwards_ray_data_context_config` only ever sets the three, so it stays green either way.

python/michelangelo/workflow/tasks/tabular_native_transform/task.py
```diff
@@ -0,0 +1,400 @@
...
+
+    Returns:
+        A ``(transformed_datasets, updated_transform_spec,
+        computed_feature_stats)`` tuple.
+    """
+    transformed_datasets: dict[str, DatasetVariable] = {}
+    feature_stats = dict(initial_feature_stats) if initial_feature_stats else {}
+
+    dataset_order = [name for name in PREFERRED_DATASET_ORDER if name in datasets]
+    dataset_order.extend(name for name in datasets if name not in dataset_order)
```
kechen2013, line 271: Worth pinning the train-first ordering with a test? Collapsing these two lines to `dataset_order = list(datasets)` leaves all 83 tests green. Since `feature_stats` is threaded through the loop and only computed when empty, whichever split runs first is the one the stats come from — hand it `{"validation": ..., "train": ...}` and normalization would quietly fit on validation. Behavior is right today, just nothing catches it if it drifts.

python/michelangelo/workflow/schema/ray_data_io.py
```diff
@@ -94,3 +96,65 @@ class DataloadingConfig:
...
+        min_rows_per_file: Min rows per output file. Buffers small blocks
+            into fewer, larger files, reducing write task overhead and
+            storage metadata calls. ``None`` uses Ray's default.
+
+    Example:
+        >>> WriteConfig(max_rows_per_file=1_000_000)
+        WriteConfig(max_rows_per_file=1000000, min_rows_per_file=None)
+    """
+
+    max_rows_per_file: int | None = None
```
kechen2013, line 159: Wondering if these need a Ray floor guard? `pyproject.toml` resolves `ray = ">=2.41.0,<2.52.0"` on Python 3.9, but `max_rows_per_file` only lands in `Dataset.write_parquet` at 2.48 (`min_rows_per_file` at 2.43) — so 2.41–2.47 are all in range without it. It doesn't fail loudly either: `write_parquet` ends in `**arrow_parquet_args`, so the kwarg is swallowed and handed to PyArrow's `make_write_options`, which raises `TypeError: unexpected parquet write option: max_rows_per_file` from inside the write task — after the whole transform has already run. `parquet_io.py` already gates the 2.50 read-API change the same way.

python/michelangelo/uniflow/plugins/ray/parquet_io.py
```diff
@@ -0,0 +1,88 @@
...
+    if per_dataset_overrides is not None:
+        override = (
+            per_dataset_overrides.get(dataset_name)
+            if dataset_name is not None
+            else None
+        )
+        if override is not None:
+            read_kwargs["override_num_blocks"] = override
+        else:
+            read_kwargs.pop("override_num_blocks", None)
```
kechen2013, line 77: Worth reshaping the test fake for this branch? Deleting the `else: read_kwargs.pop(...)` leaves all 11 tests green — `_FakeParquetReadConfig` declares `override_num_blocks_per_dataset` but not `override_num_blocks`, so the key this drops is never present and both tests that exist for it assert `{} == {}`. The shipped `ParquetReadConfig` is the exact inverse (has `override_num_blocks`, no per-dataset map), so nothing in the repo reaches this block at all right now. A fake carrying both fields would cover the case it's actually for — train's `override_num_blocks` leaking onto validation.

python/michelangelo/workflow/schema/tabular_native_transform.py
```diff
@@ -0,0 +1,147 @@
...
+        'INCREMENTAL'
+    """
+
+    INVALID = "INVALID"
+    BASE = "BASE"
+    INCREMENTAL = "INCREMENTAL"
+
+
+@dataclass
+class IncrementalTrainingConfig:
```
kenns29, line 57: Incremental training config should not be under native transform

python/michelangelo/workflow/schema/tabular_native_transform.py
```diff
@@ -0,0 +1,147 @@
...
+    "IncrementalTrainingConfig",
+    "ParquetReadConfig",
+    "RayDataContextConfig",
+    "TabularNativeTransformConfig",
+    "TrainingType",
+    "WriteConfig",
+]
+
+
+class TrainingType(str, Enum):
```
kenns29: same TrainingType should not be under native transform, because other tasks can use it too, i.e. trainer

python/michelangelo/workflow/tasks/tabular_native_transform/task.py
sallycr: Fixed in 283587b1: forwarded object_store_memory_limit and wait_for_min_actors_s to set_ray_data_context (all 5 RayDataContextConfig knobs are now wired up), and updated test_forwards_ray_data_context_config/test_calls_set_ray_data_context_with_defaults to cover all 5.

python/michelangelo/workflow/tasks/tabular_native_transform/task.py
sallycr: Fixed in 283587b1: added test_train_processed_before_validation_regardless_of_dict_order, which passes {validation: ..., train: ...} (validation first in dict order) and asserts the first native_transform call carries train's underlying Ray Dataset via identity check, not validation's. Confirmed it catches the regression by temporarily collapsing dataset_order to list(datasets) locally -- test fails as expected, restored the real ordering logic and it passes.

python/michelangelo/workflow/schema/ray_data_io.py
sallycr: Fixed in 283587b1: added a Ray version floor guard (>=2.48 for max_rows_per_file, >=2.43 for min_rows_per_file) that raises ConfigurationError up front rather than letting the kwarg get silently swallowed by write_parquet and fail later from inside a write task. Added test_max_rows_per_file_below_ray_floor_raises / test_min_rows_per_file_below_ray_floor_raises (mocking ray.__version__) to cover it.

python/michelangelo/uniflow/plugins/ray/parquet_io.py
sallycr: Fixed in 283587b1: added override_num_blocks to _FakeParquetReadConfig (mirroring the real ParquetReadConfig) and set it alongside override_num_blocks_per_dataset in the two 'no match' tests, so they now prove the pop() actually drops a leaked value rather than trivially asserting {} == {}. Confirmed by temporarily deleting the else branch locally -- both tests fail as expected, restored the branch and they pass.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Investigated moving TrainingType out -- trainer already has its own TrainingType (lib/trainer/torch/pytorch_lightning/schema.py), but it's a different shape: plain Enum with int members (BASE_MODEL_TRAINING=0, INCREMENTAL_TRAINING=1) vs. this PR's str Enum with INVALID/BASE/INCREMENTAL. Trainer's version is actively consumed across lightning_trainer.py/_private/util.py/_private/callbacks.py and re-exported publicly, so reshaping either one to match the other is a breaking API change, not a same-PR fix. This does confirm a previously-recorded architectural deferral's trigger condition (extract a shared lib/common/schema.py once a second OSS consumer needs these training-mode types) is now firing -- documented in our internal design tracking with a recommendation to schedule that extraction (reconciling both shapes) as its own follow-up PR. Left this PR's TrainingType self-contained for now since it doesn't block merge. Happy to prioritize the follow-up if that's preferred instead.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Same tradeoff as the TrainingType comment below -- IncrementalTrainingConfig here is intentionally a scoped-down version of trainer's shared IncrementalTrainingMetadata/Spec (it says so in its own docstring), not an accidental duplicate. Moving it to a shared location means reconciling it with trainer's version rather than just relocating it, which is the same follow-up as the TrainingType extraction. Kept it here for now since it's self-contained and doesn't block merge; can fold both into the same follow-up PR.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Renamed to `TrainingTypeConfig` (commit 418e11b9) — this is a config-layer value parsed directly from pipeline configuration for this task, distinct from any runtime-layer training-mode value a trainer task might construct for its own training loop. They're not the same concept despite the name collision, so rather than reusing/merging into trainer's `TrainingType`, I renamed this one to make that distinction explicit and avoid implying they're interchangeable. Docstring updated to say so.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Kept `IncrementalTrainingConfig` here (no name collision with trainer's own incremental-training types, which are named differently), but strengthened its docstring to be explicit that this is a config-layer type scoped to this task's own pipeline configuration — intentionally distinct from any runtime-layer spec a trainer task might build for itself, and intentionally narrower than a more general shared configuration (which would carry fields like `load_optimizer_weights`/`fused_model_submodule` that don't apply here). Let me know if you think it still needs to move somewhere else, but I don't think it should be unified with trainer's types since they operate at different layers.

python/michelangelo/workflow/schema/native_transform.py
```diff
@@ -0,0 +1,158 @@
+"""Configuration dataclasses for the ``tabular_native_transform`` workflow task.
```
sallycr, line 1: change this file name from tabular_native_transform.py to native_transform.py.

python/michelangelo/workflow/schema/tabular_native_transform.py
```diff
@@ -0,0 +1,158 @@
...
+    "IncrementalTrainingConfig",
+    "ParquetReadConfig",
+    "RayDataContextConfig",
+    "TabularNativeTransformConfig",
+    "TrainingTypeConfig",
+    "WriteConfig",
+]
+
+
+class TrainingTypeConfig(str, Enum):
```
sallycr: move to common.py as it's shared across tasks.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: move to common.py as it's used across tasks.

python/michelangelo/workflow/schema/tabular_native_transform.py
```diff
@@ -0,0 +1,158 @@
...
+            an ``INCREMENTAL`` run.
+        INCREMENTAL: Reuse (and optionally selectively refit) a base run's
+            fitted transform spec and feature statistics.
+
+    Example:
+        >>> TrainingTypeConfig.INCREMENTAL.value
+        'INCREMENTAL'
+    """
+
+    INVALID = "INVALID"
```
hwaminkim: This field will be removed. (no changes to DEFAULT)

python/michelangelo/workflow/schema/native_transform.py
sallycr: Done in 556dc0bf: renamed workflow/schema/tabular_native_transform.py to native_transform.py, and updated every import site (task.py, _private/incremental_training.py, __init__.py, test files) accordingly.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Done in 556dc0bf: moved TrainingTypeConfig into a new workflow/schema/common.py, since it's a config-layer type meant to be shared across workflow tasks rather than scoped to this one task's schema module. native_transform.py now imports it from .common and re-exports it in its own __all__ for convenience.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Done in 556dc0bf: moved IncrementalTrainingConfig into the same new workflow/schema/common.py alongside TrainingTypeConfig, for the same reason -- it's a config-layer type meant to be shared across tasks rather than scoped to this one.

python/michelangelo/workflow/schema/tabular_native_transform.py
sallycr: Done in 556dc0bf: removed the INVALID member from TrainingTypeConfig (now just BASE/INCREMENTAL). IncrementalTrainingConfig.training_type now defaults to None instead of TrainingTypeConfig.INVALID, so the field's default value is unchanged in behavior -- is_incremental/is_baseline already only match on the INCREMENTAL/BASE values, so None (like INVALID before it) is 'neither' either way. Added a regression test pinning that default behavior.

kenns29 (approved):

hwaminkim (approved):
LGTM

PR #1920, opened 2026-08-24
feat(tuner): add a Ray Tune helper for search on RayTask clusters

### LINKED ISSUE #1695: Ray tune plugin for uniflow hyperparameter tuning
Getting started overview's ecosystem mapping table frames hyperparam tuning (optuna, ray tune) as "uniflow tasks with ray tune", i.e., HPO is positioned as a plugin slot rather than a built in subsystem. No such plugin ships yet.

Since the ray plugin already handles cluster provisioning for uniflow tasks, a ray tune plugin can compose with it rather than introduce new infrastructure - a tune task type that takes a search space, scheduler (ASHA to start) and trial budget, fans trials out on the provisioned ray cluster, and reports the best trial's parameters and checkpoint as task o/p downstream tasks can consume. Dependency light (behind the existing ray extra plus tune), a sample pipeline under examples, and tests against a local ray session so CI needs no cluster.

Building HPO as a first class subsystem (heavier, and against the mapping table's own framing), or starting with optuna (no cluster reuse story; could follow the same plugin shape later if there is demand)

**Component**
Which part of the project does this affect?
- [x] Python SDK (`michelangelo` package)
- [ ] `ma` CLI
- [ ] API server / gRPC
- [ ] UI (Studio)
- [x] Documentation
- [ ] Other

**Additional context**
Would you accept a PR along these lines or is someone already working on HPO?

apurvapatkeshwar:
## Summary

Closes #1695.

Adds `michelangelo.lib.tuner.ray_tune`: a `TuneParam` dataclass plus a `tune()` function that run a `ray.tune` search on the Ray runtime the calling process is already connected to. Inside a `RayTask`-decorated Uniflow task that is the provisioned cluster - `RayTask.pre_run()` has already called `ray.init()` by the time the task body executes - so **one cluster runs all trials**:

```python
import michelangelo.uniflow.core as uniflow
from michelangelo.uniflow.plugins.ray import RayTask
from michelangelo.lib.tuner.ray_tune import TuneParam, tune

def objective(config: dict) -> None:
    loss = train_once(lr=config["lr"], depth=config["depth"])
    ray.tune.report({"val_loss": loss})

@uniflow.task(config=RayTask(head_cpu=2, worker_cpu=4, worker_instances=4))
def tune_step(data_url: str) -> dict:
    return tune(TuneParam(
        trainable=objective,
        param_space={"lr": ray.tune.loguniform(1e-4, 1e-1), "depth": ray.tune.randint(2, 10)},
        metric="val_loss", mode="min", num_samples=20,
    ))
```

The result dict (`best_config`, `best_metrics`, `checkpoint_path`, `num_trials`, `num_errors`) is plain data any downstream task can consume, mirroring the XGBoost trainer's contract.

Today's only sweep mechanism is the hand-written loop in `workflow-patterns.md`, which provisions **one Ray cluster per configuration** - N clusters for an N-point sweep. This helper is the "compose with the ray plugin rather than introduce new infrastructure" option #1695 proposed: ASHA by default (early-stops weak trials), `"fifo"` or any `TrialScheduler` instance accepted, per-trial resources via `ray.tune.with_resources`, checkpoints wherever the caller's `RunConfig` points (pass `create_run_config()` for UniFlow-managed storage, same as the trainers).

## Design notes

- **Library, not a new plugin layer.** The three-layer plugin machinery (Starlark + Go worker) exists to add cluster-provisioning capability; tune needs none - it runs inside an already-initialized Ray driver. This follows the shape #1793 established for the XGBoost trainer, and #1695's own preference for composing with the existing ray plugin. If per-trial resource semantics ever need cluster-shape awareness, that would be the trigger to revisit.
- **Builds on #1801** (now merged): GPU trials only make sense once `RayTask(worker_gpu=...)` actually reaches the cluster spec, which is what #1801 fixed.
- **Dependency-light**: the new `tuner` extra reuses already-declared optional deps (`ray`, `pandas`, `pyarrow`); no new package, no version changes.
- `tune_config_kwargs` passes anything else through to `ray.tune.TuneConfig` verbatim, but keys owned by first-class `TuneParam` fields are rejected rather than silently overridden.

## Testing

- 21 fully mocked unit tests (`ray.tune.Tuner` patched; no cluster): dataclass validation, scheduler resolution, Tuner wiring, resources wrapping, result contract, error propagation. **100% line and branch coverage** on the production module, so the 90% diff-coverage gate has headroom.
- 2 real-Ray integration tests in a separate module using the same CI skip guard as `lib/trainer`'s `test_auto_resume_integration.py` (skipped in CI, run locally / via `MICHELANGELO_RUN_RAY_INTEGRATION_TESTS=1`): a 4-point grid search on a local Ray runtime finds the known-best config, and the default ASHA path runs sampled trials to completion. Both pass locally on Ray 2.56.1 (23/23 total, ~15s).
- `ruff format` + `ruff check` clean with the repo config.
- `python/poetry.lock` regenerated for the new extra (`poetry check --lock` clean); the delta is extras-marker-only - no package or version changes.
- Docs: `overview.md`'s HPO row updated from "hand-written sweeps" to the helper (the row #1788 corrected - this PR ships the capability that makes the stronger claim true), and `workflow-patterns.md`'s sweep section gained a callout steering genuine HPO workloads here.

### DIFF
--- a/docs/getting-started/overview.md
+++ b/docs/getting-started/overview.md
@@ -56,7 +56,7 @@ If you're coming from other ML platforms, here's how familiar concepts map to Mi
 | **Data Preparation** | Pandas, Spark notebooks | **MA Studio Data Prep** or **Uniflow tasks** with Ray/Spark * |
 | **Experiment Tracking** | MLflow, Weights & Biases | **Model Registry** with automatic versioning |
 | **Model Training** | Custom scripts, Kubeflow Pipelines | **MA Studio Training** (UI) or **CanvasFlex/Uniflow workflows** (code) |
-| **Hyperparameter Tuning** | Optuna, Ray Tune | **Uniflow tasks** with hand-written sweeps * |
+| **Hyperparameter Tuning** | Optuna, Ray Tune | **Uniflow tasks** with the Ray Tune helper (`michelangelo.lib.tuner`) * |
 | **Model Storage** | S3 buckets, model registries | **Michelangelo AI Model Registry** with metadata & plugin storage |
 | **Batch Inference** | Airflow + custom scripts | **Uniflow tasks** with Ray for offline inference * |
 | **Online Serving** | TorchServe, TensorFlow Serving | **Deployment to inference server** with Triton Inference Server * |
--- a/docs/user-guides/ml-pipelines/workflow-patterns.md
+++ b/docs/user-guides/ml-pipelines/workflow-patterns.md
@@ -157,6 +157,8 @@ def hyperparameter_sweep(data_url: str, learning_rates: list[float]):
 
 Note that the for-loop pattern runs each task call sequentially. To run all of them in parallel, see [Parallel batch run](#parallel-batch-run) below.
 
+Also note that each task call in the loop provisions its own Ray cluster. For a genuine hyperparameter search (many configurations, early stopping), prefer the Ray Tune helper in `michelangelo.lib.tuner.ray_tune`: a single `RayTask`-decorated step runs all trials on one provisioned cluster via `ray.tune`, and returns the best trial's parameters and checkpoint for downstream tasks.
+
 ---
 
 ## Concurrent run
--- a/python/michelangelo/lib/tuner/__init__.py
+++ b/python/michelangelo/lib/tuner/__init__.py
@@ -1,0 +1,5 @@
+"""Hyperparameter tuning helpers.
+
+Subpackages wrap tuning backends for use inside Uniflow tasks; see
+``michelangelo.lib.tuner.ray_tune`` for the Ray Tune helper.
+"""
--- a/python/michelangelo/lib/tuner/ray_tune/__init__.py
+++ b/python/michelangelo/lib/tuner/ray_tune/__init__.py
@@ -1,0 +1,19 @@
+"""Ray Tune helper for hyperparameter search on a provisioned Ray cluster.
+
+Public surface re-exported below:
+
+* :func:`tune` -- runs a ``ray.tune.Tuner`` search on the Ray runtime the
+  calling process is already connected to and returns a small result dict.
+* :class:`TuneParam` -- dataclass holding the search configuration (trainable,
+  search space, metric, scheduler, trial budget, ...).
+"""
+
+from michelangelo.lib.tuner.ray_tune.tuner import (
+    TuneParam,
+    tune,
+)
+
+__all__ = [
+    "TuneParam",
+    "tune",
+]
--- a/python/michelangelo/lib/tuner/ray_tune/tests/__init__.py
+++ b/python/michelangelo/lib/tuner/ray_tune/tests/__init__.py
@@ -1,0 +1,1 @@
+"""Tests for the Ray Tune helper library."""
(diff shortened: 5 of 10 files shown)


paulzim (approved):
## Review: #1920 — feat(tuner): add Ray Tune helper for hyperparameter search

**Verdict: Approve**

Adds `michelangelo.lib.tuner.ray_tune` — a thin `TuneParam` dataclass + `tune()` wrapper around `ray.tune.Tuner`, matching the existing shape of `michelangelo.lib.trainer` (param dataclass in, small result dict out). Runs all trials on the single Ray cluster a `RayTask`-decorated step already provisions, instead of the hand-written for-loop pattern that spins up one cluster per configuration. New optional `tuner` extra (`ray`, `pandas`, `pyarrow`); `michelangelo.lib.tuner`'s own `__init__.py` stays import-free so pulling in the parent package doesn't require Ray to be installed.

Verified:
- Confirmed the PR's actual diff is exactly one commit (`a448d980`) by fetching `upstream/main` fresh and diffing against it — the worktree's local `main` was stale and an earlier pass against it falsely pulled in unrelated already-merged commits (dependency bumps, an `examples`-repo doc repoint from #1749). None of that is part of this PR; the real diff is limited to the new `lib/tuner` package, one line in `pyproject.toml`, marker-only `poetry.lock` churn, and two small doc edits.
- Built a scratch venv, installed the `tuner` extra plus `pytest`/`ruff`, and ran the suite for real: 21/21 unit tests pass (`ray.tune.Tuner`/`with_resources` mocked — dataclass validation, scheduler resolution, Tuner wiring, resources wrapping, result-dict contract, checkpoint-`None` mapping, error propagation from `get_best_result()`) and both real-Ray integration tests pass (`MICHELANGELO_RUN_RAY_INTEGRATION_TESTS=1`, actual local Ray runtime, real grid search and real ASHA random search both complete and pick the right best config).
- `ruff check` and `ruff format --check` both clean on the new files.
- `pyproject.toml`/`poetry.lock` change is genuinely just the new `tuner` extras marker (`ray`/`pandas`/`pyarrow` plus content-hash) — no package or version changes, matching the PR description. The two-entry `tuner = ["pandas", "pyarrow", "ray", "ray"]` in the lock's duplicate-looking `ray` entry is pre-existing behavior from `ray`'s two python-version-gated entries in `pyproject.toml` (same pattern already present in the `example` extra) — not introduced by this PR.
- Doc edits (`overview.md`'s HPO row, a 2-line steering note in `workflow-patterns.md`) are accurate against the new module and reference real, existing symbols (`create_run_config()` in `uniflow/plugins/ray/run_config.py`; `RayTask.pre_run()`/`post_run()` actually call `ray.init()`/`ray.shutdown()` as the module docstring claims).

**Two non-blocking notes:**
1. The commit message says `poetry.lock` was "deliberately untouched" (citing an xgrammar/PyPI pin concern), but the commit does contain 181 changed lines in the lock file — verified above to be a legitimate, harmless marker-only regen, not a discrepancy in outcome. Just a stale line in the commit message worth tidying if there's a follow-up push.
2. The linked issue (#1695) asked for "a sample pipeline under examples" alongside the helper itself; this PR doesn't include one. `python/examples/` was migrated out of this repo into `michelangelo-ai/michelangelo-examples` before this PR was opened, so a companion example (if wanted) would need to land there instead — worth a quick check with the author on whether that's planned as a fast-follow, similar to how `xgboost_train` was ported over.

Both notes are cosmetic/process, not correctness issues — code and tests are solid.

PR #1927, opened 2026-08-24
fix: forward an explicitly requested disk equal to the old 512Gi default

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Follow-up to #1801, addressing the review note there: a user explicitly requesting exactly `512Gi` of disk was the one value that never reached the pod spec.

The compatibility guard in #1801 forwarded the disk parameter only when it differed from the shipped `512Gi` default, so that honoring the previously-dropped default would not suddenly attach a large ephemeral-storage request to every existing RayTask. The side effect was that an explicit request for exactly `512Gi` was indistinguishable from the default and silently dropped.

This PR ships an empty-string default instead (same effective behavior: no request) and forwards any non-empty value verbatim:

- `RAY_DEFAULT_HEAD_DISK` / `RAY_DEFAULT_WORKER_DISK` shipped defaults change from `"512Gi"` to `""`.
- The default-equality guard becomes a simple non-empty check, so every explicit `head_disk`/`worker_disk` parameter value forwards, `512Gi` included.
- A deployment-level `RAY_DEFAULT_*_DISK` environment override now also reaches the pod spec, which is what configuring it asks for.

**Why?**

Users should be able to request `512Gi` -- with the old guard, that exact value was silently ignored while `511Gi` and `513Gi` both worked. The empty-string default keeps the intended compatibility behavior (tasks that say nothing about disk get no request, exactly as before #1801 and after it) without making any real value unrequestable.

**How did you test it?**

- Full ray plugin suite: 117 tests pass.
- Regression test added: an explicit `head_disk="512Gi"` / `worker_disk="512Gi"` renders `ephemeral-storage: 512Gi` on both containers (this fails on the previous guard).
- Updated tests: no disk parameter renders no ephemeral-storage request on head or workers; a deployment-level `RAY_DEFAULT_*_DISK` value now reaches the pod spec.
- `ruff check` and `ruff format --check` clean on the touched files.

**Potential risks**

- A deployment that explicitly set `RAY_DEFAULT_*_DISK` in the environment and relied on it doing nothing will now see that request applied to RayTask pods. Setting the variable is a statement of intent, so this is the fix working as designed, but it is a visible change for such deployments.
- Tasks that pass exactly `512Gi` today (expecting it to work) will start getting the request they asked for.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

An explicit RayTask disk request of exactly `512Gi` is now honored (it was previously indistinguishable from the shipped default and dropped), and `RAY_DEFAULT_HEAD_DISK` / `RAY_DEFAULT_WORKER_DISK` deployment overrides now reach the pod spec.

**Documentation Changes**

None -- parameter semantics are unchanged; only the previously-dropped values now take effect.

### DIFF
--- a/python/michelangelo/uniflow/plugins/ray/task.star
+++ b/python/michelangelo/uniflow/plugins/ray/task.star
@@ -15,12 +15,12 @@ RAY_ENV = {
 }
 RAY_DEFAULT_HEAD_CPU = os.environ.get("RAY_DEFAULT_HEAD_CPU", "8")
 RAY_DEFAULT_HEAD_MEMORY = os.environ.get("RAY_DEFAULT_HEAD_MEMORY", "32Gi")
-RAY_DEFAULT_HEAD_DISK = os.environ.get("RAY_DEFAULT_HEAD_DISK", "512Gi")
+RAY_DEFAULT_HEAD_DISK = os.environ.get("RAY_DEFAULT_HEAD_DISK", "")
 RAY_DEFAULT_HEAD_GPU = os.environ.get("RAY_DEFAULT_HEAD_GPU", "0")
 
 RAY_DEFAULT_WORKER_CPU = os.environ.get("RAY_DEFAULT_WORKER_CPU", "8")
 RAY_DEFAULT_WORKER_MEMORY = os.environ.get("RAY_DEFAULT_WORKER_MEMORY", "32Gi")
-RAY_DEFAULT_WORKER_DISK = os.environ.get("RAY_DEFAULT_WORKER_DISK", "512Gi")
+RAY_DEFAULT_WORKER_DISK = os.environ.get("RAY_DEFAULT_WORKER_DISK", "")
 RAY_DEFAULT_WORKER_GPU = os.environ.get("RAY_DEFAULT_WORKER_GPU", "0")
 RAY_DEFAULT_WORKER_INSTANCES = os.environ.get("RAY_DEFAULT_WORKER_INSTANCES", "1")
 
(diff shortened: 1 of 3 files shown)


python/michelangelo/uniflow/plugins/ray/task.star
```diff
@@ -164,12 +164,15 @@ def task(
...
-        # default (512Gi) has never reached the pod spec, so honoring it now
-        # would suddenly attach a large ephemeral-storage request to every
-        # existing RayTask; an explicit parameter or env override still wins.
-        _head_disk_request = _head_disk if _head_disk != RAY_DEFAULT_HEAD_DISK else None
-        _worker_disk_request = _worker_disk if _worker_disk != RAY_DEFAULT_WORKER_DISK else None
+        # Forward any non-empty disk value. The shipped default is now the
+        # empty string ("no explicit request"), which matches the historical
+        # behavior where the old 512Gi default never reached the pod spec --
+        # and it lets a user explicitly request any value, 512Gi included
+        # (the old default-equality guard silently dropped exactly that one
```
KyleKoon: nit: I don't think we need all this context in a comment. Consider simplifying

KyleKoon (approved):

apurvapatkeshwar:
@KyleKoon you approved this 2 weeks ago
It's green locally, but the fork CI run needs an approve-and-run click to show it
Could you please kick it off?

PR #1934, opened 2026-08-25
fix(python): allow shape=[] (true scalar) in Triton schema validation

krishpatel9:
**TL;DR:** Wiring the OSS assembler into a real scalar-feature pipeline surfaced two Triton-packaging bugs (schema validation and dependency-file walking) — fixed here, and verified with a green end-to-end sandbox run. (A third, independent bug found in the same effort — pusher artifact handling — is split out into #1948.)

## Summary

**Updated per review feedback** — the original approach (padding a scalar `ModelSchemaItem`'s empty shape to `[1]` via `normalize_scalar_shapes()`) was wrong: `item.shape` excludes the batch dimension, which Triton applies separately and flexibly (`config.pbtxt`'s `dims` never includes it; batching is controlled independently via `max_batch_size`/`dynamic_batching`). Hardcoding a batch-like dimension onto every scalar feature's schema contradicts that design.

The corrected fix makes `shape=[]` a legitimate, fully-supported case throughout validation instead of rejecting or padding it away:

- `validate_model_schema_item` no longer rejects an empty shape — a feature with no non-batch dimensions at all is a valid true scalar, not an error state.
- Removed `normalize_scalar_shapes()` entirely, along with its call sites in both `TorchTritonPackager` and `CustomTritonPackager`.
- Reverted the `get_sample_data()` reshape-to-`(1,)` change from an earlier revision of this PR — it was only needed to match schemas that `normalize_scalar_shapes` had padded to `[1]`. With the schema staying `shape=[]`, the original behavior (leave scalar sample values as 0-D, unreshaped) is correct.
- The rest of the pipeline was already architected batch-dim-agnostically and needed no changes: `validate_shape`'s `[-1, *shape]` batch handling and `_add_batch_dimension`'s single `unsqueeze(0)` both already treat an empty non-batch shape correctly.
- Also fixes the example pipeline's dependency-file walking: `find_dependency_files_internal` (the packager's static dependency-file walker) can reach unrelated stdlib/third-party modules and crash when a caller doesn't scope it with `prefixes` (e.g. importing `numpy.f2py` calls `sys.exit()`; importing CPython's own `test` package raised `AttributeError` from `torch.distributed.config`). No changes to `dependency_files.py` itself — the actual fix is scoping the walk via `include_import_prefixes` in the example pipeline's `assembler.py`, matching the convention every internal caller already follows.

**Note:** wiring the assembler into this same pipeline also surfaced an independent bug in `ModelPusherPlugin` (the next task downstream, `push_step`) — split into #1948 since it's unrelated to schema validation. Both were needed together to get the full pipeline green end-to-end (see Test plan).

## Test plan
- [x] `pytest michelangelo/lib/model_manager/ michelangelo/workflow/tasks/tabular_trainer/` — all passing (603+ tests across the affected areas).
- [x] **Ran the full pipeline end-to-end in a local k3d sandbox**, on this PR's final code combined with #1948's final code (`feature_prep` -> `preprocess` -> `train` -> `assembler` -> `push_step`, using michelangelo-ai/michelangelo-examples#36): run `run-20260827-033805-ad974087` `SUCCEEDED`, model `model-20260827-034145-8cdd3aec` registered at `models:/california-housing/model-20260827-034145-8cdd3aec/0`, both raw and deployable Triton packages uploaded to S3, train/validation datasets pushed.


python/michelangelo/lib/model_manager/packager/custom_triton/custom_triton_packager.py
```diff
@@ -146,6 +147,8 @@ def create_model_package(
         if not model_schema:
             raise ValueError("model_schema is required")
 
+        model_schema = normalize_scalar_shapes(model_schema)
```
kenns29: we don't want to do this. Here is the reason:
1) In principle, the schema should only reflect the schema for the actual dimension of the feature, whereas the batch dimension is to be left as a flexible input, we shouldn't hardcode a batch dimension for the user
2) In Triton, the config.pbtxt file doesn't require adding the batch dimension to the schema when dynamic batching is on, and when dynamic batching is off, batching is to be handled as part of the model and we shouldn't introduce additional dimension on top of that

kenns29 (approved):

PR #1945, opened 2026-08-26
feat(api): default EnvironmentLabel on PipelineRun create and update

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Adds a `PipelineRun` API hook that sets `api.EnvironmentLabel` to a configured default (or `api.UnspecifiedEnvironment` when unconfigured) whenever the label is absent from the request, on both `BeforeCreate` and `BeforeUpdate`. The label is never overwritten when already present. `BeforeUpdate` is intentionally narrow — unlike the sibling Model hook, PipelineRun has no source resource to inherit the label from, so it only ever fills in the default.

**Why?**
`PipelineRun`s created or updated without an explicit environment label previously had no default applied, making downstream filtering/reporting by environment unreliable. This brings `PipelineRun` create- and update-time behavior in line with the existing Model hook, which already defaults the label on both create and update.

**How did you test it?**
Added unit tests covering both `BeforeCreate` and `BeforeUpdate`: label absent with a configured default → label set to the default; label absent with no configured default → falls back to `api.UnspecifiedEnvironment`; label already present → preserved, not overwritten. Ran `go test ./components/pipelinerun/apihook/...` (passing), `gofmt -l .` (clean), and `golangci-lint run ./...` (no new findings).

**Potential risks**
Low. The hook only ever sets a label when absent; it does not modify any other field or reject any request. Existing callers that already send the label are unaffected.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A


npuppala111 (approved):
LGTM!

sallycr:
Sandbox-validated against a live k3d cluster, using both `michelangelo-examples` pipelines (`xgb_train`, `pytorch_train`), run sequentially (one closed before the next started).

**Baseline (before this PR):** created a `PipelineRun` against the currently-deployed apiserver — no `michelangelo/environment` label present, confirming the "before" state.

**After deploying this PR's apiserver build:**
- `xgb_train`: `PipelineRun` created with `michelangelo/environment: development` set immediately at create time (matches the sandbox's configured default). Waited for it to close — `CloseStatus: COMPLETED` via Cadence. No disruption from the new hook.
- `pytorch_train`: run only after `xgb_train` fully closed. Same result — label defaulted at create time, `CloseStatus: COMPLETED`.

**Result: PASS** for both pipelines — `BeforeCreate`'s defaulting works as designed and doesn't affect pipeline execution or completion.

One secondary, non-blocking observation while probing `BeforeUpdate` separately: a `kubectl label pipelinerun ... foo=bar` (PATCH) against a pre-existing label-less `PipelineRun` didn't appear to backfill the environment label, though this wasn't conclusively confirmed (the run in question was relocated to terminal storage before a `kubectl replace`-based follow-up could complete), and no real caller of the Update RPC path was found elsewhere in `go/` outside the generated proto/hook itself. Not a blocker for this PR's `BeforeCreate` behavior — just worth a closer look if `BeforeUpdate` correctness becomes load-bearing later.

Issue #1949, opened 2026-08-26
CVE Scan workflow failing on main: stale Go toolchain + base-image CVEs, and a noisier-than-intended trigger cadence


sallycr:
## Summary

[Run 33020130736](https://github.com/michelangelo-ai/michelangelo/actions/runs/33020130736) of the `CVE Scan` workflow failed across all 4 `scan` matrix jobs (`ui`, `apiserver`, `controllermgr`, `worker`). Investigated the failure and also the workflow's trigger cadence, which fires more often than its `schedule` cron alone would suggest.

## Root cause #1: real CRITICAL/HIGH CVEs, `exit-code: 1` is working as configured

This isn't a workflow bug — Trivy found real CVEs and the job is intentionally configured to fail on any CRITICAL/HIGH match. Downloaded the run's SARIF artifacts and inspected them directly:

- **`ui`** (base: `nginx:alpine`, Alpine 3.24.1): 10 unique CVEs, all in the base image's `openssl` package. Highest: `CVE-2026-14456` (Trivy security-severity 8.0, DoS via unbounded memory growth in QUIC server).
- **`apiserver` / `controllermgr` / `worker`** (base: `gcr.io/distroless/base-debian12`, statically-linked Go binaries): 36 results each, in three groups:
  - 15 `glibc` CVEs (severity 2.0–5.5) — OS-level, from the Debian 12 base image, several are long-standing/likely-unfixable-upstream (`CVE-2010-4756`, `CVE-2018-20796`, a few `CVE-2019-101002x`).
  - ~7 `openssl` CVEs (severity 2.0–8.0) — also base-image-level.
  - **14 Go standard-library CVEs, mostly severity 8.0** (`net/http`, `crypto/tls`, `crypto/x509`, `encoding/asn1`, `encoding/xml`, `html/template`, `mime`, `net/url`, `golang.org/x/net/idna`, `golang.org/x/net/dns/dnsmessage`) — these come from the **Go toolchain** the binaries were compiled with. `go/go.mod` currently pins `go 1.24.0` / `toolchain go1.24.3`, which predates the fixes for all 14 of these.

The Go-toolchain cluster is the most actionable and time-sensitive finding — bumping `go/go.mod`'s toolchain directive past the fix versions would clear the largest, most current group of findings.

## Root cause #2: trigger cadence is not "once a day" — it fires on nearly every merge to `main`

`.github/workflows/cve-scan.yml` triggers on:
```yaml
on:
[...]

PR #1950, opened 2026-08-26
ci: bump Trivy to v0.74.0 in CVE Scan workflow

sallycr:
## Summary
- The `CVE Scan` workflow pins `aquasecurity/trivy-action`'s `version` input to `v0.72.0`. Trivy itself flags this as stale in every run's output: "Version 0.74.0 of Trivy is now available, current version is 0.72.0".
- Bumps both occurrences (the SARIF-producing scan step and the table-output summary step) to `v0.74.0`.

Partial follow-up to #1949 — this addresses the tool-version staleness noted in that investigation. It does not address the underlying CRITICAL/HIGH CVE findings themselves (stale Go toolchain, base-image OS packages) or the trigger-cadence noise, both tracked separately in that issue.

## Test plan
- [ ] Manually trigger `CVE Scan` via `workflow_dispatch` after merge and confirm it runs successfully against Trivy v0.74.0 with no change in behavior other than the newer scanner version.

### DIFF
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -36,7 +36,7 @@ jobs:
           output: trivy-${{ matrix.service }}.sarif
           severity: CRITICAL,HIGH
           exit-code: '1'
-          version: v0.72.0
+          version: v0.74.0
 
       - name: Upload SARIF to GitHub Security tab
         uses: github/codeql-action/upload-sarif@v3
@@ -53,7 +53,7 @@ jobs:
           format: table
           severity: CRITICAL,HIGH
           exit-code: '0'
-          version: v0.72.0
+          version: v0.74.0
 
       - name: Upload Trivy SARIF report as artifact
         uses: actions/upload-artifact@v4


paulzim (approved):
## Review: #1950 — ci: bump Trivy to v0.74.0 in CVE Scan workflow

**Verdict: Approve**

One commit (`9157f715`), one file (`.github/workflows/cve-scan.yml`), two lines changed — both `version:` inputs to the `aquasecurity/trivy-action` steps, `v0.72.0` → `v0.74.0`.

### What was verified

- **The staleness claim is real, not paraphrased.** Pulled the actual logs from the run the PR references (33020130736) rather than trusting the PR body: `grep`'d for the message and found it verbatim, twice per matrix job, across all 4 jobs (`ui`, `apiserver`, `controllermgr`, `worker`) — `Version 0.74.0 of Trivy is now available, current version is 0.72.0`. Exact match to what's quoted in the PR description.
- **Both occurrences in the file are updated**, and there are only two (`git show` diff confirms it — the SARIF-producing step and the table-summary step, nothing else pinned to `v0.72.0`).
- **Scope matches the PR's own claim.** Read the full workflow: this only touches the Trivy binary version pin (the `version:` input), not the `aquasecurity/trivy-action` action version itself (`@v0.36.0`, untouched) or any of the CRITICAL/HIGH findings, trigger cadence, or Slack-notify logic that #1949 separately tracks. Correctly scoped as a partial, single-purpose follow-up.
- **Issue #1949 context checked directly** (not just taken from the PR body): it's a real, open issue from the same author, and it does correctly describe the underlying CVE findings and trigger-cadence problems as separate, undone follow-up work — this PR doesn't overclaim resolving those.

### Important — CI is red, but not because of this PR

`gh pr checks 1950` shows `bazel-build` and `dirty-check` failing. Root-caused this rather than assuming it's related:

- `dirty-check`'s gazelle run shows `proto/api/BUILD.bazel` is missing a generated dependency: `+ "@com_google_protobuf//:wrappers_proto"`. `bazel-build` fails downstream of the same gap — protoc can't resolve `Import "google/protobuf/wrappers.proto" was not found`.
- Traced the actual `import "google/protobuf/wrappers.proto";` in `proto/api/list.proto` to commit `55df4fdd` ("feat(ui): support pipeline run resume from + retry (#1858)"), already merged to `main` — the BUILD.bazel file was never regenerated with `gazelle` after that import was added.
- Confirmed this is repo-wide, not specific to this branch: an entirely unrelated open PR (`craig/multi-sandbox`) is hitting failures in the same CI window, and this branch's own base is current `upstream/main` — the bug is already on `main` regardless of what a PR touches.

This means **every PR right now** will show `bazel-build`/`dirty-check` red until someone runs `gazelle` and commits the regenerated `proto/api/BUILD.bazel` on `main`. Worth flagging to the team directly — it's blocking required checks repo-wide, not just here. Doesn't reflect on #1950's own change, which never touches `proto/`.

### CI (relevant checks only)

`filter`, `license/cla`, `update-release-draft`, `python-build` all pass. `bazel-build`/`dirty-check` fail for the unrelated pre-existing `main` breakage above. `isDraft: false`, `mergeable: MERGEABLE`.

### Summary

Small, correctly scoped, verified against actual run logs rather than the PR's own description. Safe to merge once the unrelated `main`-wide gazelle/BUILD.bazel breakage is fixed (or if that check isn't required for merge).

PR #1951, opened 2026-08-26
ci: run CVE Scan on schedule only, not on every main-branch release

sallycr:
## Summary
- `CVE Scan` currently triggers via `workflow_run` after **Dev Release** and **UI Release** complete, both of which fire on `push: branches: [main]` — i.e. on essentially every PR merge to `main` — plus once nightly via **Nightly Build**, plus its own weekly `schedule` cron.
- Since Dev Release / UI Release / Nightly Build all publish to the same mutable `:latest` tag, most of these `workflow_run`-triggered scans re-check image content that didn't meaningfully change, and re-fire the same Slack failure notification (`notify-on-failure`) for a CVE condition that won't clear until someone actually patches the underlying image/toolchain — see #1949 for the current findings.
- This PR drops the `workflow_run` triggers so `CVE Scan` runs on its `schedule` (weekly) and `workflow_dispatch` only. Also removes the now-always-true `if: github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'` guard on the `scan` job, since there's no longer a `workflow_run` event to guard against.

Trade-off: this trades "detect a new CVE as soon as a release publishes" for a large reduction in duplicate Slack noise. Anyone can still get an on-demand scan via `workflow_dispatch`.

Partial follow-up to #1949 — addresses the trigger-cadence noise specifically (fix option 5 in that issue), not the underlying CVE findings, which need separate follow-up work.

## Test plan
- [ ] Manually trigger `CVE Scan` via `workflow_dispatch` after merge and confirm it still runs the full `scan` matrix successfully.
- [ ] Confirm (over the following week) that `CVE Scan` no longer appears in the Actions history triggered by a `workflow_run` event, only by `schedule`/`workflow_dispatch`.

### DIFF
--- a/.github/workflows/cve-scan.yml
+++ b/.github/workflows/cve-scan.yml
@@ -1,10 +1,6 @@
 name: CVE Scan
 
 on:
-  workflow_run:
-    workflows: ["Dev Release", "UI Release", "Nightly Build"]
-    types: [completed]
-    branches: [main]
   schedule:
     - cron: '0 6 * * 1'
   workflow_dispatch:
@@ -21,7 +17,6 @@ env:
 jobs:
   scan:
     runs-on: ubuntu-24.04
-    if: github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'
     strategy:
       fail-fast: false
       matrix:


paulzim (approved):
## Review: #1951 — ci: run CVE Scan on schedule only, not on every main-branch release

**Verdict: Approve**

One commit (`88c5f8de`), one file (`.github/workflows/cve-scan.yml`). Drops the `workflow_run` trigger block (after Dev Release / UI Release / Nightly Build) and the now-dead `if: github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'` guard, leaving `schedule` (weekly) + `workflow_dispatch` as the only triggers. Same author and same #1949 root issue as #1950, addressing a different sub-problem (trigger cadence, option 5 in that issue) — the two are independent, non-overlapping edits to the same file.

### What was verified

- **The "fires on essentially every merge" claim is real, checked against actual run history, not just the workflow source.** Pulled the last 15 `CVE Scan` runs: every single one is `event: workflow_run`, spanning just ~27 hours (2026-08-25 20:31 → 2026-08-26 23:26), 13 of 15 `failure`. Not one `schedule` or `workflow_dispatch` run in that window — the noise is exactly as bad as described.
- **Trigger sources confirmed directly**, not taken on faith from #1949: `dev-release.yml` triggers on `push: branches: [main]` (paths-ignore `docs/**` only — so any Go/proto/helm/JS/Python merge fires it), `ui-release.yml` triggers on `push: branches: [main], paths: [javascript/**]`, `nightly.yml` runs on its own daily cron independent of merges. All three really do funnel into `CVE Scan` under the old config.
- **The dropped `if:` guard really is dead code post-change, not a behavior change in disguise.** With `workflow_run` removed from `on:`, `github.event_name` can never equal `workflow_run` again, so `github.event.workflow_run.conclusion == 'success' || github.event_name != 'workflow_run'` always evaluates true via the right-hand clause — removing it is a pure no-op cleanup.
- **No leftover references.** Grepped the full file post-change for `workflow_run`: zero remaining hits. The `notify-on-failure` job's `needs: scan` / `if: failure()` logic is untouched and doesn't depend on trigger type, so failure notification still works correctly under `schedule`/`workflow_dispatch`.
- **Coexistence with #1950 checked.** Both PRs touch `cve-scan.yml` but in disjoint regions — #1950 changes the two `version: v0.72.0` → `v0.74.0` lines inside the `steps:` block, this PR changes only the `on:` block and the job's top-level `if:`. Confirmed via `git show` on both: no overlapping lines, so either merge order applies cleanly without a conflict.

### Trade-off, called out honestly

The PR is upfront that this trades "detect a new CVE within minutes of a release" for far less Slack noise, and that `workflow_dispatch` remains available for an on-demand check. Given the current signal is 13-failures-in-15-runs of the same unpatched, already-tracked findings (per #1949), the trade is reasonable — the fast-detection benefit was already fully consumed by the first failure days ago; every run since has been a re-confirmation, not new information.

### Summary

Correctly diagnosed, correctly scoped, and backed by real run-history evidence rather than just the workflow source. Safe to merge in either order relative to #1950.

PR #1952, opened 2026-08-26
ci: pin base images by digest and track them via Dependabot

sallycr:
## Summary
- `docker/service.Dockerfile` and `docker/ui.Dockerfile`/`docker/ui-prebuilt.Dockerfile` pulled `gcr.io/distroless/base-debian12` and `nginx:alpine` as floating tags with no digest pin — so each build silently picked up whatever those tags resolved to at build time, and Dependabot's `docker` ecosystem (which needs a pinned reference to diff against) had nothing to track. `.github/dependabot.yml` also had no `docker` entry at all, so nothing was watching these Dockerfiles either way.
- Pins both images to their current digest (`image:tag@sha256:...`), making builds reproducible.
- Adds a `docker` package-ecosystem entry for `/docker` to `dependabot.yml` so future base-image updates show up as routine PRs instead of only surfacing via a red CVE-scan investigation.

Partial follow-up to #1949 — addresses root cause #1's OS-level (glibc/openssl) CVE cluster by giving it an update path going forward. It does **not** retroactively clear the CVEs already present in the currently-pinned digests (some of the glibc ones are old and marked will-not-fix upstream — those need a `.trivyignore` with justification, not a base-image bump), and it does not touch the separate Go-toolchain-CVE cluster from #1949, which needs a `go.mod` version bump rather than a Dockerfile change.

## Test plan
- [ ] Confirm `docker build -f docker/service.Dockerfile ...` and the UI image build still succeed with the pinned digests.
- [ ] After merge, confirm Dependabot opens its first `docker` PR on the next weekly run (or trigger a manual Dependabot check) to validate the new ecosystem entry is picked up correctly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -65,3 +65,11 @@ updates:
     groups:
       website-minor-and-patch:
         update-types: ["minor", "patch"]
+
+  - package-ecosystem: "docker"
+    directory: "/docker"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 3
+    commit-message:
+      prefix: "chore(docker)"
--- a/docker/service.Dockerfile
+++ b/docker/service.Dockerfile
@@ -1,7 +1,10 @@
 # Dockerfile for Michelangelo services.
 
 # Distroless: https://github.com/GoogleContainerTools/distroless
-FROM gcr.io/distroless/base-debian12
+# Pinned by digest (rather than the floating `latest` tag) so Dependabot can
+# track and propose base-image updates -- see .github/dependabot.yml's
+# "docker" entry.
+FROM gcr.io/distroless/base-debian12:latest@sha256:fabbf1c0c357a3d42550111351daed089b20a2c954df13ee2fcff60602515e84
 
 # Path to the service binary built by the BAZEL_TARGET.
 # The path must be relative to the repository root.
--- a/docker/ui-prebuilt.Dockerfile
+++ b/docker/ui-prebuilt.Dockerfile
@@ -1,5 +1,8 @@
 # Simple UI Dockerfile - expects pre-built assets
-FROM nginx:alpine
+# Pinned by digest (rather than the floating `alpine` tag) so Dependabot can
+# track and propose base-image updates -- see .github/dependabot.yml's
+# "docker" entry.
+FROM nginx:alpine@sha256:db35bfc6b2951e7f8a72db5db120288c127ffaeeb4a6d4b95a26fead017d5913
 
 # Copy pre-built UI assets
 COPY javascript/app/dist /usr/share/nginx/html
--- a/docker/ui.Dockerfile
+++ b/docker/ui.Dockerfile
@@ -26,7 +26,10 @@ RUN yarn setup
 RUN yarn workspace @michelangelo/app build
 
 # Production image
-FROM nginx:alpine
+# Pinned by digest (rather than the floating `alpine` tag) so Dependabot can
+# track and propose base-image updates -- see .github/dependabot.yml's
+# "docker" entry.
+FROM nginx:alpine@sha256:db35bfc6b2951e7f8a72db5db120288c127ffaeeb4a6d4b95a26fead017d5913
 
 # Copy built app to nginx
 COPY --from=builder /workspace/javascript/app/dist /usr/share/nginx/html


paulzim (approved):
# PR #1952 review — ci: pin base images by digest and track them via Dependabot

**Author:** sallycr (Sally Lee) · **Branch:** `ci/pin-base-images-and-dependabot-docker` → `main` · **1 commit, +20/-3, 4 files** · Partial follow-up to #1949.

## What it does

`docker/service.Dockerfile` FROM'd `gcr.io/distroless/base-debian12` and `docker/ui.Dockerfile`/`docker/ui-prebuilt.Dockerfile` FROM'd `nginx:alpine` as floating tags with no digest pin, and `.github/dependabot.yml` had no `docker` ecosystem entry — so nothing tracked or could propose updates for these base images. The PR pins all three `FROM` lines to `image:tag@sha256:...` and adds a `docker` package-ecosystem entry for `/docker` to `dependabot.yml`, matching the style of the existing entries (weekly, PR limit 3, `chore(docker)` prefix).

Explicitly scoped as a partial follow-up to #1949's root cause #1 (OS-level glibc/openssl CVE cluster) — doesn't retroactively clear CVEs already in the pinned digests (some are will-not-fix upstream, need a `.trivyignore`) and doesn't touch the separate Go-toolchain-CVE cluster (needs a `go.mod` bump).

## Verification performed

- **Confirmed both pinned digests are correct, not typos or stale**, by independently querying the real registries rather than trusting the PR's claim:
  - `nginx:alpine` → fetched an anonymous Docker Hub token, queried `registry-1.docker.io/v2/library/nginx/manifests/alpine` → `docker-content-digest: sha256:db35bfc6b2951e7f8a72db5db120288c127ffaeeb4a6d4b95a26fead017d5913` — **exact match** to what's pinned in both `ui.Dockerfile` and `ui-prebuilt.Dockerfile`.
  - `gcr.io/distroless/base-debian12:latest` → fetched an anonymous GCR token, queried `gcr.io/v2/distroless/base-debian12/manifests/latest` → `docker-content-digest: sha256:fabbf1c0c357a3d42550111351daed089b20a2c954df13ee2fcff60602515e84` — **exact match** to what's pinned in `service.Dockerfile`.
- Validated `.github/dependabot.yml` with `yaml.safe_load` — structurally valid, and the new `docker`/`/docker` entry's formatting (schedule, PR limit, commit-message prefix convention `chore(docker)`) is consistent with every other entry in the file.
- Confirmed `directory: "/docker"` is correct: `ls docker/*.Dockerfile` shows all three changed Dockerfiles live directly in `/docker`, so Dependabot's docker ecosystem will scan all of them from this one entry.
- Checked for other unpinned floating-tag references to these same images elsewhere in the repo that the PR might have missed: found one, `python/michelangelo/cli/sandbox/resources/mlflow.yaml` still uses bare `nginx:alpine`. Verified this is legitimately out of scope, not an oversight — `.github/workflows/cve-scan.yml` (the workflow that #1949 traces back to) only scans the four *published* `ghcr.io/michelangelo-ai/{apiserver,worker,controllermgr,ui}:latest` images; the sandbox's `mlflow.yaml` is a local-only dev Pod manifest that's never built, published, or CVE-scanned, and Dependabot's `docker` ecosystem doesn't recurse into non-Dockerfile K8s manifests anyway. Not a finding.
- Confirmed no `BUILD.bazel` file references any of the three Dockerfiles (`grep` came up empty) — these images are built via plain `docker build`, entirely outside Bazel, so this change has zero interaction with the Bazel/gazelle graph.
- **`gh pr checks 1952` shows `bazel-build` and `dirty-check` both FAILING** — investigated rather than taking at face value, since a Dockerfile/dependabot-only diff has no plausible mechanism to break either (confirmed by the BUILD.bazel grep above). The actual failures are unrelated to this PR's content:
  - `dirty-check`: "BUILD.bazel files are not up to date. Please run gazelle before commit."
  - `bazel-build`: `proto/api/BUILD.bazel:5:14` — `Import "google/protobuf/wrappers.proto" was not found or had errors`, failing `//go/api/crd:go_default_test`.
  - Both runs are from 2026-08-26 (when the PR was opened), checked out against `pull/1952/merge` into `main` at that time. Checked `main`'s history since then: commit `5e646bd6`, **`fix(build): regenerate proto/api/BUILD.bazel with gazelle (#1953)`** — landed the very next day, one PR-number later — and confirmed it's already an ancestor of current `upstream-main`. This is a pre-existing, already-fixed staleness bug in `main`'s own `proto/api/BUILD.bazel`, unrelated to Dockerfiles/dependabot and already resolved upstream; it just hasn't re-run against this PR since. **Not a defect in this PR** — but merge should wait for a rebase/re-run against current `main` to get a clean, non-stale CI signal, since the currently-visible check state is red.

## Findings

None blocking the PR's own content. One process note: **CI is currently red, but for a reason entirely unrelated to and already fixed independently of this PR's changes (#1953, merged to `main` 2026-08-27)** — rebasing (or just re-running CI against current `main`) should turn both `bazel-build` and `dirty-check` green with no code changes needed.

## Verdict: **Approve**, pending a green CI run

The change itself is correct and low-risk: both digests independently verified against live registries, config change matches established repo conventions exactly, and the PR is honest and precise about what it does and doesn't fix. Ask the author to rebase onto current `main` (or re-run checks) to clear the stale, unrelated `bazel-build`/`dirty-check` failures before merge — no code change required on their end.

PR #1862, opened 2026-08-19
docs(api-reference): add gRPC service reference for all 15 API services

zhoward-1:
**What type of PR is this? (check all applicable)**
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Adds real gRPC service reference documentation for all 15 Michelangelo AI API services, covering 90 RPCs (6 per service), and gives it a proper home:

- New `docs/api-reference/` section with two subsections: **Python SDK Reference** (moved from the previous `docs/api-reference/index.md` stub, unchanged) and **gRPC Service Reference** (`docs/api-reference/grpc/`) — `conventions.md`, `models.md`, `pipelines.md`, `serving.md`, `jobs.md`. Each service section documents the full 6-RPC CRUD pattern (Create, Get, Update, Delete, DeleteCollection, List) with method signatures and request/response field tables.
- `conventions.md` explains the shared CRUD-plus-list pattern, the Required-column methodology, and how fields with no proto comment are shown — once, instead of re-explaining it in all four reference files.
- `operator-guides/jobs/index.md` and `operator-guides/serving/index.md` keep their existing short gRPC service summary tables (the right altitude for operators) and now link forward to the full reference; `operator-guides/index.md` links to it under Architecture & Reference.

This went through a 3-agent doc review (technical accuracy + audience/placement) before landing here. That review found the content technically accurate against the proto source but flagged it as placed and finished like a rough draft, which this revision fixes:
- Replaced 150 unedited `"No description provided in source."` placeholder rows with `—`, explained once in `conventions.md`.
- Fixed an inconsistent Required-column disclaimer across files (same fix, via `conventions.md`).
- Fixed a `sidebar_position: 5` collision between `models.md` and `pipelines.md`.
- Added "Next Steps" cross-links between the reference pages and related operator-guides content.
- Moved the reference out of `operator-guides/` into `api-reference/`, since it's contributor/integrator-facing field reference (no config, no troubleshooting) rather than "what to configure/enable/monitor."

**Why?**

Operators and integrators using the Michelangelo AI gRPC API had no reference documentation — only proto files. These pages give a navigable, human-readable reference that maps directly to the proto source without inventing behavior, placed alongside the Python SDK reference where contributor-facing API docs belong.

**How did you test it?**

[...]

PR #1956, opened 2026-08-27
docs(api-reference): generate Python SDK Reference from docstrings

### LINKED ISSUE #1955: Fix docstring formatting in 11 modules so pydoc-markdown renders them cleanly
**What's broken**

`pydoc-markdown` (used to generate the [Python SDK Reference](https://github.com/michelangelo-ai/michelangelo/blob/main/docs/api-reference/python-sdk/index.md)) misrenders the `Args`/`Raises` sections in 11 modules — 110 corrupted lines total. Example, from `uniflow/plugins/pandas/io.py`:

```
- `storage_options` - Optional fsspec storage options forwarded to
  ``fsspec.core.url_to_fs``. Use this to pass credentials or
  endpoint overrides for remote filesystems (e.g.
- ``2 - "...", "secret": "..."}`` for S3).
```

**Root cause**

`pydoc-markdown`'s `GoogleProcessor` parses `Args`/`Raises` blocks line by line, looking for a `word: description` pattern to detect a new parameter. When a multi-line argument description contains an inline `key: value`-style example (e.g. a dict literal wrapped in RST double-backticks), a continuation line gets misread as the start of a new parameter.

Minimal repro:

```python
def example_fn(storage_options):
    """Example function with a broken-style docstring.

    Args:
        storage_options: Optional options forwarded to a helper.
            Use this to pass credentials or endpoint overrides
            for remote filesystems (e.g.
            ``{"key": "...", "secret": "..."}`` for S3).
    """
```

renders as:

```
- `storage_options` - Optional options forwarded to a helper. ...
- ```{"key"` - "...", "secret": "..."}`` for S3).
```

Wrapping the example in a fenced code block instead of inline double-backticks avoids the misparse:

```python
    """Example function with a fixed-style docstring.

    Args:
        storage_options: Optional options forwarded to a helper.
            Use this to pass credentials or endpoint overrides
            for remote filesystems, e.g.:
            ```
            {"key": "...", "secret": "..."}
            ```
    """
```

**Fix needed**

In each file below, find the `Args`/`Raises` description that contains an inline `key: value`-style example and move it into a fenced code block:

- `python/michelangelo/uniflow/plugins/ray/run_config.py`
- `python/michelangelo/uniflow/plugins/ray/io.py`
- `python/michelangelo/uniflow/plugins/pandas/io.py`
- `python/michelangelo/workflow/tasks/pusher/task.py`
- `python/michelangelo/workflow/tasks/pusher/registry.py`
- `python/michelangelo/workflow/variables/metadata.py`
- `python/michelangelo/workflow/variables/types.py`
- `python/michelangelo/lib/native_transform/torch/utils.py`
- `python/michelangelo/lib/native_transform/torch/id_hash_tokenizer.py`
- `python/michelangelo/lib/native_transform/torch/base_layers.py`
- `python/michelangelo/lib/trainer/torch/utils.py`

**Note for future docstrings:** any Google-style docstring with a multi-line Args/Raises description that includes an inline colon-containing example (dict literals, `key: value` examples) will hit this same bug. Worth keeping such examples in fenced code blocks generally, not just in these 11 files.

**Once fixed:** re-run `bun run docs:api` from `website/` — the corresponding pages will start rendering clean and can be linked from `docs/api-reference/python-sdk/index.md` (see the "coming soon" modules there).

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Finishes the Python SDK Reference that #1862 stubbed out:
- Added `pydoc-markdown.yml` at the repo root, scoped to the 21 modules already listed in the stub (`docs/api-reference/python-sdk/index.md`).
- Fixed the `docs:api` script in `website/package.json` — it called `python -m pydoc_markdown`, which fails (`pydoc_markdown` is a package, not runnable via `-m`); fixed to call the `pydoc-markdown` CLI entrypoint.
- Generated reference pages for 10 modules under `docs/api-reference/python-sdk/reference/`, with `_category_.json` files so Docusaurus's autogenerated sidebar picks them up and groups them by area.
- Updated `index.md` to link those 10 pages. The other 11 modules hit a `pydoc-markdown` rendering bug in their docstrings and are listed as plain text with a "coming soon" note — tracked separately in #1955, which is a `python/michelangelo/` source change out of scope for this docs-only PR.

A follow-up commit (`71c9f963`) applies the results of a documentation review over the generated output. Corrections to the pages themselves:
- **`with_overrides()` was documented incorrectly.** The docstring claimed the `config` argument is merged field-by-field; the implementation fully replaces it (`decorator.py`: `config=config or self._config`), so unset fields silently reset to `None`. Corrected, plus a caveat that `image_spec` is never carried forward to the returned `TaskFunction`.
- **Rendering corruption affected 4 of the 10 pages assumed clean**, not only the 11 excluded modules. Investigating it turned up *two* distinct bugs rather than the single one #1955 describes — full analysis and repro posted [in that issue](https://github.com/michelangelo-ai/michelangelo/issues/1955#issuecomment-5443819315):
  - A renderer bug (`pydoc_markdown/util/misc.py`, `escape_except_blockquotes`) mangles the 11th and later inline code span in any docstring into `` `<span #1>`<digit> ``. It hit `ray/task.md` — silently renaming the public fields `worker_instances`, `breakpoint`, and `runtime_env` — and `data_collate_functions.md`. Restored the real names.
  - The `GoogleProcessor` colon misparse that #1955 documents also fires inside `Example:` blocks, not just `Args:`/`Raises:`, turning example code into parameter bullets in `task_config.md` and `io_registry.md`. Restored both examples.
- **Removed a leaked internal Bazel target path** from `image_spec.md`, replacing it with the generic placeholder already used elsewhere on that page, and fixed an example calling `RayTask(cpu=1)` — no such field exists (`RayTask` takes `head_cpu`/`worker_cpu`), so the example raised `TypeError` as written.
- **Prefixed all 10 page titles with `michelangelo.`** to match the real import path declared by `python/pyproject.toml`; the titles previously contradicted the import statements in their own examples.
- Fixed HTML entity escaping and flattened code-block indentation across all 10 pages — several examples were not valid Python as rendered.

Framing and structure:
- `index.md` now opens with orientation and a link to Core Concepts, with regeneration instructions moved into their own section and caveated (`pydoc-markdown` is not a declared project dependency, so a fresh checkout needs it installed first).
- Disclosed that `LightningTrainer` and the XGBoost trainer are referenced but absent from the generator config, so neither has a page.
- Added source-verified examples for `RayTask`, `SparkTask`, and `SparkIO`, documented `default_io`, and resolved a dangling `ConfigurationError` reference.

**Why?**

The Python SDK Reference has been a stub since #1862 ("until then, this page lists the modules covered without linking to pages that don't exist yet") — the generator was never actually wired up or run. This makes it real for the modules that can render cleanly today, without blocking on the docstring fixes needed for the rest.

The review pass exists because auto-generated output was being treated as trustworthy by default. It wasn't: the corruption had a wider blast radius than #1955 documents and a different root cause than it records, one page carried an internal build path into public docs, and `with_overrides()` documented behavior the code does not implement.

**How did you test it?**

Ran `pydoc-markdown` against all 21 modules locally. Found 11 render with garbled `Args`/`Raises` sections and excluded those from this PR.

Re-ran the full generator into a scratch directory to classify the corruption rather than assume it. That produced the minimal repro for the renderer bug (10 code spans render clean, 11 corrupt), confirmed the `GoogleProcessor` misparse is genuinely separate by showing it survives the renderer workaround, and gave the per-module occurrence counts — all posted to #1955.

For the review pass, every corrected signature, field name, and default was checked against the source it documents (`ray/task.py`, `spark/io.py`, `decorator.py`, `task_config.py`, `io_registry.py`, `data_collate_functions.py`, `pyproject.toml`). The added examples use only fields that exist in source. Verified afterwards that all internal links in `index.md` resolve to real files, all 13 `_category_.json` files parse as valid JSON, and the directory is free of entity-escaping artifacts, leftover Sphinx role markers, and internal references.

**Potential risks**

None at runtime — documentation only, no changes to `python/michelangelo/` source.

One maintenance caveat worth knowing before merge: the review corrections are hand-edits to generated output, so `bun run docs:api` will overwrite every one of them. Nothing here is self-sustaining, and the upstream fixes fall into three buckets:

- **The rendering corruption and the entity escaping** are fixed by a one-line generator config change — `escape_html_in_docstring: false` on the docusaurus renderer — not by editing docstrings. Measured across all 21 configured modules it takes token corruption from 151 occurrences to 0 and entity escaping from 579 to 0. Deliberately not included here, because it changes output for every page and wants a Docusaurus build to confirm no docstring trips MDX on raw `<`/`{`. Details in #1955.
- **The `Example:` block misparse** in `task_config.py` and `io_registry.py` does need docstring edits — converting the RST `::` literal blocks to fenced blocks — and is the same class of fix #1955 already prescribes.
- **Two content bugs live in source docstrings** and will return verbatim on the next regeneration: `with_overrides()` in `python/michelangelo/uniflow/core/decorator.py` still describes merge semantics it does not implement, and the internal Bazel path plus the invalid `RayTask(cpu=1)` call still sit in `image_spec.py`. These are `python/michelangelo/` changes, out of scope for a docs-only PR, but they should land before anyone regenerates.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Adds generated reference pages for 10 Python SDK modules under `docs/api-reference/python-sdk/reference/`, reviewed and corrected for accuracy; the remaining 11 are tracked in #1955.


paulzim (approved):
# PR #1956: docs(api-reference): generate Python SDK Reference from docstrings

**Verdict: Approve**, with one non-blocking suggestion (config fix for page titles) and one trivial doc-copy nit.

## What I verified

This PR turns the `docs/api-reference/python-sdk/` stub (from #1862) into real generated content for 10 of the 21 modules, plus a hand-reviewed correction pass over the generated output. Given the PR's own thesis — "auto-generated output was being treated as trustworthy by default, and it wasn't" — I verified its claims the same way: against the actual `pydoc-markdown` source and the real repo files, not just the PR description.

**Tooling fix.** `website/package.json`'s `docs:api` script changes from `python -m pydoc_markdown` (broken — `pydoc_markdown` has no `__main__.py`) to
`pydoc-markdown` (the real console-script entrypoint). Confirmed against an actual
`pydoc-markdown==4.8.2` install.

**New renderer bug, independently confirmed.** The PR reports a second, previously-undocumented pydoc-markdown bug beyond the one #1955 tracks: `escape_except_blockquotes()` in `pydoc_markdown/util/misc.py` replaces each inline-code span with a placeholder token `BLOCKQUOTE_TOKEN_{i}` via plain `str.replace`. Once a docstring has 10+ spans, single-digit tokens (`BLOCKQUOTE_TOKEN_1`) become substrings of double-digit ones (`BLOCKQUOTE_TOKEN_10`, `_11`, ...), so the restore pass corrupts them. I reproduced this directly against the installed package:

```
>>> escape_except_blockquotes(' '.join(f'`field{i}`' for i in range(15)))
'`field0` ... `field9` `field1`0 `field1`1 `field1`2 `field1`3 `field1`4'
```

— which is exactly the `` `<span #1>`<digit> `` pattern described in the PR and in the linked #1955 comment. I also confirmed *why* my own earlier #1955 triage missed this: that review used the `markdown` renderer (`escape_html_in_docstring` defaults to `False` there), while this PR correctly uses the `docusaurus` renderer for real, where the same flag defaults to `True` — so the bug only fires under the renderer this repo actually needs. Good catch; this genuinely extends #1955's diagnosis rather than duplicating it.

**Content corrections, checked against source:**
- `with_overrides()`: confirmed `decorator.py`'s implementation is `config=config or self._config` (full replace), not a field merge as the *original* docstring claims — the corrected page text  matches the real behavior, including the `image_spec` caveat (verified `TaskFunction.with_overrides`  never passes `image_spec` through).
- `image_spec.md`: confirmed the source docstring contains a leaked internal path
  (`bazel://uber/ai/michelangelo/sdk/workflow/tasks/llm_feature_prep:...`) — removed from the page and replaced with the generic placeholder already used elsewhere. Confirmed `RayTask` has no `cpu` field (only `head_cpu`/`worker_cpu`), so the `RayTask(cpu=1)` → `RayTask(head_cpu=1)`  fix is correct.
- `ray/task.md`: confirmed `worker_instances`, `breakpoint`, `runtime_env` are real `RayTask`  fields and are now rendered correctly (no leftover mangled tokens).
- `pusher/exceptions.md` (commit `88218e4f`): confirmed `ConfigurationError` really is imported  and re-exported from `pusher/exceptions.py` (it's in `__all__`) — the restored fact is accurate;  the intermediate version from `71c9f963` had understated this.

**Sweep for remaining corruption across all 10 generated pages:** grepped for the mangled-token signature (`` `\S+`[0-9] ``), spurious fake-param bullets, and leftover HTML entities (`&lt;`/`&gt;`/`&quot;`/`&amp;`) — zero hits anywhere in `docs/api-reference/python-sdk/reference/`. All 12 `_category_.json` files parse as valid JSON (the PR body says 13 — minor miscount in its own test notes, not a functional issue). All 10 links in `index.md` resolve to files that exist.

**CI:** all checks pass, including `bazel-build`/`dirty-check` (the gazelle issue that hit prior docs-adjacent PRs is already resolved via the merged-in #1953 fix).

## One suggestion worth taking before merge

The 10 generated pages' `title:` frontmatter is hand-edited to read `michelangelo.uniflow.core.decorator` etc., to match the real import path. But `pydoc-markdown.yml` sets `search_path: [python/michelangelo]` with un-prefixed module names (`uniflow.core.decorator`), and the `docusaurus` renderer derives `title:` directly from that module name (`docusaurus.py`: `"title: {module_name}"`). I confirmed this by re-running the generator with `search_path: [python]` and modules prefixed `michelangelo.*` instead — it produces the correct title automatically, no hand-edit needed:

```
title: michelangelo.uniflow.core.decorator   # correct, unprompted
```

As configured today, the title hand-edit will be silently reverted back to the wrong,
import-path-inconsistent title the next time anyone runs `bun run docs:api` — and unlike the other hand-edits, this one isn't called out in the PR's "Potential risks" section (which lists the renderer bug, the `Example:` misparse, and the two source docstring bugs, but not this). Since the fix is a two-line config change with no runtime risk, it'd remove an entire class of "looks fine until the next regen" fragility. Not a blocker — the PR already frames all hand-edits as non-self-sustaining in general — but worth a fast follow-up, or fixing inline now if easy.

## Recommendation

Approve. This is an unusually well-verified docs PR — the author treated the generator's output as something to audit rather than trust, found a genuinely new upstream bug in the process, and every specific correction I checked against source held up. The title-frontmatter fragility is worth a follow-up but shouldn't block merge.

PR #1958, opened 2026-08-27
fix(python): self-heal sandbox mysql Pod if sync finds it not Running

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
`ma sandbox sync`'s fast path (used on long-lived CI runners) leaves infrastructure Pods running between runs and execs directly into the bare `mysql` Pod to refresh the schema. If that Pod has fallen out of `Running` phase (e.g. a node-level container-runtime restart, since it's a bare Pod with no Deployment/ReplicaSet to self-heal it), `kubectl exec` fails immediately and `sync` aborts before any pipeline test runs.

This adds `_ensure_mysql_pod_running()`, called right before the schema refresh: if the Pod isn't `Running`, it's deleted and reapplied from `resources/mysql.yaml`, and `sync` waits for it to become ready. Since this Pod has no PVC, a recreation wipes its data — including the Cadence `cadence`/`cadence_visibility` databases and domain registration. The `helm upgrade`/`install` step that already runs later in `sync` recreates the schema itself (the cadence subchart's own schema-server Job does this per `values-k3d.yaml`), so this PR only adds the missing piece: re-registering the Cadence domain via `_create_cadence_domain([])` when a recreation was detected.

**Why?**
This is the ad hoc fix for the mysql-Pod-vanishing failure documented in harness spec 037's addendum. An earlier attempt at this same self-heal guard was merged and then reverted on #1775 (commit `5204a37df`/`5e42bfd80`) because it revived the Pod without addressing the resulting Cadence data loss, which broke workflow dispatch on the next pipeline run. This PR carries the same Pod revival plus the missing domain-reinit step, closing that gap.

This is intentionally a stopgap, not the durable fix — it doesn't restore any state that might live in the wiped mysql data outside the `michelangelo`/Cadence schemas already handled by `sync`. A PVC-backed mysql Pod (or a fuller reinit path) remains the durable fix and is tracked separately.

**How did you test it?**
`python3 -m py_compile` and `ruff check`/`ruff format --check` on the changed file. Verified by code inspection that `_create_cadence_domain`'s `links` parameter is unused (safe to call with `[]`), that `_create_cadence_domain` treats "already registered" as success (safe to call unconditionally on every recreation), and that the Cadence schema-server Job (not this code) is what creates `cadence`/`cadence_visibility` on `helm upgrade`. Not validated against a live occurrence of the Pod actually vanishing on the persistent runner, since that failure mode isn't reliably reproducible on demand.

**Potential risks**
If the mysql Pod is momentarily transitioning phases for a benign reason (not actually unhealthy), this would still delete and reapply it, which is a no-op functionally but adds sync latency. Domain re-registration on every "Pod was recreated" event is idempotent and low-risk. The known residual risk — other cluster state that only lived in the wiped mysql data — is unchanged from before this PR and is called out in the code comments.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -355,6 +355,8 @@ def _sync(ns: argparse.Namespace):
     # Upgrade or install the control plane via Helm.
     # Infrastructure (mysql, cadence, minio, grafana, prometheus) is left running.
 
+    mysql_pod_recreated = _ensure_mysql_pod_running()
+
     _refresh_mysql_schema()
 
     _ensure_credentials_secret()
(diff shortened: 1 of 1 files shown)


sallycr:
Sandbox-validated against a live k3d cluster.

Summary:
- Reproduced the real failure mode via `kubectl delete pod mysql` (a status-only patch doesn't work — the kubelet reverts it within ~1-2s).
- `ma sandbox sync` on this branch: the guard fires correctly (`mysql Pod is not Running (phase=missing) — recreating it...`), the Cadence schema is auto-recreated by the cadence subchart's own schema-server Job, and `_create_cadence_domain([])` re-registers the domain (confirmed via `cadence domain describe` → `Status: REGISTERED`).
- The worker Pod that Helm restarted crash-looped once with the exact `EntityNotExistsError` regression that got `#1775`'s guard reverted, before the domain-reinit step finished — restarting it once afterward came up clean, confirming this PR closes that specific gap.
- Ran a real `xgb-train` pipeline afterward: Cadence dispatched and completed 42 history events across multiple Decision/Activity cycles via the recreated worker, before failing on an unrelated Ray/SSL networking issue in this local sandbox (not a Cadence/domain/schema problem).

Also hit and fixed (locally, not part of this PR) two pieces of unrelated pre-existing sandbox drift that were blocking the test: a `michelangelo-ui` field-manager ownership conflict, and a stale `apiserver` ReplicaSet stuck in `ImagePullBackOff`. Neither is caused by or related to this change.

PR #1961, opened 2026-08-27
fix(python): pusher storage keys, push_id, and deployable-archive config

krishpatel9:
**TL;DR:** Four related fixes/changes to model artifact storage in `ModelPusherPlugin`/`MinioStorageBackend`/the tabular assemblers, all found while wiring the assembler into a real pipeline:

1. Destination keys no longer echo local filenames (fixes a doubled-segment bug).
2. `push_id` dropped from storage keys entirely (accepted tradeoff — see below).
3. Directory artifacts are stored as real per-file objects by default, instead of always being tar-archived.
4. Deployable-package archiving is now an explicit, opt-in config option (`tar_deployable_package`).

## 1. Fixed destination-key construction

`ModelPusherPlugin.execute()` built each destination storage key by appending the local artifact's filename. For an artifact that had to be downloaded first, the local download directory's own name could collide with the category segment already in the key, producing a redundant doubled path (e.g. `models/<name>/<push_id>/raw/raw`). Destination keys are now fixed — never derived from any local artifact path.

## 2. Removed `push_id` from storage keys

`push_id` (a random 16-hex-char token) was embedded in every storage key specifically to prevent two pushes of the same `model_name` from overwriting each other's artifacts. In practice it added a layer of indirection with no benefit for callers using an auto-generated, already-unique `model_name` (the common case), while its only real protection applies to callers reusing a fixed `model_name` across pushes (e.g. periodic retraining under the same name).

Removed entirely from storage keys — keys are now `models/{model_name}/raw` and `models/{model_name}/deployable[/model.tar]`. `push_id` is still generated and returned in the result (for log/result correlation), but a caller reusing the same `model_name` will now overwrite the previous push's storage objects (the registry still records each push as a separate version, but their artifact URIs then point at the same, now-overwritten content). Documented as an explicit warning on `ModelPushResult.push_id`.

## 3. Real per-file directory storage in `MinioStorageBackend`

`MinioStorageBackend.upload()` previously tar-archived *every* directory artifact into a single object (`/__dir__.tar`), inconsistent with `LocalStorageBackend` (which stores directories as real directories via `shutil.copytree`).

`MinioStorageBackend` no longer auto-archives directories: uploads one object per file under the destination key as a prefix, preserving nested structure. `download()` distinguishes file vs. directory via `stat_object` rather than a key-suffix sentinel. Removes the tar-archive/extraction code path entirely (`tarfile`, path-traversal-safe extraction, the `/__dir__.tar` sentinel).

## 4. Made deployable-package archiving configurable — in the pusher, not the assembler

Added `tar_deployable_package: bool = False` to `ModelPluginConfig`. **Default**: `ModelPusherPlugin` uploads the deployable artifact as-is — if it's a directory, as loose files preserving its internal structure (e.g. `deployable/0/model.pt`, `deployable/config.pbtxt`); if already a single file, as-is. **When `True`**: a directory deployable artifact is archived into a single `deployable/model.tar` via `shutil.make_archive()` immediately before the pusher's own upload call.

The assembler (`torch_assembler()`/`custom_assembler()`) never tars — it always hands off the deployable package as loose files, exactly like the raw model. Tarring is purely the pusher's decision at final-push time, since it's the pusher that already re-uploads regardless of how the assembler staged the artifact; keeping this logic out of the assembler avoids two components independently deciding the on-disk representation of the same artifact.

## Test plan
- [x] `pytest michelangelo/workflow/tasks/pusher/` — all 128 tests passing.
- [x] `pytest michelangelo/lib/artifact_manager/` — all 38 tests passing.
- [x] `pytest michelangelo/workflow/tasks/tabular_assembler/` — all 70 tests passing.
- [x] `pytest michelangelo/lib/model_manager/ michelangelo/workflow/` — all 1176 tests passing.
- [x] `ruff format --check` / `ruff check` — clean.
- [x] **Ran the full pipeline end-to-end in a local k3d sandbox** (`feature_prep` -> `preprocess` -> `train` -> `assembler` -> `push_step`, using michelangelo-ai/michelangelo-examples#36) on this PR's final code: run `SUCCEEDED`. Confirmed in the resulting MinIO bucket:
  - `raw_artifact_uri`: `s3://default/models/<name>/raw` — no `push_id` segment; loose files (`defs/`, `dependencies/`, `metadata/`, `model/`).
  - With `ModelPluginConfig(tar_deployable_package=True)`: `deployable_artifact_uri`: `s3://default/models/<name>/deployable/model.tar` — confirmed as exactly one 40KB object in MinIO (not multiple files under that prefix), produced by the pusher's own `shutil.make_archive()` call immediately before upload.
  - With the default (`tar_deployable_package=False`, unset): `deployable_artifact_uri`: `s3://default/models/<name>/deployable` — real Triton package layout preserved as loose files (`0/model.pt`, `config.pbtxt`, `metadata/`), matching the raw model's treatment (`raw/defs/`, `raw/dependencies/`, `raw/metadata/`, `raw/model/`). No `push_id` segment in either case.

### DIFF
--- a/python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
+++ b/python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
@@ -91,6 +91,9 @@ def torch_assembler(
     include_import_prefixes = (
         config.torch.include_import_prefixes if config and config.torch else None
     )
+    archive_deployable_package = (
+        config.torch.archive_deployable_package if config and config.torch else False
+    )
     model_class = raw_model.metadata.model_class
     hyperparameters = raw_model.metadata.hyperparameters or {}
 
@@ -224,16 +227,20 @@ def torch_assembler(
             ),
         )
 
-        # Archived into a single tar -- a self-contained serving bundle,
-        # unlike the raw model, which is uploaded as loose files.
-        deployable_tar_path = shutil.make_archive(
-            os.path.join(temp_dir, "deployable_package"), "tar", model_package_path
-        )
-
         upload_prefix = f"tabular_assembler/{uuid.uuid4().hex}"
-        deployable_uri = storage_backend.upload(
-            deployable_tar_path, f"{upload_prefix}/deployable/model.tar"
-        )
+        if archive_deployable_package:
+            # A single tar -- a self-contained serving bundle -- rather than
+            # the default loose files, only when explicitly requested.
+            deployable_tar_path = shutil.make_archive(
+                os.path.join(temp_dir, "deployable_package"), "tar", model_package_path
+            )
+            deployable_uri = storage_backend.upload(
+                deployable_tar_path, f"{upload_prefix}/deployable/model.tar"
+            )
+        else:
+            deployable_uri = storage_backend.upload(
+                model_package_path, f"{upload_prefix}/deployable"
+            )
         raw_uri = storage_backend.upload(raw_model_package_path, f"{upload_prefix}/raw")
 
     e2e_schema = fuse_e2e_schema(
--- a/python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
+++ b/python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
@@ -91,8 +91,8 @@ def torch_assembler(
     include_import_prefixes = (
         config.torch.include_import_prefixes if config and config.torch else None
     )
-    archive_deployable_package = (
-        config.torch.archive_deployable_package if config and config.torch else False
+    tar_deployable_package = (
+        config.torch.tar_deployable_package if config and config.torch else False
     )
     model_class = raw_model.metadata.model_class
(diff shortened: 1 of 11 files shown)


python/michelangelo/workflow/tasks/tabular_assembler/torch/assembler.py
```diff
@@ -224,9 +228,19 @@ def torch_assembler(
         )
 
         upload_prefix = f"tabular_assembler/{uuid.uuid4().hex}"
-        deployable_uri = storage_backend.upload(
-            model_package_path, f"{upload_prefix}/deployable"
-        )
+        if tar_deployable_package:
+            # A single tar -- a self-contained serving bundle -- rather than
+            # the default loose files, only when explicitly requested.
+            deployable_tar_path = shutil.make_archive(
```
kenns29: the tar option should be done in pusher, assembler doesn't tar the bundle

badcount (approved):

PR #1967, opened 2026-08-28
fix(ui): use fieldSelector on pipeline_name for Pipeline detail Runs tab

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

The Pipeline detail page's Runs tab (`detail.ts`) was filtering PipelineRuns using `labelSelector: 'pipeline.michelangelo/name=...'`. This label is written by nothing in the platform (confirmed by full-tree grep -- the only two hits are the filter itself and a comment in `resume-run-fields.tsx` documenting it as broken), so the Runs tab returned zero results for every pipeline, always.

Replaced the `labelSelector` with `fieldSelector: 'pipeline_name=${page.metadata.name}'`. `pipeline_name` is the indexed column the metadata store generates from the `spec.pipeline` `ResourceIdentifier` on `PipelineRun` (proto index key `"pipeline"`, sub-key `"name"` -- see `go/kubeproto/util/index.go:buildSubKey`). This field is populated on every `PipelineRun` by construction, and the Go API handler already parses `FieldSelector` via `fields.ParseSelector` (`handler.go:431`).

Also updated a comment in `resume-run-fields.tsx:168-170` that referenced the broken label as evidence it was not a viable alternative -- the broader point (that client-side filtering is still needed because `ListOptionsExt` can be silently ignored) remains valid, but the specific reference to the phantom label is corrected.

Added a unit test (`detail.test.ts`) asserting that the Runs page queries via `fieldSelector` on `pipeline_name` and no longer uses a `labelSelector`.

**Why?**

The Runs tab on every Pipeline detail page was completely non-functional -- it always showed zero results regardless of how many runs existed for the pipeline. The root cause was filtering on a label key (`pipeline.michelangelo/name`) that no code path in the platform ever writes.

**How did you test it?**

- Added `detail.test.ts` with assertions on the query config shape (fieldSelector present, labelSelector absent).
- Could not run `vitest` in this environment (Node 22 vs required Node 24, `node_modules` not installed). Tests should be verified in CI.
- Confirmed correct `fieldSelector` syntax by tracing the codegen: `go/kubeproto/util/index.go:214-222` shows `ResourceIdentifier` indexes produce sub-columns `{key}_namespace` and `{key}_name`; for proto index key `"pipeline"` this yields column `pipeline_name`. Existing usage of this exact column name in `resume-run-fields.tsx:53` (`fieldName: 'pipeline_run.pipeline_name'`) and `rpc/__tests__/services.test.ts:81` confirms it.

**Potential risks**

Low. This changes a completely broken filter (always returns empty) to a working one. The `pipeline_name` column is already indexed and used by other query paths. If a deployment uses a custom storage backend that doesn't support `FieldSelector`, the tab would continue returning all runs unfiltered (same as current `ListOptionsExt` fallback behavior documented in `resume-run-fields.tsx`), which is strictly better than the current zero-results behavior.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Fix: the Runs tab on Pipeline detail pages now correctly shows runs belonging to the pipeline. Previously it always showed zero results due to filtering on a label that was never written.

**Documentation Changes**

N/A -- inline code comments updated as part of the fix.


sallycr:
Screenshot 2026-08-28 at 4 54 30 PM

craigmarker:
Added a commit on top (490f72e3) that corrects the selector and completes the backend fix:

- **fieldSelector → labelSelector**: `pipeline_name` field selectors only resolve through the metadata-storage-backed list path, which is off by default. On the default etcd-backed path there's no field indexer registered for it in the Go codebase, so this would silently stop filtering (return every run, unscoped) on the common deployment rather than erroring — the risk this PR's description already flagged. Reverted `detail.ts` to a `labelSelector` on `pipelinerun.michelangelo/pipeline-name`, which is stamped unconditionally at creation time regardless of storage backend.
- **Backend stamping completed**: that label was previously only set on trigger-created runs, under a different key than what the frontend queried for. Manually-created runs (the common path, via `apiHook.BeforeCreate`) got no pipeline-name label at all. Added a shared `PipelineNameLabelName` constant (`go/api/api.go`) and stamp it from `BeforeCreate`, so every creation path now agrees with the frontend's label key.
- **Test**: replaced `detail.test.ts`'s query-shape assertions with an integration test in `pipeline.test.tsx` that renders the real Pipeline detail route and asserts a run carrying the label renders in the table.
- Also trimmed a stale parenthetical in `resume-run-fields.tsx`'s comment that referenced the fieldSelector approach, since it no longer applies.

`yarn typecheck`, `yarn lint --quiet`, `yarn test packages/core/config/entities/pipeline`, and `go test ./components/pipelinerun/apihook/... ./worker/workflows/trigger/... ./api/...` all pass on top of this commit.

This duplicates the backend half of #1986, which was opened independently against the same root cause — Craig's deciding what to do with that PR now that this one covers both halves of the fix.

craigmarker:
Update per Craig: sticking with this PR's original fieldSelector approach, not the labelSelector correction from my previous comment/commit. Two commits now on top of your original one:

- \`490f72e3\` (superseded by the next commit, keeping it rather than rewriting history) added a labelSelector-based backend fix.
- \`b97945bc\` reverts that backend work and restores the original \`fieldSelector: 'pipeline_name=...'\` in \`detail.ts\`, since Craig wants to keep this approach as-is.

What \`b97945bc\` adds on top of your original fix:
- A comment at the \`fieldSelector\` call site documenting the known limitation your PR description already flagged: this filtering only takes effect when `EnableMetadataStorage` is enabled (off by default) — on the default etcd-backed path there's no field indexer registered for `pipeline_name`, so the query returns every run unfiltered.
- Replaces \`detail.test.ts\`'s query-shape assertions with an integration test in \`pipeline.test.tsx\` that renders the real Pipeline detail route and asserts a run reaches the table via the interpolated fieldSelector.

`yarn typecheck`, `yarn lint --quiet`, `yarn test packages/core/config/entities/pipeline`, and `go test ./components/pipelinerun/apihook/... ./worker/workflows/trigger/... ./api/...` all pass on the current HEAD.

My separate PR #1986 (which took the labelSelector route) is now closed in favor of this one.

craigmarker:
**With pipeline run**
Screenshot 2026-08-31 at 3 21 22 PM

**Resume from still works**
Screenshot 2026-08-31 at 3 22 08 PM

donovanlopez (approved):

PR #1973, opened 2026-08-29
fix: filter Pipeline by type using a label selector, not a field selector

badcount:
The UI's phase-entity-view filter built a field selector (`pipeline_type in (...)`) for the pipeline service, but Kubernetes field selectors (k8s.io/apimachinery/pkg/fields) only ever support `=`/`==`/`!=` — never the set-based `in (...)` operator. Every pipeline-type filtered list call failed server-side with "can't understand '...)'".

PipelineRun/TriggerRun already solved this correctly by filtering on the michelangelo/SourcePipelineType label (which supports `in (...)` via labels.Parse), inherited from their source Pipeline. Add the same label stamping to Pipeline itself via a new PipelineAPIHook, and switch the UI to use a label selector for all three resource kinds.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Breaking Changes**

- [ ] No breaking changes

**Migration guide**

**Release notes**

**Documentation Changes**


krishpatel9 (approved):

go/components/inferenceserver/backends/interface.go
```diff
@@ -32,4 +33,9 @@ type Backend interface {
 	IsHealthy(ctx context.Context, logger *zap.Logger, kubeClient client.Client, inferenceServerName string, namespace string) (bool, error)
 	// CheckModelStatus reports whether a model is loaded and ready for inference.
 	CheckModelStatus(ctx context.Context, logger *zap.Logger, kubeClient client.Client, httpClient *http.Client, apiServerURL string, inferenceServerName string, namespace string, modelName string) (bool, error)
+	// LoadModel stages a packaged model into the inference server's model repository and
+	// triggers the server to load it. modelPackage is the raw bytes of a tar archive whose
+	// members are already laid out as the backend's model-repository entry (e.g. for Triton:
+	// config.pbtxt, a numbered version directory, etc.) for the model named modelName.
+	LoadModel(ctx context.Context, logger *zap.Logger, kubeClient client.Client, restConfig *rest.Config, httpClient *http.Client, apiServerURL string, inferenceServerName string, namespace string, modelName string, modelPackage []byte) error
```
aritra0227, line 40: This is not utilizing the sidecar pattern. In this case, the controller is directly loading the model within the inferenceserver. 
I think we want to rely on the sidecar to perform the actual loading.

go/components/inferenceserver/clientfactory/factory.go
```diff
@@ -189,6 +189,16 @@ func (f *remoteClientFactory) GetDynamicClient(ctx context.Context, cluster *v2p
 	return dynamicClient, nil
 }
 
+// GetRESTConfig returns a *rest.Config for the given ClusterTarget. Unlike the other
+// client types, this isn't cached: it's used rarely (e.g. building a clientset for pod
+// exec), and rebuilding it is cheap relative to a TLS handshake.
+func (f *remoteClientFactory) GetRESTConfig(ctx context.Context, cluster *v2pb.ClusterTarget) (*rest.Config, error) {
```
aritra0227, line 195: This function becomes redundant when we rely on the sidecar to actually load the models.

PR #1977, opened 2026-08-31
fix(blobstore): construct the GCS client eagerly when a gcs config section is present

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Follow-up to review feedback on #1837 (https://github.com/michelangelo-ai/michelangelo/pull/1837#discussion_r3834692496): the GCS blob store client is now constructed eagerly at dependency-injection time whenever the application config contains a `gcs` section, so misconfiguration fails at service startup instead of on the first `gs://` read.

- `newConfig` records whether a `gcs` section was present in the config (even an empty one, which still declares intent to use GCS with Application Default Credentials). The flag is an unexported field, so it cannot be set from YAML.
- `newClient` constructs the underlying `storage.Client` immediately when the section is present and returns the construction error, which fails fx application startup. Without a `gcs` section the client stays lazy, exactly as before.
- Added unit tests for both paths plus fx-level module tests proving that a broken `gcs` section fails app construction while an app without the section starts cleanly.

**Why?**

Review feedback on #1837 asked why the client is not created eagerly to catch failures at deployment time. Unconditional eager construction would resolve credentials at startup in every deployment -- including s3- or minio-only installs with no GCP credentials, which would then fail to boot. Constructing eagerly only when the operator has declared a `gcs` config section gives the fail-fast behavior where it is wanted and keeps the zero-cost lazy path everywhere else.

**How did you test it?**

- `go test ./base/blobstore/... -count=1` -- all packages pass, including 6 new tests: presence detection (full, empty, and absent `gcs` sections), eager construction success (anonymous mode), eager construction failure (unreadable credentials file), and fx-level startup tests consuming the `blobstore_clients` group the same way the worker and controller manager binaries do.
- `go test ./base/blobstore/gcs/... -count=1 -race` -- passes.
- `go build ./...`, `go vet ./base/blobstore/gcs/...`, `gofmt -l` -- clean.
- All new tests are hermetic: they use anonymous mode or a deterministically unreadable credentials file, so they do not depend on ADC being present or absent in the environment.

**Potential risks**

A deployment that already declares a `gcs` section pointing at broken configuration (for example an unreadable `credentialsFile`) but never actually reads `gs://` URIs would previously start "successfully" and now fails at startup. That is the intended fail-fast behavior, but it is a startup-behavior change for such deployments; the fix on their side is to correct or remove the stale `gcs` section. Deployments without a `gcs` section are unaffected.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

GCS blob store: when a `gcs` config section is present, the client is now constructed at startup so credential or endpoint misconfiguration fails the deployment immediately instead of on the first `gs://` read. Deployments without a `gcs` section keep the lazy construction path and are unaffected.

**Documentation Changes**

None -- behavior is documented in the package godoc comments.

### DIFF
--- a/go/base/blobstore/gcs/client.go
+++ b/go/base/blobstore/gcs/client.go
@@ -18,12 +18,14 @@ var _ blobstore.BlobStoreClient = (*gcsBlobClient)(nil)
 
 // gcsBlobClient is a client for Google Cloud Storage.
 //
-// The underlying storage.Client is created lazily on the first Get call
-// rather than at construction time: storage.NewClient resolves credentials
-// (Application Default Credentials unless configured otherwise) eagerly, so
-// eager construction would fail application startup on deployments that
-// have no GCP credentials and never read gs:// URIs. Lazy construction
-// keeps the gs scheme registered everywhere at no cost to non-GCS users.
+// By default the underlying storage.Client is created lazily on the first
+// Get call rather than at construction time: storage.NewClient resolves
+// credentials (Application Default Credentials unless configured otherwise)
+// eagerly, so eager construction would fail application startup on
+// deployments that have no GCP credentials and never read gs:// URIs. Lazy
+// construction keeps the gs scheme registered everywhere at no cost to
+// non-GCS users. Deployments that declare a gcs config section opt in to
+// eager construction at startup instead (see newClient in module.go).
 type gcsBlobClient struct {
 	config Config
 	scheme string
--- a/go/base/blobstore/gcs/BUILD.bazel
+++ b/go/base/blobstore/gcs/BUILD.bazel
@@ -23,7 +23,12 @@ go_test(
     srcs = [
         "client_test.go",
         "config_test.go",
+        "module_test.go",
     ],
     embed = [":go_default_library"],
-    deps = ["@org_uber_go_config//:go_default_library"],
+    deps = [
+        "//go/base/blobstore:go_default_library",
+        "@org_uber_go_config//:go_default_library",
+        "@org_uber_go_fx//:go_default_library",
+    ],
 )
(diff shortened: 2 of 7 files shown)


go/base/blobstore/gcs/client.go
```diff
@@ -18,12 +18,14 @@ var _ blobstore.BlobStoreClient = (*gcsBlobClient)(nil)
 
 // gcsBlobClient is a client for Google Cloud Storage.
 //
-// The underlying storage.Client is created lazily on the first Get call
-// rather than at construction time: storage.NewClient resolves credentials
-// (Application Default Credentials unless configured otherwise) eagerly, so
-// eager construction would fail application startup on deployments that
-// have no GCP credentials and never read gs:// URIs. Lazy construction
-// keeps the gs scheme registered everywhere at no cost to non-GCS users.
+// By default the underlying storage.Client is created lazily on the first
```
uzbq, line 21: If the client is not configured, how the client can be created lazily?

apurvapatkeshwar: `storage.NewClient` doesn't need any michelangelo config to construct
With no `gcs:` section, `ensureClient` passes no options and it falls back to Application Default Credentials (the GKE workload identity metadata server, `GOOGLE_APPLICATION_CREDENTIALS`, or gcloud user creds)
So the zero value `Config{}` is a working config wherever ADC resolves
The `gcs:` section only exists to override ADC - an explicit key file, a custom endpoint, or anonymous for public buckets and emulators.

That's why configured is the opt-in signal rather than any individual field
Declaring the section at all is a statement of intent to use gcs, so a broken key path should fail startup
Without the section a `gs://` read is unexpected but still works if ADC resolves, and if it doesn't, the error surfaces on that read wrapped as failed to create gcs client (same as before this PR)

`TestAbsentSectionStaysLazyEndToEnd` and `TestConfiguredSectionFailsFastEndToEnd` pin both halves
Happy to move this into the doc comment on `newClient` if it's worth stating there rather than only in review

Issue #1662, opened 2026-08-02
Release-branch merge-back squash-merges, breaking tag ancestry for changelog generation


sallycr:
## Summary

`release/vX.Y` -> `main` merge-back PRs (e.g. #1597 for `release/v0.6`) are being **squash-merged**, which means the actual tagged release commit (e.g. `v0.6.0` at `227d691d`) is never an ancestor of `main`'s history.

## Evidence

```
$ git merge-base --is-ancestor v0.6.0 HEAD && echo yes || echo no
no
$ git merge-base --is-ancestor v0.5.0 HEAD && echo yes || echo no
no
$ git merge-base --is-ancestor v0.4.0 HEAD && echo yes || echo no
no
$ git log -1 --format="%H %P" <merge-back-commit-0a5321bf>
0a5321bf... 555dd1af...   # single parent = squash, not a real 2-parent merge
```

## Why this matters

`git-cliff` (used by `changelog.yml`) relies on tag reachability from the branch being changelog'd to bucket commits into per-version sections. Direct tag-to-tag ranges (`git log vA..vB`) still work regardless of ancestry with `HEAD` — confirmed while manually fixing #1658/#1659's changelog — but any tooling or manual process that assumes `git describe`/ancestry-based version resolution against `main` (or a fresh release branch cut from `main`, like `release/v0.7`) will silently misbehave. This is a contributing factor (not the sole cause) to the `CHANGELOG.md` missing its `[0.6.0]` section — the primary cause there was #1611 never merging, but even a merged #1611 wouldn't fix `git describe`-style ancestry checks against `main`.

## Suggested fix

Either:
1. Stop squash-merging release-branch merge-backs — use a real merge commit (or fast-forward if possible) so tags remain ancestors of `main`, or
2. If squash-merging is intentional (e.g. to keep `main`'s history linear), re-tag the equivalent commit on `main` after the squash-merge lands (e.g. `git tag v0.6.0-on-main <squash-commit> && git push origin v0.6.0-on-main`), or
3. Confirm this doesn't actually matter in practice for any tooling beyond changelog generation, and only fix `changelog.yml`'s specific range-based generation to always use tag-to-tag ranges (not ancestry/ ` --latest` defaults) — this is what the manual fix for #1659 relied on.

Related: #1658 (v0.7.0-rc.1 cut), the `[0.6.0]` changelog section fix in #1659.

PR #1978, opened 2026-08-31
docs(contributing): document the tag-ancestry / merge-back invariant

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Adds a "Tag ancestry and merge-backs" subsection to `CONTRIBUTING.md`'s release-process section, documenting that release tags are never ancestors of `main` (release-branch merge-backs are squash-merged) and the rule that follows from it: tooling that scopes work between releases must use explicit tag-to-tag ranges, never ancestry checks against `main`.

**Why?**

This started as a combined fix (workflow code + docs) that also patched `changelog.yml`'s tag-ancestry fallback -- paulzim reviewed and approved that version. Since then #2064 landed a cleaner fix for the same underlying bug (the release-branch-deletion case that broke v0.10.0-rc.1/v0.10.0), so the workflow-code half here is now fully superseded and conflicts with `main`. Rather than let the whole PR go stale over a diff that's moot, re-scoping to just the docs half: `CONTRIBUTING.md` currently has zero coverage of the tag-ancestry invariant (checked against current `main`), even though it's a real, previously-hit gotcha (#1662) that #2064 shows can bite twice. Worth writing down once so the next person doesn't have to rediscover it via a broken release.

**How did you test it?**

- Re-verified the core claim against current `main`: `git merge-base --is-ancestor v0.8.0 main` still returns false.
- Confirmed `#1662` and `#2046` (both referenced in the new text) are real, closed issues describing the exact bugs mentioned.
- Rewrote the paragraph describing `changelog.yml`'s current fallback mechanism to match what's actually on `main` post-#2064 (tag-based `git merge-base`, not the branch-based fallback the original version of this PR shipped) -- that wording had gone stale by the time I got back to it.
- Confirmed the new subsection sits under the correct parent heading (`## How to Release`, alongside the existing `### Patch releases` / `### Release candidates` subsections) and doesn't need a Table of Contents entry, since sibling `###` subsections aren't listed there either.
- `docs-check.yml` doesn't run against `CONTRIBUTING.md` (its `paths:` filter is `docs/**`, `website/**`, `CHANGELOG.md` only), so there's nothing to validate there beyond reading the rendered markdown by eye.

**Potential risks**

None -- documentation only, no code changed.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Not release-note-worthy (docs only).

**Documentation Changes**

This PR *is* the documentation change -- adds the "Tag ancestry and merge-backs" section to `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -303,6 +303,25 @@ If soak testing finds a problem with an RC:
 
 For a fix to an already-released version, skip the RC flow: cherry-pick the fix to the release branch, update `CHANGELOG.md`, and tag `vX.Y.Z` directly — `release.yaml` publishes the patch the same way it does a final release.
 
+### Tag ancestry and merge-backs
+
+Merge-back PRs (`release/vX.Y` → `main`) are squash-merged, so the tagged
+release commits are **never ancestors of `main`** (`git merge-base
+--is-ancestor v0.8.0 main` returns false — see #1662). This is a known,
+accepted property of the branching model, with one rule attached:
+
+- Any tooling that scopes work between releases must use **explicit
+  tag-to-tag ranges** (`git log vA..vB`), which are well-defined regardless
+  of ancestry. Never resolve versions against `main` via `git describe`,
+  `git merge-base --is-ancestor <tag> main`, or git-cliff's ancestry-based
+  `--latest` — on this repo they will silently misbehave.
+
+`changelog.yml` already follows this rule: it picks the preceding final tag
+by commit timestamp and, when that tag is not a real ancestor, falls back to
+*that tag's own* fork point off `main` (via `git merge-base`) rather than
+hunting for the release branch it came from — a fallback that keeps working
+even after the branch is deleted post-merge (see #2046).
+
 ## Creating a Pull Request
 
 If you want to fix a bug or propose a new feature you’ll do this through creating a Pull Request.


paulzim (approved):
# PR #1978 review — fix(ci): use an explicit tag-to-tag changelog range when tag ancestry is broken

**Author:** apurvapatkeshwar (Apurva Patkeshwar) · **Branch:** `fix/changelog-ancestry-1662` → `main` · **1 commit, +32/-2, 2 files** · Closes #1662.

## What it does

`.github/workflows/changelog.yml`'s "Determine changelog range" step has a defensive fallback for a tag that lands directly on `main` (against the documented release policy) instead of a `release/vX.Y` branch. That fallback previously just emitted an empty range, deferring to git-cliff's `--latest`, which derives its range from tag ancestry — exactly what #1662 found this repo's squash-merged merge-backs break (every release tag is confirmed non-ancestor of `main`). The fix: when a preceding final tag exists, use an explicit `preceding..current` range (`git log A..B` is well-defined regardless of ancestry) and only fall back to `--latest` when there's truly no preceding tag (first release ever). `CONTRIBUTING.md` gains a short "Tag ancestry and merge-backs" section documenting the invariant and the rule (tooling must use tag-to-tag ranges, never `git describe`/ancestry checks against `main`).

The `if [ "$CURRENT_BRANCH" = "main" ]` block is the *only* thing touched — the two other paths (ancestor found → tag-to-tag; release-branch fork-point fallback) are unchanged.

## Verification performed

- Read the full "Determine changelog range" step in context, not just the diff, to confirm the new `if [ -n "$CANDIDATE" ]` branch reuses the already-computed `CANDIDATE`/`CURRENT_TAG` correctly and that `exit 0` still fires for both new sub-cases.
- **Empirically tested the core claim against this repo's real tag history** (not just trusting the PR's prose), since "is `A..B` actually well-scoped when `A` is confirmed not an ancestor of `B`" is the load-bearing question here:
  - `git merge-base --is-ancestor v0.8.0 upstream-main` → false (confirmed non-ancestor, matching the CONTRIBUTING.md claim).
  - `git log --oneline v0.8.0..upstream-main` → **140** commits (out of 1343 total reachable from `upstream-main`) — reasonably scoped, not the whole history.
  - `git merge-base --is-ancestor v0.9.0 upstream-main` → false.
  - `git log --oneline v0.9.0..upstream-main` → **48** commits — a plausible count for the ~1 week since v0.9.0 (tagged 2026-08-24) given this repo's activity level.
  - This confirms the PR's technical premise: because release branches fork from `main` and share a long common prefix with it, `A..B` correctly excludes that shared prefix even without `A` being a literal ancestor of `B` — it does not degenerate into "re-include everything," which was my main concern going in.
- `python3 -c "import yaml; yaml.safe_load(...)"` → valid YAML.
- Downloaded and ran `actionlint` v1.7.12 directly against the modified file (the PR claims this was run locally; I verified it independently rather than trusting the claim) → clean, exit 0.
- `bash -n` on the extracted step script → clean.
- `gh pr checks 1978`: all real checks pass (`bazel-build`, `dirty-check`, `filter`, `python-build`, `update-release-draft`); the three advisory checks (Go API Surface, Helm Values Key Diff, Proto Breaking Changes) correctly skip for a workflow/docs-only change.
- Confirmed the changed path is genuinely dead-in-practice today: tags are only ever cut on `release/vX.Y` branches per the documented process, so `CURRENT_BRANCH = "main"` should not occur for a real release — this is purely a defensive/correctness improvement for a mis-dispatched tag, as the PR states.

## Findings

None. This is a small, precisely-scoped fix: it touches only the one remaining ancestry-dependent path (confirmed via the PR's own stated sweep of `.github/workflows/` and `scripts/`, which I have no reason to doubt given how narrow and self-contained the change is), the new logic is correct and matches real repo data, and the documentation addition accurately describes verified repository behavior.

## Verdict: **Approve**

Clean fix for a real, previously-reported bug (#1662), correctly scoped, backed by the author's own fixture-driven testing plus my independent empirical verification against actual tag/commit history, clean lint, and passing CI.

apurvapatkeshwar:
@paulzim approved and green - ready to merge whenever you have a moment

apurvapatkeshwar:
@paulzim Heads up - the scope changed materially since your approval, so flagging before this merges on a stale review

The workflow code related half you reviewed (`changelog.yml` ancestry fallback) is now superseded by #2064, which landed a cleaner fix for the same bug in the meantime. I dropped that chunk entirely - it now conflicts with main and there's nothing left for it to do. What's left is just the `CONTRIBUTING.md` addition: documenting the tag-ancestry/merge-back invariant you'd already independently verified against real tag history in your review.

I also corrected its last paragraph, which described the old branch-based fallback - it now matches #2064's actual mechanism. Would appreciate a fresh look at the docs-only version before it merges

sallycr (approved):

PR #1979, opened 2026-08-31
fix(sandbox): record a pod-reachable API endpoint in the compute Cluster CRD

### LINKED ISSUE #1901: Sandbox Cluster CRD records unreachable API address (0.0.0.0), so no KubeRay RayCluster is ever created

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Fixes #1901: `_create_compute_cluster_crd` copied the API server address straight out of `k3d kubeconfig get` into the `Cluster` CRD. That is the host-side endpoint (for example `https://0.0.0.0:59896`), which the controllermgr cannot reach from inside the cluster, so every KubeRay `RayCluster` creation failed with `connection refused` and no `rayclusters.ray.io` object was ever created.

The endpoint now goes through a small helper, `_cluster_endpoint_for_crd`:

- The sandbox cluster itself maps to the in-cluster service endpoint, `https://kubernetes.default.svc:443` -- the fix the issue reporter verified live.
- For any other k3d compute cluster, a known-unroutable host (`0.0.0.0`, `127.0.0.1`, `localhost`, `host.docker.internal`) maps to the cluster's server container on the shared docker network, `https://k3d-<name>-server-0:6443`, which pods can resolve (per the issue's reachability table) and whose port is stable across cluster recreations, unlike the host-mapped port that re-randomises on every `k3d cluster create`.
- A genuinely routable address passes through unchanged.

**Why?**

The controllermgr runs inside the cluster; `0.0.0.0` is a bind address there, not a destination, and the `host.docker.internal` name the old code comment expected does not resolve from pods either. Writing either into the CRD breaks every Ray-backed pipeline task on a fresh sandbox.

**How did you test it?**

- 4 new unit tests on the helper: sandbox cluster maps to `kubernetes.default.svc:443`; each unroutable host maps to the k3d server container endpoint; a routable address passes through verbatim; a malformed server URL raises. Full sandbox CLI suite passes (10 tests).
- `ruff check` and `ruff format --check` clean on both touched files.
- The in-cluster endpoint itself is the fix the issue reporter verified by patching a live sandbox (their "Verified fix" section), and the `k3d-<name>-server-0:6443` reachability is from the same report's table.

**Potential risks**

- A deployment whose kubeconfig legitimately reports a loopback address that is somehow reachable from pods (for example a host-network setup outside k3d) would now be rewritten to the k3d container name. The sandbox CLI only creates k3d clusters, so this combination does not arise from `ma sandbox` itself.
- The multi-cluster (non-sandbox) mapping is grounded on the issue's reachability table rather than an end-to-end run here; the sandbox-itself path is the reporter-verified one.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Sandbox: the compute `Cluster` CRD now records an API endpoint reachable from inside the cluster (in-cluster service endpoint for the sandbox itself, the k3d server container address for other local clusters), so KubeRay `RayCluster` objects are actually created and Ray-backed pipeline tasks run.

**Documentation Changes**

None -- no user-facing parameter changes; the sandbox just works as documented.

### DIFF
--- a/docs/getting-started/sandbox-setup.md
+++ b/docs/getting-started/sandbox-setup.md
@@ -97,6 +97,8 @@ Docker containers need to communicate with services on your host machine. Verify
    ```
 3. If missing, add it to the end of the file and save.
 
+> **Note:** this only fixes host-side resolution. It does not make `host.docker.internal` (or any other host address) reachable from inside pods — pod-to-host and pod-to-pod traffic requires the sandbox's k3d/docker-network addressing (see `_cluster_endpoint_for_crd` in `sandbox.py`), not `/etc/hosts` edits.
+
 ### Install Python dependencies
 
 From the repository you cloned, install the Michelangelo AI Python packages:
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -1937,6 +1937,34 @@ def _ensure_namespace_exists(namespace: str):
         print(f"Created namespace '{namespace}' in the sandbox cluster.")
 
 
+def _cluster_endpoint_for_crd(cluster_name: str, server_url: str) -> tuple[str, str]:
+    """Pick an API endpoint for the Cluster CRD that is reachable from inside pods.
+
+    `k3d kubeconfig get` returns the host-side endpoint (for example
+    `https://0.0.0.0:59896`), which the controllermgr cannot reach from inside
+    the cluster. Returns a `(host, port)` pair that pods can actually dial.
+    """
+    if cluster_name == _michelangelo_sandbox_kube_cluster_name:
+        # The sandbox cluster itself: use the in-cluster service endpoint.
+        return "https://kubernetes.default.svc", "443"
+
+    import re
+
+    match = re.search(r"(https://[^:]+):(\d+)", server_url)
+    if not match:
+        raise ValueError(
+            f"Could not extract cluster host and port from server URL: {server_url}"
+        )
+    host, port = match.groups()
+
+    unroutable = {"0.0.0.0", "127.0.0.1", "localhost", "host.docker.internal"}
+    if host.removeprefix("https://") in unroutable:
+        # Another k3d cluster on the shared docker network: its server
+        # container name resolves from pods and its port is stable.
+        return f"https://k3d-{cluster_name}-server-0", "6443"
+    return host, port
+
+
 # Given a cluster name, create a Cluster CRD in the sandbox cluster
 def _create_compute_cluster_crd(cluster_name: str):
     """Create a Cluster CRD for the Ray jobs cluster in the sandbox cluster."""
@@ -1954,16 +1982,7 @@ def _create_compute_cluster_crd(cluster_name: str):
     # Extract server URL from clusters[0].cluster.server
     server_url = kubeconfig_data["clusters"][0]["cluster"]["server"]
 
-    # Extract host and port from server URL
-    # Example: "https://host.docker.internal:52910"
-    import re
-
-    match = re.search(r"(https://[^:]+):(\d+)", server_url)
-    if not match:
-        raise ValueError(
-            f"Could not extract cluster host and port from server URL: {server_url}"
-        )
-    host, port = match.groups()
+    host, port = _cluster_endpoint_for_crd(cluster_name, server_url)
 
     # Create Cluster CRD manifest
     cluster_crd = {
--- a/python/michelangelo/cli/sandbox/sandbox.py
+++ b/python/michelangelo/cli/sandbox/sandbox.py
@@ -30,6 +30,17 @@ _michelangelo_sandbox_kube_cluster_name = "michelangelo-sandbox"
 _cadence_domain = "default"
 _default_compute_kube_cluster_name = "michelangelo-compute-0"
 
+# Hosts that `k3d kubeconfig get` may return but that are not reachable from
+# inside pods (host-side loopback/docker-gateway addresses).
+_unroutable_cluster_hosts = {
+    "0.0.0.0",
+    "127.0.0.1",
+    "localhost",
+    "host.docker.internal",
+}
+# Port the k3d server container listens on within the shared docker network.
+_k3d_server_port = "6443"
+
 # Path to the Michelangelo Helm chart (relative to this file)
 _chart_dir = Path(__file__).parent.parent.parent.parent.parent / "helm" / "michelangelo"
 
@@ -1957,11 +1968,10 @@ def _cluster_endpoint_for_crd(cluster_name: str, server_url: str) -> tuple[str, 
         )
     host, port = match.groups()
 
-    unroutable = {"0.0.0.0", "127.0.0.1", "localhost", "host.docker.internal"}
-    if host.removeprefix("https://") in unroutable:
+    if host.removeprefix("https://") in _unroutable_cluster_hosts:
         # Another k3d cluster on the shared docker network: its server
         # container name resolves from pods and its port is stable.
-        return f"https://k3d-{cluster_name}-server-0", "6443"
+        return f"https://k3d-{cluster_name}-server-0", _k3d_server_port
     return host, port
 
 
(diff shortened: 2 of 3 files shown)


sallycr:
## Automated review from the OSS agent team (Advocate pass)

**Verdict: PASS** (no MUST findings)

The fix is well-reasoned and grounded directly in issue #1901's reachability table. The new `_cluster_endpoint_for_crd` helper cleanly isolates the previous inline regex-parsing logic and adds the three-way branch (sandbox cluster → in-cluster service endpoint, known-unroutable host → k3d server container DNS name, otherwise pass through). This addressing pattern (`k3d-<cluster_name>-server-0:6443`) is also already used and documented elsewhere in this same file for `_create_inference_compute_cluster`'s cross-cluster reachability — this fix is consistent with an existing, working convention rather than inventing a new one.

**Test coverage:** good — genuinely meaningful, not just line coverage. The four new tests hit all three logical branches (sandbox-cluster mapping, all four unroutable hosts, pass-through of a routable host) plus the malformed-URL error path, asserting on both host and port.

**Docstrings/errors:** the helper's docstring explains the host-side vs. pod-side reachability distinction clearly for contributors unfamiliar with the k3d docker-network setup; the `ValueError` includes the offending `server_url`, which is actionable for debugging.

### Two non-blocking SHOULD suggestions

1. **Docs caveat** — `docs/getting-started/sandbox-setup.md:89-96` tells readers to add `host.docker.internal` to `/etc/hosts`, with no note that this only helps host-side resolution and doesn't make pod-to-host or pod-to-pod traffic work — exactly the confusion behind issue #1901. Not required for this PR (that doc's original purpose is unrelated to Ray-cluster reachability), but a one-line caveat there would preempt the next person from trying to "fix" pod reachability by editing `/etc/hosts`.
2. **Maintainability** — `unroutable = {"0.0.0.0", "127.0.0.1", "localhost", "host.docker.internal"}` and the hardcoded `"6443"` are local literals inside the function, while the same addressing scheme is referenced elsewhere in the file. Consider hoisting these into module-level constants (matching the existing `_michelangelo_sandbox_kube_cluster_name` convention) so the two sites can't drift independently. Purely cosmetic.

No correctness or edge-case concerns found.

sallycr:
## Follow-up: draft patches for the two SHOULD suggestions above

These aren't postable as GitHub inline "suggested change" comments since neither touched file/line range is part of this PR's diff (the docs file isn't touched by this PR at all, and the constants insertion point predates the diff hunk) — posting as plain patches instead, for you to take or leave.

### 1. Docs caveat — `docs/getting-started/sandbox-setup.md`

Add a note after the existing `/etc/hosts` steps so the next person doesn't try to "fix" pod reachability by editing it:

```diff
 3. If missing, add it to the end of the file and save.
+
+> **Note:** this only fixes host-side resolution. It does not make `host.docker.internal` (or any other host address) reachable from inside pods — pod-to-host and pod-to-pod traffic requires the sandbox's k3d/docker-network addressing (see `_cluster_endpoint_for_crd` in `sandbox.py`), not `/etc/hosts` edits.
```

### 2. Hoist literals to module-level constants — `python/michelangelo/cli/sandbox/sandbox.py`

(Checked `_create_inference_compute_cluster` — its `6443`/`k3d-<cluster_name>-server-0` reference is inside a comment string only, not executable code, so nothing to hoist there; just the new helper.)

**Add constants** near the file's other module-level names (e.g. after `_default_compute_kube_cluster_name`):

```diff
 _cadence_domain = "default"
 _default_compute_kube_cluster_name = "michelangelo-compute-0"
+
+# Hosts that `k3d kubeconfig get` may return but that are not reachable from
+# inside pods (host-side loopback/docker-gateway addresses).
+_unroutable_cluster_hosts = {"0.0.0.0", "127.0.0.1", "localhost", "host.docker.internal"}
+# Port the k3d server container listens on within the shared docker network.
+_k3d_server_port = "6443"
```

**Use them in `_cluster_endpoint_for_crd`:**

```diff
-    unroutable = {"0.0.0.0", "127.0.0.1", "localhost", "host.docker.internal"}
-    if host.removeprefix("https://") in unroutable:
+    if host.removeprefix("https://") in _unroutable_cluster_hosts:
         # Another k3d cluster on the shared docker network: its server
         # container name resolves from pods and its port is stable.
-        return f"https://k3d-{cluster_name}-server-0", "6443"
+        return f"https://k3d-{cluster_name}-server-0", _k3d_server_port
     return host, port
```

Naming follows the file's existing convention (`_michelangelo_sandbox_kube_cluster_name`, `_cadence_domain`, `_default_compute_kube_cluster_name` — lowercase snake_case, leading underscore, module scope).

Both are optional/cosmetic per the original review — feel free to take, adapt, or skip.

PR #1980, opened 2026-08-31
fix(sandbox): back the mysql Pod with a data volume so restarts keep the databases

### LINKED ISSUE #1908: Sandbox MySQL has no data volume, so any restart destroys the Cadence schema and all metadata

apurvapatkeshwar:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Fixes #1908: the sandbox mysql Pod (`python/michelangelo/cli/sandbox/resources/mysql.yaml`) declared no volumes at all, so all data lived in the container's writable layer and any restart -- VM restart, `k3d cluster stop && start`, pod recreation -- destroyed the `cadence`, `cadence_visibility`, and `michelangelo` databases. The Cadence tier then wedged indefinitely on `wait-for-schema`, because the schema Job is already `Complete` and never re-runs.

This mounts `/var/lib/mysql` from a `hostPath` volume (`/shared/mysql-data`, `DirectoryOrCreate`) -- exactly the pattern the sandbox's minio Pod already uses for `/data` (`/shared/minio-data`), which the issue notes survived every restart in the reporter's testing.

**Why?**

`ma sandbox stop` / `ma sandbox start` are documented lifecycle commands; they should return a working environment, not one whose entire database tier is gone. This is the issue's suggested fix, kept deliberately minimal: it does not convert the bare Pod to a Deployment/StatefulSet, both to stay one-concern and because the in-flight #1958 adds sync-time self-healing that operates on the Pod by name -- the two changes compose (self-heal revives the Pod, this volume makes the revival lossless).

**How did you test it?**

- YAML parses (`yaml.safe_load_all`) and the Pod spec's `volumeMounts`/`volumes` structure verified programmatically to match the minio precedent.
- The mysql:8.0 image's entrypoint skips database initialization when the data directory is non-empty, so a restarted Pod picks up the existing databases -- standard image behavior, the same contract minio relies on.
- Not exercised end-to-end here (no local k3d): the restart scenario is the issue's own reproduction; with the volume, step 4's `Unknown database 'cadence'` cannot occur since the datadir persists.

**Potential risks**

- Stale data now persists across `ma sandbox delete`/`create` cycles on the same host if the k3d volume mount recurs, where before every mysql Pod started fresh. `_refresh_mysql_schema` already drops and recreates the `michelangelo` database on every sync, which covers the platform schema; Cadence keeps its history, which is the point of the fix.
- Disk usage on the host node grows with retained history (same profile as the existing minio hostPath).

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Sandbox: the mysql Pod now stores its data on a hostPath volume, so the Cadence and platform databases survive VM restarts, `k3d cluster stop/start`, and Pod recreation.

**Documentation Changes**

None -- documented lifecycle commands (`ma sandbox stop`/`start`) now behave as documented.

### DIFF
--- a/python/michelangelo/cli/sandbox/resources/mysql.yaml
+++ b/python/michelangelo/cli/sandbox/resources/mysql.yaml
@@ -19,6 +19,9 @@ spec:
           value: temporal
       # Note: Cadence auto-setup will create cadence/cadence_visibility databases
       # Ingester will create michelangelo database via mysql-ingester.yaml Job
+      volumeMounts:
+        - mountPath: /var/lib/mysql
+          name: mysql-data
       readinessProbe:
         exec:
           command:
@@ -49,6 +52,11 @@ spec:
         timeoutSeconds: 5
         successThreshold: 1
         failureThreshold: 3
+  volumes:
+    - name: mysql-data
+      hostPath:
+        path: /shared/mysql-data
+        type: DirectoryOrCreate
 ---
 apiVersion: v1
 kind: Service


paulzim (approved):
# PR #1980 review — fix(sandbox): back the mysql Pod with a data volume so restarts keep the databases

**Author:** apurvapatkeshwar (Apurva Patkeshwar) · **Branch:** `fix/sandbox-mysql-data-volume` → `main` · **1 commit, +8/-0, 1 file** · Fixes #1908.

## What it does

The sandbox's `mysql.yaml` declares a bare `Pod` with no volumes at all, so `cadence`, `cadence_visibility`, and `michelangelo` all lived in the container's writable layer — any restart (VM restart, `k3d cluster stop`/`start`, pod recreation) wiped them. Because Cadence's schema-setup Job is a one-shot `Job` that never re-runs once `Complete`, losing the data (but not the `Complete` status) wedges the whole Cadence tier on `wait-for-schema` forever, per the exact repro in #1908.

The fix mounts `/var/lib/mysql` from a `hostPath` volume (`/shared/mysql-data`, `DirectoryOrCreate`) — the identical pattern the sandbox's minio Pod already uses for `/data` (`/shared/minio-data`). Deliberately minimal: doesn't convert the bare Pod to a Deployment/StatefulSet, both to stay one-concern and because the in-flight #1958 (self-heals the mysql Pod on `sync` if not `Running`) operates on the Pod by name — the two compose (#1958 revives the Pod, #1980 makes the revival lossless).

## Verification performed

- Read the full diff and the post-patch `mysql.yaml` — `volumeMounts`/`volumes` block correctly cross-references `name: mysql-data` on both sides; confirmed valid YAML and correct linkage via `yaml.safe_load_all`.
- Read `minio.yaml` and confirmed the PR's central claim is literally true: identical `hostPath`/`DirectoryOrCreate` idiom, identical `/shared/<service>-data` naming convention, already working in production use.
- Confirmed the official `mysql:8.0` entrypoint's documented behavior — it skips `mysql_install_db`/init when `/var/lib/mysql` is already non-empty — is the correct, standard mechanism this fix relies on.
- Read #1908 in full: the repro (kill/restart mysql pod → `wait-for-schema` init containers stuck → `ERROR 1049 (42000): Unknown database 'cadence'`) and root-cause explanation match the PR's description exactly. This is also independently corroborated by this reviewer's own prior real-world sandbox notes documenting the same MySQL/restart data-loss failure mode on a live Mac sandbox, hit before this PR existed.
- Read `_delete()`/`_create()` in `sandbox.py`: `k3d cluster create` is invoked with no `-v`/`--volume` flag, so `/shared/` is never bind-mounted to the true host filesystem — it lives only in the k3d node container's own writable layer, which `k3d cluster delete` destroys. The PR's own "Potential risks" section already hedges this correctly ("stale data now persists across delete/create... *if the k3d volume mount recurs*") rather than asserting it as fact — and based on the code plus this reviewer's own prior documented observation that a full sandbox delete+create on a live sandbox loses all volume state, that condition does not actually hold. Net effect: the fix delivers exactly the restart-survival behavior it's designed for (`stop`/`start`, pod recreation, VM restart) without the caveated delete/create risk actually materializing — a strictly better outcome than even the PR's own conservative risk framing suggests.
- Confirmed #1958 (referenced as the complementary in-flight self-heal PR) is real, open, and titled consistent with the PR's description — the composability claim checks out.
- `kubectl apply --dry-run=client` against the manifest isn't possible from this environment (no live cluster reachable, and client-side dry-run still needs API discovery from a real apiserver) — expected environment limitation, not a PR defect. Structural validation via `yaml.safe_load_all` is the appropriate substitute here, and mirrors exactly the testing method the PR author used.
- `gh pr checks 1980`: all real checks pass (`python-build`, `Python Run Ruff on Changed Files`, `filter`, `update-release-draft`, `bazel-build`, `dirty-check`); advisory checks (Go API Surface, Helm Values Key Diff, Proto Breaking Changes) correctly skip for a sandbox-resource-only YAML change.

## Findings

None. Small, surgical, correctly-targeted fix for a real, previously-hit bug, following an established working pattern exactly, with an honestly and conservatively hedged risk disclosure.

## Verdict: **Approve**

Minimal, well-scoped fix that mirrors a proven existing pattern (minio's hostPath) to close a real, reproducible, previously-documented bug (#1908) affecting documented sandbox lifecycle commands. No blocking findings.

apurvapatkeshwar:
@paulzim this is approved/green too; please consider merging :)

sallycr:
## Sandbox validation: PASS

Validated end-to-end on a live k3d sandbox by directly reproducing the bug on the unpatched manifest, then confirming the fix.

**1. Reproduced the bug (baseline, unpatched `mysql.yaml`, matches `main`):**
- Confirmed no `volumes`/`volumeMounts` on the live `mysql` Pod.
- Had 14 real rows in `michelangelo.pipeline_run`, plus `cadence`/`cadence_visibility`/`michelangelo` databases present.
- Deleted and recreated the `mysql` Pod (simulating a restart/pod recreation) with the unpatched manifest.
- Result: `ERROR 1049 (42000): Unknown database 'michelangelo'` — all three databases and the pipeline history were gone. This matches the issue's reported `Unknown database 'cadence'` symptom exactly.

**2. Applied this PR's patched `mysql.yaml`:**
- Deleted the pod again and applied the patched manifest — pod came up with the `hostPath` volume at `/shared/mysql-data` mounted to `/var/lib/mysql`, confirmed via `kubectl get pod mysql -o jsonpath='{.spec.volumes}'`.
- First boot on the new volume was necessarily empty (fresh hostPath dir), so I ran `ma sandbox sync` once to recreate schema — this matches the issue's own documented recovery path.
- **Definitive persistence test:** deleted and recreated the `mysql` Pod a second time, now that the `hostPath` volume actually had real data in it. Result: `cadence`, `cadence_visibility`, and `michelangelo` all survived intact, and the Cadence tier (frontend/history/matching/worker) stayed `Running` throughout without needing to restart or re-wedge on `wait-for-schema`.

**Conclusion:** the fix works as designed — a `mysql` Pod restart/recreation with this patch no longer destroys the Cadence or platform metadata schema. No blocking concerns.

PR #1985, opened 2026-08-31
docs(scheduler): operator guide for the Kueue backend and updated extension guide

apurvapatkeshwar:
Stacked on #1984 -- review only the last commit; the earlier commits belong to #1984 and its predecessors.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Documentation for the Kueue scheduler backend shipped in the previous PRs of this series (michelangelo-ai/enhancements#19).

- New operator guide, "Kueue Scheduler Backend": how the delegation design works (validate at enqueue, label at dispatch, Kueue's own RayCluster integration admits on the compute cluster), the one-ClusterQueue-per-cluster / one-LocalQueue-per-project topology, step-by-step enablement (Kueue install, `scheduler_type` on the Cluster spec, Helm values), a configuration reference table, observability (`Queued` condition, `KueueQueueNotFound`/`KueueQueueCheckFailed`, metrics), compatibility notes (v1beta2/Kueue 0.15+ default with `apiVersion: v1beta1` for older estates, why RayJobs are never queue-labeled, the `waitForPodsReady` caveat for RayClusters), per-step rollback, and the sandbox walkthrough.
- "Extend the Job Scheduler" guide: Option 2 (custom JobQueue) previously sketched a hypothetical `KueueScheduler` that created Kueue `Workload` objects from the control plane. That sketch is replaced with the real, shipped package and the actual `jobs.scheduler.backend` factory switch, plus guidance on when to wrap the default scheduler versus fully replace it.
- Operator-guides index: added the new page to the Jobs & Compute table.

No release-notes entry needed beyond the previous PR's -- the changelog is generated from conventional commits.

**Why?**

The extension guide's Kueue sketch now contradicts the shipped implementation (hand-crafted Workloads versus delegation to Kueue's integrations), and operators enabling the backend need a single page covering topology, configuration, failure modes, and rollback.

**How did you test it?**

`bun install --frozen-lockfile && bun run build` in `website/` succeeds with Docusaurus link checking in throw mode (a broken internal link fails the build). Config keys, condition names, reasons, and metric names in the guide were checked against the code they document.

**Potential risks**

None -- documentation only.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

None (documentation).

**Documentation Changes**

This PR is the documentation change for the Kueue scheduler backend series.


paulzim (approved):
**Approve**

Reviewed only this PR's own commit (`9dddda75`) — per the PR description, it's stacked on #1984 and the earlier commits (`909fd8e6`, `df461c648`, `3fb810b9`) belong to that stack, not to this PR. Isolated the diff via `git log upstream/main..pr-1985` + `git show --stat 9dddda75`, and reconciled the insert/delete counts against the PR-level per-file stats to confirm the isolation is complete and correct.

What changed: two docs, both additive/edits to prose, no code.

- **New:** `docs/operator-guides/jobs/kueue-scheduler-backend.md` — operator guide for the Kueue scheduler backend (delegation model, queue topology/naming, enablement steps, config reference, observability, compatibility notes, rollback, sandbox walkthrough).
- **Edited:** `docs/operator-guides/jobs/extend-michelangelo-batch-job-scheduler-system.md` — replaces an outdated *hypothetical* `KueueScheduler` code sketch (fictional `kueueclient.Interface`, direct `Workload` creation, a `"kueue"`/`"volcano"` string-based factory) with accurate references to the real shipped `go/components/jobs/scheduler/kueue` package and an updated, now-accurate illustrative snippet.
- **Nav:** `docs/operator-guides/index.md` gains one link to the new guide, correctly placed next to the other jobs docs.

I independently verified every specific technical claim in both docs against the actual shipped code (not just the PR's own "checked against the code" claim) — everything checked out with zero discrepancies:

- `proto/api/v2/cluster.proto`: `SCHEDULER_TYPE_KUEUE = 2` enum value and `scheduler_type = 8` field — match.
- `go/base/config/config.go`: `SchedulerConfig{Backend, Kueue}` / `KueueConfig{LocalQueueTemplate, LocalQueueOverrides, APIVersion}` — match the doc's config-path claims exactly.
- Defaults: doc correctly identifies these live in Go, not Helm (`values.yaml` has them as empty strings) — `DefaultKueueAPIVersion = "v1beta2"` (`kueue/localqueues.go`) and `DefaultLocalQueueTemplate = "ma-{project}"` + the override→template→default resolution logic in `ResolveLocalQueueName` (`jobs/common/utils/utils.go`) — match.
- Label/condition/reason/metric strings: `kueue.x-k8s.io/queue-name`, `KueueQueueNotFound`, `KueueQueueCheckFailed` (`jobs/common/constants/constants.go`), `ClusterAdmitted`/`AwaitingAdmission` (`ray/cluster/controller.go`), `kueue.queue_not_found_count`/`kueue.queue_check_failed_count` under the `kueuescheduler` scope (`scheduler/kueue/kueue.go`) — all match.
- Security claim ("user-supplied labels can never pick the queue"): `mapper_test.go`'s `TestMapLabels`/`TestMapRayClusterKueueLabels` explicitly forge a `kueue.x-k8s.io/queue-name` label on the input and assert only the server-resolved value survives — this is genuinely enforced and tested, not just asserted in prose.
- RayJob exemption claim: matches the code comment in `k8sengine/ray.go` verbatim in substance.
- Sandbox CLI claims: `ma sandbox demo kueue` (`_create_kueue_demo_crs`) and `--compute-cluster-name` flag both genuinely exist in `sandbox.py`.
- Extension guide's rewritten illustrative snippet (`_backendDefault`/`_backendKueue` constants, `provide(in ProvideIn) (JobQueue, error)` factory switch, `kueue.NewKueueJobQueue(kueue.Params{...})` wrapping the default scheduler): matches `go/components/jobs/scheduler/module.go` exactly, down to the error message format for an unrecognized backend value.

Three non-blocking notes, none of which should hold up merging this PR itself:

1. **Merge order:** #1984 (which this stacks on) is still open/unmerged, so this PR can't land on `main` until #1984 and its own predecessors merge first. Not a defect, just a sequencing note for whoever merges.
2. **`Proto Breaking Changes (advisory)` shows FAILURE**, but it's unrelated to this PR and to this PR's own commit (which touches no proto files). Root-caused it: `proto/api/v2/cached_output.proto` has an import (`k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto`) that buf can't resolve — confirmed this already exists byte-for-byte on `upstream/main`, predating this PR entirely. The workflow (`.github/workflows/api-surface-check.yml`) actually has a filter meant to suppress exactly this class of noise (`grep -v "imported file does not exist\|cannot find.*in this scope"`), but the pattern doesn't match buf's actual wording (`import "...": file does not exist` — no "imported" prefix), so it leaks through as a false positive. That then trips a second, separate bug: the job tries to post an advisory PR comment via `actions/github-script` using `secrets.GITHUB_TOKEN`, which GitHub Actions forces to read-only for `pull_request`-triggered runs from external contributors regardless of the workflow's declared `permissions: pull-requests: write` — the POST 403s, and the resulting unhandled promise rejection is what actually turns the job red (not a real breaking-change finding). Worth a follow-up issue for the grep pattern and the missing k8s.io well-known-types import in the buf module config, but out of scope for a docs-only PR.
3. Minor hygiene: the branch includes an odd merge commit (`b75e1613`, "Merge remote-tracking branch 'localmain/main' into merge-1985") — confirmed via the GitHub API that it's genuinely part of the PR's own commit list, not local noise. Harmless, just a bit unusual to see in the history.

Didn't re-run the Docusaurus build (`bun install --frozen-lockfile && bun run build`) myself — no bun/node toolchain available in this environment — but `validate-docs` and the PR's own stated test method both passed, and I have no reason to doubt the claim given everything else checked out.

Solid, accurate documentation PR. Approve, pending #1984 landing first.

PR #1986, opened 2026-08-31
fix(ui): show PipelineRuns on the Pipeline Detail page

craigmarker:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

The Pipeline Detail page's Runs tab and PipelineRun creation
disagreed on which label identifies a run's owning pipeline, so runs
never showed up. Stamp that label consistently on every creation
path, and point the frontend at the same key.

**Why?**

Trigger-created runs were labeled differently than what the frontend
queried for, and manually-created runs (the common "Run" button path)
got no pipeline-name label at all. Either way, the Runs tab came up
empty.

I considered a `fieldSelector` on `spec.pipeline.name` instead of a
label, since `PipelineRun` has a proto index annotation on that
field. But field selectors only resolve through the metadata-storage
list path, which is off by default — so they'd silently fail to
filter in most deployments (the same class of bug #1973 hit). A label
works regardless of storage backend.

**How did you test it?**

Added two Go unit tests covering the new stamping behavior; existing
Go and JS test suites pass, along with `go build`, `yarn typecheck`,
and `yarn lint`.

I also verified end-to-end in a local sandbox: reproduced the empty
Runs tab against the unfixed apiserver
(`02-runs-still-empty-before-label-fix.png`), confirmed the frontend
fix alone by manually labeling a run
(`03-runs-visible-after-label-fix.png`), then deployed the real fix
and confirmed a run created through the UI is auto-labeled and shows
up with no manual intervention
(`05-real-run-visible-full-e2e.png`).

**Potential risks**

Low — the label is additive (never overwrites an existing value) and
follows the same stamping pattern already used for `ownerRef` and
`SourcePipelineType` on the same hook. PipelineRuns created before
this fix won't be backfilled.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/go/api/api.go
+++ b/go/api/api.go
@@ -74,6 +74,11 @@ const (
 	// SourcePipelineManifestTypeLabelName is the Kubernetes label key that identifies
 	// the pipeline manifest type (e.g. PIPELINE_MANIFEST_TYPE_ASL).
 	SourcePipelineManifestTypeLabelName = "pipeline.michelangelo/PipelineManifestType"
+
+	// PipelineNameLabelName is the Kubernetes label key that identifies the
+	// owning Pipeline's metadata.name on a PipelineRun, used to filter runs
+	// down to a single pipeline (e.g. on the Pipeline detail page).
+	PipelineNameLabelName = "pipelinerun.michelangelo/pipeline-name"
 )
 
 // DefaultContextTimeout defines the default timeout for the context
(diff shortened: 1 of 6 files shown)


craigmarker:
Consolidating this fix into #1967 instead — going with fieldSelector per Craig's call, documented with its EnableMetadataStorage caveat. Closing this one in favor of that PR.

PR #1988, opened 2026-08-31
feat(auth): swap the bespoke Auth contract for the k8s.io/apiserver authenticator and authorizer interfaces

apurvapatkeshwar:
Stacked on #1852 -- review only the last commit; the earlier commit belongs to #1852.

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

The bespoke `Auth`/`DummyAuth` contract is retired in favor of the upstream `k8s.io/apiserver` interfaces, per the design in michelangelo-ai/enhancements#20:

- `go/auth` now exposes `authenticator.Token` and `authorizer.Authorizer` (as `authapi.TokenAuthenticator` / `authapi.Authorizer` aliases), provided together through an `AuthBundle` (`fx.Out`). The shipped default, `NewAllowAllBundle`, preserves the permissive behavior: an always-authenticated anonymous identity plus `authorizerfactory.NewAlwaysAllowAuthorizer()`.
- The generated handlers switch to authenticate-then-authorize: `authapi.Authenticate` resolves the caller's `user.Info` from the bearer token (absent token passed through — the authenticator decides), and `authapi.Authorize` builds an `authorizer.AttributesRecord{User, Verb, APIGroup: "michelangelo.api", Resource, Namespace, ResourceRequest: true}` and denies unless the decision is allow. Verbs are the lowercased actions (`DeleteCollection` -> `deletecollection`); resources are the CRD plurals (including `modelfamilys`, which is what the ModelFamily CRD actually declares). Unknown action or kind fails closed.
- `go/kubeproto/templates/crd_svc.tmpl` emits the new block in all six verb handlers; all 15 `proto-go/api/v2/*.pb.kubeyarpc.go` files regenerated.
- `go/cmd/apiserver/main.go` wires `auth.AuthModule` (the allow-all bundle at this point in the series; the config-driven factory arrives with the k8s-rbac PR).

**Why?**

Adopting the interfaces `kube-apiserver` itself composes gives every later piece of #20 (OIDC, TokenReview, SubjectAccessReview, caches) a drop-in home instead of a bespoke seam, and makes the auth boundary testable against the upstream contracts.

**How did you test it?**

- `go test ./auth/` — unit tests for the helpers (bearer extraction, attribute mapping, deny-unless-allow, fail-closed on unknowns) plus cross-checks that `APIGroup` matches `v2pb.GroupVersion.Group` and the resource map matches the CRD plurals decoded from `v2pb.YamlSchemas`.
- An end-to-end allow-all test builds a real generated handler (`v2pb.NewClusterServiceHandler`) with the default bundle and asserts Create/Get/List reach the inner handler with no token on the context — the retired `DummyAuth` behavior, pinned.
- `tools/gen-proto-go.sh` followed by `git diff --exit-code`: regeneration is byte-stable.
- `go build ./...`; `bazel test //go/auth/...`.

**Potential risks**

Behavior-preserving under the shipped default by construction (allow-all before and after, covered by the e2e test). Any out-of-tree implementation of the old `Auth` interface no longer compiles — see Breaking Changes.

**Breaking Changes**

- [ ] No breaking changes

**Migration guide**

Out-of-tree `auth.Auth` implementations: implement `authenticator.Token` (identity) and `authorizer.Authorizer` (decisions) from `k8s.io/apiserver` instead, and provide them via `auth.AuthBundle`. `UserAuthenticated`/`UserAuthorized` call sites map to `auth.Authenticate` / `auth.Authorize`. Runtime behavior of a stock deployment is unchanged.

**Release notes**

The API server's auth extension point is now the upstream `k8s.io/apiserver` authenticator/authorizer contract. Default behavior (allow all) is unchanged.

**Documentation Changes**

None in this PR; the operator guide is corrected at the end of the series.


apurvapatkeshwar:
@andriika #1988–#1992 stack implements the enh#20 design you reviewed, reworked per your two comments
Flagging it's ready when you have the bandwidth

PR #1993, opened 2026-08-31
feat(ui): add Last Updated and Environment columns to pipeline run list

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
- Reordered/expanded the pipeline run list's columns from `Name, Created, Pipeline, Started by, State` to `Pipeline run name, Pipeline, Last Updated, Created, Environment, Started by, State`.
- Renamed the "Name" column to "Pipeline run name".
- Added a "Last Updated" column backed by a new `getCrdLastUpdatedSeconds()` utility (sibling to the existing `getCrdUpdatedSeconds()`) that reads the `michelangelo/UpdateTimestamp` label — set on *any* update to the resource, not just spec changes — with the same `creationTimestamp` fallback. A pipeline run's `status` changes continuously as steps execute while its `spec` is effectively immutable after creation, so the existing spec-only utility would render a value frozen at creation time for nearly every run.
- Added an "Environment" column reusing the existing `readEnvironmentLabel()` utility.
- Kept the existing "Created" column, moved to sit immediately after "Last Updated", reusing the `RUN_CREATED_COLUMN` constant now exported from `shared.ts`.
- Refactored `run/shared.ts` to export each of `SHARED_RUN_CELL_CONFIG`'s four entries individually (`RUN_CREATED_COLUMN`, `RUN_PIPELINE_COLUMN`, `RUN_STARTED_BY_COLUMN`, `RUN_STATE_COLUMN`), with `SHARED_RUN_CELL_CONFIG` itself composed from those same exports so its value/order is unchanged — this lets the list page pick an interleaved subset of shared columns without touching the shared array's contents (still used as-is by the run detail page and by two other entities' sub-tables).

**Why?**
The run list's column set didn't surface when a run was last touched (beyond its creation time) or which environment it ran in, both of which are useful at a glance when scanning a list of runs. Created was kept (rather than dropped) since it remains useful alongside the new Last Updated column, placed right after it.

**How did you test it?**
- `run.test.tsx`'s `describe('Run list page', ...)` block: one test asserting all 7 column headers render in the target order, and one test asserting Last Updated falls back to `creationTimestamp` when the update-timestamp label is absent, Created always renders from `creationTimestamp`, and Environment renders blank when its label is absent.
- Confirmed the existing `Run detail page` tests still pass unmodified after the `shared.ts` refactor.
- `yarn test`, `yarn typecheck`, `yarn lint`, and `prettier --check` all pass for the changed files (pre-existing failures elsewhere in the repo are unrelated — they come from generated protobuf/RPC client files that aren't checked in, reproducible on a clean checkout of `main` with no changes from this PR).

**Potential risks**
None beyond the column reorder itself — Created is retained, so no downstream consumer relying on its presence is affected.

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/javascript/packages/core/config/entities/run/__tests__/run.test.tsx
+++ b/javascript/packages/core/config/entities/run/__tests__/run.test.tsx
@@ -35,6 +35,7 @@ describe('Run list page', () => {
       'Pipeline run name',
       'Pipeline',
       'Last Updated',
+      'Created',
       'Environment',
       'Started by',
       'State',
@@ -59,7 +60,10 @@ describe('Run list page', () => {
                       'michelangelo/UpdateTimestamp': '1700000000000000',
                       'michelangelo/environment': 'production',
                     },
-                    creationTimestamp: { seconds: 1650000000 },
+                    // Distinct from run-without-labels' creationTimestamp below so this row's
+                    // Created value (which always reads creationTimestamp) doesn't collide with
+                    // the other row's Last Updated fallback (which also reads creationTimestamp).
+                    creationTimestamp: { seconds: 1660000000 },
                   },
                   spec: { actor: { name: 'jsmith' }, pipeline: { name: 'prediction-pipeline' } },
                   status: { state: 3 },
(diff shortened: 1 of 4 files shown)


sallycr:
Screenshot 2026-08-31 at 12 47 00 PM

craigmarker (approved):

PR #1995, opened 2026-08-31
feat(ui): add Environment field to run pipeline dialog

sallycr:
- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

- Added a required Environment field (Development / Production) to the "Start new pipeline
  run" dialog, positioned right after the read-only pipeline selector. Selecting an option
  writes it to the run's `michelangelo/environment` metadata label — the same label the
  pipeline run list already reads to render its Environment column, so this is the first
  write-side use of an existing read-side convention.
- Reordered the dialog's fields so Description now renders before the "Select run to resume
  from" section (previously reversed).
- Fixed a left-alignment mismatch between the "Select run to resume from" and "Set Up
  Notifications" section titles: `CollapsibleBox`'s `Header` style override used a shorthand
  `padding: 0`, which doesn't reliably cancel baseUI's own longhand `Header` padding when
  style overrides are merged. Replaced it with explicit longhand zero values
  (`paddingTop`/`paddingBottom`/`paddingLeft`/`paddingRight`), which fixes the indent for
  every `CollapsibleBox` consumer, not just this dialog.
- Added an optional `labels?: Record<string, string>` field to `PipelineRun.metadata` (mirrors
  the existing shape on `ModelRecord.metadata`) and exported `ENVIRONMENT_LABEL_KEY` from
  `environment-utils.ts` (previously module-private) so the new field and the existing
  `readEnvironmentLabel()` reader share one source of truth for the label key.

**Why?**

The dialog had no way to pick which environment a pipeline run targets, and its field order
didn't group related fields together. The alignment fix removes a visible inconsistency
between two section headers in the same dialog.

**How did you test it?**

- Added/updated tests in `create-pipeline-run-form.test.tsx`: field render order, both
  Environment options render, submission is blocked until an environment is selected, and the
  selected value is submitted as `metadata.labels['michelangelo/environment']`.
- Updated every existing submit-path test in that file to select an environment first, since
  the field is now required and react-final-form blocks submission on a sync validation
  error — all previously-passing tests (notifications, description, resume-from) still pass
  unmodified in behavior, only gaining the new required selection step.
- Ran the full suite: `yarn test`, `yarn lint`, `yarn typecheck` (project-level `tsc -b` has
  pre-existing, unrelated failures from missing generated proto stubs in this environment,
  confirmed identical before and after this change via `git stash` comparison; the
  Vitest-scoped typecheck used for this package is clean). `prettier --check` passes.
- No new test was added for the `collapsible-box.tsx` padding fix — `CollapsibleBox`'s own
  test file has no precedent for asserting on style/padding values anywhere in the component's
  11 existing tests, so this is left to manual/visual verification, consistent with existing
  practice.

**Potential risks**

Low. The Environment field is new and required, so any caller of this dialog must now select
an environment before a run can be submitted — this is an intentional behavior change (see
Breaking Changes) but is UI-scoped and does not affect any other entry point for creating a
pipeline run. The label write only adds a key to an already-generic `metadata.labels` map, so
it does not change the request's wire shape.

**Breaking Changes**

- [ ] No breaking changes

`PipelineRun.metadata` gains an optional `labels` field and `ENVIRONMENT_LABEL_KEY` becomes a
named export — both purely additive TypeScript surface changes in `packages/core`, not covered
by the categories above. The only user-facing behavior change is that the run dialog now
requires an Environment selection before submitting.

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/javascript/packages/core/components/box/collapsible-box.tsx
+++ b/javascript/packages/core/components/box/collapsible-box.tsx
@@ -71,7 +71,10 @@ export const CollapsibleBox: React.FC<CollapsibleBoxProps> = ({
           },
           Header: {
             style: {
-              padding: 0,
+              paddingTop: 0,
+              paddingBottom: 0,
+              paddingLeft: 0,
+              paddingRight: 0,
             },
           },
           Content: {
--- a/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
+++ b/javascript/packages/core/config/entities/pipeline/__tests__/create-pipeline-run-form.test.tsx
@@ -34,6 +34,15 @@ describe('CreatePipelineRunForm', () => {
     return <CreatePipelineRunForm record={data} onClose={() => setMounted(false)} />;
   }
 
+  /** Environment is required, so every submit-path test must pick one first. */
+  async function selectEnvironment(
+    user: ReturnType<typeof userEvent.setup>,
+    dialog: HTMLElement,
+    label: 'Development' | 'Production'
+  ) {
+    await user.click(within(dialog).getByRole('radio', { name: label }));
+  }
+
   it('submits pipeline run with correct data structure and closes dialog', async () => {
     const user = userEvent.setup();
     const mockResponse = { pipelineRun: { metadata: { name: 'created-run' } } };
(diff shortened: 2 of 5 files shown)


sallycr:
Screenshot 2026-08-31 at 3 27 23 PM

craigmarker (approved):

PR #2004, opened 2026-09-01
feat(model): inherit pipeline-type, owner, and pipeline-name/revision from source PipelineRun

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Added three new behaviors to the Model API hook, building on the refactored `resolveSourcePipelineRun` shape from #2003:

1. **Pipeline-type label default/override**: defaults `SourcePipelineTypeLabelName` to `PIPELINE_TYPE_TRAIN`; overrides from the source PipelineRun's own label when present.
2. **Owner/actor copy**: copies `Owner.Name`/`ProxyUser` from `PipelineRun.Spec.Actor` when set.
3. **Pipeline-name/revision label copy**: stamps `ModelSourcePipelineName` from `PipelineRun.Spec.Pipeline.Name`; resolves the `Revision` via a second `Get` and stamps `ModelSourcePipelineRevision` with `RevisionId` (fallback `GitRef`). Revision `Get` failure is non-fatal (log-and-skip).

New constants in `go/api/api.go`: `ModelSourcePipelineName`, `ModelSourcePipelineRevision`, `DefaultSourcePipelineType`.

**Why?**

These behaviors are needed for feature parity so that Models carry the same provenance metadata (pipeline type, owner identity, source pipeline/revision) that callers expect.

**How did you test it?**

- 13 new unit tests covering all four test-case templates per behavior (default-when-absent, inherit-from-source, preserve-explicit, source-lacks-label) plus revision-specific cases (RevisionId vs GitRef fallback, Get failure non-fatal, revision absent).
- All 9 existing `EnvironmentLabel` tests pass unchanged.
- `gofmt -l .` and `golangci-lint run ./...` clean.

**Potential risks**

Models will now have additional labels stamped on create/update. Existing Models without these labels are unaffected until their next update, at which point the hook will apply defaults.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

### DIFF
--- a/go/api/api.go
+++ b/go/api/api.go
@@ -74,6 +74,19 @@ const (
 	// SourcePipelineManifestTypeLabelName is the Kubernetes label key that identifies
 	// the pipeline manifest type (e.g. PIPELINE_MANIFEST_TYPE_ASL).
 	SourcePipelineManifestTypeLabelName = "pipeline.michelangelo/PipelineManifestType"
+
+	// ModelSourcePipelineName records the name of the Pipeline whose
+	// PipelineRun produced a Model.
+	ModelSourcePipelineName = "model.michelangelo/source-pipeline-name"
+
+	// ModelSourcePipelineRevision records the revision identifier (RevisionId
+	// or GitRef fallback) of the Revision used by the PipelineRun that
+	// produced a Model.
+	ModelSourcePipelineRevision = "model.michelangelo/source-pipeline-revision"
+
+	// DefaultSourcePipelineType is the default value for
+	// SourcePipelineTypeLabelName when no source PipelineRun provides one.
+	DefaultSourcePipelineType = "PIPELINE_TYPE_TRAIN"
 )
 
 // DefaultContextTimeout defines the default timeout for the context
--- a/go/components/model/apihook/BUILD.bazel
+++ b/go/components/model/apihook/BUILD.bazel
@@ -19,6 +19,7 @@ go_test(
     srcs = ["apihook_test.go"],
     embed = [":go_default_library"],
     deps = [
+        "//go/api:go_default_library",
         "//go/api/apimocks:go_default_library",
         "//proto/api:go_default_library",
         "//proto/api/v2:go_default_library",
--- a/go/components/model/apihook/apihook.go
+++ b/go/components/model/apihook/apihook.go
@@ -21,10 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// RegisterModelAPIHook registers the API hook that defaults and inherits the
-// environment label on Models at create/update time. defaultEnv is the
-// operator-configured default (empty when unconfigured, in which case
-// api.UnspecifiedEnvironment is used instead).
+// RegisterModelAPIHook registers the API hook that defaults and inherits
+// labels and owner metadata on Models at create/update time. defaultEnv is
+// the operator-configured default environment (empty when unconfigured, in
+// which case api.UnspecifiedEnvironment is used instead).
 func RegisterModelAPIHook(logger *zap.Logger, apiHandler api.Handler, defaultEnv string) {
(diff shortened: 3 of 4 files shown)


sallycr:
## Sandbox validation

Validated this branch (on top of #2003) in a local k3d sandbox against both `michelangelo-examples` pipelines (`xgb_train`, `pytorch_train`), across three real pipeline runs and a full sandbox recreate in between.

**Unit tests and the unconditional defaults work correctly**: `go test ./components/model/apihook/...` passes, and every resulting `Model` correctly gets `michelangelo/SourcePipelineType: PIPELINE_TYPE_TRAIN` and the environment label set.

**However, the headline feature — inheriting owner/pipeline-name/revision from the source `PipelineRun` — never activates with any pipeline in the current codebase.** `Model.Spec.SourcePipelineRun` has no producer anywhere today: `ModelPusherPlugin.execute()` → `registry.register_model(...)` has no `source_pipeline_run` parameter, and nothing else in `go/` or `python/` ever sets that field. Confirmed empirically — `Model` objects from all three runs only ever carry the unconditional defaults; `model.michelangelo/source-pipeline-name` and `model.michelangelo/source-pipeline-revision` are always absent, and owner is never copied from the actor.

This is the same gap already root-caused for the Studio UI's Model detail page (a consumer reading `spec.sourcePipelineRun` that no producer ever populates). As currently scoped, this PR adds inheritance logic that is correct but functionally inert until something sets `Model.Spec.SourcePipelineRun` on create/update.

Suggest either scoping this PR explicitly as inert prerequisite plumbing (verified by unit test only), or pairing it with a change that actually populates `SourcePipelineRun` from the calling pipeline context so the new behavior is observable end-to-end.

Happy to share more detail on the sandbox setup/repro if useful.

sallycr:
## Sandbox validation - PASS

Validated in a local k3d Michelangelo sandbox by building this branch's `apiserver` binary/image and rolling it out over `deployment/michelangelo-apiserver`, then running a real `michelangelo-examples` `california-housing` **pytorch_train** pipeline through it.

**Also validated together with #2034** (Python SDK `source_pipeline_run` registration), which produces the real `SourcePipelineRun` reference this hook consumes - the two changes were tested as one end-to-end provenance chain rather than in isolation.

**Run:** `california-housing/run-20260905-021805-6869daea` - pipeline reached `SUCCEEDED` (all 5 steps passed) against the `apiserver` build from this branch (head SHA `6c01a77b0`, includes #2003's base changes).

**Resulting model** (`california-housing/model-20260905-022144-059d3711`) confirms all three behaviors added by this PR fired correctly from the real `PipelineRun`:

```yaml
metadata:
  labels:
    michelangelo/SourcePipelineType: PIPELINE_TYPE_TRAIN
    model.michelangelo/source-pipeline-name: pytorch-train
spec:
  owner:
    name: sally.lee
```

- **Pipeline-type label**: defaulted/inherited to `PIPELINE_TYPE_TRAIN` (Behavior #1/#2).
- **Owner/actor copy**: `spec.owner.name` copied from the triggering `PipelineRun.Spec.Actor` (`sally.lee`, the account that ran `ma pipeline run`) rather than left unset (Behavior #3).
- **Pipeline-name label**: `model.michelangelo/source-pipeline-name` stamped from `PipelineRun.Spec.Pipeline.Name` (Behavior #4, name half).

No `model.michelangelo/source-pipeline-revision` label on this run - expected, since a manually-triggered `ma pipeline run` doesn't set `PipelineRun.Spec.Revision`, so the revision-copy path correctly no-ops rather than producing a bogus label. Didn't get a chance to exercise the revision-copy path itself (would need a run with `Spec.Revision` set, e.g. via the cron trigger path) - flagging as untested rather than failing, since the unit tests already cover it directly.

No regressions observed in the rest of the pipeline or in existing `EnvironmentLabel` behavior.

_Ad hoc sandbox validation, not a formal CI run._

Issue #1867, opened 2026-08-19
pytorch_train example crashes on Apple Silicon: DDP strategy not supported on MPS accelerator


sallycr:
**Describe the bug**

`michelangelo-examples`' `pytorch_train` California Housing pipeline crashes when run via the documented local-run command on an Apple Silicon Mac. PyTorch Lightning's `accelerator="auto"` default detects the physical MPS device, but `michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py`'s `_train_loop_per_worker()` never overrides `accelerator`, while `_resolve_strategy()` always defaults to `RayDDPStrategy` when the caller doesn't override it. Lightning explicitly rejects the combination of an MPS accelerator with any DDP-family strategy.

Root-caused while re-validating #1753/#1754's `dot_path` fix end-to-end on a real Mac — that fix works correctly (no `AssertionError`, confirmed), which is what allowed execution to reach this next, separate crash.

**To Reproduce**

1. `pip install "michelangelo-examples[california-housing]"`
2. `python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline`
3. See error (below) once execution reaches the actual training step (after Spark preprocessing succeeds)

**Expected behavior**

The tutorial's documented "fastest way to iterate" local-run command should complete without a manual code change, on both Apple Silicon and CUDA-equipped dev machines — matching `xgb_train` (the other California Housing pipeline), which already runs clean locally.

**Actual behavior**

```
File ".../michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py", line 1272, in _train_loop_per_worker
    trainer = pl.Trainer(**trainer_kwargs)
...
ValueError: You set `strategy=<ray.train.lightning._lightning_utils.RayDDPStrategy object at ...>` but strategies from the DDP family are not supported on the MPS accelerator. Either explicitly set `accelerator='cpu'` or change the strategy.
```

[...]

PR #2005, opened 2026-09-01
fix(trainer): default to SingleDeviceStrategy when world size is 1

KyleKoon:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Added `RaySingleDeviceStrategy` and updated `_resolve_strategy()` in the Lightning trainer utilities to auto-select it when `world_size == 1`, instead of unconditionally defaulting to `RayDDPStrategy`.

- New `_resolve_single_device()` helper picks the best available device (TPU > MPS > CUDA > CPU), mirroring [Lightning's own auto-detection order](https://github.com/Lightning-AI/pytorch-lightning/blob/411eec98d50368d700c45edd29d9c20b21e7be17/src/lightning/fabric/utilities/device_parser.py#L209-L221).
- New `RaySingleDeviceStrategy(SingleDeviceStrategy)` class passes the resolved device to Lightning so the strategy and accelerator agree.
- `_resolve_strategy(None)` now checks `ray.train.get_context().get_world_size()`: returns `RaySingleDeviceStrategy` when 1, `RayDDPStrategy` otherwise.
- `_resolve_strategy("single_device")` is supported as an explicit option.
- `_prepare_trainer_for_ray()` bypasses Ray's strategy whitelist check for `RaySingleDeviceStrategy` (same pattern as the existing FSDP2/ModelParallel bypass).

**Why?**

On Apple Silicon Macs, PyTorch Lightning auto-detects the MPS accelerator, but `RayDDPStrategy` (a DDP-family strategy) is incompatible with MPS. This causes a `ValueError: strategies from the DDP family are not supported on the MPS accelerator` when running single-worker training locally (e.g. `python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline`). Since single-worker training doesn't need DDP, using `SingleDeviceStrategy` is the correct choice and also enables MPS acceleration.

**How did you test it?**

- Reproduced the original `ValueError` on Apple Silicon by running `python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline` before the fix.
- Verified the fix runs the full `pytorch_train` pipeline end-to-end locally on Apple Silicon (`feature_prep` → `preprocess` → `train` → `assembler` → `push_step`), with Lightning correctly using MPS (`GPU available: True (mps), used: True`).
- Verified the fix does not affect the sandbox (k3d/Linux) path: `xgb_train` and `pytorch_train` both pass on a k3d sandbox where `world_size > 1` and DDP is used as before.

**Potential risks**

- `_resolve_single_device()` imports `XLAAccelerator` from `lightning.fabric` — wrapped in a `try`/`except ImportError` to handle environments without XLA installed.
- Multi-worker training is unaffected: `_resolve_strategy(None)` only selects `RaySingleDeviceStrategy` when `world_size == 1`; multi-worker runs continue to use `RayDDPStrategy`.
- Ray's `prepare_trainer` is bypassed for `RaySingleDeviceStrategy` (it hard-rejects unknown strategies). The only check it performs that we replicate is validating `RayLightningEnvironment` as the cluster environment plugin.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Fixed `ValueError` when running Lightning-based training locally on Apple Silicon (MPS). Single-worker training now auto-selects `SingleDeviceStrategy` instead of `RayDDPStrategy`, enabling MPS GPU acceleration on Macs.

**Documentation Changes**

N/A

### DIFF
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -326,18 +326,12 @@ def _prepare_trainer_for_ray(trainer: pl.Trainer) -> pl.Trainer:
 
 
 def _resolve_single_device() -> torch.device:
-    """Pick the best available single device: TPU > MPS > CUDA > CPU.
+    """Pick the best available single device: MPS > CUDA > CPU.
 
-    Mirrors Lightning's own auto-detection order:
+    Mirrors Lightning's own auto-detection order (minus TPU, which the
+    trainer SDK does not currently support):
     https://github.com/Lightning-AI/pytorch-lightning/blob/411eec98d50368d700c45edd29d9c20b21e7be17/src/lightning/fabric/utilities/device_parser.py#L209-L221
     """
-    try:
-        from pytorch_lightning.accelerators import XLAAccelerator
-
-        if XLAAccelerator.is_available():
-            return torch.device("xla")
-    except ImportError:
-        pass
     if torch.backends.mps.is_available():
         return torch.device("mps")
     if torch.cuda.is_available():
@@ -348,9 +342,9 @@ def _resolve_single_device() -> torch.device:
 class RaySingleDeviceStrategy(SingleDeviceStrategy):
     """Ray glue for single-device training.
 
-    Resolves the best available device (TPU, then MPS on Apple Silicon, then
-    CUDA if Ray assigned one, then CPU) so that Lightning's auto-detected
-    accelerator and the strategy agree on the device.
+    Resolves the best available device (MPS on Apple Silicon, then CUDA if
+    Ray assigned one, then CPU) so that Lightning's auto-detected accelerator
+    and the strategy agree on the device.
     """
 
     def __init__(self, **kwargs: Any):
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -332,7 +332,7 @@ def _resolve_single_device() -> torch.device:
     https://github.com/Lightning-AI/pytorch-lightning/blob/411eec98d50368d700c45edd29d9c20b21e7be17/src/lightning/fabric/utilities/device_parser.py#L209-L221
     """
     try:
-        from lightning.fabric.accelerators.xla import XLAAccelerator
+        from pytorch_lightning.accelerators import XLAAccelerator
 
         if XLAAccelerator.is_available():
             return torch.device("xla")
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/_private/util.py
@@ -27,7 +27,7 @@ from pytorch_lightning.plugins import (
     LayerSync,
     Precision,
 )
-from pytorch_lightning.strategies import Strategy
+from pytorch_lightning.strategies import SingleDeviceStrategy, Strategy
 from ray.train.lightning import (
     RayDDPStrategy,
     RayDeepSpeedStrategy,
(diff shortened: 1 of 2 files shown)


sallycr:
Thanks for taking this on directly — validated locally and did a design review against our own strategy-only fix attempt for the same issue (#1867). Summary below; happy to discuss any of it.

## Local end-to-end validation (real Apple Silicon hardware)

Ran the exact documented local-run command against this branch:
`PYTHONPATH=. python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline`

Full pipeline completed cleanly: `feature_prep` → `preprocess` (Spark) → `train` → `assembler` → `push_step`, no errors.

- `GPU available: True (mps), used: True` — `RaySingleDeviceStrategy` correctly selected for `world_size=1`.
- 5 training epochs completed with steadily decreasing loss (`val_loss` 1.45 → 0.83).
- Model registered successfully (`model-<ts>-<hex>` v1).

This confirms the fix works exactly as claimed in the PR description.

## Design review

**Real bug — TPU/XLA branch is likely non-functional.** `_resolve_single_device()` returns a bare `torch.device("xla")` for the TPU case and hands it to plain `RaySingleDeviceStrategy` (a `SingleDeviceStrategy` subclass). But Lightning's own `_choose_strategy()` special-cases XLA to build `SingleDeviceXLAStrategy` instead — a different subclass that calls `xm.xla_device()` to get a real XLA device handle and wires up `XLACheckpointIO`/`XLAPrecision`. `_check_and_init_precision()` only assigns `XLAPrecision` when the strategy is `SingleDeviceXLAStrategy`/`XLAStrategy`, so on real TPU hardware this would likely produce a strategy with the wrong device-allocation path and the wrong precision plugin. Worth either dropping the TPU branch (michelangelo has no TPU support today) or routing it through a `RaySingleDeviceXLAStrategy(SingleDeviceXLAStrategy)`.

**A real strength worth calling out**: resolving CUDA devices via `ray.train.torch.get_device()` (rather than deferring to Lightning's own `"auto"`/`_choose_strategy()` CUDA branch) is more correct when `CUDA_VISIBLE_DEVICES` is shared across co-located workers — Ray's `get_device()` maps `ray.get_gpu_ids()` against the shared env var to resolve the actual assigned index, where Lightning's own auto-detection would see all node GPUs and could incorrectly fall through to a DDP-multi-device pick even for a genuinely single-worker trial. Nice catch, and a real point in favor of hand-constructing the strategy rather than universally deferring to `"auto"`.

**Scope note**: this changes the default for *every* single-worker run on *every* platform (not just MPS) — e.g. today's single-GPU CUDA fine-tuning jobs would now also get `RaySingleDeviceStrategy` instead of `RayDDPStrategy`. Given the CUDA correctness point above this seems like a deliberate and reasonable generalization, but the PR's own "How did you test it" section only covers MPS-local and k3d/Linux-DDP (`world_size > 1`) — a single-worker CUDA run isn't mentioned. Worth confirming that path got some verification (even manual) before merge, since it's the most common production topology this PR's default now touches.

**Duplicated device-priority logic**: `_resolve_single_device()`'s TPU > MPS > CUDA > CPU order reimplements `lightning.fabric.utilities.device_parser._select_auto_accelerator()`. Reasonable given the CUDA-correctness tradeoff above, but it's copy of upstream logic that can silently drift if Lightning changes its own priority/adds a new accelerator kind. A version-pin comment near wherever `pytorch-lightning` is pinned (pointing back at this function) would help future upgraders remember to re-diff it.

**Tests**: the new unit tests are solid for `_resolve_single_device()`'s own branch logic, but they mock `ray.train.get_context()`, `torch.backends.mps/cuda.is_available()`, etc. end-to-end, so they assert the function *calls* `RaySingleDeviceStrategy`/returns the right device — not that a real `pl.Trainer(strategy=RaySingleDeviceStrategy())` actually initializes correctly through `_AcceleratorConnector`. That's exactly the boundary where the TPU/XLA issue above lives, and it wouldn't be caught by any test in this diff. A real (non-mocked) CPU-device `Trainer` smoke test — one real training step — would backstop the manual MPS verification without needing MPS/CUDA/TPU hardware in CI.

None of the above blocks the core MPS fix, which I independently verified works end-to-end. The TPU/XLA gap seems worth a fix or an explicit "out of scope, michelangelo has no TPU support" note before merge.

sallycr (approved):
Looks good to me. thx for adding MSP

PR #2006, opened 2026-09-01
feat(raycluster): terminate cluster missing from remote and mark all terminal states immutable

dragod812:
## Summary
Intent:
- The RayCluster controller monitors a launched cluster by fetching its status
  from the remote compute cluster. When the backing RayCluster has been deleted
  from the remote, GetJobClusterStatus returns NotFound and the controller
  requeued every 10s indefinitely, spinning on the log line
  "cluster not found on remote cluster, requeue" (controller.go:199).
- A cluster that has vanished from the remote is a terminal, non-recoverable
  condition: there is nothing left to monitor or clean up. The controller
  should record it and stop, not loop forever.
- Separately, RayCluster never marked itself immutable on reaching *any*
  terminal state (unlike RayJob/SparkJob, which do). A cluster that finished
  normally (killed, succeeded, failed) converged to Killed=TRUE/TERMINATED via
  the existing cleanup flow but stayed a live, mutable object in K8s/etcd
  indefinitely instead of being archived by the ingester.

Changes:
- On NotFound from the remote status fetch, mark the RayCluster as a terminal
  failure via the status subresource in one write (State=FAILED; Succeeded=FALSE
  with a reason/message; Killing=FALSE, Killed=TRUE). Setting the full terminal
  condition set directly (rather than just Succeeded=FALSE) means cleanupCluster
  will never issue a pointless DeleteJobCluster against a remote resource that's
  already confirmed gone.
- Added `isClusterFullyTerminal` + `markImmutableIfTerminal`: once a RayCluster's
  conditions show Succeeded decided and cleanup converged (Killing=FALSE,
  Killed=TRUE), it gets the `michelangelo/Immutable` annotation via a metadata
  Update. The ingester then moves the object to metadata storage and removes it
  from ETCD, so reconciliation stops permanently. This is checked after every
  `processClusterTermination` call, so it now covers **all** terminal paths
  (user/spec-driven kill, natural success/failure, creation failure) — not just
  NotFound-on-remote. It's a no-op while cleanup is still in flight (e.g.
  Killing=TRUE but Killed not yet TRUE); a later reconcile picks it back up.
- `markClusterFailedImmutable` (NotFound-on-remote) now delegates to
  `markImmutableIfTerminal` instead of duplicating the freeze logic, so there's
  a single place that decides when it's safe to freeze an object.
- Added unit test coverage: immutability assertions on all cleanup-convergence
  cases and the NotFound case, explicit non-immutability assertions on
  transient/in-flight states (guarding against freezing before cleanup actually
  runs), and a new multi-reconcile test proving the normal kill path converges
  to immutable over successive reconciles.

---

<sub>Generated by the 🪄 [pr-create](https://sg.uberinternal.com/code.uber.internal/uber-code/devexp-agent-marketplace/-/blob/claude-code/plugins/dev/uber-dev/skills/pr-create/SKILL.md) skill in devexp-agent-marketplace</sub>

## Test Plan

**Unit:** `bazel test //go/components/ray/cluster:go_default_test` passes — covers immutability across all terminal paths and non-immutability on transient/in-flight states.

**Live E2E** (OSS `k3d-michelangelo-sandbox`, `controllermgr` run locally, compute == control plane):

1. **Not-found-on-remote:** deleted a running cluster's backing KubeRay `ray.io` CR → controller logged `cluster not found on remote compute cluster, marking as failed`, set `State=FAILED / Succeeded=FALSE (ClusterNotFoundOnRemote) / Killing=FALSE / Killed=TRUE` + `michelangelo/Immutable`, then stopped requeuing.
2. **Success termination:** terminated a cluster with `TERMINATION_TYPE_SUCCEEDED` → `State=TERMINATED / Succeeded=TRUE / Killed=TRUE` + immutable; `cleanupCluster` deleted the KubeRay CR.
3. **Uniflow pipeline (Cadence `remote-run`):** ran a multi-task Ray pipeline end-to-end — workflow COMPLETED and all 9 task RayClusters reached `Succeeded=TRUE / Killed=TRUE` + `michelangelo/Immutable=true`.

Both terminal outcomes exercised — `Succeeded=TRUE` (success/kill) and `Succeeded=FALSE` (not-found) — 11 clusters total, each frozen immutable on reaching a terminal state.

### DIFF
--- a/go/components/ray/cluster/controller.go
+++ b/go/components/ray/cluster/controller.go
@@ -115,6 +115,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{RequeueAfter: requeueAfter}, err
 		}
 		logger.Info("processed cluster termination")
+
+		// Once termination has fully converged (kill/cleanup finished, or there was
+		// nothing to clean up), freeze the object so the ingester archives it and
+		// reconciliation stops for good. No-op if cleanup is still in progress; a later
+		// reconcile will pick this back up once it does.
+		if err := r.markImmutableIfTerminal(ctx, &rayCluster); err != nil {
+			logger.Error(err, "failed to mark cluster immutable after termination")
+			return ctrl.Result{RequeueAfter: requeueAfter}, err
+		}
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


go/components/ray/cluster/controller.go
```diff
@@ -556,6 +568,75 @@ func (r *Reconciler) getClusterStatus(ctx context.Context, log logr.Logger, assi
...
+	if err := retry.OnError(retry.DefaultRetry, jobsutils.IsRetriableError, func() error {
+		latest := &v2pb.RayCluster{}
+		if err := r.Get(ctx, cluster.Namespace, cluster.Name, &metav1.GetOptions{}, latest); err != nil {
+			return err
+		}
+		if utils.IsImmutable(latest) {
+			return nil
+		}
+		utils.MarkImmutable(latest)
+		return r.Update(ctx, latest, &metav1.UpdateOptions{})
```
badcount: This update will update spec + status, meaning your current reconcile will not able to update it right? because the revision is outdated?

dragod812: The main reason we use "Update" instead of "UpdateStatus" is to mark the CR immutable. The immutable label exists in metadata. 
Once its marked immutable no one should be able to update it.
I have made some changes to mark it immutable on terminal states for Raycluster.

go/components/ray/cluster/controller.go
```diff
@@ -556,6 +577,104 @@ func (r *Reconciler) getClusterStatus(ctx context.Context, log logr.Logger, assi
...
+	if err := retry.OnError(retry.DefaultRetry, jobsutils.IsRetriableError, func() error {
+		latest := &v2pb.RayCluster{}
+		if err := r.Get(ctx, cluster.Namespace, cluster.Name, &metav1.GetOptions{}, latest); err != nil {
+			return err
+		}
+		if utils.IsImmutable(latest) || !isClusterFullyTerminal(latest) {
+			return nil
+		}
+		utils.MarkImmutable(latest)
+		return r.Update(ctx, latest, &metav1.UpdateOptions{})
```
dragod812, line 624: @badcount 
The main reason we use "Update" instead of "UpdateStatus" is to mark the CR immutable. The immutable label exists in metadata. 
Once its marked immutable no one should be able to update it. I have made some changes to mark it immutable on terminal states for RayCluster.

badcount: Will we call updateStatus after the markImmutableIfTerminal is called in the controller?

badcount (approved):

PR #2007, opened 2026-09-01
docs(api-reference): add trainer modules to the Python SDK reference

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Adds the two trainers to the Python SDK reference that #1956 established, plus the two supporting modules they reference:

| Module | Covers |
|---|---|
| `lib.trainer.torch.pytorch_lightning.lightning_trainer` | `LightningTrainer`, `LightningTrainerWithStateDict`, `LightningTrainerParam` |
| `lib.trainer.torch.pytorch_lightning.schema` | `TransferLearningSpec`, `IncrementalTrainingSpec`, `ModelSpec`, `TrainingObserver`, `ExperimentStore` |
| `lib.trainer.torch.pytorch_lightning.experiment_store` | `FsspecExperimentStore` |
| `lib.trainer.xgboost.xgboost_trainer` | `XGBoostTrainer`, `XGBoostTrainerParam` |

The index's **Trainer** section links all four, and the note saying the XGBoost trainer and `LightningTrainer` "aren't in the current generator config and have no page yet" is removed, since that is no longer true.

**Why?**

`XGBoostTrainer` has had no documentation anywhere in `docs/` since it merged in #1793, and `LightningTrainer` — the SDK's primary training entry point — had no reference page either, even though other pages refer to it. The generated index was itself flagging both as missing.

**How did you test it?**

- Every rendered bullet name was checked back against the source docstrings; 52/52 trace to real `Attributes:`/`Args:` entries. No prose was invented for these pages.
- Ran the `docs-check.yml` relative-link rule (`grep -rEn '\]\(\.\.?/' docs/ --include='*.md'` minus the skip patterns) — clean, both for the new files and repo-wide.
- Verified every relative link added to `index.md` resolves to a file on disk.

**Potential risks**

Documentation only — no code paths, no build or deployment changes.

---

**Two things worth a reviewer's attention**

**1. The committed reference pages are not reproducible from `bun run docs:api`.** Running the documented command on current `main` (pydoc-markdown 4.8.2, the latest release) does not reproduce the 10 pages already committed. The committed pages have Sphinx roles converted to Markdown (`` :func:`x` `` → `` `x` ``), a `michelangelo.` prefix on `title:`, `_category_.json` sidebar files, and at least one `**Example**:` code block that does not exist in the source docstring. Stock generation produces none of those. So the existing pages were generated once and then polished by hand.

That matters because the index tells contributors to regenerate after editing docstrings. Doing so today would strip the hand-polish, revert roles to raw `:func:` text that renders literally on the docs site, and drop the sidebar files.

**2. The Google-style section parser corrupts these docstrings.** On the four modules here, stock generation mangled 31 lines across all four pages into indexed placeholders — e.g. `` - ```label_column`2 - "reg:squarederror", "eta": 0.1}``). `` and `` - ``0 - Extra keyword arguments... ``. The trigger is an attribute description containing a colon inside a literal, such as `params`' example value `{"objective": "reg:squarederror", "eta": 0.1}`, which the parser reads as the start of a new field.

I rebuilt the affected `Attributes:`/`Args:`/`Returns:` blocks directly from the source docstrings rather than hand-patching the corrupted output, which is why the bullet-name verification above was worth running.

Net effect: this PR follows the existing convention (generate, then correct), but that convention is undocumented and silently lossy. Worth a follow-up issue to either pin a working toolchain and make generation truly reproducible, or state plainly that these pages are hand-maintained and stop advertising a regeneration command that damages them. Happy to file that separately.

**Also unblocked by this PR:** re-running generation on current `main` yields 21 pages against the 10 committed, because #1955's docstring fixes made 11 previously-`skip_empty_modules`-skipped modules renderable and nobody re-ran the generator afterward. I deliberately left those 11 out of this PR — they have never been reviewed and would need the same correction pass. Flagging so it is a decision rather than an oversight.

**Breaking Changes**

- [x] No breaking changes

**Documentation Changes**

- `pydoc-markdown.yml` — 4 trainer modules added to the loader
- `docs/api-reference/python-sdk/index.md` — Trainer section links the new pages; stale "no page yet" note removed; note added on DDP/FSDP/FSDP2 strategy selection
- `docs/api-reference/python-sdk/reference/lib/trainer/{xgboost,torch/pytorch_lightning}/` — 4 new pages + 2 `_category_.json`

### DIFF
--- a/docs/api-reference/python-sdk/index.md
+++ b/docs/api-reference/python-sdk/index.md
@@ -13,10 +13,7 @@ Start here if you're authoring pipelines with the Python SDK.
 > for what a task, a workflow, and Uniflow actually are before diving into the raw
 > signatures below.
 
-Some modules below aren't linked yet — coming soon. Two modules referenced elsewhere
-in this reference, the XGBoost trainer and `LightningTrainer` itself, aren't in the
-current generator config and have no page yet either; only the Torch collate-function
-utilities under **Trainer** are covered so far.
+Some modules below aren't linked yet — coming soon.
 
 ## Uniflow — Tasks & Workflows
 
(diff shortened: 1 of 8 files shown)


zhoward-1:
Closing in favor of a hand-written guide — see below for the reasoning and where the review findings went.

A doc review of this PR found four defects in the generated pages. All four originate in the source docstrings rather than the generator, so they can't be fixed at the docs layer: generated pages *are* the docstrings. They're now tracked in #2009 with suggested fixes and file:line references.

- `schema.md` — `fused_model_submodule` on both warm-start specs names a non-public counterpart system and points readers at "the PR that ports it"
- `schema.md` — `TrainingObserver`'s `Example::` block loses its code fence *and all indentation*, leaving uncopyable Python
- `lightning_trainer.md` — `profiler_sink` documents `_private.util.…`, which is both wrong (it's `_private.profiler`) and non-public
- `lightning_trainer.md` — a stray `#### num_epochs` heading whose body is `type: ignore[assignment]  # sentinel replaced in __post_init__`

The two modules carrying all four defects (`schema`, `lightning_trainer`) are also the two most useful ones here, so scoping them out would leave the PR without its main value.

Instead, `LightningTrainer` is covered by a hand-written guide at `docs/user-guides/train-and-deploy-models/distributed-training.md`, which documents the same surface — every `LightningTrainerParam` field, DDP/FSDP/FSDP2/DeepSpeed selection, warm starts, auto-resume, and getting trained weights back out — without inheriting the docstring problems.

One thing worth carrying forward from this PR regardless: `pydoc-markdown` isn't pinned in `pyproject.toml`, so generation isn't reproducible — version 4.8.2 produces materially different output from what's on this branch, from identical source. Noted at the end of #2009.

PR #2008, opened 2026-09-01
fix(python): correct and de-internalize public trainer docstrings

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Corrects four problems in public docstrings under
`python/michelangelo/lib/trainer/torch/pytorch_lightning/`. These docstrings are
the generation source for the Python SDK Reference, so their contents ship to the
public docs site.

- **`fused_model_submodule` (both warm-start specs)** — described an internal
  counterpart system by name and told readers to "see the PR that ports it", which
  an external reader cannot find. Reworded to state what the field does today
  (nothing — it is reserved). The default-value divergence is preserved as a
  maintainer code comment on the field itself, so the information isn't lost.
- **`TrainingObserver` example** — used a bare ``Example::`` literal block, which
  the generator fails to convert: the rendered page lost its code block and *all*
  indentation, leaving Python that cannot be copied. Switched to the
  parenthesized form already used twice in this same file, which converts
  correctly.
- **`profiler_sink`** — pointed at `_private.util.comet_profiler_sink`. The sinks
  actually live in `_private.profiler`, and `_private` is not a public import
  path. Now references the public re-export.
- **`num_epochs`** — carried a trailing `# sentinel replaced in __post_init__`
  comment that the generator emitted as a stray documented attribute exposing an
  internal sentinel. Annotating `_UNSET` as `Any` removes the need for the
  `type: ignore`, and the explanation moves to the sentinel's own definition.

**Why?**

#2007 adds these modules to the generated Python SDK Reference. Reviewing that PR
surfaced these issues in the generated output, but every one of them originates in
the source docstrings rather than the generator — so they are fixed here, at the
source, rather than hand-patched in generated markdown that the next regeneration
would overwrite.

Two of the four are correctness issues rather than style: the `profiler_sink`
module path is simply wrong, and the internal-system reference is a trademark and
public-docs concern.

**How did you test it?**

- `ruff check` and `ruff format --diff` pass on both changed files.
- Confirmed `schema.py` still imports on Python 3.11 and that every dataclass
  field and default is unchanged (`TransferLearningSpec` and
  `IncrementalTrainingSpec` constructed and asserted field-by-field).
- Verified the example fix empirically with a controlled A/B: generated the page
  from the pre-change and post-change docstring using the same pydoc-markdown
  version. Before, every line of the example rendered at zero indentation
  (a paragraph, with `{metrics}` exposed as body text); after, the block renders
  at 4/8-space indentation as a proper code block.
- Confirmed the stray `num_epochs` attribute no longer appears in generated output.

No unit tests added — the change is comments, docstrings, and one annotation on a
module-private sentinel.

**Potential risks**

None. `git diff` contains no logic changes: the only non-comment edits are adding
`Any` to a `typing` import and annotating the module-private `_UNSET` sentinel,
neither of which affects runtime behavior.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

This PR *is* the documentation change — it corrects the docstrings that generate
the Python SDK Reference. No `docs/` files are touched here; #2007 should be
regenerated on top of this so the corrections land in the published reference.

### DIFF
--- a/python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
+++ b/python/michelangelo/lib/trainer/torch/pytorch_lightning/lightning_trainer.py
@@ -34,7 +34,7 @@ import os
 import uuid
 from contextlib import contextmanager
 from dataclasses import asdict, dataclass, field
-from typing import TYPE_CHECKING, Callable
+from typing import TYPE_CHECKING, Any, Callable
 
 import ray
 import torch
@@ -69,7 +69,10 @@ CHECKPOINT_PATH_KEY = "checkpoint_path"
 # public ``Result.from_path`` is unimplemented and ``can_restore`` /
 # ``resume_from_checkpoint`` are deprecated.
 _CHECKPOINT_MANAGER_SNAPSHOT_FILENAME = "checkpoint_manager_snapshot.json"
-_UNSET = object()
+# Sentinel distinguishing "num_epochs never set" from an explicit None; replaced
+# with the real default in ``LightningTrainerParam.__post_init__``. Typed as Any
+# so it can stand in for the annotated field types it defaults.
+_UNSET: Any = object()
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


zhoward-1:
Closing deliberately — not abandoning the findings.

The decision is to avoid source changes here for now. All four docstring fixes from this PR are captured in #2009 so they aren't lost, including the verification evidence (a controlled A/B showing the `Example::` block renders with zero indentation today vs. a proper code block after the fix).

Worth calling out one item that outlives this decision: the `profiler_sink` docstring at `lightning_trainer.py:129-130` points at `_private.util.comet_profiler_sink`, but the sinks actually live in `_private/profiler.py` — and `_private` isn't a public path regardless. That's wrong in the source, not just in generated docs, so anyone reading it in their IDE gets a path that doesn't resolve. Tracked as item 3 in #2009.

This branch stays pushed if anyone wants to revive it; CI was fully green (8 passed, 0 failed).

PR #2013, opened 2026-09-01
feat(inferenceserver): POC -- containerBuildTemplate + CI-built per-project image

krishpatel9:
**TL;DR:** POC for an alternative to #2012's `pythonDependencies` — a project supplies a plain `requirements.txt` (still no Dockerfile), and CI builds a per-project Triton image from the existing shared Dockerfile. For comparison, not both meant to land.

## Summary

- **`go/components/inferenceserver/backends/triton.go`** — adds `tritonImage()`, wiring up `ServingSpec.container_build_template` (already in the proto, previously unused anywhere in Go — confirmed nothing on `main` read it). Falls back to the stock image when unset.
- **`docker/triton-serving.Dockerfile`** — parameterized with an `ARG EXTRA_PIP_PACKAGES`, defaulting to its existing fixed `torch==2.4.1 transformers==4.44.2` so the shared-default build (`build-triton-image.yaml`) is unaffected.
- **`.github/workflows/build-example-triton-images.yaml`** (new) — discovers every `python/examples/*/inferenceserver/requirements.txt`, builds a per-project image from the same shared Dockerfile with that project's deps as the build-arg, and pushes it as `ghcr.io/<owner>/<project>-triton-serving`. Mirrors `build-image.yaml`'s existing `PROJECT_EXTRA` pattern for training-time images.
- **`examples/bert_cola/inferenceserver/`** — wired up as a live example: `requirements.txt` (the dependency list) + `inferenceserver.yaml` (`containerBuildTemplate` pointing at the resulting image) + `deployment.yaml`.

## Comparison with #2012 (`pythonDependencies` + init container)

| | This PR (`containerBuildTemplate`) | #2012 (`pythonDependencies`) |
|---|---|---|
| Customer writes a Dockerfile? | No — plain `requirements.txt` | No — plain YAML list |
| Pod cold start | Fast (deps baked into image) | Slower (pip install every pod start) |
| Needs live PyPI access at pod-start? | No | Yes |
| Non-Python/system-level deps (apt, CUDA) | Yes (same Dockerfile mechanism) | No (Python-only) |
| Time to pick up a dependency change | A CI build + registry push, then redeploy | Instant — edit YAML, re-apply |
| New infra required | None (reuses training's existing build/registry/CI pattern) | None (pure Kubernetes primitives) |

## Test plan
- [x] `go build`/`go test ./components/inferenceserver/...` — all green, including new tests for `tritonImage()`'s default and override behavior.
- [x] `gofmt`/`go vet` — clean.
- [x] Verified the new workflow's discovery + `requirements.txt`-to-build-arg logic locally.
- [x] All new/changed YAML parses correctly.

### DIFF
--- a/.github/workflows/build-example-triton-images.yaml
+++ b/.github/workflows/build-example-triton-images.yaml
@@ -1,0 +1,96 @@
+name: Build example Triton serving images
+
+# Each example under python/examples/<name>/inferenceserver/ that ships its own
+# requirements.txt gets its own Triton serving image, built from the same
+# shared docker/triton-serving.Dockerfile as the default image (just with a
+# different EXTRA_PIP_PACKAGES build-arg) -- no per-project Dockerfile, no
+# per-project workflow file. Mirrors how build-image.yaml builds a per-project
+# task image from a pyproject.toml extras group via a PROJECT_EXTRA build-arg.
+on:
+  push:
+    branches: [main]
+    paths:
+      - "python/examples/*/inferenceserver/requirements.txt"
+      - "docker/triton-serving.Dockerfile"
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  discover:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.find.outputs.matrix }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Find example Triton dependency manifests
+        id: find
+        run: |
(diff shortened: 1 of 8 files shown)


krishpatel9:
Superseded by #2012, which now resolves \`python_dependencies\` to a shared, content-addressed image built by CI -- the same cold-start/offline properties this PR's \`containerBuildTemplate\` approach was going for, but without the per-project \`requirements.txt\` + CI-matrix coupling this PR introduced. Closing this in favor of that.

PR #2015, opened 2026-09-02
feat(python): support file-backed task kwargs

SuhasAnumukonda:
## Summary
Intent:
- Avoid process argument-size failures when task kwargs are stored externally.
- Keep the runner contract storage-agnostic through fsspec.

Changes:
- Accept exactly one of --kwargs and --kwargs-file while preserving inline callers.
- Read file-backed kwargs as UTF-8 and decode them with the existing Uniflow codec.
- Cover local and memory filesystems, CLI validation, safe errors, codec parity, and oversized subprocess input.

## Test Plan
- Targeted runner suite: 14 passed.
- Uniflow core suite excluding Docker-dependent file_sync tests: 75 passed.
- Ruff formatting and lint checks passed for every changed Python file.
- Built michelangelo-0.9.0-py3-none-any.whl successfully.
- The full core attempt reached 106 passed and 6 environment-only failures because the borrowed local virtualenv lacks the optional docker package; all failures were in test_file_sync.py.

---

<sub>Generated by the 🪄 [pr-create](https://sg.uberinternal.com/code.uber.internal/uber-code/devexp-agent-marketplace/-/blob/claude-code/plugins/dev/uber-dev/skills/pr-create/SKILL.md) skill in devexp-agent-marketplace</sub>

### DIFF
--- a/python/michelangelo/uniflow/core/run_task.py
+++ b/python/michelangelo/uniflow/core/run_task.py
@@ -5,6 +5,8 @@ import argparse
 import logging
 import sys
 
+import fsspec
+
 from michelangelo.uniflow.core.codec import decoder
 from michelangelo.uniflow.core.utils import LOGGING_FORMAT, import_attribute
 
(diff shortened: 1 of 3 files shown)


kenns29 (approved):

python/michelangelo/uniflow/core/run_task.py
```diff
@@ -19,16 +21,20 @@ def main():
     p = argparse.ArgumentParser()
     p.add_argument("--task", required=True, type=str)
     p.add_argument("--args", required=True, type=_decode_arg)
-    p.add_argument("--kwargs", required=True, type=_decode_arg)
+    kwargs_group = p.add_mutually_exclusive_group(required=True)
+    kwargs_group.add_argument("--kwargs", type=_decode_arg)
```
kenns29: try to run `_decode_arg` and `_decode_kwargs_file` outside of `add_argument` because add_argument suppress the stacktrace when there is error

SuhasAnumukonda: Moved all Uniflow CLI decoding outside argparse type callbacks in dc53c3554d9ee38efcb4ba0331bb5849e61e98b0. parse_args now validates only CLI structure; _decode_arg and _decode_kwargs_file run afterward, so their original chained exceptions retain the full traceback. Added regression coverage for inline and file-backed failures; Ruff and all 16 focused tests pass.

[addressed by agent]

PR #1964, opened 2026-08-28
feat(ui): reorder phase tabs (Triggers before Runs) and rename Pipeline Runs to Runs

sallycr:
## Summary
- Reorders each phase's tab bar so **Triggers** comes before **Runs**:
  - Train & Evaluate: Pipelines, Triggers, Runs, Trained Models, Evaluations, Notebooks
  - Experiment & Productionize: Pipelines, Triggers, Runs
  - Deploy & Predict: Pipelines, Triggers, Runs, Targets, Deployments
- Renames the **"Pipeline Runs"** tab/label to **"Runs"** — a single change in `RUN_ENTITY_CONFIG.name` picked up by every phase that includes it, plus the Pipeline detail page's own "runs" sub-tab for consistency.
- `data.ts` (Prepare & Analyze Data) and `monitor-debug.ts` (Monitor & Debug) are untouched — the former has no Triggers tab to reorder against, the latter has no tabs yet.

## Test plan
- [x] `yarn test` — 165 files / 1552 tests passed
- [x] `yarn typecheck` — clean
- [x] `yarn lint --quiet` — clean
- [x] `yarn format --write` — clean
- [x] OSS compliance scan — PASS
- [x] Sandbox: rebuilt the UI image (`--no-cache`), redeployed to k3d, confirmed via the deployed bundle that `"Pipeline Runs"` no longer appears anywhere and `"Runs"` is present.
- [ ] No headless browser available in this environment to screenshot the rendered tab bar — verification rests on the `entities` array order (which drives tab render order directly) plus the passing test suite.

PR #2016, opened 2026-09-02
docs(ui): update entity config examples for the Pipeline Runs → Runs rename

zhoward-1:
## Summary

PR #1964 renamed `RUN_ENTITY_CONFIG.name` from `'Pipeline Runs'` to `'Runs'`. Two UI configuration docs quote that config verbatim — one of them literally labelled `// From: config/entities/run/run.ts` — and both still show the old value, so the copy-pasteable examples no longer match [the source](https://github.com/michelangelo-ai/michelangelo/blob/main/javascript/packages/core/config/entities/run/run.ts#L82-L88).

**What changed (4 lines, 2 files):**
- `entity-configuration-reference.md` — two `RUN_ENTITY_CONFIG` code blocks, plus the `name` property's examples list, which would otherwise contradict the two code blocks around it
- `configuration-api.md` — one `RUN_ENTITY_CONFIG` code block

**Deliberately not touched:** the `### Pipeline Runs` heading in `docs/operator-guides/operations/monitoring.md`. That section documents the `pipelinerun_*` metrics, which describe the PipelineRun resource rather than the renamed UI tab, and renaming it would break the heading anchor for no gain.

## Context

Found by the `ma-oss-docquality` audit harness. `entity-configuration-reference.md` was also 160 days stale; `configuration-api.md` was inside the freshness window and wrong anyway, which is a decent illustration that mtime and accuracy are separate things.

## Test plan
- [x] `grep -rn "Pipeline Runs" docs/contributing/` returns nothing
- [x] Remaining values match `run.ts` L82–88 exactly
- [ ] docs-check CI (no link changes — text-only edits inside code fences and one bullet)

### DIFF
--- a/docs/contributing/dev/ui/configuration/configuration-api.md
+++ b/docs/contributing/dev/ui/configuration/configuration-api.md
@@ -46,7 +46,7 @@ See [`javascript/packages/core/config/phases`](https://github.com/michelangelo-a
 // Example: config/entities/run/run.ts
 export const RUN_ENTITY_CONFIG: PhaseEntityConfig = {
   id: 'runs',
-  name: 'Pipeline Runs',
+  name: 'Runs',
   service: 'pipelineRun',
   state: 'active',
   views: [RUN_LIST_CONFIG, RUN_DETAIL_CONFIG],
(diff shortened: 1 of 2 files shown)


sallycr (approved):
thx

PR #2026, opened 2026-09-03
docs: revert getting-started.md tutorial to pytorch_train (MPS fix merged)

sallycr:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Reverts `docs/user-guides/getting-started/getting-started.md`'s tutorial back to `michelangelo-examples`' `pytorch_train` (PyTorch Lightning), undoing #1871's temporary revert to the in-core XGBoost example. This is a clean `git revert` of #1871's merge commit — only that one file changes.

**Why?**

#1871 reverted the tutorial to `xgb_train` because `pytorch_train`'s documented local-run command crashed on Apple Silicon Macs with a DDP/MPS incompatibility (issue #1867). That's now fixed and merged in #2005 (`fix(trainer): default to SingleDeviceStrategy when world size is 1`), so the original rationale for #1871 no longer applies and the tutorial can safely go back to teaching `pytorch_train`.

**How did you test it?**

- Confirmed #2005 is merged to `main` (`8d789c836`) before opening this revert.
- Re-ran the restored tutorial's exact documented local-run command against a fresh `main` checkout on real Apple Silicon hardware: `feature_prep → preprocess → train → assembler → push_step` completed cleanly, `GPU available: True (mps), used: True`, model registered successfully.
- Separately simulated a genuine first-time user: fresh venv, `pip install "michelangelo-examples[california-housing]"` from PyPI (not an editable/local install), then `python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline` with no `PYTHONPATH` workaround, exactly as the doc instructs. Also completed cleanly end-to-end with MPS training and a registered model — confirms `michelangelo-examples`' `michelangelo>=0.8.0` floor pin doesn't clobber a locally-installed dev copy of `michelangelo` carrying the fix, as long as the reader follows the doc's own step order (core repo's `poetry install -E example` before the `pip install michelangelo-examples[...]` step).

**Potential risks**

None beyond a doc/tutorial content change. No code changes.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Yes — restores the getting-started tutorial's flagship example to `pytorch_train`, now that #2005 fixes the Apple Silicon crash that motivated #1871's temporary revert.

### DIFF
--- a/docs/user-guides/getting-started/getting-started.md
+++ b/docs/user-guides/getting-started/getting-started.md
@@ -9,7 +9,7 @@ By the end of this guide you'll have a working pipeline that looks like this:
 ```
 dataset_cols
     └─▶ feature_prep()    # Ray task: downloads data, splits into train/validation
-            └─▶ train()   # Ray task: trains XGBoost model on the splits
+            └─▶ train()   # Ray task: trains a PyTorch Lightning model on the splits
 ```
 
 Each step runs as an isolated, containerized task. Michelangelo AI handles data passing between them, caches intermediate results, and retries on transient failures — your code stays plain Python.
@@ -28,7 +28,6 @@ Each step runs as an isolated, containerized task. Michelangelo AI handles data 
 * Java 17 with `JAVA_HOME` set — required for the Spark preprocessing step. Java 21 is not compatible with PySpark 3.5 + Hadoop 3.3 (`getSubject is not supported` error). On macOS: `brew install openjdk@17` then `export JAVA_HOME=$(brew --prefix openjdk@17)/libexec/openjdk.jdk/Contents/Home`
 * For remote runs: Docker and access to a Kubernetes cluster (or use the [local sandbox](../../getting-started/sandbox-setup.md))
 * [Create a project](./project-management-for-ml-pipelines.md)
-* **macOS only**: XGBoost requires OpenMP (`libomp.dylib`), which is not installed by default. Run `brew install libomp` before installing dependencies.
 
 ## Environment setup
 
@@ -141,12 +140,9 @@ def train_workflow(dataset_cols: str):
         train_dv=train_dv,
         validation_dv=validation_dv,
         params={
-            "objective": "reg:squarederror",
-            "colsample_bytree": 0.3,
-            "learning_rate": 0.1,
-            "max_depth": 5,
-            "alpha": 10,
-            "n_estimators": 10,
+            "max_epochs": 5,
+            "batch_size": 64,
+            "precision": "32",
         },
     )
     return result
@@ -184,12 +180,33 @@ The context provides three methods:
 
 Then run it:
 
+The `feature_prep`/`train` workflow above is exactly what the `pytorch_train` pipeline in [`michelangelo-examples`](https://github.com/michelangelo-ai/michelangelo-examples) implements — this tutorial's example now lives there, in its own pip-installable package, rather than in this repo. Install and run it:
+
 ```bash
-cd michelangelo/python
-PYTHONPATH=. poetry run python examples/pipelines/california_housing_xgb/california_housing_xgb.py
+pip install "michelangelo-examples[california-housing]"
+python -m michelangelo_examples.california_housing.pipelines.pytorch_train.pipeline
 ```
 
-Local runs execute everything in your Python interpreter with zero infrastructure setup. This is the fastest way to iterate on your workflow logic.
+Local runs execute everything in your Python interpreter with zero additional infrastructure setup (beyond the one-time `pip install` above) — the fastest way to iterate on your workflow logic.
+
+> **Why a separate repo?** Heavier example pipelines with their own dependency
+> footprint (here, Spark + PyTorch Lightning) live in
+> [`michelangelo-examples`](https://github.com/michelangelo-ai/michelangelo-examples)
+> to keep this core repo's own dependencies lean. Lighter in-repo examples
+> like `bert_cola` (`python/examples/bert_cola/`) are unrelated to this
+> tutorial's pipeline but still run directly from this checkout:
+> ```bash
+> cd michelangelo/python
+> PYTHONPATH=. poetry run python examples/bert_cola/bert_cola.py
+> ```
+>
+> `pytorch_train` also ships a separate, lighter-weight local-only trainer
+> (`python -m michelangelo_examples.california_housing.pipelines.pytorch_train`,
+> no `.pipeline` suffix) that trains directly with plain PyTorch
+> Lightning — no Ray, Spark, or Uniflow context at all. That script is a
+> standalone quick-start, not the `ctx.run(train_workflow)` pattern this
+> tutorial teaches; use the `.pipeline`-suffixed command above to run the
+> actual workflow you just wrote.
 
 ## Step 4: Run remotely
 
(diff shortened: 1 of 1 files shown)


badcount (approved):

paulzim:
# PR #2026 — second-pass review — docs: revert getting-started.md tutorial to pytorch_train (MPS fix merged)

**Author:** sallycr · **Branch:** `sallycr/revert-1871-restore-pytorch-train-tutorial` → `main` · **1 commit, +61/-35, 1 file** · Reverts #1871. Existing review: one Approve (no comment body) from `badcount`.

## What it does

Reverts `docs/user-guides/getting-started/getting-started.md` back to the `pytorch_train` (PyTorch Lightning) tutorial from `michelangelo-examples`, undoing #1871's temporary revert to the in-core XGBoost example. #1871 was motivated by #1867 (`pytorch_train` crashed on Apple Silicon with DDP-on-MPS); #2005 fixed that, so the original rationale is gone.

## The mechanical revert and its premise: both verified clean

- **It is a byte-exact revert.** `git diff <#1871-merge>^1 pr-2026 -- docs/user-guides/getting-started/getting-started.md` returns **empty output** — the restored file is identical to the pre-#1871 state, character for character.
- **#1871 touched exactly one file**, so "only that one file changes" is accurate — this is a complete revert, not a partial one that could leave inconsistency behind.
- **#2005 is genuinely merged and genuinely fixes the stated cause.** Merge commit `8d789c836`, merged 2026-09-03. Read the commit: it adds `RaySingleDeviceStrategy` and makes `_resolve_strategy()` select it when `world_size == 1` instead of unconditionally using `RayDDPStrategy`, with device preference TPU > **MPS** > CUDA > CPU. That is precisely the DDP-on-MPS crash in #1867.
- **No other doc depends on the xgb tutorial.** Grepped all of `docs/` for `california_housing_xgb`: the only other hit is a historical CHANGELOG entry. Reverting creates no cross-doc inconsistency.
- **Everything the restored text points at externally is real.** `michelangelo-examples` is public; `src/michelangelo_examples/california_housing/pipelines/pytorch_train/` exists with `pipeline.py`, `pipeline.yaml`, `README.md`, and `__main__.py` (which confirms the doc's "no `.pipeline` suffix" standalone-trainer note); `config/project.yaml` exists and really does register the `california-housing` project/namespace; the repo root really does have a `Dockerfile`; the PyPI package `michelangelo-examples` really does expose a `california-housing` extra.
- **The pipeline.yaml in the doc is byte-identical to the real one** in the examples repo — namespace, name, annotation, and `filePath` all match exactly.
- **`ghcr.io/michelangelo-ai/michelangelo-examples:california-housing` exists and is publicly pullable** (anonymous ghcr manifest query → HTTP 200).
- **Every CLI flag the doc teaches is valid.** `run.py` declares `"args": ["-n", "--namespace"]` and `["--name"]` for `ma pipeline run`; `context.py:250` declares `--environ`/`--env` with `nargs="*"`, so the four repeated `--environ KEY=VALUE` flags work.
- **All 8 relative links resolve** against current `main`, and the `#workflow-constraints` anchor matches a real heading.
- **The version math in the PR's test narrative holds.** The extra pins `michelangelo>=0.8.0`; the core repo is at `0.9.0`, so pip is already satisfied and won't clobber the local dev install. The author's tested step order is sound.
- **Dropping the macOS `libomp` prerequisite is correct** — the tutorial no longer uses XGBoost. (`poetry install -E example` still pulls `xgboost`, but nothing imports it, and libomp is an import-time not install-time requirement.)
- **CI is fully green** (`filter`, `license/cla`, `update-release-draft`, `validate-docs`).

So: the revert decision is well-founded and mechanically exactly right. Everything below is about the **content being restored**, which a byte-exact revert by definition does not re-examine — and which two weeks of `main` drift and a first-run read now put in a different light.

**One calibration note on the green CI:** I read `.github/workflows/docs-check.yml`. `validate-docs` runs a website typecheck, a changelog sync, and a grep for relative links missing a `.md` extension. It does **not** verify that referenced file paths exist, that external links resolve, or that documented commands work. Green CI here does not cover findings 1–3 below.

## Findings

### 1. The tutorial's Step 5 tells you to run commands from a repo it never tells you to clone (first-run blocker)

Step 3 installs the example with `pip install "michelangelo-examples[california-housing]"`. Then:

- Step 4: `# From the michelangelo-examples repo checkout:` → `docker build ... .`
- Step 5: `# From the michelangelo-examples repo:` → `ma project apply -f src/michelangelo_examples/california_housing/config/project.yaml`
- Step 5: `ma pipeline apply -f src/michelangelo_examples/california_housing/pipelines/pytorch_train/pipeline.yaml`

A reader who followed the tutorial literally has a **pip install**, not a checkout. There is no `src/michelangelo_examples/...` directory to point `-f` at. (I checked the packaging: hatchling with `packages = ["src/michelangelo_examples"]` does ship the YAML inside the wheel, but at `<site-packages>/michelangelo_examples/...` — never at `src/...`.) So both `ma ... apply -f` commands fail for anyone who didn't independently guess they needed to `git clone`.

Worth noting this is a **regression against current `main`**, not just a pre-existing wart: on `main` the whole tutorial runs out of the single core repo the reader cloned in "Environment setup," so Steps 4 and 5 are self-consistent. The revert splits the tutorial across two repos and only instructs the reader to obtain one of them.

Cheapest fix: add a clone step before Step 4.
```bash
git clone https://github.com/michelangelo-ai/michelangelo-examples.git
cd michelangelo-examples
```
Alternatively point `-f` at raw GitHub URLs so the pip-only path stays viable.

### 2. "exactly what the `pytorch_train` pipeline implements" is not accurate

Line 183 (new in this PR, no equivalent on `main`):

> The `feature_prep`/`train` workflow above is exactly what the `pytorch_train` pipeline in `michelangelo-examples` implements

I read the real `pipeline.py`. Its `train_workflow` is **five** tasks, not two:

```
feature_prep (Ray, .with_overrides) → preprocess (Spark, .with_overrides) → train → assembler → push_step
```

And the training call doesn't match either. The tutorial teaches:

```python
result = train(train_dv=train_dv, validation_dv=validation_dv,
               params={"max_epochs": 5, "batch_size": 64, "precision": "32"})
```

The real task is `def train(pr: PreprocessResult, feature_columns: list[str]) -> ModelVariable` — **no `params` argument at all**. Those three values are real, but they live hardcoded inside nested `TabularTrainerConfig` / `LightningTrainerConfig` / `LightningTrainerKwargs` / `BatchIterConfig`, not in a flat dict a caller passes.

So a reader runs "the workflow you just wrote" and watches five steps go by, including a Spark stage and a model push, with a training task whose signature they've never seen. The doc contradicts itself here too — the "Complete example" section further down correctly advertises Spark preprocessing and task overrides.

Simplified teaching code is completely fine; the word **"exactly"** is what breaks. Suggest: *"The `feature_prep`/`train` workflow above is a simplified version of the `pytorch_train` pipeline in `michelangelo-examples`, which adds Spark preprocessing, model assembly, and a registry push."*

### 3. "Then run it:" now dangles, with a competing second lead-in

Rendered, Step 3 currently reads:

> Then run it:
>
> The `feature_prep`/`train` workflow above is exactly what … **Install and run it:**
>
> ```bash …```

"Then run it:" ends in a colon and is followed by a paragraph, not a command; two sentences later a second lead-in ("Install and run it:") introduces the actual block. On `main` this is clean — "Then run it:" sits directly above the code block. The inserted paragraph orphaned it. Fix is one deleted line.

### 4. The removed "Sandbox tip" leaves Step 4 and Step 5 silently contradicting each other

The revert deletes this from `main` without replacement:

> **Sandbox tip:** the `michelangelo/uniflow-image` annotation controls which Docker image runs your tasks. For a k3d sandbox, build the image from Step 4 … and import it with `k3d image import`. The `ghcr.io/...` image is published by CI for production deployments.

After the revert, Step 4 has the reader build `michelangelo-examples:local` and `k3d image import` it — and then Step 5's `pipeline.yaml` points at `ghcr.io/michelangelo-ai/michelangelo-examples:california-housing`, quietly ignoring the image they just built, with nothing explaining why. The ghcr image is real and public (verified above), so nothing is *broken*; the reader is just left wondering what the last five minutes were for. The deleted tip answered exactly this question and is worth restoring in updated form.

### 5. Minor: which interpreter is `pip install` landing in?

"Environment setup" creates a Poetry venv and says "activate it directly or run commands via `poetry run`." Step 3 then uses a bare `pip install` and a bare `python -m ...` — neither activated nor `poetry run`-prefixed. The author's tested path is correct and the version floor math works out, but the doc never tells the reader how to be *on* that path; a literal reader can easily install into system Python. Suggest `poetry run pip install ...` and `poetry run python -m ...`, or an explicit activation line.

### 6. Minor: "rather than in this repo" is now imprecise

`python/examples/pipelines/california_housing_xgb/` **still exists** on current `main`. Precisely: the *PyTorch* variant lives in `michelangelo-examples` while a California Housing *XGBoost* variant remains in-repo. The blanket "this tutorial's example now lives there … rather than in this repo" reads as though California Housing left the repo entirely.

### 7. Minor: #1867 is still open

The issue this whole revert is predicated on resolving is **still `OPEN`** — #2005 merged the code fix without closing it. Since this PR is the user-facing completion of that fix, it's a natural place to add `Closes #1867`.

### 8. Nit: flag style

`ma pipeline run -n california-housing --name pytorch-train` mixes short and long forms. Both are valid (verified in `run.py`); `main` used `--namespace ... --name ...`. Cosmetic only.

### Not a finding (checked and cleared)

- The long `>` blockquotes aren't a style violation — there are **zero** `:::` admonitions anywhere in `docs/user-guides/getting-started/`, so `>` is this directory's own convention. The "Why a separate repo?" aside is still heavy (~18 rendered lines, three topics, one of which the text itself calls "unrelated to this tutorial's pipeline") at the exact moment a reader wants to run their first pipeline — worth trimming, but that's taste, not correctness. Its `bert_cola` path and `__main__` entry point are both still valid on current `main`.

## Verdict: **Approve the revert; recommend a short follow-up commit**

The decision to revert is correct, the premise is verified true, and the mechanics are flawless — byte-exact, single-file, no cross-doc fallout. Nothing here argues for holding the revert itself.

But findings **1 and 2** are things a new user hits inside the first five minutes: Step 5 cannot be executed as written by a reader who followed Step 3, and the doc promises the pipeline is "exactly" the workflow they just wrote when it's roughly twice the size with a different training signature. Since a clean revert has real value, the author's call between two options:

1. Push a small follow-up commit on this PR fixing 1–4 (a clone step, one softened sentence, one deleted line, one restored tip), or
2. Merge the clean revert as-is and open an immediate follow-up — acceptable, but 1 and 2 shouldn't sit on `main` long in the project's flagship getting-started page.

Either way, worth adding `Closes #1867`.

Issue #1608, opened 2026-07-27
release-promote.yml can never succeed: pushes directly to a protected release branch (bypass_actors: [])


sallycr:
## Summary

`release-promote.yml` has a 0/5 historical success rate (checked via `gh run list --workflow release-promote.yml`, spanning `release/v0.4` and `release/v0.6`). It is not flaky — it is structurally incapable of succeeding as written.

## Root cause

The "Strip RC suffix and set final version" step does:
```yaml
run: |
  bash scripts/version-bump.sh "${{ steps.vars.outputs.version }}"
  git add -A
  git commit -m "chore: release ${{ steps.vars.outputs.version }}"
  git push origin "${{ steps.vars.outputs.branch }}"
```
directly against the release branch (e.g. `release/v0.6`), using `secrets.GITHUB_TOKEN`.

That branch is covered by a ruleset requiring pull requests for all changes (`ruleset_id` visible via `gh api repos/michelangelo-ai/michelangelo/rules/branches/release%2Fv0.6`), and that ruleset's `bypass_actors` is empty:
```
gh api repos/michelangelo-ai/michelangelo/rulesets/19126671 | jq '.bypass_actors, .current_user_can_bypass'
[]
"never"
```
No actor — not `github-actions[bot]`, not a PAT, not even a repo admin — can push directly to a protected branch here. The initial *creation* of a release branch (by `release-cut.yml`) succeeds because branch creation isn't covered by the "require pull request" rule type, but every subsequent push to that branch, by anyone, is rejected with `GH013`.

## Fix needed

`release-promote.yml` needs to open a PR for the version-strip commit (mirroring what `release-cut.yml`'s cherry-pick/correction flows already have to do) and either:
- pause for a human to review+merge before continuing (breaks full automation, but matches the ruleset's `required_approving_review_count: 1`), or
- if full automation is wanted, get an org admin to add a bypass actor (e.g. a dedicated release-bot App) to the ruleset — a policy decision, not something to do unilaterally in a workflow fix.

## Workaround used for v0.6.0

Promoted manually: branched off `release/v0.6`, ran `scripts/version-bump.sh`, opened PR #1607, human-reviewed/merged it, then tagged `v0.6.0` and manually triggered the same artifact-publish workflows `release-promote.yml` would have. Same pattern used previously for `v0.4.0` (see harness `needs_human_review.md`).

Issue #2031, opened 2026-09-04
Release v0.10.0


sallycr:
## Pre-Release

- [x] All CI checks pass on the release branch
- [x] Version numbers updated across all components:
  - [x] `python/pyproject.toml`
  - [x] `javascript/packages/core/package.json`
  - [x] `javascript/packages/rpc/package.json`
  - [x] `website/package.json`
  - [x] `helm/michelangelo/Chart.yaml` (both `version` and `appVersion`)
- [x] `CHANGELOG.md` updated via `git cliff`
- [x] Release notes drafted (follows three-layer template: summary, categorized changes, compatibility matrix)
- [x] Breaking changes reviewed and documented -- N/A, no `BREAKING CHANGE:` commits in this release
- [x] Compatibility matrix updated in release notes -- N/A, no breaking changes this release

## Release Candidate

- [x] Release branch created: `release/v0.10`
- [x] RC tag pushed (`v0.10.0-rc.1`)
- [x] RC artifacts published and accessible:
  - [x] Python wheel on PyPI (PEP 440: `0.10.0rc1`)
  - [x] Go service containers on ghcr.io
  - [x] UI container on ghcr.io
  - [x] npm packages (`@michelangelo-ai/core`, `@michelangelo-ai/rpc`)
  - [x] Helm OCI chart on ghcr.io
- [x] RC announcement posted (GitHub Discussions or relevant channel)
- [x] Soak period complete -- promoted after 4/7 days at explicit maintainer direction, no P0/P1 found
- [x] No P0/P1 issues reported against RC

## Final Release

- [x] Release tag pushed (`v0.10.0`)
- [x] All artifacts published and accessible:
  - [x] Python wheel on PyPI (`michelangelo==0.10.0`)
  - [x] Go service containers on ghcr.io (tagged `v0.10.0` + `latest`)
  - [x] UI container on ghcr.io
  - [x] npm packages (latest dist-tag)
  - [x] Helm OCI chart on ghcr.io
- [x] GitHub Release created with generated release notes
- [x] Announcement posted
- [x] Release branch merged back to main -- PR #2032, merged
- [ ] Next development version bumped on main -- N/A, no separate next-dev-version bump practice on `main`

---

**v0.10.0 shipped 2026-09-08.** See the closing comment below for the full artifact-verification summary and links.

sallycr:
## Release test — v0.10.0-rc.1: **PASS**

[...]

PR #2032, opened 2026-09-04
Merge back release/v0.10 to main

sallycr:
Merge-back for the v0.10.0 release. Tracking issue: https://github.com/michelangelo-ai/michelangelo/issues/2031. Marked ready for review once the release is promoted to final.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,90 +8,6 @@ All notable changes to this project will be documented in this file.
 ### Bug Fixes
 
 
-- Reword ma-design-interview skill to reduce duration (#1714)
-
-
-- **build:** Bump Go SDK to 1.26.3 + rules_go to 0.57.0 (CVE Scan fix, step 1/2) (#1567)
-
-
-- **python:** Restore ColumnConfig.shape optional default (#1762)
-
-
-- **ci:** Skip JavaScript coverage PR comment on fork pull requests (#1756)
-
-
-- **python:** Wire up workflowClient.executionUrlFormat for the sandbox (#1761)
-
-
-- **core:** Correct Model list columns for description, environment, type, last updated (#1673)
-
-
-- **storage/mysql:** Validate criterion field name before SQL interpolation (#1707)
-
-
-- **storage/mysql:** Implement directUpdate for the generic Update RPC (#1645)
-
-
-- **ui:** Block entity navigation for disabled/coming-soon phases in menu drawer (#1791)
-
-
-- **release:** Mark -rc tags as prereleases in GitHub releases (#1678)
-
-
-- **api:** Log blob-storage delete failure only when it fails (#1679)
-
-
-- **helm:** Drop placeholder signKey annotation from chart metadata (#1768)
-
-
-- **api:** Use status.Error for non-format error messages (#1782)
-
-
-- **core:** Exempt barrel files and test infra from react-refresh export rule (#1807)
-
-
-- **go:** Sync paused TriggerRun schedules (#1827)
-
-
-- **python:** Stop shipping test files in wheels and sdists (#1832)
-
-
-- Create TriggerRun schedules paused (#1828)
-
-
-- Make TriggerRun deletion idempotent (#1688)
-
-
-- **ui:** Tighten breadcrumb menu drawer close button and icons (#1799)
-
-
-- **core:** Stop capitalizing entity/phase names (#1846)
-
-
-- **core:** Title-case Data Sources, Evaluations, and Notebooks entity names (#1857)
-
-
-- **python:** Use __spec__.name for dot_path() under python -m (#1754)
-
-
-- **go:** Pipelinerun revision resolve and pipeline latest revision (#1797)
-
-
-- **python:** Render <crd> get <name> as table and fix -o yaml OrderedDict crash (#1822)
-
-
-- **python:** Rename evaluation_strategy to eval_strategy in bert_cola example (#1883)
-
-
-- **docs,examples:** Correct Lightning trainer imports and parameters (#1787)
-
-
-- Plumb RayTask gpu, disk and object store memory into the cluster spec (#1801)
-
-
-- **release:** Align website version and add it to version-bump.sh (#1849)
-
-
 - **ci:** Pin release.yaml's Go setup to go.mod instead of a hardcoded version (#1897)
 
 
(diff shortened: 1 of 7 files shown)


paulzim (approved):
# PR #2032 — Merge back release/v0.10 to main

**Author:** sallycr (+ github-actions[bot] co-authors) · **`release/v0.10` → `main`** · **5 commits, +264/−10, 7 files** (true PR diff, merge-base to head — see note on methodology below) · Tracking issue: #2031 (closed). Existing reviews: none.

## Verdict: **Approve**

Textbook merge-back: version bump, changelog, one release-tooling bugfix. Every number in it cross-checks against the repo's own tooling and git history, and the one claim I couldn't take at face value — that `release-promote.yml` can't push directly to a protected branch — checks out against the actual ruleset config and the workflow's own failure history.

## What it does

Five squash-merged commits, applied in order: bump to `0.10.0-rc.1`, regenerate `CHANGELOG.md` for the RC, fix a bad changelog range for the RC (#2045), promote to final `0.10.0` (#2048), regenerate `CHANGELOG.md` for the final release (#2049). Net diff from the merge-base: `CHANGELOG.md` gains a `[0.10.0]` section, and six version-carrying files move `0.9.0` → `0.10.0`.

## A methodology note before the findings

`git diff upstream-main..pr-2032` is the wrong comparison for this PR and initially made the change look 10x larger than it is — it includes 11 commits that landed on `main` independently after `release/v0.10` branched off (including PR #1751's California-housing deletion, which shows up as spuriously "re-added" files). The actual PR diff is `git diff $(git merge-base upstream-main pr-2032)..pr-2032`, which reproduces GitHub's own reported `+264/−10, 7 files` exactly. Everything below uses that.

## Verified

**Version bump is complete and exactly matches the project's own authoritative file list.** `scripts/version-bump.sh` hardcodes six `VERSION_FILES`: `python/pyproject.toml`, `javascript/packages/{core,rpc}/package.json`, `javascript/app/package.json`, `website/package.json`, `helm/michelangelo/Chart.yaml` (plus its `appVersion` field). All six are bumped `0.9.0` → `0.10.0` in this diff, one line each, nothing missing, nothing extra bundled in. Running the script's own `--check` against the PR head confirms: `All versions aligned: 0.10.0`. A repo-wide grep for leftover `0.9.0` version strings in any `.json`/`.toml`/`.yaml` returns nothing.

**`python/pyproject.toml` changes by exactly one line** — the version bump only. No stray dependency payload rode in on this merge (relevant since `main` and `release/v0.10` had diverged for a while and #1751's dependency cleanup landed on `main` in the interim — that content correctly does *not* appear in this diff since the release branch never had it to begin with).

**The `CHANGELOG.md` range fix (#2045) actually worked — verified independently, not by trusting the commit message.** I extracted every PR number referenced in the new `[0.10.0]` section and compared it against `git log --oneline --no-merges v0.9.0..v0.10.0`: both give **74 unique PR numbers**, and they match. Zero overlap with the existing `[0.9.0]` section's 97 PR numbers. That directly confirms the fix produced a correct, non-duplicated range — this is the exact failure class (content from an old release silently re-appearing in a new one) that bit a past release PR via an overwritten `ExperimentStore` entry, and it isn't present here.

**The root-cause issue for the underlying bug is real and well-filed.** Commit `f90d43f1`'s message says it only fixes the symptom and files a separate issue for `changelog.yml`'s fallback logic. That issue is **#2046** (open): a clear repro (`release/v0.9` auto-deleted after its own merge-back, so the branch-based fallback skips to `release/v0.8`, pulling in 164 commits instead of 72) and two concrete suggested fixes (resolve via tag instead of branch, or derive `PREV_BRANCH` from tag history like the existing `CANDIDATE` search does). Correctly scoped out of this PR.

**Checked a claim I couldn't take on faith: "release-promote.yml can't push directly to the protected release branch."** Confirmed three ways:
- GitHub's ruleset API shows both `release/*` and `main` require PRs with ≥1 approval, `bypass_actors: null` (no exemptions).
- Every one of the 5 historical `release-promote.yml` runs (3× `release/v0.4`, 2× `release/v0.6`) **failed**, all at the identical step: `git push origin release/vX.Y` → `remote: error: GH013 ... Changes must be made through a pull request.`
- There is **no recorded run at all** for `v0.7` through `v0.10` — the team has apparently stopped even attempting the automated dispatch and gone straight to the manual flow this commit describes, for at least 5 cycles running.

This is pre-existing, already tracked in open issue **#1608** ("release-promote.yml can never succeed"), filed around the `v0.6` timeframe. Not this PR's problem to fix — it doesn't touch `release-promote.yml` — but it does mean `CONTRIBUTING.md`'s "What each workflow does automatically" section for `release-promote.yml` describes a flow that, per the actual run history, has never once completed unattended. See non-blocking note below.

**Soak-period and process claims check out against `CONTRIBUTING.md`.** The doc confirms: draft merge-back PR opens at RC cut and stays open through soak; promotion is dispatched "once the RC has soaked and no P0/P1 issues were found"; soak minimum is "1 week for minor releases." The commit's "day 4 of 7, cut short at user's explicit request" is consistent with the timestamps (RC cut `2026-09-04T15:23Z`, promotion commit `2026-09-08T15:56Z` ≈ 4 days) and is explicit that it's a deliberate override, not a silent skip. (The literal string "REL-005" doesn't appear anywhere in the repo — it reads like informal internal shorthand for this now-routine manual flow. Harmless, but see note below.)

**CI is green.** All checks pass except `Build (linux/amd64)`, still in flight at review time — its `linux/arm64` counterpart in the same matrix already passed, so this is an ordinary in-progress multi-arch build, not a signal of anything.

## Non-blocking notes

**1. `release-promote.yml`'s automation is now fiction, and it's been that way for 5+ cycles.** Not a defect in this PR, but worth surfacing since this PR's own commit message is the artifact that documents the workaround: the "automatic" path in `CONTRIBUTING.md` has never completed end-to-end since `v0.4`, per the run history above. Two options going forward, whenever someone has a cycle to spend on it: actually fix #1608 (a `bypass_actors` entry for the promote workflow's identity, or restructure it to open a PR like `changelog.yml` already does), or — cheaper — update `CONTRIBUTING.md` to describe the manual PR flow as the real primary path instead of "what the workflow does automatically," and give it a name that's actually written down somewhere (right now "REL-005" only exists in this commit message).

**2. Nothing else.** The version bump is exhaustively correct, the changelog is provably non-duplicated, and the one substantive engineering claim in the whole PR (the range-fix root cause) is filed properly. This is as clean as a merge-back gets.

## Summary

Every checkable claim in this PR checks out against the repo's own tooling, ruleset config, and workflow run history — including the one claim that turned out to reveal a real, if already-tracked, gap elsewhere (`release-promote.yml`). Nothing here should hold up the merge.

PR #2034, opened 2026-09-04
Populate Model.Spec.SourcePipelineRun from pipeline-run identity env vars

sallycr:
## Summary

The Go-side model API hook already reads `Model.Spec.SourcePipelineRun`  (a`ResourceIdentifier{namespace, name}` proto field, added in #2003/#2004) to derive provenance labels on a model -- pipeline type, owner, pipeline name/revision. But nothing in the Python SDK ever sets that field: the pipeline worker injects the run's identity into every task pod's environment (`MA_NAMESPACE`, `MA_PIPELINE_RUN_NAME`), yet no registration path reads it back. Every model registered through the SDK today carries an unset `SourcePipelineRun`, making #2004's label-inheritance behavior effectively dead code for SDK-registered models.

## Changes

- New `lib/shared/pipeline_run/` -- a `None`-safe accessor (`get_source_pipeline_run()`) that reads the two environment variables and returns a `SourcePipelineRun(name, namespace)` reference, or `None` when running outside a pipeline (local dev, tests). Cached process-globally since these values are fixed for the pod's lifetime. Lives in the cross-cutting `lib/shared/` package (alongside the existing `json_data`/`model_fuser`/`model_metadata` utilities) rather than under `model_manager`, since reading pipeline-run identity isn't specific to model registration.
- `ModelRegistryClient.register_model()` gains a new optional `source_pipeline_run` parameter, threaded through `InMemoryRegistryClient` (silently ignored, matching the existing convention for `schema`) and `APIRegistryClient` (written into `model.spec.source_pipeline_run` under a truthiness guard, matching the existing pattern for `deployable_artifact_uri`/`description`).
- `ModelPusherPlugin.execute()` derives the value automatically and passes it on every registration call -- no new field on `ModelPluginConfig`. This is system-derived provenance, not something a task author should be able to override with an arbitrary value.

## Test plan

- 123 new/extended unit tests covering: both/only-name/only-namespace/neither env var set, empty-string run name, process-global caching, silent ignoring by `InMemoryRegistryClient`, proto field assertions (with and without namespace) by `APIRegistryClient`, and the pusher plugin passing the same value to every registry in a multi-registry fan-out.
- 100% coverage on the new module and every touched code path.
- Full `lib/model_manager` and `workflow/tasks/pusher` regression suites pass (13 pre-existing, unrelated failures reproduced identically on a clean `main` checkout -- fixture/environment issues in unrelated packager/loader tests, untouched by this change).
- `ruff format`/`ruff check` clean.

### DIFF
--- a/python/michelangelo/lib/model_manager/registry/api_client.py
+++ b/python/michelangelo/lib/model_manager/registry/api_client.py
@@ -58,6 +58,9 @@ if TYPE_CHECKING:
     from collections.abc import Mapping
 
     from michelangelo.api.v2.services.gen.model import ModelService as _ModelServiceType
+    from michelangelo.lib.model_manager.registry.pipeline_run import (
+        SourcePipelineRun,
+    )
 
 _logger = logging.getLogger(__name__)
 
@@ -195,6 +198,7 @@ class APIRegistryClient(ModelRegistryClient):
         schema: dict[str, Any] | None = None,
         labels: Mapping[str, str] | None = None,
         metadata: Mapping[str, Any] | None = None,
+        source_pipeline_run: SourcePipelineRun | None = None,
     ) -> RegisteredModel:
         """Register a model via ``create_model``, falling back to ``update_model``.
 
@@ -210,6 +214,10 @@ class APIRegistryClient(ModelRegistryClient):
             labels: String key-value pairs stored in ``model.metadata.labels``.
             metadata: Arbitrary JSON-serializable key-value pairs stored under
                 the annotation ``michelangelo.io/metadata``.
+            source_pipeline_run: Optional identity of the pipeline run that
+                produced this model. When provided, stored as
+                ``model.spec.source_pipeline_run.name`` (and ``.namespace``,
+                when set). Left unset when ``None``.
 
         Returns:
             :class:`~michelangelo.lib.model_manager.registry.client.RegisteredModel`
@@ -228,6 +236,7 @@ class APIRegistryClient(ModelRegistryClient):
                 kind=kind,
                 labels=labels,
                 metadata=metadata,
+                source_pipeline_run=source_pipeline_run,
             )
             try:
                 _logger.info("Calling create_model for '%s'.", name)
@@ -302,6 +311,7 @@ class APIRegistryClient(ModelRegistryClient):
         kind: str | None,
         labels: Mapping[str, str] | None,
         metadata: Mapping[str, Any] | None,
+        source_pipeline_run: SourcePipelineRun | None = None,
     ) -> model_pb2.Model:
         model = model_pb2.Model()
         model.metadata.name = name
@@ -313,6 +323,10 @@ class APIRegistryClient(ModelRegistryClient):
         if description:
             model.spec.description = description
         model.spec.kind = _convert_model_kind(kind)
+        if source_pipeline_run:
+            model.spec.source_pipeline_run.name = source_pipeline_run.name
+            if source_pipeline_run.namespace:
+                model.spec.source_pipeline_run.namespace = source_pipeline_run.namespace
 
         # Kubernetes label values are capped at 63 characters and restricted
         # to alphanumerics/dashes/underscores/dots. Callers built on top of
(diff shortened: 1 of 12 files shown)


krishpatel9 (approved):

sallycr:
## Sandbox validation - PASS

Validated against a real `michelangelo-examples` `california-housing` **pytorch_train** pipeline run on a local k3d sandbox (Cadence/Uniflow, Ray + Spark), with this branch's Python SDK changes built into the pipeline's task image (installed over the PyPI-pinned `michelangelo` version).

**Also validated together with #2004** (Go apiserver `Model` API hook), deployed into the same sandbox, to exercise the full provenance chain this PR enables.

**Run:** `california-housing/run-20260905-021805-6869daea` - all 5 steps (`feature_prep`, `preprocess`, `train`, `assembler`, `push_step`) reached `PIPELINE_RUN_STEP_STATE_SUCCEEDED`; pipeline run terminal state `SUCCEEDED`.

**Resulting model** (`california-housing/model-20260905-022144-059d3711`) confirms this PR's SDK change works end-to-end:

```yaml
spec:
  source_pipeline_run:
    namespace: california-housing
    name: run-20260905-021805-6869daea
```

i.e. `push_step` -> `ModelPusherPlugin.execute()` -> `get_source_pipeline_run()` correctly read the worker-injected `MA_NAMESPACE`/`MA_PIPELINE_RUN_NAME` env vars and threaded them through `register_model()` into the proto field, exactly as designed. No regressions observed in the rest of the pipeline.

_Ad hoc quick-smoke-test-style sandbox validation, not a formal CI run._

python/michelangelo/lib/model_manager/registry/api_client.py
```diff
@@ -313,6 +323,10 @@ def _build_model_proto(
         if description:
             model.spec.description = description
         model.spec.kind = _convert_model_kind(kind)
+        if source_pipeline_run:
```
sallycr, line 326: Change to set run name and namespace when source_pipeline_run and source_pipeline_run.namespace exists.

python/michelangelo/lib/shared/pipeline_run/pipeline_run.py
```diff
@@ -0,0 +1,85 @@
...
+
+_ENV_PIPELINE_RUN_NAME = "MA_PIPELINE_RUN_NAME"
+_ENV_NAMESPACE = "MA_NAMESPACE"
+
+
+class _Unset:
+    """Sentinel distinguishing "not yet read" from "read, and it was None"."""
+
+
+_cache: SourcePipelineRun | _Unset | None = _Unset()
```
sallycr: we don't need cache here as it's reading env vars.

sallycr:
Pushed 9b05d913d addressing [this comment](https://github.com/michelangelo-ai/michelangelo/pull/2034/files#r3939194537): removed the process-global cache from `get_source_pipeline_run()` — it now just reads the two env vars on every call, no `_Unset` sentinel/module-global state. Dropped the now-unused `_reset_source_pipeline_run_cache()` test hook and its call sites accordingly.

Re-verified: unit tests (120 passed, incl. a new no-caching regression test replacing the old cache-behavior tests) and a fresh sandbox run of `pytorch_train` (`california-housing/run-20260905-030137-c34a7212`, `SUCCEEDED`) — resulting model `model-20260905-030526-8972c9f4` still carries the correct `spec.source_pipeline_run` (namespace/name), confirming no regression from the simplification.

python/michelangelo/lib/model_manager/registry/api_client.py
sallycr: This is intentional, not an oversight -- `SourcePipelineRun.namespace` is deliberately optional (see its docstring: `None` means "the run's namespace wasn't available, fall back to the referencing resource's own namespace"). The current code sets `.name` whenever `source_pipeline_run` is given at all, and `.namespace` additionally when present, so a namespace-less run still gets its provenance recorded (just without a namespace override).

Collapsing to a single `if source_pipeline_run and source_pipeline_run.namespace:` guard would drop `.name` entirely whenever only `MA_PIPELINE_RUN_NAME` is set without `MA_NAMESPACE` -- a real case `get_source_pipeline_run()` produces on purpose. There's an existing regression test covering exactly this: `test_source_pipeline_run_without_namespace_leaves_namespace_unset` in `api_client_test.py`, which asserts `.name` is still set with an empty `.namespace`.

Let me know if you had a different change in mind, or if this reasoning resolves it.

PR #2047, opened 2026-09-07
feat(triggerrun): auto catch-up for cron triggers with a past start time

badcount:
A cron trigger only ever scheduled forward from the moment it was created. Covering a historical window meant hand-creating a separate BatchRerun trigger. CronSchedule.start_time now backfills the missed occurrences through the trigger's own schedule.

The obvious approach does not work: ScheduleSpec.StartAt only filters out times before it, and CatchupWindow recovers from server downtime rather than a backdated start. Temporal's schedule backfill is the mechanism that actually replays a historical range, so createScheduleForCron keeps the handle it previously discarded and issues one Backfill for [start_time, now].

Overlap is overridden to BUFFER_ALL for that request only; the schedule's steady-state SKIP is unchanged, so no existing trigger behaves differently. BUFFER_ALL is also what holds the max_concurrency budget: peak parallelism is (concurrent occurrences) x max_concurrency, and serializing the replay pins the first term to 1. It also replays in chronological order, which matters for date-partitioned pipelines where a day depends on the one before.

Deriving DS from workflow.Now() would have made the feature useless. A backfill runs every replayed action at once, so all caught-up runs would have landed on today's date instead of the date each one stands for - verified against a real server, 24/24 actions wrong. logicalTimestamp now reads the occurrence from the RFC3339 suffix the engine puts on a scheduled action's workflow ID, falling back to the execution clock when absent, which covers manually started runs and all of Cadence.

Guardrails: a nil start_time must not decode to the Unix epoch and request a 56-year backfill, so the timestamp is nil-checked before conversion; a start older than 30 days is rejected rather than clamped, so a mistyped year is visible instead of quietly enqueueing thousands of runs. Cadence has no schedule-backfill equivalent and rejects start_time outright rather than starting a trigger that silently never catches up.

start_time applies at trigger creation only.

Note for rollout: CRDs are generated by the apiserver from these protos, so the new field is silently pruned by the API server until the apiserver ships this change - controllermgr alone is not enough.

Verified end to end against a Temporal-backed sandbox: a backdated hourly trigger produced four runs, each carrying its own occurrence (18:00/19:00/20:00/21:00Z) rather than the 21:16-21:31Z wall clock they actually executed at, all SUCCEEDED.

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Breaking Changes**

- [ ] No breaking changes

**Migration guide**

**Release notes**

**Documentation Changes**


proto/api/v2/trigger_run.proto
```diff
@@ -34,6 +34,29 @@ import "michelangelo/api/v2/notification.proto";
...
+  // - Ignored when the trigger is created paused, since a paused schedule takes no
+  //   actions. Resuming it later does not replay the window.
+  // - Requires the Temporal workflow provider; rejected under Cadence, which has no
+  //   schedule-backfill equivalent.
+  //
+  // If the schedule is created but the replay fails, the trigger stays running and the
+  // missed window is reported in TriggerRunStatus.error_message. It is not retried
+  // automatically, because a backfill whose outcome is unknown cannot be re-issued
+  // without risking duplicate runs for the same period.
+  google.protobuf.Timestamp start_time = 2;
```
yangfan0328, line 59: Let's hold the proto change until we have design discussions, where to put start and end timestamp is still a question mark.

yangfan0328:
We may merge the workflow-client changes now with no proto change. API side and execution plane changes still need design discussions. Separate the workflow-client changes to a new PR if you want to accelerate the process.

yangfan0328:
@badcount — I've commandeered this PR to land it on the API shape we agreed after the design discussion, rather than asking you for another round. Pushed to your branch as one commit on top of a merge from `main` (no history rewritten, your commits are intact):

- `spec.catchup` (field 12) replaces `CronSchedule.start_time`; the catch-up start is read from `spec.start_timestamp`.
- `start_timestamp` / `end_timestamp` now map to Temporal `StartAt` / `EndAt`, and the backfill is bounded at `min(now, EndAt)`.
- Occurrence-count cap (500) instead of the 30-day age cap; future start is allowed (idles the schedule).
- API-hook validation, including rejecting a future-dated backfill window at create time.

Everything you built around the engine — the retained handle, `Backfill` with `BUFFER_ALL`, `CatchUpError` keeping the trigger live, the nil guard and the logical-timestamp fix — is kept as-is. Details and the follow-up list are in the updated description; the design itself is [here](https://personal.uberinternal.com/yangfan/triggerrun-schedule-window.html).

Please take a look when you have a moment, especially the Cadence behaviour change in "Potential risks".

PR #2055, opened 2026-09-08
fix(ray): persist RayJob immutable annotation and fix always-dirty status check

dragod812:
## Problem

Part of the reconcile-storm fix. In `RayJob.Reconcile` (`go/components/ray/job/controller.go`), RayJobs rewrote their status on every reconcile, and terminal jobs never became immutable — so they were reconciled forever and the ingester never collected them.

## Fix

- **Always-dirty status check.** `originalRayJob` (a `*RayJob` from `DeepCopy`) was compared to `rayJob` (a value) via `reflect.DeepEqual`, which is unequal on type alone — so status was rewritten (firing a fresh watch event) every pass. Now compares `.Status` to `.Status`, matching the `RayCluster` controller.
- **Dropped annotation.** `utils.MarkImmutable` sets an annotation (metadata), but the block only did `Status().Update` (status subresource), so the marker never reached etcd. Now persisted with a plain `Update`.
- **Marker gated behind the dirty-check.** Once a terminal job's status settles, the dirty-check goes false and the marker was skipped forever. It now runs independently, on every reconcile while terminal-but-mutable.
- **Hardened write.** Factored into `markImmutableIfTerminal`: it re-fetches the latest object, retries on conflict, and the caller requeues if the write ultimately fails — so a conflict (from the preceding status update or a concurrent writer) no longer strands a terminal job as mutable.

Out of scope: the cluster controller's own fixes and the NotFound-on-remote case (#2006, now on main).

## Verification

- `bazel build //go/components/ray/...` and `bazel test //go/components/ray/...` pass.
- `controller_test.go`: all three terminal states assert immutability on the RayJob re-fetched from the fake client; `NoOpSkipsStatusWrite` proves no write occurs when nothing changed; `RetriesImmutableAnnotationWhenStatusUnchanged` proves a settled terminal job still becomes immutable.
- Local OSS sandbox e2e: missing-cluster job → FAILED + immutable, happy-path job → SUCCEEDED + immutable, `resourceVersion` flat after terminal (no status-write storm).

### DIFF
--- a/go/components/ray/job/controller.go
+++ b/go/components/ray/job/controller.go
@@ -97,19 +97,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			res.RequeueAfter = requeueAfter
 			return res, err
 		}
+	}
 
-		if isTerminalRayJobState(rayJob.Status.State) && !utils.IsImmutable(&rayJob) {
-			// Annotations are metadata, so Status().Update above never persists
-			// them. Persist with a plain Update, and do it after the status
-			// write using the resourceVersion that refreshed: doing it first
-			// would bump resourceVersion and hand back the pre-update status in
-			// the response, clobbering the status we just persisted.
-			utils.MarkImmutable(&rayJob)
-			if err := r.Update(ctx, &rayJob); err != nil {
-				// Log only: don't turn an already-successful status write into
-				// a failed reconcile over a best-effort annotation.
-				logger.Error(err, "failed to persist immutable annotation")
-			}
+	// Checked independently of the dirty-check above -- and on every
+	// reconcile, not just the pass that first reached a terminal state --
+	// so a job stuck terminal-but-mutable (e.g. a prior annotation write
+	// failed, or it lands on the same terminal status every pass, as with a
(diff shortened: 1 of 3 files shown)


go/components/ray/job/controller.go
```diff
@@ -102,6 +99,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
+	// Checked independently of the dirty-check above -- and on every
+	// reconcile, not just the pass that first reached a terminal state --
+	// so a job stuck terminal-but-mutable (e.g. a prior annotation write
+	// failed, or it lands on the same terminal status every pass, as with a
+	// permanently missing cluster spec) keeps getting retried instead of
+	// being skipped forever once its status stops changing.
```
craigmarker: did we collect any telemetry on what contributed to jobs becoming stuck terminal-but-mutable?

craigmarker (approved):

go/components/ray/job/controller.go
dragod812: I think all of the jobs were terminal and mutable before this change. Previously, we were marking the entity immutable and not updating the Entities metadata, if you see we were only updating the Status not the metadata on line 94.

PR #2059, opened 2026-09-10
docs(ui): document lowercase-canonical entity name convention

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Brings the UI configuration docs in line with the lowercase-canonical entity `name`
convention that #2040 established, and documents the convention itself.

- `entity-configuration-reference.md` — `name: 'Runs'` → `'runs'` in both code
  examples; property table and Examples list switched to lowercase; new paragraph
  explaining that `formatEntityName(name, 'nav')` applies Title Case at render time
  for nav chrome only; a `:::note` clarifying this does **not** apply to phase `name`.
- `configuration-api.md` — same value fix, plus a `name` bullet calling out that
  entity `name` is lowercase while the phase `name` shown directly above it on the
  same page stays Title Case.

**The stale example strings were the smaller half of this.** The `name` field's
*described semantics* were also wrong. The docs called it "Display name shown in
navigation and headers." It is now the lowercase canonical value, title-cased only
when rendered into nav chrome (breadcrumbs, left-nav, phase tabs, page H1s).
Everywhere else the lowercase form *is* the display form — which is what makes it
read correctly mid-sentence ("no runs found").

**Why?**

#2040 ("fix(core): title-case entity names in remaining nav consumers") moved nav
casing to render time and lowercased all six entity configs, but shipped with
`Documentation Changes: N/A`. That left these docs describing the previous
convention.

This is the second time these examples have gone stale in a week: #2016 corrected
them for the `Pipeline Runs` → `Runs` rename on 2026-09-03, and #2040 invalidated
that correction on 2026-09-09. Documenting the *rule* rather than only patching the
strings is the part meant to stop a third round.

The failure mode is quiet, which is why it's worth spelling out in the docs:
`formatEntityName` leaves existing capitals untouched, so it's idempotent. A
contributor who copies the old Title Case example gets something that looks correct
in the tabs and breaks sentence case everywhere else — no error, no visual cue in the
place they'd think to check.

**How did you test it?**

Verified against `origin/main`:

- All six entity configs are lowercase: `deployments`, `trained models`, `pipelines`,
  `runs`, `targets`, `triggers`. Every `name` value shown in the docs now matches its
  source file.
- `formatEntityName` (`core/utils/string-utils.ts:115`) title-cases each
  space-separated word only when passed `'nav'`; its docstring states the lowercase
  canonical value is the correct display form elsewhere. Applied in `breadcrumb-bar`,
  `phase-entity-list`, `phase-entity-view` tabs, and `phase-card`.
- Phase configs confirmed **unaffected** — `config/phases/train.ts` and `data.ts` keep
  Title Case (`'Train & Evaluate'`, `'Prepare & Analyze Data'`), and
  `phase-configuration-reference.md` already documents that correctly, so it is
  untouched here.
- Swept every doc referencing `PhaseEntityConfig` / `ENTITY_CONFIG`;
  `types-and-patterns.md:57` was already correct (`Display name (plural, lowercase)`).
  No Title Case entity examples remain anywhere under `docs/`.
- Relative link and `#name` anchor in the new `configuration-api.md` bullet resolve.

**Potential risks**

None — documentation only. No code, config, or schema changes.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

This PR is the documentation change — two files under
`docs/contributing/dev/ui/configuration/`. No content removed.

One open question for UI owners: `formatEntityName` is now load-bearing for anyone
authoring an entity config, but it is only described in these guides. If the casing
convention shifts again, this doc is the thing that goes stale. A short doc comment
on `PhaseEntityConfig.name` pointing at the render-time rule would keep the
convention next to the type contributors actually autocomplete against.

### DIFF
--- a/docs/contributing/dev/ui/configuration/configuration-api.md
+++ b/docs/contributing/dev/ui/configuration/configuration-api.md
@@ -46,7 +46,7 @@ See [`javascript/packages/core/config/phases`](https://github.com/michelangelo-a
 // Example: config/entities/run/run.ts
 export const RUN_ENTITY_CONFIG: PhaseEntityConfig = {
   id: 'runs',
-  name: 'Runs',
+  name: 'runs',
   service: 'pipelineRun',
   state: 'active',
   views: [RUN_LIST_CONFIG, RUN_DETAIL_CONFIG],
@@ -54,6 +54,9 @@ export const RUN_ENTITY_CONFIG: PhaseEntityConfig = {
 ```
 
 - **id**: URL routing (`/<phase>/runs`)
+- **name**: lowercase canonical form — nav title-cases it at render time via
+  `formatEntityName`. Note this differs from phase `name` above, which is stored
+  Title Case. See [Entity Configuration Reference](./entity-configuration-reference.md#name)
 - **service**: Maps to protobuf service name
 - **views**: Supported views (list, detail)
 
--- a/docs/contributing/dev/ui/configuration/entity-configuration-reference.md
+++ b/docs/contributing/dev/ui/configuration/entity-configuration-reference.md
@@ -13,7 +13,7 @@ Entities define data models and their properties within a phase. Each entity rep
 | Property | Type | Description | Required |
 |----------|------|-------------|----------|
 | `id` | `string` | Unique identifier within the phase, used in URL routing | ✅ Yes |
-| `name` | `string` | Display name for the entity (plural, lowercase recommended) | ✅ Yes |
+| `name` | `string` | Lowercase canonical display name (plural); title-cased at render time in nav | ✅ Yes |
 | `service` | `string` | Name of the protobuf service this entity maps to | ✅ Yes |
 | `state` | `PhaseEntityState` | Controls entity availability | ✅ Yes |
 | `views` | `ViewConfig[]` | Array of view configurations (list, detail, form) | ✅ Yes |
@@ -33,7 +33,7 @@ The `state` property controls individual entity behavior within a phase:
 // From: config/entities/run/run.ts
 export const RUN_ENTITY_CONFIG: PhaseEntityConfig = {
   id: 'runs',
-  name: 'Runs',
+  name: 'runs',
   service: 'pipelineRun',
   state: 'active',
   views: [RUN_LIST_CONFIG, RUN_DETAIL_CONFIG],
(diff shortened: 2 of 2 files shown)


paulzim (approved):
# PR #2059 — docs(ui): document lowercase-canonical entity name convention

**Author:** zhoward-1 (+ Claude Opus 5 co-author) · **`docs/entity-name-lowercase-convention` → `main`** · **1 commit, +25/−7, 2 files** (docs only, matches GitHub exactly — no merge-base correction needed, this branch is only 1 commit deep). Existing reviews: none.

## Verdict: **Approve**

Docs-only PR fixing stale examples left behind by #2040, plus documenting the actual rule so it doesn't go stale a third time. Every factual claim in the PR body and commit message — the six entity values, the `formatEntityName` behavior and call sites, the phase/entity distinction, the "nothing else is stale" sweep — checks out exactly against the current source. I also found one thing worth flagging: the "open question for UI owners" in the PR description is already answered by code that landed the day before #2040.

## What it does

Two files under `docs/contributing/dev/ui/configuration/`:
- `entity-configuration-reference.md`: `name: 'Runs'` → `'runs'` in both code examples, property table and Examples list switched to lowercase, new paragraph explaining `formatEntityName(name, 'nav')` title-cases at render time for nav chrome only, and a `:::note` clarifying this doesn't apply to phase `name`.
- `configuration-api.md`: same value fix, plus a bullet contrasting entity `name` (lowercase) against the phase `name` shown directly above it on the same page (which stays Title Case).

## Verified

**All six `PhaseEntityConfig.name` values are exactly as claimed.** Grepped every entity config under `javascript/packages/core/config/entities/`: `deployments`, `trained models`, `pipelines`, `runs`, `triggers`, `targets` — all lowercase, matching the PR's list precisely. (Nested entity configs defined inline inside phase files — `evaluations`, `notebooks`, `data sources` in `train.ts`/`data.ts` — are lowercase too, consistent with the same convention, though not part of what this PR documents.)

**`formatEntityName`'s behavior matches the doc's new prose exactly.** `string-utils.ts:115` — title-cases each space-separated word only when called with `casing: 'nav'`, returns the input unchanged otherwise. Confirmed idempotent on already-capitalized input (capitalizes `word[0]`, leaves the rest of the word untouched) — which is the exact mechanism the new doc text warns about ("pre-casing... looks right in the tabs and breaks sentence case everywhere else").

**Call sites match the test plan's list exactly.** Non-test callers: `breadcrumb-bar.tsx`, `phase-entity-list.tsx`, `phase-card.tsx`, `phase-entity-view.tsx` — four files, all four named in the PR's "How did you test it?" section, nothing extra or missing.

**Phase vs. entity distinction is real, not just asserted.** `config/phases/*.ts` — `'Train & Evaluate'`, `'Prepare & Analyze Data'`, etc. — all still stored Title Case, confirmed unaffected by #2040. `phase-configuration-reference.md`'s existing `name` row ("Display name shown in navigation") is accurate as-is for phase config (no render-time transform exists for it), so the PR is correct to leave that file untouched.

**The "swept everywhere, nothing else stale" claim holds.** Grepped all of `docs/` for the old Title Case example strings and for any lingering "Display name shown in navigation and headers" phrasing (the old, wrong entity description) — zero hits outside what this PR already fixes. `types-and-patterns.md:57` is confirmed already-correct at that exact line (`// Display name (plural, lowercase): "trained models"`) and correctly left alone.

**The relative link resolves.** `entity-configuration-reference.md#name` — the target file exists at that relative path, and its `### \`name\`` heading (line 52) slugs to `#name` under standard heading-anchor rules. The `:::note` admonition syntax matches existing usage elsewhere in the docs (`documentation-guide.md`), not a one-off invention.

**CI is green** (`validate-docs` passes; the three advisory checks skip as expected for a docs-only change).

## Non-blocking notes

**The PR's own "open question for UI owners" is already resolved — worth telling the author so they don't chase it further.** The description asks for "a short doc comment on `PhaseEntityConfig.name` pointing at the render-time rule... next to the type contributors actually autocomplete against." That comment already exists: `studio-types.ts:65-78` has a full JSDoc block on `PhaseEntityConfig.name` — "Plural, lower case, and not pre-cased — nav call sites apply Title Case via `formatEntityName` at render time" — with `@example` lines for `trained models`, `feature consistency`, etc. It landed in **#2029** ("render breadcrumb entity names in Title Case"), which merged 2026-09-09, one day before #2040. Nothing to act on in this PR — the type-level doc and this markdown doc now just happen to say the same thing in two places, which is fine — but worth a one-line reply on the PR so the "open question" doesn't turn into unnecessary follow-up work.

## Summary

Small, accurate, thoroughly cross-checked docs fix. Every value, file reference, and behavioral claim in the PR matches the current source exactly, and the one loose thread the author flagged themselves turns out to already be tied off elsewhere.

PR #2064, opened 2026-09-11
ci(changelog): fix range fallback to survive release-branch deletion

### LINKED ISSUE #2046: changelog.yml's range fallback breaks once a previous release branch is deleted
## Summary

`changelog.yml`'s "Determine changelog range" step (added in #1561 to work around squash-merge breaking tag ancestry — see also #1662) has a fallback path that silently computes the wrong range once a previous cycle's release branch no longer exists.

## Mechanism

For a tag on a branch other than `main`, the range logic:
1. Finds the most recent *final* tag and checks if it's a real git ancestor of the current tag.
2. Because this repo only allows squash merges (`allow_merge_commit: false`), a release branch's merge-back PR squash-merges into `main`, so the release tag's commit is never actually an ancestor of `main` afterward (documented, permanent — #1662).
3. So it falls back to: find the *previous* release branch (immediately preceding the current one, sorted) and use its fork-point off `main`:
   ```bash
   PREV_BRANCH=$(git branch -r --list 'origin/release/v*' \
     | sed 's#^[* ]*origin/##' \
     | sort -V \
     | grep -B1 -x "$CURRENT_BRANCH" | head -1)
   ```
   This only searches **currently-existing** remote branches.

## Repro

`release/v0.9` was auto-deleted after its merge-back PR (#1925) merged, since this repo has `delete_branch_on_merge: true`. When `v0.10.0-rc.1` was tagged on `release/v0.10`, the fallback's branch search skipped straight past the now-deleted `release/v0.9` to `release/v0.8` — computing a changelog range of `release/v0.8`'s fork point `..v0.10.0-rc.1` (164 commits) instead of the correct `release/v0.9`'s fork point `..v0.10.0-rc.1` (72 commits). This silently re-included all of `v0.8.0`'s and `v0.9.0`'s changelog content into `v0.10.0-rc.1`'s release notes and `CHANGELOG.md` — both the GitHub Release page and the auto-merged changelog PR (#2030) shipped with the wrong content until manually corrected (see #2045).

This is likely to recur every cycle now that branch auto-delete is enabled and merge-back PRs are actually getting merged promptly (a recent improvement) — previously, merge-back PRs often sat unmerged/draft for a long time, which is presumably why this hadn't surfaced before.

## Suggested fix

Don't rely on the previous release branch still existing. Some options:
- Use the previous release branch's **tag** instead (e.g. `v0.9.0` or `v0.9.0-rc.1`, whichever exists) to compute the fork point via `git merge-base origin/main <tag>` — tags aren't deleted by branch cleanup, so this survives the branch being gone.
- Or derive `PREV_BRANCH` purely from tag history (sort real final tags by commit timestamp, same technique already used for the `CANDIDATE` search earlier in the same step) rather than from `git branch -r`.

Related: #1561, #1662.

sallycr:
## Summary
- The changelog workflow's "Determine changelog range" step falls back to searching currently-existing `origin/release/v*` branches to find the previous cycle's fork point off `main`.
- Since `delete_branch_on_merge: true` deletes a release branch immediately after its merge-back PR merges, the search silently skips past the deleted branch to the one before it — widening the changelog range by a full extra release cycle.
- This has already caused two real incidents that required hand-patching: `v0.10.0-rc.1`'s release notes/CHANGELOG (fixed via #2045) and `v0.10.0` final's changelog PR (fixed via #2049), both because `release/v0.9` was gone by the time the workflow ran.
- Fix: reuse `$CANDIDATE` — the most recent preceding **final tag**, which the step already computes one step earlier — instead of hunting for the release branch. Tags are never deleted by branch cleanup, so `git merge-base origin/main "$CANDIDATE"` reproduces the correct fork point regardless of whether the branch still exists. This will recur on every cycle going forward (not an edge case) now that branch auto-delete + prompt merge-backs are standard practice, so the fix removes the dependency entirely rather than patching around it again.

Fixes #2046

## Test plan
- [x] `actionlint .github/workflows/changelog.yml` returns 0
- [x] Verified algebraically that `git merge-base origin/main v0.9.0` returns `a567b2fb057112eecdab8bf0de3650a5334c148e`, matching the previously hand-verified-correct fork point from #2045/#2049
- [x] Same-branch patch-release path (ancestor check) is unchanged, so no regression there
- [x] First-release-ever path (`CANDIDATE` empty) still falls back to `--latest`, behavior-preserving
- [ ] Live-verify on the next real RC or final cut that `RELEASE_NOTES.md`/`CHANGELOG.md`'s commit count matches an independently computed `git log <range> --oneline | wc -l`

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -101,28 +101,26 @@ jobs:
           # main, which is always a real ancestor since main only grows
           # forward, and correctly scopes to "what's new since that branch
           # point" without re-including anything already released.
-          CURRENT_BRANCH="${{ steps.branch.outputs.base }}"
-
-          if [ "$CURRENT_BRANCH" = "main" ]; then
-            echo "range=" >> "$GITHUB_OUTPUT"
-            echo "Tag is on main, not a release branch, and the preceding tag isn't an ancestor; falling back to --latest"
+          # CANDIDATE (found above) is not an ancestor of CURRENT_TAG -- expected
+          # whenever a release-branch boundary was crossed, since this repo is
+          # squash-merge-only and a release tag's commit is therefore never an
+          # ancestor of anything on main afterward (#1662). The correct range start
+          # is the fork point of CANDIDATE's own release cycle off main, which
+          # equals merge-base(main, CANDIDATE) as long as cycles are sequential
+          # (the next cycle's branch is cut only after the previous one's merge-back
+          # has landed on main -- the same assumption the old branch-based fallback
+          # already made). Using CANDIDATE's *tag* instead of hunting for its
+          # release *branch* survives that branch being deleted after its
+          # merge-back PR merges (delete_branch_on_merge: true) -- see #2046.
+          if [ -n "$CANDIDATE" ]; then
+            START=$(git merge-base origin/main "$CANDIDATE")
+            echo "range=${START}..${CURRENT_TAG}" >> "$GITHUB_OUTPUT"
+            echo "Changelog range: ${START} (${CANDIDATE}'s fork point from main) .. ${CURRENT_TAG}"
             exit 0
           fi
 
-          PREV_BRANCH=$(git branch -r --list 'origin/release/v*' \
-            | sed 's#^[* ]*origin/##' \
-            | sort -V \
-            | grep -B1 -x "$CURRENT_BRANCH" | head -1)
-
-          if [ -z "$PREV_BRANCH" ] || [ "$PREV_BRANCH" = "$CURRENT_BRANCH" ]; then
-            echo "range=" >> "$GITHUB_OUTPUT"
-            echo "No previous release branch found (first release); falling back to --latest"
-            exit 0
-          fi
-
-          START=$(git merge-base origin/main "origin/$PREV_BRANCH")
-          echo "range=${START}..${CURRENT_TAG}" >> "$GITHUB_OUTPUT"
-          echo "Changelog range: ${START} (${PREV_BRANCH}'s fork point from main) .. ${CURRENT_TAG}"
+          echo "range=" >> "$GITHUB_OUTPUT"
+          echo "No previous final tag found (first release); falling back to --latest"
 
 
 


sallycr:
## Verification results

Full report: `specs/070/verify.md` in the harness repo. **Overall verdict: PASS.**

### Diff vs. spec
- Matches `specs/070/spec.md`'s "Changes needed" line-for-line, including the explanatory comment.
- `steps.branch.outputs.base` confirmed still referenced downstream (`base:` input for the CHANGELOG.md PR) — not removed.

### Independent re-checks
- `actionlint .github/workflows/changelog.yml` → exit 0, clean.
- `git merge-base origin/main v0.9.0` → `a567b2fb057112eecdab8bf0de3650a5334c148e`, matches the hand-computed fork point used to fix #2045/#2049.
- WF-002 (no unpinned actions touched) and SEC-001 (no hardcoded secrets) both pass.

### Regression scenarios checked
- **Same-branch patch release inside an existing release branch** (e.g. a hypothetical `v0.10.1` cut on `release/v0.10` after `v0.10.0`) — **live-simulated**, not just diff-inspected: created a local-only commit + tag off `origin/release/v0.10`, ran the step's actual bash logic by hand. `git merge-base --is-ancestor v0.10.0 v0.10.1-test` returned true, so the step takes the unchanged ancestor-check fast path (`range=v0.10.0..v0.10.1-test`) and never reaches this PR's new fallback code. Confirms the fix is correctly inert for the common patch-release path — no regression. Test tag/worktree cleaned up, nothing pushed to any remote.
- **First release ever** (`$CANDIDATE` empty) — confirmed the new code's fallthrough preserves the old `range=`/`--latest` behavior.

### Known deferred item
- "Live-verified on the next real RC/final cut" (spec's last acceptance criterion) can't be checked until that actually happens — not a merge blocker, tracked separately for the next real tag cut.

No issues found. Ready for human merge review.

paulzim (approved):
# PR #2064 — ci(changelog): fix range fallback to survive release-branch deletion

**Author:** sallycr · **`fix/changelog-range-fallback` → `main`**, fully rebased on current `main` (merge-base = `main` tip exactly) · **1 commit, +17/−19, 1 file** (`.github/workflows/changelog.yml`) · Fixes #2046. Existing reviews: none.

## Verdict: **Approve**

This is the direct fix for the root-cause issue I identified and cleared as "correctly scoped out" during the PR #2032 review. I re-implemented the algorithm independently and ran it against all 9 final tags in the project's real history (not just the two incidents the PR cites) — every range comes out correct, with zero skips or duplication. Both checked test-plan boxes verify exactly as claimed once run outside the worktree's fork-remote naming quirk (see note below).

## What it changes

Replaces the "hunt for the previous release *branch*, then `git merge-base origin/main origin/$PREV_BRANCH`" fallback with "reuse `$CANDIDATE`" (the most recent preceding **final tag**, already computed one step earlier in the same script for the primary ancestor-check path) and `git merge-base origin/main "$CANDIDATE"`. Tags survive `delete_branch_on_merge: true`; branches don't — that's the whole fix. It also drops the old `CURRENT_BRANCH = "main"` early-exit special case, which no longer appears anywhere in the new logic.

## Verified

**The merge-base algebra checks out exactly as claimed — with one worktree-specific gotcha worth flagging.** Running `git merge-base origin/main v0.9.0` in the pr-review worktree gives a *different* hash than the PR claims, because this worktree's `origin` remote points at a personal fork (`paulzim/michelangelo`), not `michelangelo-ai/michelangelo` — so `origin/main` there is a stale, unrelated ref. Substituting the actual upstream ref (`git merge-base upstream-main v0.9.0`) reproduces the PR's claimed `a567b2fb057112eecdab8bf0de3650a5334c148e` exactly. In the real workflow run (which executes inside `michelangelo-ai/michelangelo` itself), `origin` *is* upstream, so this isn't a PR bug — just a trap for reviewing it from a fork-based worktree. Noting it so the next `changelog.yml` review doesn't get a false alarm from it.

**`actionlint .github/workflows/changelog.yml` returns 0** — confirmed independently (no `actionlint` job runs in this repo's own CI, so this wasn't otherwise checked). Built `github.com/rhysd/actionlint@latest` via `go install` and ran it directly against the file; clean, exit 0.

**Ran the new algorithm against every one of the 9 real final tags in the repo's history (v0.3.0 through v0.10.0), not just the two hand-verified incidents (#2045/#2049) the PR cites:**

| Tag | Result |
|---|---|
| v0.3.0 | no CANDIDATE → falls back to `--latest` (first release ever) |
| v0.3.1 | CANDIDATE=v0.3.0, real ancestor → tight range `v0.3.0..v0.3.1` |
| v0.4.0 | CANDIDATE=v0.3.1, real ancestor → tight range `v0.3.1..v0.4.0` |
| v0.5.0 | CANDIDATE=v0.4.0, not ancestor → fork-point range |
| v0.6.0 – v0.10.0 | same pattern as v0.5.0, each one clean |

Every range chains cleanly into the next with no gap and no re-inclusion — this is a stronger check than the PR's own test plan attempts (which only algebraically checks the v0.9.0→v0.10.0 hop). v0.10.0's computed range matches the PR's claimed hash exactly.

**The removed `CURRENT_BRANCH == "main"` special case is not a regression.** That path only mattered for a tag not found on any currently-existing `release/v*` branch. Two ways of confirming the new code handles that identically:
- Per `CONTRIBUTING.md`'s "How to Release" table, every tag — RC or final — is created only *after* `release/vX.Y` is cut, via `release-cut.yml`/`release-promote.yml`, and patch releases are explicitly "cherry-pick to the release branch... tag `vX.Y.Z` directly" — never on `main`. So the only real-world way to hit the old `CURRENT_BRANCH == "main"` branch was the very first release ever, before any release branch existed.
- The v0.3.0 row above (the actual first release in this repo's history) proves the new `CANDIDATE`-empty check produces the identical fallback (`--latest`) without needing the old special case at all — it was fully subsumed, not just coincidentally compatible.

**Patch-release path is provably untouched, not just behaviorally unchanged.** The ancestor-check block (`if [ -n "$CANDIDATE" ] && git merge-base --is-ancestor ...`) that handles the patch case sits *above* the diff's line range entirely (diff starts at old line 101; the ancestor check is lines 85–89 in the current file) — zero characters of it changed. The v0.3.1 and v0.4.0 rows above additionally confirm it still produces correct tight ranges against real data.

**The bug this fixes is currently live, not hypothetical.** `git ls-remote --heads upstream release/v0.9` returns nothing (deleted, exactly as #2046 describes); `release/v0.10` still exists (its merge-back PR — #2032 — hasn't merged yet). The old branch-based fallback would misfire on `v0.9`'s deletion today if re-run; the new tag-based one, confirmed above, doesn't.

**CI is green** (bazel-build, python-build, dirty-check, CLA, filter all pass; the three "advisory" checks skip as expected for a workflow-only change).

## Non-blocking notes

**Nothing.** The fix is minimal, correctly scoped to exactly the bug in #2046, and the dead-code removal alongside it is provably safe rather than just plausible.

## Summary

Every claim in the test plan and PR description checks out against independently-run tooling and the repo's actual tag history — including edge cases (first release, patch releases) the PR's own test plan didn't explicitly re-verify. Nothing here should hold up the merge.

PR #2081, opened 2026-09-14
feat(workflow): add ModelVariable.to_artifact() to bridge to StorageBackend

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Adds `ModelVariable.to_artifact(storage_backend, *, destination_key=None) -> ModelArtifact` in `python/michelangelo/workflow/variables/_private/model.py`. It downloads the variable's persisted weights (via fsspec, from wherever `UF_STORAGE_URL` put them) and re-uploads them through the given `StorageBackend`, returning a `ModelArtifact` whose path is addressable by that backend's own `download()`.

**Why?**

`train_tabular()` returns a `ModelVariable`, but downstream tasks like `torch_assembler()` require a `StorageBackend`-addressable `ModelArtifact`. `ModelVariable.path` is written directly via fsspec against `UF_STORAGE_URL`, independent of any pipeline's `StorageBackend` instance — and `LocalStorageBackend.download()` only accepts URIs it produced itself. Today, callers hand-roll this download+reupload themselves at each call site (see `michelangelo-examples`' `california_housing/pipelines/pytorch_train/assembler.py`, which does this in ~6 lines). This method centralizes that bridging logic as a single reusable call, following the same method-per-storage-concern convention `ModelVariable` already uses for `save_*`/`load_*`.

This was investigated and designed via the `ma-oss-api-features` harness in response to https://github.com/michelangelo-ai/michelangelo-examples/issues/37 (which also proposed decorating `train_tabular`/`torch_assembler`/`push` directly as `@uniflow.task`s in core — that direction was evaluated and rejected: it would couple every direct call to these functions, including tests and notebooks, to real `TaskConfig.pre_run()`/`post_run()` lifecycle, e.g. `ray.init()`/`ray.shutdown()` for `RayTask`-configured calls).

**How did you test it?**

Added 7 new tests to `python/michelangelo/workflow/variables/tests/variables_test.py`: 5 mocked-fsspec unit tests (unsaved-variable `ValueError`, directory vs. single-file dispatch, custom destination key) and 2 end-to-end round-trip tests against a real `LocalStorageBackend` (upload then re-download through the returned `ModelArtifact.path`). Full existing suite (119 tests total in this file) passes. `ruff check`/`ruff format --check` clean on both changed files (one pre-existing, unrelated `ruff` finding elsewhere in the test file, confirmed present on `main` before this branch).

**Potential risks**

None expected — this is a pure addition. No existing method's signature or behavior changes; the only other diff hunk is a redundant module-level `import uuid` hoisted out of an existing method body. No other in-repo caller of `ModelVariable` was found that this could affect.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Adds `ModelVariable.to_artifact()`, a new public method for converting an in-pipeline `ModelVariable` into a `StorageBackend`-addressable `ModelArtifact`.

**Documentation Changes**

N/A — this is a new method on an existing class; its own docstring documents usage and the constraint it addresses. No wiki/README references existed to update.


sallycr:
Closing without merging.

Further design work on the storage-access interface (following up on this PR) concluded that `torch_assembler()` doesn't need a `StorageBackend`-addressable input at all — its `download()`/`upload()` calls are generic file copies with no backend-specific logic, so it can read/write directly via `fsspec` instead. That removes the reason `to_artifact()` needed to exist: bridging `ModelVariable`'s fsspec path into a `StorageBackend`-addressable `ModelArtifact` before assembly.

Under the revised design, `StorageBackend` is only needed once — at the actual publish-to-registry step (`push()`) — and `torch_assembler()`/the trainer→assembler handoff stays fsspec-native throughout. A separate, complementary `ModelVariable.download(local_path)` method (for callers who just want local bytes, unrelated to this bridging problem) is being carried forward instead.

PR #2087, opened 2026-09-14
feat(pipeline): stamp michelangelo/PipelineType label on Revision

donovanlopez:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
For pipeline revisions, to know which type of Pipeline it is, we want to stamp the `michelangelo/PipelineType` on the revision.

**Why?**
When calling the ListRevision endpoint, it will make filtering for pipeline revisions easier.

**How did you test it?**
Unit tests

**Potential risks**
N/A

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A


craigmarker:
as far as I know, we should be able to use field selector on ListRevision to filter by spec.base_type. What am I missing?

https://github.com/michelangelo-ai/michelangelo/blob/main/proto/api/v2/revision.proto#L26

donovanlopez:
> as far as I know, we should be able to use field selector on ListRevision to filter by spec.base_type. What am I missing?
> 
> https://github.com/michelangelo-ai/michelangelo/blob/main/proto/api/v2/revision.proto#L26

The `michelangelo/PipelineType` is used to filter by the phase type (Train, Deploy, Retrain, etc). I've updated the PR description because I think I didn't write the correct use of `michelangelo/PipelineType`.

craigmarker (approved):

PR #2088, opened 2026-09-14
feat(rpc): register RevisionService and unpack registered Any payloads

donovanlopez:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Registers RevisionService so we can call ListRevision and unpacks Any payloads so we can access the `content` blob from Revision which would be the snapshotted base CR (currently this PR is focusing on the pipeline revisions).

**Why?**
We want to be able to list pipeline revisions on the UI so we need to be able to unpack it from revisions.

**How did you test it?**
Unit tests

**Potential risks**
N/A

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A


javascript/packages/rpc/request.ts
```diff
@@ -58,16 +59,28 @@ function toPlainObject(value: unknown): unknown {
...
   if (!('$typeName' in value) || value.$typeName !== 'google.protobuf.Any') return undefined;
   // cast: the $typeName check above identifies this as a google.protobuf.Any message
   const any = value as Any;
-  if (any.typeUrl !== TYPED_STRUCT_TYPE_URL) return undefined;
-  return toJson(TypedStructSchema, fromBinary(TypedStructSchema, any.value));
+  if (any.typeUrl === TYPED_STRUCT_TYPE_URL) {
+    return toJson(TypedStructSchema, fromBinary(TypedStructSchema, any.value));
+  }
+  const schema = typeRegistry.getMessage(any.typeUrl.replace(TYPE_URL_PREFIX, ''));
+  if (!schema) return undefined;
```
craigmarker, line 84: Were you witnessing falsy schema in practice? My understanding is that we have a guard for this behavior at https://github.com/michelangelo-ai/michelangelo/blob/main/javascript/packages/rpc/services.ts#L67.

javascript/packages/rpc/__tests__/request.test.ts
```diff
@@ -50,3 +52,47 @@ it('handles arrays containing objects with protobuf internals', async () => {
 
   expect(await request('GetPipelineRun', {} as never)).toEqual({ items: [{ x: 1 }, { x: 2 }] });
 });
+
+it('unpacks a registered Any payload (Pipeline) into a plain object', async () => {
```
craigmarker, line 62: Consider a test case that packs the Any payload (Pipeline) with spec.manifest.conent (TypedStruct)

javascript/packages/rpc/request.ts
donovanlopez: this is because `schema` was getting `Type 'undefined' is not assignable to type 'DescMessage'.ts` since `typeRegistry.getMessage` can return `undefined`.

craigmarker (approved):

PR #2091, opened 2026-09-15
feat(ui): Add pipeline revision detail page

donovanlopez:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**
Currently, if we click on a revision on the revision list page, we still call `GetPipeline` to populate the detail page's information. This PR now adds the `revisionId` query param to designate that we're looking at a revision from which we will call `GetRevision` instead. When fetching recent pipeline runs, we will filter by the revision name instead of pipeline name.

https://github.com/user-attachments/assets/37d01b91-f8f6-4f22-8416-06dd5c247a6c

**Why?**
Users will now be able to see the revision details instead of just the pipeline.

**How did you test it?**
Unit tests and manually

**Potential risks**
N/A

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A


javascript/packages/core/config/entities/pipeline/detail.ts
```diff
@@ -32,13 +36,26 @@ export const PIPELINE_DETAIL_CONFIG: DetailViewConfig = {
...
             operation: {
-              criterion: [
-                {
-                  fieldName: PIPELINE_RUN_PIPELINE_NAME_FIELD,
-                  operator: CRITERION_OPERATOR_EQUAL,
-                  matchValue: '${page.metadata.name}',
-                },
-              ],
+              // A revision view lists only the runs pinned to that Revision CR; the live view
+              // lists every run of the pipeline regardless of revision.
```
craigmarker: nit: either don't need comment or clarify that "live" view is a view that doesn't support revisions or something.

javascript/packages/core/router/entity-detail-route.tsx
```diff
@@ -96,16 +106,22 @@ export function EntityDetailRoute({ phases = PHASES }: { phases?: Record<string,
...
+    | Record<string, unknown>
+    | undefined;
   const resolvedDetailViewConfig = resolver(detailViewConfig, { page: entityData });
   return (
     <DetailView
       subtitle={entityConfig!.name}
       title={entityId}
       onGoBack={handleReturnToEntityList}
-      actions={entityConfig!.actions}
+      actions={isRevisionView ? undefined : entityConfig!.actions}
```
craigmarker: this is probably the wrong decision—would want some way for revision view to have its own actions. Consider leveraging action conditional disabling.

javascript/packages/core/utils/revision-utils.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
craigmarker, line 1: this file is mixed on its stance for how generic the code is. `buildRevisionName` says it accepts any service, but formatRevisionId only claims to support git ref revisions.

javascript/packages/core/router/entity-detail-route.tsx
donovanlopez: do we know what kind of actions we would want the revision view to have?

javascript/packages/core/utils/revision-utils.ts
```diff
@@ -1,24 +1,24 @@
 /**
```
craigmarker, line 1: flagging here, something is still off. I think revision naming convention should be a configuration-level concern. so any entity that supports revisioning also declares its naming convention

craigmarker (approved):

PR #2104, opened 2026-09-15
feat(ui): Add more fields to inference server form

donovanlopez:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Add fields for cluster targets and initialization for the inference server create form. 

Currently, the InferenceServer CRD does not reference the Cluster CRD. This will be implemented in the future but for this form, we are fetching the clusters via `ListCluster` and then fitting it to [clusterTarget](https://github.com/michelangelo-ai/michelangelo/blob/main/python/michelangelo/cli/sandbox/demo/inference-multicluster/inferenceserver.yaml#L27-L40) on the inference server.

https://github.com/user-attachments/assets/636e5e77-1317-40fb-8921-0da05660b89e

**Why?**
We should be able to set the cluster and other fields for an inference server when creating.

**How did you test it?**
Unit tests and manually

**Potential risks**
N/A

**Breaking Changes**

- [x] No breaking changes

**Release notes**
N/A

**Documentation Changes**
N/A


javascript/packages/core/components/form/fields/select/build-select-overrides.ts
```diff
@@ -11,9 +11,12 @@ export function buildSelectOverrides(
   // When rendered inside modals, dropdown options can overflow past the modal body into its backdrop.
   // name is forwarded to the internal <input> so form-focus libraries (e.g. final-form-focus) can
   // match this control to its field error by name.
+  // BaseUI Tag truncates its text at ~100px by default, which hides most resource names.
+  // Let multi-select tags grow to their content instead.
   const base = {
     Popover: { props: { ignoreBoundary: true } },
     Input: { props: { name } },
+    Tag: { props: { overrides: { Text: { style: { maxWidth: 'none' } } } } },
```
donovanlopez, line 19: let me know if this shouldn't be overrided. it looked harder to read when truncated but hovering over the value will show the full string so it's not impossible to read it

image

javascript/packages/core/config/entities/targets/create-inference-server-form.tsx
```diff
@@ -60,10 +86,11 @@ export const CreateInferenceServerForm = ({ onClose }: CreateActionComponentProp
         numInstances: 1,
       },
     },
+    clusterIds: clusters.length === 1 ? [clusters[0].metadata.name] : [],
```
craigmarker: From AI, makes sense to me. what do you think?

clusterIds makes initialValues depend on the async cluster query. When ListCluster resolves after the dialog mounts, React Final Form reinitializes and can discard a name or initialization values entered while the cluster select was loading.

javascript/packages/core/config/entities/targets/inference-server-cluster-fields.tsx
```diff
@@ -0,0 +1,84 @@
...
+ */
+export function toClusterTargets(
+  clusterIds: string[],
+  clusters: RegisteredCluster[]
+): ClusterTarget[] {
+  const byName = new Map(clusters.map((cluster) => [cluster.metadata.name, cluster]));
+  const targets: ClusterTarget[] = [];
+  for (const clusterId of clusterIds) {
+    const rest = byName.get(clusterId)?.spec?.cluster?.value?.rest;
+    if (!rest?.host) continue;
```
craigmarker: should this filtering be done at a higher level? eg in useRegisteredClusters? otherwise there's a misalignment between what the user sees and can actually select.

what am i missing

javascript/packages/core/components/form/fields/select/build-select-overrides.ts
craigmarker, line 19: i'm not sure we want maxWidth: none, but it does look better to see options.

craigmarker: no need to block on this, though. let's consider this in a follow up

craigmarker (approved):

javascript/packages/core/config/entities/targets/inference-server-cluster-fields.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/** Lists the compute clusters operators have registered in the control plane. */
+export function useRegisteredClusters() {
+  const query = useStudioQuery<ClusterListResult>({
+    queryName: 'ListCluster',
+    serviceOptions: { namespace: CLUSTER_REGISTRY_NAMESPACE },
+  });
+  const clusters = useMemo(() => query.data?.clusterList.items ?? [], [query.data]);
+  return { clusters, isLoading: query.isLoading };
```
craigmarker: interestingly, I don't think the memo does all that much ... since a new object is always returned?

two things:

1. consider if memoization is needed
2. considering doing the destructuring at the rpc handler level

javascript/packages/core/config/entities/targets/inference-server-cluster-fields.tsx
donovanlopez: that makes sense, i'll filter it on useRegisteredClusters

javascript/packages/core/config/entities/targets/create-inference-server-form.tsx
donovanlopez: i removed this from the initialValues so we don't have to deal with this problem. we'll just make the user always have to pick the cluster target

javascript/packages/core/components/form/fields/select/build-select-overrides.ts
donovanlopez: yeah i'm not sure where else we use this. It might long ugly elsewhere but i agree that we can revisit later

Issue #2109, opened 2026-09-17
Pipeline Revision snapshots are missing parent lineage, content hygiene, and owner-based cleanup; PipelineRun creation has no production-environment guard


sallycr:
**Is your feature request related to a problem? Please describe.**

Pipeline's revision-snapshotting controller (`go/components/pipeline/controller.go`, `snapshotRevision`) already stamps a `michelangelo/PipelineType` label on every Revision it creates (#2087), but several other pieces of revision-snapshot correctness and cleanup are still missing:

1. `RevisionSpec` already declares a `parent` field (`proto/api/v2/revision.proto`), but `snapshotRevision` never populates it, so Pipeline's Revision objects carry no lineage/fork-chain relationship to whatever Revision preceded them.
2. `snapshotRevision` marshals the live Pipeline object directly into the Revision's `content` field without stripping `ObjectMeta.ManagedFields` first, so an unbounded, ever-growing `ManagedFields` blob rides along in every immutable Revision snapshot.
3. `snapshotRevision` sets no `OwnerReferences` on the Revision objects it creates, so when a Pipeline is deleted, its Revisions have nothing tying them to it — they are never cleaned up by Kubernetes garbage collection and accumulate indefinitely.
4. Separately, `go/components/pipelinerun/apihook` has no equivalent of a production-environment safety check: today a PipelineRun create against a "production"-labeled environment succeeds no matter what revision/branch it pins or what Pipeline it targets, even one still on a feature branch.

**Describe the solution you'd like**

1. In `snapshotRevision`, when the Pipeline currently has a `Status.LatestRevision`, set the new Revision's `Spec.Parent` to that prior identifier (captured before the reconcile overwrites the pointer for main/master commits), so Revisions form a lineage chain rather than being unrelated, disconnected snapshots.
2. Deep-copy the Pipeline and zero its `ObjectMeta.ManagedFields` before marshaling it into the Revision's `Content` field, so `ManagedFields` growth doesn't ride along in every snapshot.
3. After constructing each Revision, stamp the owning Pipeline as its controller owner reference (the same `cascadedelete.StampOwnerRefOnCreate` helper `go/components/pipelinerun/apihook/apihook.go` already uses for PipelineRun → Pipeline ownership), so Kubernetes garbage collection cleans up a Pipeline's Revisions once the Pipeline itself is deleted.
[...]

PR #2110, opened 2026-09-17
feat(pipeline): revision parent lineage and owner-based cascade delete

sallycr:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

In the Pipeline controller's `snapshotRevision`:
- Populates the new Revision's `Spec.Parent` with the Pipeline's prior latest revision, forming a lineage chain across snapshots instead of leaving each Revision disconnected from its predecessor. The prior revision is captured from a pre-reconcile snapshot of the Pipeline's status, because by the time the Revision is built for a main/master commit, the in-memory `Status.LatestRevision` has already been advanced to point at the very revision being created — reading it at that point would make a revision its own parent.
- Strips `ObjectMeta.ManagedFields` from a deep copy of the Pipeline before marshaling it into the Revision's `Content` field, since `ManagedFields` is server-managed, can grow unboundedly across updates, and carries no meaning as part of a point-in-time content snapshot.
- Stamps the owning Pipeline as the Revision's controller owner reference at creation time (via the existing `cascadedelete.StampOwnerRefOnCreate` helper, the same one already used for PipelineRun -> Pipeline ownership), so Kubernetes garbage collection now cleans up a Pipeline's Revisions when the Pipeline itself is deleted.

**Why?**

Revisions created by this controller previously had no lineage relationship to each other, could carry an unbounded `ManagedFields` blob in every snapshot, and were never cleaned up (no owner reference was ever set, so nothing GCs them) once their source Pipeline was deleted — leading to orphaned Revision objects accumulating over time.

**How did you test it?**

- Added `TestReconcile_RevisioningEnabled_SetsParentLineage` (two-reconcile scenario asserting the second Revision's `Spec.Parent` points at the first, not at itself), `TestReconcile_RevisioningEnabled_FeatureBranch_ParentStillSet`, `TestReconcile_RevisioningEnabled_StripsManagedFields`, and `TestReconcile_RevisioningEnabled_StampsOwnerRef` to `controller_test.go`.
- Confirmed all pre-existing tests in the package still pass unmodified.
- `gofmt -l .` clean; `golangci-lint run ./...` clean (no findings in changed files).
- `bazel run //:gazelle` (expected `BUILD.bazel` deps diff only, for the new `cascadedelete`/`runtime` imports).
- `bazel build //...` — full repo build passes.
- `bazel test //...` — all 101 test targets pass, including `//go/components/pipeline:go_default_test`.

**Potential risks**

- `Spec.Parent` is an additive field on an existing proto message; low risk unless some other consumer assumes it's always empty for Pipeline-kind Revisions (none found).
- The owner-reference stamp is best-effort and non-fatal on failure (matches the existing helper's contract); it does enforce a single-controller-owner invariant per Kubernetes semantics, so if anything else ever tries to stamp a different owner on a Pipeline-sourced Revision it would now conflict (nothing else does today).

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A

Part of #2109. This PR covers the Spec.Parent lineage, ManagedFields stripping, and cascade-delete items only — not the PipelineRun production-run guard item, which is separate follow-up work.


go/components/pipeline/controller.go
```diff
@@ -206,19 +214,36 @@ func isTerminatedState(state v2pb.PipelineState) bool {
...
+	contentSource := pipeline.DeepCopy()
+	contentSource.ObjectMeta.ManagedFields = nil
+	content, err := pbtypes.MarshalAny(contentSource)
 	if err != nil {
 		return fmt.Errorf("marshal pipeline content: %w", err)
 	}
 
+	var parent *apipb.ResourceIdentifier
+	if priorLatestRevision.GetName() != "" {
+		parent = priorLatestRevision
```
npuppala111: @sallycr I think this sets Parent correctly on the first reconcile, but then overwrites it with a self-reference on the next one. I think the lineage chain would still be broken unless the parent is preserved rather than recomputed on the update path

go/components/pipeline/controller_test.go
```diff
@@ -240,6 +241,161 @@ func TestReconcile_LatestRevisionOnlyOnDefaultBranch(t *testing.T) {
...
+
+	_, err = reconciler.Reconcile(context.Background(), req)
+	require.NoError(t, err)
+
+	secondRev := &v2pb.Revision{}
+	require.NoError(t, reconciler.Get(context.Background(), "test-namespace", "pipeline-test-pipeline-def987654321", &metav1.GetOptions{}, secondRev))
+	require.Equal(t, &apipb.ResourceIdentifier{
+		Name:      "pipeline-test-pipeline-abc123456789",
+		Namespace: "test-namespace",
+	}, secondRev.Spec.Parent, "second revision must chain to the first, not self-reference")
```
npuppala111: @sallycr I think this is why the overwrite case slips through — each test reconciles once per commit, so the second pass over the same commit is never exercised. Could we add a case that reconciles the same commit twice and asserts Spec.Parent is unchanged and not self-referential?

Issue #621, opened 2025-11-25
[Model Configmap Provider] Add appropriate locking mechanism to safely access shared configmap for inference-servers


aritra0227:
Each inference-server has its own configmap. The deployment controller should be able to make concurrent updates to the shared configmap.

aritra0227:
complete in https://github.com/michelangelo-ai/michelangelo/pull/2111

PR #2111, opened 2026-09-17
fix(go): unload models per deployment and per cluster on delete

aritra0227:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Two fixes to model unloading on Deployment delete.

1. **Unload from every target cluster.** `CleanupActor` iterates the `deployment.michelangelo.ai/target-clusters` annotation and removes the model using a per-cluster client from `ClientFactory`. `Retrieve` checks per cluster too, so the condition only goes TRUE once every cluster is clear.

2. **Scope model config entries by deployment.** Entries carry a `deployment_name` and are identified by the `(deployment, model)` pair rather than model name alone. `RemoveModelFromConfig` and `CheckModelExists` take a deployment name and only match that deployment's entry. The sync sidecar collapses entries by model name, so a model stays loaded while any deployment still claims it.

Model config writes are now serialized by a per-ConfigMap `keyedmutex` plus `retry.RetryOnConflict` (closes TODO #621). `ModelCleanupActor` checks the model config instead of querying Triton, so it no longer waits on a model another deployment keeps loaded.

**Why?**

Deleting a Deployment leaked the model on remote clusters: the model config is a per-cluster ConfigMap, but cleanup only mutated the control plane's copy, and `Retrieve` shared the blind spot — it released the finalizer while remote clusters were still serving, so nothing ever reclaimed those GPUs.

Keying entries by model name alone was wrong in the other direction. Two Deployments serving the same model share one entry, so deleting either one unloaded the model out from under the survivor, breaking live traffic.

Concurrency made both worse: two Deployments on one InferenceServer read-modify-write the same ConfigMap, and the loser of the 409 race stranded its rollout in `ROLLOUT_FAILED`, since the conditions engine does not retry.

**How did you test it?**

- Unit tests: per-cluster unload, deployment-scoped add/remove/exists, and three concurrency tests asserting simultaneous adds and removes each keep the other deployment's entry. `./tools/bazel test //go/... --build_tests_only` passes.
- End-to-end in k3d, both topologies, two Deployments sharing one model on one InferenceServer:
  - Single cluster and multi-cluster (2 clusters): before delete each model config holds one entry per Deployment, both serve `200`.
  - After deleting one Deployment: only its entry disappears in every cluster's ConfigMap, the survivor keeps serving `200` and stays `READY` in Triton across a full sidecar cycle, the deleted route returns `404`, and the CR is gone with no stuck finalizer.
  - The earlier single-Deployment multi-cluster case still unloads everywhere, and the InferenceServer stays `SERVING`.

**Potential risks**

Cleanup now depends on reaching every target cluster, so an unreachable cluster holds the finalizer until it recovers. That is deliberate — the alternative is the silent leak this fixes — but a delete that used to be fast can now block.

Entries written before this change have no `deployment_name` and no longer match any deployment, so cleanup will not remove them; they linger until the InferenceServer is deleted. Matching them by model name would reintroduce the bug of unloading a model another deployment is serving.

**Breaking Changes**

- [ ] No breaking changes

**Migration guide**

`ModelConfigProvider.RemoveModelFromConfig` and `common.CheckModelExists` take a `deploymentName` before the model name; `ModelConfigEntry` gains `DeploymentName`. Implementors and callers outside this repo pass the owning Deployment's name:

```go
// before
provider.RemoveModelFromConfig(ctx, logger, client, isName, namespace, modelName)
// after
provider.RemoveModelFromConfig(ctx, logger, client, isName, namespace, deployment.GetName(), modelName)
```

Model config ConfigMaps written before this change carry entries without `deployment_name`. They are inert and are not removed by cleanup; delete them by hand, or leave them until the InferenceServer goes away. Sidecars tolerate the new field in either direction, so controller and sidecar do not need to upgrade together.

**Release notes**

Fixed: deleting a Deployment now unloads its model from every target cluster, and only its own model config entry — a model shared with another Deployment stays loaded. Model config entries are now keyed by deployment and model name.

**Documentation Changes**

N/A — no user-facing interface change.


go/components/deployment/plugins/oss/cleanup/cleanup.go
```diff
@@ -36,22 +34,32 @@ func (a *CleanupActor) GetType() string {
...
+	// Cleanup is only complete once all of them are clear.
 	targets, err := common.ReadTargetClustersAnnotation(deployment)
 	if err != nil {
 		return conditionUtils.GenerateFalseCondition(condition, "UnableToReadTargetClusters", fmt.Sprintf("Unable to read target-clusters annotation: %v", err)), nil
 	}
+
+	// The model config is per-cluster, so cleanup is only complete once every cluster has
+	// dropped the entry. Checking only the control plane would release the finalizer while
+	// remote clusters keep serving the model.
+	for _, target := range targets {
```
badcount, line 50: Where is the check to see there is no same model has been used for different deployment, so when we delete one deployment it doesn't affect others?

badcount (approved):

go/components/deployment/plugins/oss/cleanup/cleanup.go
aritra0227: resolved.

Issue #2009, opened 2026-09-01
Trainer docstrings produce defective Python SDK Reference output


zhoward-1:
Four problems in public docstrings under `python/michelangelo/lib/trainer/torch/pytorch_lightning/`. These docstrings are the generation source for the Python SDK Reference (`pydoc-markdown.yml`), so each one becomes a defect in the published reference the moment those modules are added to the generator config.

Filed to preserve the findings from a review of #2007. Both #2007 and #2008 were closed deliberately — see [Context](#context) below.

## 1. `fused_model_submodule` exposes an internal system and an unfindable reference

`schema.py:249-258` (`IncrementalTrainingSpec`) and `schema.py:290-299` (`TransferLearningSpec`) both read:

> Schema-only in OSS today — carried through for forward compatibility with internal Michelangelo's warm-start config shape […] reconcile this divergence deliberately once OSS implements the stripping behavior (see the PR that ports it).

Three issues: it says bare "Michelangelo" (external-facing content is required to say "Michelangelo AI"); it describes a non-public counterpart system and its differing default; and "see the PR that ports it" points an external reader at something they cannot locate.

**Suggested fix:** state what the field does today (nothing — it is reserved), and keep the default-value divergence as a maintainer `#` comment on the field rather than in the public docstring.

## 2. `TrainingObserver` example renders as broken, uncopyable Python

`schema.py:28` uses a bare `Example::` RST literal block. pydoc-markdown fails to convert it: the rendered page loses the code block **and all indentation**, leaving flush-left Python that cannot be copied, with `{metrics}` / `{epoch}` sitting in body text.

Verified with a controlled A/B on a single pydoc-markdown version (4.8.2), generating from the current docstring vs. a fixed one:

| | Indentation of example lines |
|---|---|
| `Example::` (current) | `0` on every line — renders as a paragraph |
| `Example (observing training events)::` | `4` / `8` — renders as a code block |

The parenthesized form is already used twice in this same file (`schema.py:117`, `schema.py:132`) and converts correctly.

**Suggested fix:** one-line change at `schema.py:28` to match the parenthesized form used elsewhere in the file.

## 3. `profiler_sink` documents a module path that does not exist

`lightning_trainer.py:129-130` says:

> `:func:`~_private.util.comet_profiler_sink`` does this for Comet and `:func:`~_private.util.mlflow_profiler_sink`` does this for MLflow.

[...]

PR #2112, opened 2026-09-17
docs(examples): fix broken run commands and fabricated output in example READMEs

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Audited the six oldest README files under `python/examples/` (274–287 days untouched) against the code each one documents. Four were inaccurate and are fixed; two were already correct and are left unchanged.

| README | Verdict |
|---|---|
| `amazon_books_qwen/` | Fixed — broken run command, wrong output path |
| `gpt_oss_20b_finetune/` | Fixed — fabricated output block, wrong `cd` path |
| `llm_prediction/` | Fixed — both run commands broken, leaked personal path |
| `nomic_ai/` | Fixed — broken run command, fabricated output block |
| `model_manager/simple_custom/` | **Verified accurate — unchanged** |
| `model_manager/simple_custom_torch/` | **Verified accurate — unchanged** |

**Why?**

Three of the six documented run commands could not work as written, so anyone following them hit an immediate `ModuleNotFoundError`. Four contained "Expected Output" blocks quoting terminal text and return values that exist nowhere in the code.

**How did you test it?**

Every change is sourced from code, and the central failure was reproduced rather than assumed.

*Run commands — verified empirically.* `amazon_books_qwen` used `PYTHONPATH=examples`; `llm_prediction` (both backends) and `nomic_ai` omitted `PYTHONPATH` entirely. Running the real script-file invocation from `python/`:

```
PYTHONPATH=examples  → ModuleNotFoundError: No module named 'examples'
(no PYTHONPATH)      → ModuleNotFoundError: No module named 'examples'
PYTHONPATH="."       → import OK
```

Cause: `pyproject.toml:13-14` packages only `michelangelo`, not `examples`, and running a *script file* places the script's own directory on `sys.path`, not cwd. (Note: testing this with `python -c` gives a false pass, because `-c` *does* add cwd.) After the fix, all 7 example entrypoints resolve via `importlib.util.find_spec` under `PYTHONPATH="."`. Corrected form matches `python/examples/README.md:38-41`.

*Expected Output blocks — checked string by string against source.*

- `amazon_books_qwen`: claimed `Training completed with model saved to: /tmp/qwen_dual_encoder_model`. That path exists nowhere in the repo. Real log is `Training completed! Final train loss: …, Val loss: …` (`train.py:683-686`); real paths are `/tmp/qwen_dual_encoder_local.pt` (`train.py:689`) and `/tmp/qwen_dual_encoder_distributed.pt` (`train.py:526`). The banner lines were genuine (`amazon_books_qwen.py:63-65,87-89`) and are kept.
- `gpt_oss_20b_finetune`: claimed a result of `{'model_path': '/tmp/simple_model', 'train_loss': …, 'training_type': 'simple_local'}`. Actual return is `{"checkpoint_path": …, "mlflow_run_id": …}` (`simple_train.py:270-273`), written to `s3://default/ray_checkpoints` (`simple_train.py:137`) — no `/tmp` path appears in the directory at all. The quoted `{'loss': …, 'grad_norm': …}` lines are HuggingFace `transformers.Trainer` console format, which this codebase never uses (training goes through Lightning). The banner (`simple_workflow.py:55-57`) and the `trainable params:` line (real — `print_trainable_parameters()` at `model.py:90-91`) are kept.
- `nomic_ai`: claimed `{'model_path': '/tmp/nomic_bert_model', 'metrics': {...}}` and per-epoch perplexity. Actual return is `{'status': 'Training completed successfully'}` (`train.py:99`), saved to `./nomic_ai` (`train.py:57`); perplexity is never computed anywhere in the example.
- `llm_prediction`: claimed log lines (`Loading dataset: …`, `Prediction completed: 2/2 samples`) that match no log statement in the code; `data.py:30-53` emits no logs at all. Only `ok.` was real (`hf_prediction.py:72`, `vllm_prediction.py:74`) and is kept.

Where accurate output could not be sourced from code, the fabricated transcript was replaced with a description of the real return shape and artifact location rather than an invented substitute — none of these examples were executed (they require GPU, dataset downloads, or Kaggle credentials).

*Also fixed:* `llm_prediction/README.md:26` contained a leaked personal absolute path (`/Users/<name>/Uber/...`) as its `cd` target. `gpt_oss_20b_finetune` used `michelangelo_ai/...` (underscore; the org is `michelangelo-ai` and a clone lands at `michelangelo/python`). Both replaced with "run from the `python/` directory". Added the required poetry extra (`-E example` / `-E vllm`) to `llm_prediction` and `nomic_ai`, which previously omitted it.

*Scope note:* the same wrong command was duplicated in the code comment at `amazon_books_qwen.py:58`, so that one line is corrected too — the only non-README change.

**Potential risks**

None to runtime behaviour. The single code change is a comment. All other changes are Markdown.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Four example READMEs corrected; two verified accurate and left unchanged. No `docs/` pages affected.

**Follow-ups identified but deliberately not included**

- `nomic_ai/train.py:35` accepts a `test_data` parameter that is never referenced in the function body. The README previously advertised "evaluation" on the strength of it; that wording is removed here, but the dead parameter is a code fix and is left alone.
- `gpt_oss_20b_finetune/simple_train.py:223` still passes the deprecated `num_epochs=` into `LightningTrainerParam`. Already tracked as item 5 of #2009.
- No example was found to be dead — all six have live code and are linked from `docs/user-guides/examples/index.md`. Nothing was deleted.


paulzim (approved):
Reviewed by spot-checking every corrected claim against the actual source (not just against the old README text):

- **`PYTHONPATH="."` fix** — confirmed `pyproject.toml`'s `packages` only includes `michelangelo` (not `examples`), so the old `PYTHONPATH=examples`/no-`PYTHONPATH` forms would indeed raise `ModuleNotFoundError`. The corrected form matches the canonical pattern already used in `python/examples/README.md`.
- **amazon_books_qwen** — `train.py:684-685` prints `Training completed! Final train loss: {...}, Val loss: {...}` verbatim, and the local (`:689`) vs. distributed (`:526`) checkpoint paths match the new note exactly.
- **gpt_oss_20b_finetune** — `simple_train.py:246` prints `✅ Training completed, MLflow run: {run_id}` and returns `{"checkpoint_path": ..., "mlflow_run_id": ...}` (`:270-272`), matching the rewritten block. `gpt2` default model and `print_trainable_parameters()` call also confirmed.
- **llm_prediction** — `data_limit` defaults (2 for HF, 15 for vLLM) and `batch_size` (1 / 8) match `hf_prediction.py`/`vllm_prediction.py`; `print("ok.")` and the `Wrote {n} items to {out_path}` log line both confirmed in source.
- **nomic_ai** — `wikitext-2-raw-v1`, 512-token default, `./nomic_ai` save dir, and `{'status': 'Training completed successfully'}` return value all confirmed in `data.py`/`train.py`.
- Leaked personal path removal (`/Users/sally.lee/Uber/...`) and the `poetry install -E example` / `-E vllm` extras both check out against `pyproject.toml`.

Every one of these matched on the first check — no fabrications found in the corrected text. Nice, methodical audit.

Two minor, non-blocking notes:
- `gpt_oss_20b_finetune/README.md` still has no trailing newline, but that predates this PR — not introduced here.
- Agree with scoping out the `nomic_ai/train.py` unused-parameter and `gpt_oss_20b_finetune/simple_train.py` deprecated-argument follow-ups (the latter already tracked in #2009).

Approve.

sallycr (approved):
thx

PR #2113, opened 2026-09-17
docs(user-guides): record verification of backfill and file-sync guides

zhoward-1:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**What changed?**

Adds `last_verified: 2026-09-17` frontmatter to the two pages in the 110-day `docs/user-guides/` stale cohort that were verified accurate. **No content lines are changed** — the diff is 8 added lines, 0 removed.

The other four pages in the cohort are deliberately **not** included; see below.

**Why?**

These pages were checked against the current implementation and found correct, so they need a verification record rather than an edit.

**How did you test it?**

*First, the frontmatter mechanism itself* — Docusaurus 3.9.2 tolerates the unknown `last_verified` key, confirmed two independent ways:

1. `DocFrontMatterSchema` applies `.unknown()` (`website/node_modules/@docusaurus/plugin-content-docs/lib/frontMatter.js:42`). Executing the real `validateDocFrontMatter` against `{last_verified: '2026-09-17'}` passes, alone and alongside known keys. A negative control (`sidebar_position: 'not-a-number'`) is correctly **rejected**, proving the validator actually runs rather than no-opping.
2. A full `bun run build` of the site with the key in place: exit 0, no frontmatter warning, all 134 pages emitted, the affected page rendered with its H1 and body intact, and `last_verified` absent from the rendered HTML (consumed as frontmatter, not printed).

*Then the page content:*

**`backfill.md` — accurate.** Verified: `@uniflow.task()` / `@uniflow.workflow()` / `create_context` (`uniflow/core/decorator.py:338,433`, `core/context.py:61-119`); `SparkTask` import path (`plugins/spark/task.py:43`); `ma pipeline dev-run -f … --env KEY=VALUE` including `_process_env_variables` (`cli/mactl/plugins/entity/pipeline/dev_run.py:43-131,360`); `STARLARK_TIME` set into pod env as `"unix:" + executionTs` (`go/components/pipelinerun/actors/executeworkflow.go:562-563`).

The page's central claim — that setting `STARLARK_TIME` makes Uniflow's time utilities return the historical timestamp — needs a note, because a repo-local search makes it look broken: nothing in this repo reads the variable, and `uniflow/core/lib/time.py:21-28` returns `py_time.time()` unconditionally. The consumer is **external**: `github.com/cadence-workflow/starlark-worker` (`go/go.mod:7`), whose `plugin/time/module.go` implements `getEffectiveTime` honoring `STARLARK_TIME`, with a `force_system_time` argument to opt out. The documented behaviour holds for the `dev-run` path the page describes. Flagging this explicitly so the next person auditing this page doesn't reach the wrong conclusion from a repo-only grep.

**`file-sync-testing-flow-runbook.md` — accurate.** Verified: `remote-run --storage-url/--image/--file-sync` (`core/context.py:199-253`); `ma pipeline dev-run --file-sync --file` (`dev_run.py:43-131`); "patches an existing image rather than building one" (`remote_run.py:32,66-72,237-243`, `dev_run.py:282-295`); git-aware diff behaviour with and without a build SHA (`file_sync.py:210-245`); untracked files included via `git ls-files --others --exclude-standard` and `.gitignore` respected (`file_sync.py:248-255`); `UF_FILE_SYNC_TARBALL_URL` set at `remote_run.py:71-72` / `dev_run.py:335` and read at `file_sync.py:571,648`; MinIO troubleshooting URL and credentials (`cli/sandbox/resources/minio.yaml:38-40`). One post-cutoff change to `dev_run.py` (2026-06-08) added a log statement only — no behaviour change.

**Potential risks**

None. Frontmatter-only addition, verified to build.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

Two `docs/user-guides/` pages gain a `last_verified` key. No prose changes.

**Excluded from this PR — the other four pages in the cohort are not accurate**

Handled separately, since a verification stamp should not be applied to a page with a known error:

- `cache-and-pipelinerun-resume-form.md` — documents `ma pipeline run --revision <name>`. That flag does not exist on this command; the required flag is `--name` (`cli/mactl/plugins/entity/pipeline/run.py:63`, enforced at `:267`). The example fails argparse as written. The same wrong flag also appears in `pipeline-management.md:273,279`, outside this cohort.
- `reference-system.md` — its "Cross-Framework Data Passing" section documents automatic Ray↔Spark dataset conversion. No such conversion exists: `unref` dispatches on the type recorded at *write* time with no knowledge of the consuming task's framework (`uniflow/core/ref.py:60-79`), and no bridging code exists in `uniflow/plugins/`.
- `type-system.md` — the codec content is correct (exactly 5 codecs, matching `core/codec.py:614-622`), but all 15 appendix samples use bare `@uniflow.task()`, while `config` is a required positional parameter (`core/decorator.py:338`). Every appendix sample raises `TypeError` if copy-pasted.
- `running-uniflow.md` — accurate apart from naming `apply_local_diff`, a symbol that does not exist anywhere in the codebase (nearest real artifacts: the `PipelineRunSpec.local_diff` proto field, and an `apply_to_local_diff` mention in `plugins/ray/task.star:60`). Cosmetic, but it is an error, so no stamp.


paulzim (approved):
Reviewed the frontmatter-only diff and independently re-verified the content claims in both pages (not just trusting the PR description):

- **`STARLARK_TIME`** — confirmed Go sets it as `"unix:" + executionTs` (`executeworkflow.go:563`), matching the doc's `STARLARK_TIME=unix:<seconds>` format. I went looking for the Python-side consumer that makes `uniflow_time()` actually respect it and couldn't find one anywhere in this repo — `python/michelangelo/uniflow/core/lib/time.py`'s `time()` is a plain `py_time.time()` call, and `executeworkflow_test.go` only asserts the env var gets wired into the pod spec, not that it changes runtime behavior. This lines up with the PR description's own disclosure that this specific mechanism "relies on logic in an external repository" — good catch flagging that rather than glossing over it. Only suggestion: it'd be worth a one-line note in `backfill.md` itself (not just the PR description) so a future reader doesn't assume the whole mechanism is auditable from this repo.
- **CLI flags** — `--env` (`dev_run.py:69`), `-f`/`--file` (`dev_run.py:56`, `crd.py:838`), and `--file-sync` (present in `file_sync.py`/`remote_run.py`/`dev_run.py`) all confirmed as real flags.
- **`SparkTask`** import path confirmed (`uniflow/plugins/spark/task.py:43`).
- **MinIO sandbox credentials** — `minioadmin`/`minioadmin` and port `9091` both confirmed against `sandbox/resources/minio-credentials.yaml` and `minio.yaml`.
- **All 8 outbound links** from both pages (`pipeline-running-modes.md`, `pipeline-management.md`, `type-system.md`, `sandbox-setup.md`, `running-uniflow.md`, `cache-and-pipelinerun-resume-form.md`, `set-up-triggers.md`, `model-registry-guide.md`) resolve to real files.
- Docusaurus version claim (`3.9.2`) matches `website/package.json` exactly. Couldn't independently verify the `frontMatter.js:42` `.unknown()` claim since `node_modules` isn't installed in this checkout, but that's a low-risk claim — a real site-build failure from an unrecognized frontmatter key would surface immediately in CI, so that's the practical gate here.
- No `last_verified` precedent exists elsewhere in `docs/` — this PR is introducing the convention, not following one. Not a blocker, just flagging in case the docs team wants to weigh in on the pattern before it spreads to the other pages in this cohort.

Approve.

PR #2114, opened 2026-09-17
feat(go): support a custom Triton serving image on InferenceServer

aritra0227:
- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`InferenceServer.spec.initSpec.servingSpec.image` now selects the model server container image, carrying a URI, pull policy, and pull secret names.

The Triton backend resolves the image from the spec, then an operator-configured default (`inferenceServer.triton.defaultImage` in Helm), then the compiled-in `defaultTritonImage`. Validation rejects a reference without an explicit tag or digest, and an unsupported pull policy.

That compiled-in default moves from `nvcr.io/nvidia/tritonserver:25.01-py3` to `ghcr.io/michelangelo-ai/triton-serving`, which adds the ML framework dependencies the stock Triton image omits. It is the single place the default is recorded; `inferenceServer.triton.defaultImage` ships empty so it reads as an override rather than a second copy of the pin.

`build-triton-image.yaml` now publishes that image for arm64 as well as amd64, building each platform on its native runner and merging the digests, as `build-examples-image.yaml` already does. The previous workflow restricted itself to amd64 on the premise that the base image is amd64-only, which is not the case.

Because the image is a private GHCR package, `ma sandbox demo inference` and `ma sandbox demo inference-multicluster` now write a `dockerconfigjson` Secret into every cluster the InferenceServer targets, sourced from `GHCR_TOKEN`, `GITHUB_TOKEN`, or the GitHub CLI's token, and both demo manifests name it in `imagePullSecrets`.

The image source is a `oneof` holding a single `uri` member today. A build-from-Dockerfile variant can be added to it later without disturbing the pull policy and pull secret fields, which apply however the image is produced. The generated CRD schema is unaffected: oneof members render as ordinary sibling properties.

**Why?**

The image was hardcoded, so running a patched or internally mirrored Triton meant editing and redeploying the controller. Operators need the default set per environment, and individual InferenceServers need to override it.

Requiring a tag or digest keeps a workload from silently running an implicit `:latest`, where a registry push would change what is serving without any change to the resource.

**How did you test it?**

Unit tests cover image validation, including tag and digest forms, registry ports, and pull policy values. `./tools/bazel test //go/... //proto/... --build_tests_only` passes.

End-to-end in the k3d sandbox, across four InferenceServers covering both topologies and both resolution paths:

| InferenceServer | Topology | `image.uri` | Resolved image | Result |
|---|---|---|---|---|
| `img-single-default` | single-cluster | unset | `triton-serving:sha-e9b9955` | SERVING |
| `img-single-custom` | single-cluster | `tritonserver:23.04-py3` | `tritonserver:23.04-py3` | SERVING |
| `img-mc-default` | 2 clusters | unset | `triton-serving:sha-e9b9955` on both | SERVING |
| `img-mc-custom` | 2 clusters | `tritonserver:23.04-py3` | `tritonserver:23.04-py3` on both | SERVING |

The default pulled from the private registry using the Secret the sandbox helper creates (10.2 GB in 5m30s), resolved to the arm64 variant on the aarch64 nodes, and reported `torch 2.4.1` / `transformers 4.44.2` inside the container. Helm renders `defaultImage: ""` by default and honours `--set inferenceServer.triton.defaultImage`.

**Potential risks**

The default serving image changes for any Triton-backed InferenceServer that does not set one. The new image is the stock Triton release plus torch and transformers, so the server itself is the same 25.01 build, but it is a different registry and a much larger pull on first schedule.

The image is a private GHCR package. Clusters outside the sandbox need their own pull credentials; a missing secret surfaces as `ImagePullBackOff` on the pod rather than as a validation failure. The same applies to any secret named in `imagePullSecrets`, which this PR does not create or replicate outside the sandbox demos.

Editing `image.uri` on a live InferenceServer does not restart the workload, since `createTritonDeployment` returns early when the Deployment exists. The field takes effect at creation.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Added: `InferenceServer.spec.initSpec.servingSpec.image` sets the model server container image, with a cluster-wide override at `inferenceServer.triton.defaultImage`.

Changed: the default Triton serving image is now `ghcr.io/michelangelo-ai/triton-serving`, published for amd64 and arm64.

**Documentation Changes**

N/A — new optional field, documented inline in the proto and Helm values. Follow-up tracked in #2115 to replace the bespoke workload fields with a Kubernetes pod template.


proto/api/v2/inference_server.proto
```diff
@@ -14,6 +14,28 @@ import "michelangelo/api/v2/pod.proto";
...
+  oneof source {
+    // Fully-qualified reference to an image that is already pushed, including a
+    // tag or a digest. Examples: "nvcr.io/nvidia/tritonserver:25.01-py3",
+    // "myregistry.io/team/triton@sha256:6f1a...".
+    string uri = 1;
+  }
+
+  // Pull policy for the image: "Always", "IfNotPresent", or "Never". Defaults
+  // to "IfNotPresent" when empty.
+  string image_pull_policy = 2;
```
badcount, line 34: The image pull policy should not be in the ServingImage, this should be able to control in the control plane

proto/api/v2/inference_server.proto
```diff
@@ -14,6 +14,28 @@ import "michelangelo/api/v2/pod.proto";
...
+
+  // Names of kubernetes.io/dockerconfigjson Secrets used to pull from a private
+  // registry. Each must already exist in the InferenceServer's namespace in
+  // every target cluster.
+  repeated string image_pull_secrets = 3;
```
badcount, line 39: this should also be in env config rather than built with inferenceServer

helm/michelangelo/values.yaml
```diff
@@ -377,6 +377,12 @@ inferenceServer:
     serviceNamespace: default
     portName: http
 
+  triton:
+    # defaultImage is used by Triton-backed InferenceServers that don't set
+    # spec.initSpec.servingSpec.image. Empty means the controller's built-in
+    # image.
+    defaultImage: "nvcr.io/nvidia/tritonserver:25.01-py3"
```
badcount: The default image should be pointing to github michelangelo-ai triton image

badcount (approved):

proto/api/v2/inference_server.proto
aritra0227: resolved offline. This seems to be standard in oss.

proto/api/v2/inference_server.proto
aritra0227: resolved offline. This seems to be standard pattern in oss.

helm/michelangelo/values.yaml
aritra0227: resolved offline. This seems to be standard pattern in oss.

PR #2116, opened 2026-09-18
ci: build the Triton serving image for arm64 as well as amd64

aritra0227:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

`build-triton-image.yaml` now publishes `ghcr.io/michelangelo-ai/triton-serving` for `linux/arm64` as well as `linux/amd64`. Each platform builds on its native runner and the two digests merge into a single manifest, the same structure `build-examples-image.yaml` already uses.

The workflow previously pinned `platforms: linux/amd64`, with a comment explaining that the multi-arch handling was unnecessary because the base image is amd64-only:

```yaml
# nvcr.io/nvidia/tritonserver is amd64-only, so this doesn't need the
# multi-arch QEMU/buildx-merge dance used by build-examples-image.yaml.
```

That premise is wrong. `nvcr.io/nvidia/tritonserver:25.01-py3` publishes both `linux/amd64` and `linux/arm64`.

The trigger also now includes the workflow's own path, so a change to the build itself republishes the image.

**Why?**

Every previously published tag is amd64-only, so containerd rejects the image on an arm64 node. Local k3d sandboxes on Apple Silicon run `aarch64` nodes and cannot use it at all. The stock `nvcr.io/nvidia/tritonserver` image works there today only because it is genuinely multi-arch, so the restriction applied here is what made our derived image unusable.

This blocks pointing the InferenceServer controller's default serving image at this package, which #2114 does.

**How did you test it?**

Dispatched this workflow from the branch. Both platform jobs and the merge job succeeded, publishing `linux/amd64, linux/arm64` under `sha-e9b9955` and `25.01-py3-torch`:

```
$ docker manifest inspect ghcr.io/michelangelo-ai/triton-serving:25.01-py3-torch
  linux/amd64
  linux/arm64
```

Pulled in a k3d sandbox on an `aarch64` node: the image resolved to the arm64 variant, Triton started, and `torch 2.4.1` / `transformers 4.44.2` imported inside the container.

The arm64 job failed once on a transient GHCR push error (`error writing layer blob: not_found`) and succeeded on re-run, with the amd64 job unchanged.

**Potential risks**

Low. Nothing consumes this image from the controller yet on `main`, so republishing it changes no running workload. The amd64 manifest is built from the same Dockerfile as before.

Build time roughly doubles in wall clock terms for the arm64 job, which installs torch natively on an `ubuntu-24.04-arm` runner.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Changed: `ghcr.io/michelangelo-ai/triton-serving` is now published for amd64 and arm64.

**Documentation Changes**

N/A — CI-only change, documented inline in the workflow and Dockerfile.


.github/workflows/build-triton-image.yaml
```diff
@@ -45,16 +63,78 @@ jobs:
...
+        working-directory: /tmp/digests
+        run: |
+          docker buildx imagetools create \
+            $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
+            $(printf '${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@sha256:%s ' *)
+
+      - name: Inspect final image
+        run: |
+          docker buildx imagetools inspect \
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.meta.outputs.version }}
```
aritra0227, line 140: this is the exactly the same merge methodology used in [.github/workflows/build-examples-image.yaml](https://github.com/michelangelo-ai/michelangelo/blob/main/.github/workflows/build-examples-image.yaml#L111)

badcount (approved):

PR #2121, opened 2026-09-18
fix(ci): scope NODE_AUTH_TOKEN to the npm-publish job, not just its final step

sallycr:
Found blocking the `v0.11.0-rc.1` cut: `npm-publish.yml`'s `actions/setup-node@v7` (bumped from v4 by #2076, a routine Dependabot PR) resolves the registry's `${NODE_AUTH_TOKEN}` placeholder as soon as `cache: yarn` + `registry-url` are both set at the Setup Node.js step — before the step-scoped env on the later 'Publish to npm' step supplies it. Fails with `Failed to replace env in config: ${NODE_AUTH_TOKEN}` at the `yarn cache dir` call inside setup-node.

Never caught by #2076's own CI run or any push since, because every prior `push`-triggered run of this workflow skipped the Setup Node.js step entirely (the 'changed' guard only runs it when a package.json version actually changed in that push) — `v0.11.0-rc.1`'s cut was the first real exercise of this step since the v4→v7 bump.

Fix: move `NODE_AUTH_TOKEN` to job-level `env:` so it's in scope for every step, not just the final publish step. actionlint clean.


paulzim (approved):
Confirmed the underlying bug: `actions/setup-node@v7`'s yarn-cache logic does validate the registry's `${NODE_AUTH_TOKEN}`-embedded `.npmrc` as soon as `cache: yarn` + `registry-url` are both set on that step — before the old step-scoped `env:` on "Publish to npm" would ever supply it. So job-scoping the token does fix the reported failure, and the YAML parses cleanly (checked with `yaml.safe_load` — no `actionlint` binary available in this checkout to verify further).

One follow-up worth raising before merge: moving `NODE_AUTH_TOKEN` to job-level `env:` puts it in scope for all 7 steps, including two that didn't have it before and that execute non-first-party code:

- **Install dependencies** — `yarn setup` (`javascript/package.json`) runs `yarn install --frozen-lockfile` across the full workspace. No `enableScripts: false` or equivalent guard exists, so any transitive dependency's postinstall script now runs with the npm publish token in its environment.
- **Build** — runs the workspace build/bundler, same exposure.

Previously the token was scoped only to "Publish to npm," which runs just `npm view`/`npm publish` — minimal, first-party commands. Widening this hands a publish-capable token to a much larger, less-audited surface (dependency install + build tooling), which is a real supply-chain exposure increase, not just a style nit.

A narrower fix should resolve the same timing bug: add step-level `env: NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` to the "Setup Node.js" step (where the actual early-resolution happens) in addition to keeping it on "Publish to npm" — leaving Install and Build without it. That's a smaller diff than the current job-level change and keeps the token's blast radius the same as before.

Given this is blocking a release cut, not blocking on this — the job-level fix is correct and safe to ship now. Approve, with a suggested immediate follow-up PR to narrow `NODE_AUTH_TOKEN` back to step-level scope (Setup Node.js + Publish to npm only) once the release pressure is off.

Approve.

sallycr:
Good catch — pushed a narrower fix (daac2658) per your suggestion: `NODE_AUTH_TOKEN` is now step-scoped on `Setup Node.js` (where the early resolution actually happens) and `Publish to npm` only, not job-wide. Install/Build no longer see the token, same blast radius as before the `setup-node` v4→v7 bump. actionlint clean.

PR #2127, opened 2026-09-18
fix(ci): scan :main instead of stale :latest tag in CVE Scan

sallycr:
- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**What changed?**

Four workflow-YAML changes, all part of the same "dev-tip (`:main`) vs. real release (`:latest`) must never be conflated" principle:

- `.github/workflows/cve-scan.yml`: the Trivy `image-ref` for **all four services** (`apiserver`/`worker`/`controllermgr`/`ui`; both the SARIF-output step used for the Security tab and the human-readable table-output step used for the job summary) now points at `:main` instead of `:latest`. `ui-release.yml` uses the same `docker/metadata-action` tag-list shape as `dev-release.yml` (including `type=ref,event=branch`), so `ui:main` already exists and has the same reliability properties as the other three services' `:main` tags. Note `ui:main` updates less often than the other three (only rebuilds when `javascript/app/package.json`'s version changes on a plain `main` push, vs. their unconditional per-push rebuild), but is still strictly more current than `ui:latest` was.
- `.github/workflows/dev-release.yml`: `docker/metadata-action`'s config sets `flavor: latest=false` and an explicit `type=raw,value=latest,enable=...` tag entry. **Corrected in this update:** `:latest` now fires *only* on a real, non-prerelease `v*` tag push — the earlier version of this PR also fired it on every plain `main` push, which meant `:latest` and `:main` were redundant almost all the time (the exact dev-tip/release conflation this fix is meant to eliminate). (This file's matrix only covers `apiserver`/`worker`/`controllermgr`; `ui` is built by the separate `ui-release.yml`.)
- `.github/workflows/ui-release.yml` (new in this update): applies the identical fix. It previously had **no explicit `type=raw,value=latest` entry at all**, so `ui:latest` was governed purely by `docker/metadata-action`'s implicit `flavor.latest=auto`, which fires unconditionally on *any* tag push — including RC/alpha/beta tags, with no pre-release filter. Now it has the same explicit, pre-release-excluded `latest` entry (and `flavor: latest=false` to disable the implicit behavior) as `dev-release.yml`.
- `.github/workflows/integration-test.yaml` (new in this update): the `image_tag` input's default (both `workflow_dispatch` and `workflow_call` declarations) changes from `"latest"` to `"main"`, with updated help text. This input targets `ghcr.io/michelangelo-ai/examples`, built by `build-examples-image.yaml`, which already produces correct `:latest`/`:main` tags from the same build (no lag bug there) — but defaulting to `:latest` still used the "release" tag name to mean "dev tip," the same naming anti-pattern this PR fixes elsewhere. `build-examples-image.yaml` itself is out of scope for this PR (flagged as a known follow-up).

**Why?**

`CVE Scan` had failed on every run for weeks, reporting Go stdlib CVEs already fixed by newer `MODULE.bazel` `go_sdk.download` pins. Root cause: `:latest` only advanced when a `v*` git tag was pushed, so it was frozen on whatever the last tagged release build was — one confirmed case, 10 days stale — while `main` had already moved on and picked up newer CVE fixes. `:latest` also has no real consumer in this repo today (Helm chart defaults to `.Chart.AppVersion`, `values-k3d.yaml` pins `tag: main`) other than this scan, so pointing the scan at `:main` is the direct fix.

The `dev-release.yml`/`ui-release.yml` changes are a secondary hygiene fix, corrected in this update to fully separate the two tags' meanings: `:latest` should mean "most recent real release," never "current `main` tip." Standard OSS practice (many core Kubernetes-ecosystem images) keeps these strictly distinct — some skip `:latest` publication entirely rather than let it drift into meaning dev-tip. Letting `:latest` also fire on every `main` push (as an earlier version of this PR did) would have made it redundant with `:main` almost all the time, reintroducing the same conflation this PR set out to fix. `integration-test.yaml`'s default is switched for the same naming-consistency reason, even though its underlying image has no lag bug.

**How did you test it?**

- `python3 -c "import yaml; yaml.safe_load(open(...))"` against all four modified files — parses cleanly.
- `actionlint` against all four modified files — no new findings (pre-existing shellcheck/runner-label notices on unrelated lines are untouched by this PR).
- `git diff main --name-only` confirms only these four files changed.
- Post-merge follow-up: manually trigger `CVE Scan` via `workflow_dispatch` and confirm all four services' steps resolve to `:main` and no longer report the stale Go stdlib CVEs. Push a scratch-branch commit and confirm `docker/metadata-action`'s emitted tags for `dev-release.yml`/`ui-release.yml` do **not** include `latest`; push a real `v*` tag and confirm `latest` **is** emitted and matches that build's `:sha-...` digest; push a `v*-rc.*`-style tag and confirm `latest` is **absent**.

**Potential risks**

Low. No production code path is touched — no Dockerfile, Helm template, or application source changes. `:latest` no longer moving on `main` pushes (only on real release tags) has no known blast radius since nothing in the repo's deployment paths consumes `:latest` today.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

N/A

**Documentation Changes**

N/A


.github/workflows/ui-release.yml
```diff
@@ -87,6 +89,7 @@ jobs:
             type=ref,prefix=pr-,event=pr
             type=raw,value=v${{ needs.check-version-change.outputs.new-version }},enable={{is_default_branch}}
             type=raw,value=${{ steps.bare_version.outputs.version }},enable=${{ steps.bare_version.outputs.is_tag }}
+            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-rc') && !contains(github.ref, '-alpha') && !contains(github.ref, '-beta') }}
```
craigmarker, line 92: I thought ui-release was a stable release? 🤔 Or is this the old release mechanism?

badcount (approved):

paulzim (approved):
**Approve**

Verified this against `upstream/main`, it matches exactly what's described: 4 files, +11/-5.

What I checked:

- **`cve-scan.yml`**: `:latest` → `:main` for all four services (`apiserver`/`worker`/`controllermgr`/`ui`), both the SARIF and table-output Trivy steps. Confirmed via `values-k3d.yaml` that nothing in the chart's own image references depend on `:latest` (it pins `tag: main`, and defaults elsewhere come from `.Chart.AppVersion`), and a repo-wide grep turns up no other consumer of the Docker `:latest` tag — CVE Scan really was the only thing reading it.
- **`dev-release.yml` / `ui-release.yml`**: the `flavor: latest=false` + explicit `type=raw,value=latest,enable=<real-release-only condition>` pattern is correct. I pulled the actual `docker/metadata-action` README to confirm: `flavor.latest=auto` (the prior, implicit default) fires `latest` on `type=ref,event=tag` output — i.e. on *any* git tag push, RC/alpha/beta included, with no relation to default-branch pushes. That lines up exactly with the observed bug (`:latest` frozen at whatever tag last shipped, real or RC, while `:main` kept moving on every push) and rules out the "double latest tag" footgun called out in `docker/metadata-action` issue #593, since `flavor.latest` is explicitly set to `false` here rather than left on `auto`/`true` alongside the manual raw tag. The `startsWith(...) && !contains('-rc') && !contains('-alpha') && !contains('-beta')` guard correctly matches every real vs. pre-release tag currently in the repo (checked all 19 existing `v*` tags).
- **`integration-test.yaml`**: default bump from `"latest"` → `"main"` for the `image_tag` input, in both `workflow_dispatch` and `workflow_call`. Correct as a naming-consistency change. Side note (pre-existing, not introduced by this PR, not a blocker): `image_tag` isn't actually wired into any step in this job today — it's declared but unused on `main` too. Not this PR's job to fix, just flagging in case it's news.
- YAML sanity: `python3 -c "import yaml; yaml.safe_load(...)"` against all four files — parses cleanly. (`actionlint` isn't available in my environment to independently re-run, but the YAML structure and GitHub Actions expression syntax match patterns already in production use elsewhere in these same files.)
- Confirmed via `git log` that the PR went through 3 iterative commits, with the final one fixing the redundant-`:latest`-on-every-main-push issue the PR description calls out — the version under review is the corrected one.

No production code paths touched (workflow YAML only), and the risk assessment in the PR description (no known consumer of `:latest` besides the scan itself) checks out against my own grep.

.github/workflows/ui-release.yml
sallycr: @craigmarker it was not. latest moves to the any release including rc in the current setting.

sallycr: https://github.com/michelangelo-ai/michelangelo/pkgs/container/ui/versions?page=3

PR #2128, opened 2026-09-19
chore(go): bump go.uber.org/cadence to v1.4.0-rc.5

yangfan0328:
- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Dependency bump

**What changed?**

Bumps `go.uber.org/cadence` from `v1.3.1-rc.11` to `v1.4.0-rc.5` in `go/go.mod`. `go mod tidy` pulled in two transitive updates:

- `go.uber.org/yarpc` `v1.80.0` → `v1.88.0`
- `github.com/uber/cadence-idl` → `v0.0.0-20260818192101-d6d4d81fa739` (indirect)

No Go source changes.

**Why?**

`v1.4.0-rc.1+` is the first Cadence Go client line that ships the **Schedules API** (`client.ScheduleClient()`, `ScheduleSpec`, `SchedulePolicies`, `BackfillRequest`, and the `CadenceScheduleID` / `CadenceScheduleTime` / `CadenceScheduleIsBackfill` search attributes). This is a prerequisite for adding a Cadence Schedules implementation to the TriggerRun controller next to the existing Temporal Schedules path (see the discussion around #2047 on `CronSchedule.start_time` / catch-up support). Landing the bump on its own first keeps any dependency churn separate from the controller work.

Notes on the release line:
- `v1.4.0-rc.5` (2026-09-03) is the latest tag; there is no stable `v1.4.0` yet. The Cadence team has indicated a stable cut is planned; we should re-pin when it lands.
- Breaking changes listed in the 1.4.0 notes (`x/blocks` removal, `ScheduleClient.Update` becoming a describe-then-update callback, active-active IDL refactor, `compatibility.NewThrift2ProtoAdapter` signature) do not touch anything this repo uses today. `go/base/workflowclient/cadenceclient` only uses the `client.Client` surface.
- Schedules require Cadence **server ≥ v1.4.1** with the per-domain scheduler worker enabled. Nothing in this PR depends on that; it only matters once schedule code is added.

**How did you test it?**

- `go build ./...` — clean
- `go vet ./base/workflowclient/... ./components/triggerrun/...` — clean
- `go test ./base/workflowclient/... ./components/triggerrun/... ./worker/workflows/trigger/... ./worker/activities/trigger/...` — all pass (`go/worker/activities/trigger` fails at setup locally because the generated `mock/.../v2mock` package is not present outside CI; verified identical on unmodified `main`)
- `go mod tidy` produces no further diff

**Potential risks**

Low. yarpc 1.80 → 1.88 is the largest transitive jump; it is exercised by the passing cadenceclient tests. If anyone runs an embedded Cadence server (`github.com/uber/cadence/host`) in tests, note that the 1.4.0 client changed `compatibility.NewThrift2ProtoAdapter` and requires a server build compiled against it — this repo does not embed the server.

**Breaking Changes**

- [x] No breaking changes

**Release notes**

Dependency: Cadence Go client upgraded to v1.4.0-rc.5 (Schedules API now available to Go code). No behavior change.

**Documentation Changes**

None.


go/go.mod
```diff
@@ -33,10 +33,10 @@ require (
 	go.temporal.io/api v1.49.1
 	go.temporal.io/sdk v1.35.0
 	go.temporal.io/sdk/contrib/tally v0.2.0
-	go.uber.org/cadence v1.3.1-rc.11
+	go.uber.org/cadence v1.4.0-rc.5
 	go.uber.org/config v1.4.0
 	go.uber.org/fx v1.22.0
-	go.uber.org/yarpc v1.80.0
+	go.uber.org/yarpc v1.88.0
```
yangfan0328, line 39: **Why yarpc moves 1.80.0 → 1.88.0:** it is forced by cadence itself, not an opportunistic bump. `go.uber.org/cadence@v1.4.0-rc.5` declares `require go.uber.org/yarpc v1.88.0` in its own go.mod:

https://github.com/cadence-workflow/cadence-go-client/blob/v1.4.0-rc.5/go.mod#L28

(v1.3.1-rc.11 only required v1.55.0, which is why our v1.80.0 was sufficient before.) Go's minimal version selection takes the maximum of every module's declared minimum, so the graph cannot resolve to 1.80.0 anymore. Verified: pinning 1.80.0 by hand and running `go mod tidy` raises it straight back to 1.88.0, which is also what CI's tidy check in `main.yml` would enforce.

go/go.mod
```diff
@@ -125,6 +125,7 @@ require (
 	github.com/tidwall/pretty v1.2.0 // indirect
 	github.com/twmb/murmur3 v1.1.8 // indirect
 	github.com/uber-go/mapdecode v1.0.0 // indirect
+	github.com/uber/cadence-idl v0.0.0-20260818192101-d6d4d81fa739 // indirect
```
yangfan0328, line 128: **Why `cadence-idl` appears as a new `// indirect` line:** the module was already in our build graph on `main` (it is in go.sum there), but it was not listed in go.mod because none of the cadence packages we import pulled it in. With rc.5 the Schedules wiring in `go.uber.org/cadence/internal` imports `github.com/uber/cadence-idl/go/proto/api/v1`, and rc.5 requires this exact pseudo-version:

https://github.com/cadence-workflow/cadence-go-client/blob/v1.4.0-rc.5/go.mod#L21

`go mod why -m github.com/uber/cadence-idl` on this branch:
```
go/base/workflowclient/cadenceclient
go.uber.org/cadence/client
go.uber.org/cadence/internal
github.com/uber/cadence-idl/go/proto/api/v1
```

Since Go 1.17 module-graph pruning, go.mod must list every module that provides a transitively imported package (marked `// indirect` when we do not import it directly). Deleting this line and running `go mod tidy` re-adds it verbatim; go.sum alone cannot carry it.

yangfan0328 (commented):
Self-review notes on the two transitive changes, so reviewers don't have to re-derive them.

